PR #53, opened 2026-02-18
Open cmd+clicked terminal links in cmux browser panel

lawrencecchen:
## Summary
- Handle `GHOSTTY_ACTION_OPEN_URL` in the Ghostty action callback so cmd+clicking a URL in the terminal opens it in a cmux browser panel instead of the system default browser
- Reuses an existing right-side pane via `preferredBrowserTargetPane`, or creates a new horizontal split if none exists

## Test plan
- [ ] Cmd+click a URL in the terminal — should open in a browser panel, not the system browser
- [ ] If a right-side browser pane already exists, the link should open as a new tab there
- [ ] If no right-side pane exists, a new horizontal split should be created


cooper667:
Think this would be good as an option? I want clicks to open in my regular browser most of the time

PR #57, opened 2026-02-18
Fix flaky WebKit escape focus tests

lawrencecchen:
## Summary
- App-side: retry `isWebViewFocused` check with increasing delays (0.05-0.5s) for the exit-address-bar case
- Test-side: send a second Escape if the first only clears suggestions/editing state (Chrome-like two-stage escape)

## Test plan
- [x] `testEscapeLeavesOmnibarAndFocusesWebView` passes on VM
- [x] `testCmdNWorksWhenWebViewFocusedAfterTabSwitch` passes on VM
- [x] `testCmdWWorksWhenWebViewFocusedAfterTabSwitch` passes on VM
- [x] `testCmdShiftWWorksWhenWebViewFocusedAfterTabSwitch` passes on VM


lawrencecchen:
@codex review

Issue #151, opened 2026-02-20
SSH support with remote dev server port proxying


lawrencecchen:
## Summary

Support SSH connections to remote hosts, with automatic port proxying so browser panels can access services running on the remote machine.

## Motivation

Many developers work on remote dev servers (cloud VMs, containers, etc.) for GPU access, environment isolation, or team infrastructure. Currently cmux only supports local terminals. SSH support would let users open a terminal tab connected to a remote host and have browser panels automatically proxy remote ports (e.g., a dev server on remote:3000 accessible locally).

This also enables better isolation for concurrent AI agents — each agent could run in its own warm container/VM rather than sharing the local filesystem.

## Desired behavior

- Open an SSH terminal tab to a remote host
- Automatically detect ports opened on the remote host (similar to VS Code Remote)
- Proxy remote ports locally so browser panels can access them (e.g., `localhost:3000` → remote:3000)
- Surface proxied ports in the sidebar (related to https://github.com/manaflow-ai/cmux/issues/129)

## References

- Discussed in [HN thread](https://news.ycombinator.com/item?id=47079718) by iamnafets and others
- VS Code Remote SSH is the closest UX reference

PR #239, opened 2026-02-21
Add cmux ssh remote workspaces with auto port forwarding

lawrencecchen:
## Summary
- add first-class `cmux ssh` workspaces and remote session metadata in the CLI and socket API
- add daemon-backed SSH remote browser proxying, including the custom localhost alias path for remote loopback browsing
- expose remote workspace state in the app UI and socket responses
- keep the SSH shell wrapper and mux behavior stable for repeated remote sessions

## Testing
- `./scripts/reload.sh --tag ssh-only-pr`
- `CMUX_SOCKET=/tmp/cmux-debug-ssh-only-pr.sock CMUXTERM_CLI=/tmp/cmux-ssh-only-pr/Build/Products/Debug/cmux CMUX_SSH_TEST_HOST=cmux-macmini python3 tests_v2/test_ssh_remote_browser_eval_snapshot_regression.py`

Closes https://github.com/manaflow-ai/cmux/issues/151

### DIFF
--- a/Sources/Workspace.swift
+++ b/Sources/Workspace.swift
@@ -647,14 +647,107 @@ extension Workspace {
     }
 }
 
+final class WorkspaceRemoteDaemonPendingCallRegistry {
+    final class PendingCall {
+        let id: Int
+        fileprivate let semaphore = DispatchSemaphore(value: 0)
+        fileprivate var response: [String: Any]?
+        fileprivate var failureMessage: String?
+
+        fileprivate init(id: Int) {
+            self.id = id
+        }
+    }
+
+    enum WaitOutcome {
+        case response([String: Any])
+        case failure(String)
+        case missing
+        case timedOut
+    }
+
+    private let queue = DispatchQueue(label: "com.cmux.remote-ssh.daemon-rpc.pending.\(UUID().uuidString)")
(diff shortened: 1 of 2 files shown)


raghavpillai:
🙏 would love to see this

lawrencecchen:
Addressed the latest outstanding review items on the current head in https://github.com/manaflow-ai/cmux/commit/b6f0e3a3f6485908105c3196f741535001c626b4.

The two issues identified by cubic are fixed there:
- `tests_v2/test_ssh_remote_cli_metadata.py` now records cleanup IDs even when a workspace ID is resolved via `workspace_ref`, so the final cleanup pass cannot miss those workspaces.
- `tests_v2/test_ssh_remote_docker_forwarding.py` now checks the cached daemon artifact for an embedded version marker plus SHA-256 parity with the uploaded remote binary, instead of relying on a path-only assertion.

Earlier CodeRabbit and Codex findings from older commits were already addressed in https://github.com/manaflow-ai/cmux/commit/de473455388aaa69856ad71032294e48f51bc131.

Current PR head is https://github.com/manaflow-ai/cmux/commit/b6f0e3a3f6485908105c3196f741535001c626b4 and all GitHub checks are green.

PR #482, opened 2026-02-25
Fix titlebar icons invisible when sidebar closed with dark Ghostty theme in light macOS

serafimcloud:
## Problem

When using a dark Ghostty theme (e.g. Cursor Dark `#141414`) with macOS light system appearance and the sidebar **closed**, the three titlebar icons (sidebar toggle, bell, plus) are invisible — they're dark/black (from macOS light appearance) on a dark titlebar background (from Ghostty theme).

The titlebar background correctly derives its color from the Ghostty theme, but the icons have no explicit foreground color, so they inherit from the macOS system appearance.

## Solution

When the sidebar is **closed**, derive icon color from `GhosttyApp.shared.defaultBackgroundColor.isLightColor`:
- Dark Ghostty theme → light icons
- Light Ghostty theme → dark icons

When the sidebar is **open**, leave icons unchanged — they use the default system appearance color, which is correct since the sidebar background comes from the system theme, not Ghostty.

## Test plan
- [ ] Dark Ghostty theme + light macOS + sidebar **closed** → icons should be light/visible
- [ ] Dark Ghostty theme + light macOS + sidebar **open** → icons use default system color
- [ ] Dark Ghostty theme + dark macOS + sidebar closed → icons light/visible
- [ ] Light Ghostty theme + sidebar closed → icons dark/visible

### DIFF
--- a/Sources/Update/UpdateTitlebarAccessory.swift
+++ b/Sources/Update/UpdateTitlebarAccessory.swift
@@ -479,7 +479,9 @@ struct TitlebarControlsView: View {
             .background(ShortcutHintPillBackground())
     }
 
-    private var iconForegroundColor: Color {
+    private var iconForegroundColor: Color? {
+        let sidebarVisible = AppDelegate.shared?.sidebarState?.isVisible ?? false
+        guard !sidebarVisible else { return nil }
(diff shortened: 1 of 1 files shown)


lawrencecchen:
@codex review

PR #535, opened 2026-02-26
Fix workspace rename when selection is missing with single workspace

FujiwaraChoki:
## Summary

Fixes a rename edge case where `Rename Workspace` could fail (beep/no-op) when there is exactly one workspace and the explicit selected workspace id is temporarily missing or stale.

## Root Cause
Workspace rename flows relied on `selectedWorkspace`, which returns `nil` when `selectedTabId` is `nil` or invalid.
In single-workspace state, there is still an obvious rename target, but the action had no fallback.

## Changes
- Added `TabManager.selectedWorkspaceForUserAction()`:
  - Returns `selectedWorkspace` when available.
  - Falls back to the sole workspace when `tabs.count == 1`.
  - Does not guess when multiple workspaces exist.
- Updated workspace-command paths to use this fallback:
  - Command palette workspace context snapshot
  - `palette.renameWorkspace`
  - `palette.clearWorkspaceName`
  - `palette.toggleWorkspacePin`
  - AppDelegate rename prompt path (`promptRenameSelectedWorkspace`)
- Added regression tests:
  - `cmuxTests/TabManagerWorkspaceSelectionFallbackTests.swift`
  - Covers selected, single-workspace fallback, stale selection fallback, and multi-workspace no-guess behavior.

## Behavior
- Single workspace: rename now works even if explicit selection state is missing.
- Multiple workspaces: unchanged behavior (still requires explicit selection).

## Verification
- Manual: reproduced the issue and confirmed rename now works in single-workspace mode.
- Added focused unit tests for fallback selection behavior.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -4455,8 +4455,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
 
     func promptRenameSelectedWorkspace() -> Bool {
         guard let tabManager,
-              let tabId = tabManager.selectedTabId,
-              let tab = tabManager.tabs.first(where: { $0.id == tabId }) else {
+              let tab = tabManager.selectedWorkspaceForUserAction() else {
             NSSound.beep()
             return false
         }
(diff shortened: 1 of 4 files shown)


lawrencecchen:
@codex review

lawrencecchen:
@codex review

Issue #490, opened 2026-02-25
Crash on launch: exclusive access violation in windowDragHandleShouldCaptureHit


andrewcummings:
### cmux version

0.61.0 (stable, Homebrew) and nightly (manual install) — both affected

### macOS version

Sonoma (Darwin 23.6.0)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew (stable), direct download (nightly)

### Bug description

cmux crashes ~1-2 seconds after launch with a fatal Swift exclusive access violation. The crash occurs in `windowDragHandleShouldCaptureHit(_:in:)` when a `mouseMoved` event is routed to `WindowDragHandleView.DraggableView.hitTest(_:)` while SwiftUI is already modifying the same state during the initial window layout pass.

This is a race condition between window layout and mouse event handling — it affects both stable and nightly builds, regardless of bundle identifier or saved application state.

### Expected behavior

cmux should launch without crashing, even when the mouse cursor is positioned over the window during launch.

### Steps to reproduce

1. Install cmux (stable or nightly)
2. Launch from Finder or Dock (the mouse cursor will typically be over or near the window)
3. cmux crashes within ~1-2 seconds with exit code 5

**Note:** Launching from a terminal shell via the binary path directly does *not* crash — because no mouse events are routed to the window without activation. Launching via `open -a` or `osascript activate` reproduces the crash consistently.

### Shell and environment

Tested with Ghostty 1.2.3 also installed, but the crash is independent of Ghostty — it reproduces on the nightly build which has its own bundle identifier (`com.cmuxterm.app.nightly`).

### Relevant logs or crash reports

No crash report is generated in `~/Library/Logs/DiagnosticReports/`.

```text
Simultaneous accesses to 0x6000024f45e0, but modification requires exclusive access.
Previous access (a modification) started at SwiftUI`OUTLINED_FUNCTION_10 + 2364
Current access (a modification) started at:
...
7    cmux    windowDragHandleShouldCaptureHit(_:in:) + 648
8    cmux    @objc WindowDragHandleView.DraggableView.hitTest(_:) + 44
9    AppKit  -[NSView hitTest:] + 216
...
27   AppKit  -[NSTrackingArea mouseMoved:] + 96
28   AppKit  -[_NSTrackingAreaAKManager _mouseMoved:] + 436
29   AppKit  _routeMouseMovedEvent + 224
...
[...]

PR #550, opened 2026-02-26
Fix top-bar hover crash on Sonoma and Sequoia

dragonhuntr:
## Summary
- Fix crash caused by `WindowDragHandleView` capturing `mouseEntered`, `mouseExited`, `appKitDefined`, and `systemDefined` events that should be deferred — these tracking/system events were incorrectly handled as click-like events on Sonoma and Sequoia
- Fix build error by marking `enqueueLaunchServicesRegistrationWork` as `nonisolated` since it only dispatches to a background queue

## Test plan
- [x] Verified build succeeds locally
- [x] Tested hover interactions on top bar — no crash
- [x] Verify on both Sonoma and Sequoia

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -751,7 +751,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         label: "com.cmuxterm.app.launchServicesRegistration",
         qos: .utility
     )
-    private static func enqueueLaunchServicesRegistrationWork(_ work: @escaping @Sendable () -> Void) {
+    nonisolated private static func enqueueLaunchServicesRegistrationWork(_ work: @escaping @Sendable () -> Void) {
         launchServicesRegistrationQueue.async(execute: work)
     }
     private var lastSessionAutosaveFingerprint: Int?
--- a/Sources/WindowDragHandleView.swift
+++ b/Sources/WindowDragHandleView.swift
@@ -8,7 +8,7 @@ private func windowDragHandleFormatPoint(_ point: NSPoint) -> String {
 
 private func windowDragHandleShouldDeferHitCapture(for eventType: NSEvent.EventType?) -> Bool {
     switch eventType {
-    case nil, .mouseMoved?, .cursorUpdate?:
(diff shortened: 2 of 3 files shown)


andrewcummings:
Nice — I was working on this same crash this morning. I filed #490 yesterday with a detailed breakdown after digging into the root cause. The deferral expansion here looks correct for the tracking-area events (scenario 2 in the issue), but there's a third crash scenario this doesn't cover:

**Layout-triggered hitTest with stale `currentEvent`** — When clicking the "+" button to create a new tab, SwiftUI runs a layout pass that calls `DraggableView.hitTest` directly (no `sendEvent` on the stack). `NSApp.currentEvent` is a stale `.leftMouseDown` from the click, so it passes the deferral check and falls through to `isWindowDragSuppressed(window: dragHandleView.window)`, which accesses view properties during layout → exclusive access violation.

Two additional hardening suggestions:

1. **Reorder the deferral check** to run before `isWindowDragSuppressed(window: dragHandleView.window)`. Currently the deferral runs _after_ accessing `dragHandleView.window`, which can itself trigger the violation for events that should be deferred. Moving it to the top of `windowDragHandleShouldCaptureHit` ensures no view properties are touched for passive events.

2. **Add an event-dispatch guard** — a `windowDragHandleIsInEventDispatch` flag set in `cmux_sendEvent` and checked at the top of `DraggableView.hitTest`. When `hitTest` is called outside of `sendEvent` (i.e., directly by SwiftUI during layout), return `nil` immediately. This catches the stale-`currentEvent` scenario completely:

```swift
// WindowDragHandleView.swift
var windowDragHandleIsInEventDispatch = false

// In DraggableView.hitTest:
override func hitTest(_ point: NSPoint) -> NSView? {
    guard windowDragHandleIsInEventDispatch else { return nil }
    // ... existing logic
}

// AppDelegate.swift, in cmux_sendEvent:
let wasInEventDispatch = windowDragHandleIsInEventDispatch
windowDragHandleIsInEventDispatch = true
defer { windowDragHandleIsInEventDispatch = wasInEventDispatch }
```

The drag handle still works for real mouse-button events since those always go through `sendEvent` → `hitTest`. Full analysis in #490 — happy to put up a follow-up PR with these if you'd prefer.

lawrencecchen:
@codex review

PR #561, opened 2026-02-26
Add Kagi as a search provider option

tmcarr:


lawrencecchen:
Thanks for the contribution!

PR #274, opened 2026-02-21
Honor Ghostty background opacity across cmux chrome

lawrencecchen:
## Summary
- Propagate Ghostty background opacity into Bonsplit chrome updates by emitting RGBA hex when opacity is below 1.0.
- Make workspace/pane/titlebar empty-state backgrounds follow the same Ghostty color+opacity path for live updates.
- Clamp Ghostty config opacity parsing and use clear window backgrounds for non-opaque cases to avoid flattening transparency through host layers.
- Add regression coverage for `background-opacity` parsing and Bonsplit chrome hex generation.
- Bump `vendor/bonsplit` to include RGBA chrome hex parsing support.

## Testing
- `swift test` (in `vendor/bonsplit`) (passed)
- `xcodebuild -project GhosttyTabs.xcodeproj -scheme cmux-unit -destination 'platform=macOS' -only-testing:cmuxTests/GhosttyConfigTests -only-testing:cmuxTests/WorkspaceChromeColorTests -only-testing:cmuxTests/GhosttyBackgroundThemeTests test` (passed)
- `xcodebuild -project GhosttyTabs.xcodeproj -scheme cmux -configuration Debug -destination 'platform=macOS' build` (passed)

## Related
- Task: https://github.com/manaflow-ai/cmux/issues/263
- Bonsplit dependency PR: https://github.com/manaflow-ai/bonsplit/pull/6

PR #640, opened 2026-02-27
Decouple glass effect from sidebar blend mode

ariel-wiz:
## Summary

- Remove `sidebarBlendMode == "behindWindow"` requirement from glass effect checks in `GhosttyTerminalView.swift` and `ContentView.swift`
- Change `bgGlassEnabled` default from `true` to `false` (opt-in instead of silently enabled)

The glass/blur effect was gated behind `sidebarBlendMode == "behindWindow"`, but that defaults to `"withinWindow"`, making the feature effectively unreachable. This decouples the two settings so glass works independently when explicitly enabled.

## How to use

1. `defaults write com.cmuxterm.app bgGlassEnabled -bool true`
2. Set `background-opacity = 0.5` in `~/.config/ghostty/config`
3. Restart cmux — terminal shows frosted glass effect behind text

## Summary by CodeRabbit

* **Style**
  * Background glass visual effect is now disabled by default, resulting in a different initial visual presentation of the application.
  * Refined the activation mechanism for the glass effect fallback feature, making it operate independently of other display-related settings.

### DIFF
--- a/Sources/ContentView.swift
+++ b/Sources/ContentView.swift
@@ -1831,7 +1831,7 @@ struct ContentView: View {
     // Background glass settings
     @AppStorage("bgGlassTintHex") private var bgGlassTintHex = "#000000"
     @AppStorage("bgGlassTintOpacity") private var bgGlassTintOpacity = 0.03
-    @AppStorage("bgGlassEnabled") private var bgGlassEnabled = true
+    @AppStorage("bgGlassEnabled") private var bgGlassEnabled = false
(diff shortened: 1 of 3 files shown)


lawrencecchen:
@codex review

lawrencecchen:
Closing in favor of PR https://github.com/manaflow-ai/cmux/pull/274 which handles background opacity more comprehensively. Thank you for the contribution @ariel-wiz — the glass decoupling idea is good and we'll incorporate it into #274.

PR #692, opened 2026-02-28
Add workspace close TERM/KILL fallback with reentrancy guards

### LINKED ISSUE #629: Closing a workspace not shutting down processes
### cmux version

0.61.0, build 73

### macOS version

macOS Tahoe 26.3

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Bug description

Closing workspaces with active processes leaves them lingering.

### Expected behavior

I would expect that the processes were shut down when the workspace was closed, not leaving lingering processes.

### Steps to reproduce

**Steps**

- Create a new workspace
- Start something like lazygit, claude, codex or similar
- Close the workspace
- Check activity monitor
- Process is still active 

Image

trydis:
When closing a workspace whose terminal processes don't exit within 0.6s, fall back to SIGTERM → SIGKILL escalation per panel before removing the workspace from the tab list. This prevents orphaned shell processes when closing workspaces with long-running commands.

Teardown flow (closeAllPanelsForWorkspaceClose):
- For each panel: closePanel → wait 0.6s → SIGTERM process groups → wait 1s → SIGKILL remaining → wait 1s → retry closePanel
- Returns WorkspaceCloseResult with per-panel success/failure tracking
- Gate workspace removal on full teardown (allClosed == true)

Main-thread responsiveness:
- All waits use RunLoop drain (10ms ticks) instead of blocking calls so the UI stays responsive during teardown
- ttyProcessInfos uses process.isRunning + RunLoop spin instead of the blocking waitUntilExit()

Reentrancy protection:
- Widen isWorkspaceClosingTransaction to internal
- Guard all mutating socket handlers (V1 and V2) against mid-teardown workspace mutations — surface.create/close/split/move, workspace.close, pane.create all return invalid_state if the workspace is being closed
- Extract v2ResolveActiveWorkspace helper combining workspace resolution with the closing-transaction guard

Refactoring:
- Extract teardownWorkspacePanels helper in TabManager (eliminates duplicated teardown + logging block in closeWorkspace and closeWorkspaceIfRunningProcess)
- Inline processGroupsForPIDs at its sole call site — the caller already knows the target TTY, no need to scan all workspace TTYs
- Collapse identical isDetaching / isWorkspaceClosingTransaction branches in didCloseTab
- Remove unused force parameter from closeAllPanelsForWorkspaceClose

Tests:
- WorkspaceCloseResultTests: unit tests for allClosed computed property covering success, partial failure, and edge cases
- test_close_workspace_kills_processes.py: integration test verifying processes are terminated when closing a workspace via socket

Fixes #629

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures workspaces tear down all panels before removal.
  * Improves termination of lingering terminal processes during workspace shutdown.
  * Prevents operations from proceeding during workspace-close transitions, yielding more consistent error outcomes.

* **Tests**
  * Added tests validating workspace teardown semantics and that terminal processes are killed when a workspace is closed.
  * Added unit tests covering workspace-close result behavior.

### DIFF
--- a/GhosttyTabs.xcodeproj/project.pbxproj
+++ b/GhosttyTabs.xcodeproj/project.pbxproj
@@ -84,6 +84,7 @@
 					F6000000A1B2C3D4E5F60718 /* AppDelegateShortcutRoutingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F6000001A1B2C3D4E5F60718 /* AppDelegateShortcutRoutingTests.swift */; };
 					F7000000A1B2C3D4E5F60718 /* WorkspaceContentViewVisibilityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F7000001A1B2C3D4E5F60718 /* WorkspaceContentViewVisibilityTests.swift */; };
 					F8000000A1B2C3D4E5F60718 /* SocketControlPasswordStoreTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F8000001A1B2C3D4E5F60718 /* SocketControlPasswordStoreTests.swift */; };
+					F9000000A1B2C3D4E5F60718 /* WorkspaceCloseResultTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9000001A1B2C3D4E5F60718 /* WorkspaceCloseResultTests.swift */; };
 			/* End PBXBuildFile section */
 
 /* Begin PBXCopyFilesBuildPhase section */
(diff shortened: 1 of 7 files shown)


trydis:
pr-629

teamleaderleo:
cmux-reconcile: actionable



### Actionable handoff

**Next action:** Resolve the process-liveness gate before choosing this as the retained #629 implementation.

1. Add a regression where closePanel removes the panel from the map but a child survives; demonstrate the process cleanup still runs.
2. Capture child/TTY identity before teardown discards panel state; cover descendants that leave the original TTY.
3. Retain the getpgrp() exclusion in both TERM and KILL paths and carry the useful result tests into the consolidated patch.
4. Reconcile with #1072 rather than stacking two teardown implementations; address nested RunLoop waits and reentrant close in the current architecture.

**Done / closure gate:** A process-liveness test passes independently of panel removal, all close entrypoints share the behavior, and the retained patch has ownership-safe escalation. Then close the duplicate after transferring tests/credit.

**Evidence:** The reviewed waitForPanelClose returns panels[panelId] == nil, and closeAllPanelsForWorkspaceClose skips the kill fallback when that succeeds. This is a concrete gap to test, not proof that every close currently leaks.

Patch reviewed at `e4b4024738bdec8e874de4c5af8146c87e78b209`.

Proposed follow-through from source/diff review; these checks have not been run in this triage pass. Tracking: #12969.

PR #1017, opened 2026-03-06
fix: add CJK font fallback to prevent decorative font rendering

atani:
## Summary

- On macOS, Ghostty has two font fallback paths for CJK characters. `CTFontCreateForString` handles CJK Unified Ideographs (U+4E00-9FFF) and typically selects Hiragino Sans correctly. However, `CTFontCollection` with scoring handles Hiragana, Katakana, and symbols — and may select an unintended decorative font.
- The scoring prioritizes: codepoint coverage → monospace → Regular style → glyph count. Decorative fonts with monospace attributes (e.g. AB_appare from Adobe CC, or LingWai) can be scored highest depending on installed fonts.
- Add automatic `font-codepoint-map` injection based on the system's preferred language to bypass the unreliable fallback.
- The fallback is only applied when no `font-codepoint-map` is already configured by the user.

## Details

When the primary font (e.g. Menlo) lacks CJK glyphs, libghostty uses two fallback mechanisms. `CTFontCreateForString` handles CJK Unified Ideographs (U+4E00-9FFF) and usually selects Hiragino Sans W3 correctly. For Hiragana, Katakana, and CJK symbols, `CTFontCollection` scoring is used instead — and the result depends on which fonts are installed on the system.

The scoring prioritizes monospace fonts, which causes decorative fonts with monospace attributes to be selected unexpectedly. For example, AB_appare (synced via Adobe Creative Cloud) or LingWai may be chosen over appropriate system fonts like Hiragino Sans.

This fix injects `font-codepoint-map` defaults based on `Locale.preferredLanguages`:

| Language | Font |
|---|---|
| `ja` | Hiragino Sans |
| `ko` | Apple SD Gothic Neo |
| `zh-Hant` / `zh-TW` / `zh-HK` | PingFang TC |
| `zh` | PingFang SC |

Covered Unicode ranges:
- Hiragana, Katakana, CJK Symbols and Punctuation
- CJK Unified Ideographs and Extension A
- CJK Compatibility Ideographs
- Halfwidth and Fullwidth Forms
- Hangul Syllables and Jamo

If the user has any `font-codepoint-map` in their Ghostty config, the automatic fallback is skipped.

### Before
Hiragana and Katakana render in a decorative font (varies by environment — AB_appare, LingWai, etc.).

### After
Japanese text renders in Hiragino Sans, the standard macOS system font.

## Test plan

- [x] Unit tests for `preferredCJKFontFamily` (ja/ko/zh-Hant/zh-Hans/non-CJK)
- [x] Unit tests for `userConfigContainsCJKCodepointMap` (present/absent/commented/missing)
- [x] All 9 new tests pass locally

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -1005,9 +1005,12 @@ class GhosttyApp {
     }
 
     /// When the user has not configured `font-codepoint-map` for CJK ranges,
-    /// macOS Core Text may pick an inappropriate fallback font (e.g. LingWai,
-    /// a decorative calligraphic font) for CJK characters. This injects a
-    /// sensible default based on the system's preferred languages.
+    /// Ghostty's `CTFontCollection` scoring may pick an inappropriate fallback
+    /// font for Hiragana, Katakana, and CJK symbols. The scoring prioritizes
+    /// monospace fonts, so decorative fonts with monospace attributes (e.g.
+    /// AB_appare from Adobe CC, or LingWai) can be selected depending on what
+    /// is installed. This injects a sensible default based on the system's
+    /// preferred languages.
     ///
     /// See: https://github.com/manaflow-ai/cmux/pull/1017
     private func loadCJKFontFallbackIfNeeded(_ config: ghostty_config_t) {


lawrencecchen:
Thanks again for all these contributions!

PR #1078, opened 2026-03-08
fix: route OAuth/SSO URLs to system browser instead of embedded WebView

Milofax:
## Summary

- OAuth/SSO URLs (e.g., `claude /login`, GitHub OAuth, Google sign-in) now open in the **system browser** instead of cmux's embedded WebView
- Identity providers like Google [explicitly block OAuth in embedded WebViews](https://developers.google.com/identity/protocols/oauth2/policies#browsers), causing the login flow to hang at "Einen Moment bitte..." / "One moment please..."
- Two-level interception: terminal URL resolution + WKWebView navigation delegate catches redirects

## Problem

When a terminal process (e.g., Claude Code `/login`) calls `open https://claude.ai/oauth/authorize?...`, cmux routes it to the embedded browser. The OAuth flow redirects to `accounts.google.com`, which detects the embedded WebView and blocks the consent flow — the user sees a perpetual loading spinner and can never complete login.

## Fix

**`resolveTerminalOpenURLTarget`** — Detect OAuth URLs before routing to `.embeddedBrowser`:
- `/oauth`, `/oauth2`, `/o/oauth2` path patterns
- `accounts.google.com/signin/*`
- `login.microsoftonline.com`, `login.live.com`
- `github.com/login/oauth`
- `appleid.apple.com/auth`

**`decidePolicyFor navigationAction`** — Catch OAuth redirects from within the embedded browser (e.g., a website's "Sign in with Google" button triggers a redirect to Google's consent screen).

## Test plan

- [ ] Run `claude /login` in cmux terminal — should open system browser, not embedded
- [ ] Click a "Sign in with Google" link on a website in the embedded browser — should bounce to system browser
- [ ] Non-OAuth URLs still open in the embedded browser as before
- [ ] OAuth callback URLs (containing `/oauth`) correctly route to system browser

---
## Summary by cubic
Route all OAuth/SSO URLs to the system browser instead of the embedded WebView to prevent blocked consent screens and fix hanging logins (Google, Microsoft, GitHub, Apple). Applies to terminal-initiated flows like `claude /login` and in-app redirects.

- **Bug Fixes**
  - Detect and open OAuth/SSO URLs externally during terminal URL resolution and from `WKWebView` main-frame navigations; non-OAuth links stay in the embedded browser.
  - Safer detection: path-segment match for `/oauth` and `/oauth2`, http/https scheme guard, boundary-aware Microsoft hosts, handle bare-host OAuth URLs, and log failures when opening externally. Patterns covered include `accounts.google.com/signin`, `login.microsoftonline.com`, `login.live.com`, `github.com/login/oauth`, `appleid.apple.com/auth`.

<sup>Written for commit 91b6999294d375ecc9d96142cbc553b879eb3f58. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects OAuth/SSO authentication URLs and ensures they open in the system browser instead of embedded windows.
  * Cancels embedded navigation for these flows and logs the action for diagnostics.
  * Improves compatibility with common identity providers (Google, Microsoft, GitHub, Apple) while leaving non-auth links unchanged.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -287,6 +287,12 @@ func resolveTerminalOpenURLTarget(_ rawValue: String) -> TerminalOpenURLTarget? 
     if let parsed = URL(string: trimmed),
        let scheme = parsed.scheme?.lowercased() {
         if scheme == "http" || scheme == "https" {
+            if browserIsOAuthFlowURL(parsed) {
+                #if DEBUG
+                dlog("link.resolve result=external(oauth) url=\(parsed)")
+                #endif
+                return .external(parsed)
+            }
             guard BrowserInsecureHTTPSettings.normalizeHost(parsed.host ?? "") != nil else {
                 #if DEBUG
                 dlog("link.resolve result=external(invalidHost) url=\(parsed)")
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -311,6 +311,12 @@ func resolveTerminalOpenURLTarget(_ rawValue: String) -> TerminalOpenURLTarget? 
     }
 
     if let webURL = resolveBrowserNavigableURL(trimmed) {
+        if browserIsOAuthFlowURL(webURL) {
+            #if DEBUG
+            dlog("link.resolve result=external(bareHost-oauth) url=\(webURL)")
+            #endif
+            return .external(webURL)
+        }
         guard BrowserInsecureHTTPSettings.normalizeHost(webURL.host ?? "") != nil else {
             #if DEBUG
             dlog("link.resolve result=external(bareHost-invalidHost) url=\(webURL)")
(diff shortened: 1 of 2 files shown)


ss251:
This PR still fixes a live bug — I just hit it and filed #7351 with a fresh repro and a source-level root-cause writeup against current `main` (@ 9c91710).

Concrete case: `claude /login` in a cmux terminal → cmd-click the printed `https://claude.ai/oauth/authorize?…` URL (routed into the in-app browser by #5406) → "Continue with Google" succeeds on Google's side → claude.ai bounces back to the logged-out state and no OAuth code is issued. Pasting the same URL into Safari/Chrome logs in instantly.

Reading the source, this PR's approach (route OAuth/SSO URLs to the system browser) is the right fix. The failure isn't cookie persistence or dropped popups — the in-app browser correctly uses a persistent shared `WKWebsiteDataStore.default()` and implements `createWebViewWith`. It's that the auth URL loads in an **isolated, app-private WKWebView** with no existing claude.ai/Google session, under WebKit's **default ITP / cross-site cookie partitioning**, so the Google→claude.ai session-cookie handshake gets blocked. Punting auth URLs to the system browser (where the user is already signed in) sidesteps all of it, and the `http://localhost:PORT/callback` loopback is still reachable by any browser on the machine.

Status blockers right now: this branch is **conflicting with `main`** (the touched files, `Sources/GhosttyTerminalView.swift` and `Sources/Panels/BrowserPanel.swift`, have since been refactored — the routing seam is now the `.embeddedBrowser` branch around `GhosttyTerminalView.swift:3143-3169` and the `BrowserExternalOpenSettings` in `BrowserPanel.swift:461-464`), and CI never ran because the Vercel deploy is still waiting on team authorization.

Would a maintainer be willing to authorize the deploy and/or greenlight a rebase? Happy to help rebase onto current `main` if that unblocks it. This is a fairly common paper-cut for anyone signing into Claude/GitHub/Google from a cmux terminal.

PR #1192, opened 2026-03-11
feat: auto-resume AI coding agents on session restore

tt-a1i:
Hey! 👋

I've been using cmux as my daily driver and one thing that bugged me was losing my Claude Code / Codex sessions every time I restarted. The scrollback restore is great, but I'd always have to manually re-run `claude --continue` to pick up where I left off.

So I put together a lightweight approach — instead of a full daemon architecture, this just detects what agent was running at quit time and auto-resumes it on next launch.

## What it does

When you quit cmux while an AI agent is running in a terminal pane, it remembers which one. Next time you open cmux, it automatically runs the right resume command:

- Claude Code → `claude --continue`
- Codex → `codex --resume`
- Aider → `aider` (just restarts in same dir)
- Goose → `goose session resume`

## How

Pretty simple — during session save, I grab the foreground process list via `ps -t <tty>` and match against known agent binaries. The resume command gets stashed in the existing session JSON (`agentResumeCommand` field on the terminal panel snapshot). On restore, shell integration picks it up from an env var and runs it.

Kept it minimal:
- Only detects on full saves (quit/explicit), not the 8s autosave tick
- Allowlist validation in both Swift and shell layers so a tampered JSON can't inject arbitrary commands
- Optional field, so old snapshots still load fine
- If the agent binary isn't installed, silently skips

## Files changed

- `Sources/SessionPersistence.swift` — `AgentProcessDetector` + schema + env var plumbing
- `Sources/Workspace.swift` — hook into snapshot build + restore
- `Resources/shell-integration/cmux-{zsh,bash}-integration.{zsh,bash}` — `_cmux_resume_agent_once`

## Testing

- [x] Old session JSON without `agentResumeCommand` loads fine
- [x] Claude Code running → quit → relaunch → auto-resumes
- [x] Plain shell → quit → relaunch → no resume attempt
- [x] Unknown command in env var → rejected by allowlist
- [x] Agent not installed → graceful skip

Would love feedback on whether this approach makes sense or if you'd prefer a different design (e.g. storing an agent identifier instead of the raw command string).

---

*Code reviewed with GPT-5.4.*

## Summary by CodeRabbit

* **New Features**
  * Added automatic resumption of AI agent processes when terminal sessions are restored from saved scrollback history. The system detects previously-running agents and provides seamless re-engagement on session recovery.

### DIFF
--- a/Resources/shell-integration/cmux-bash-integration.bash
+++ b/Resources/shell-integration/cmux-bash-integration.bash
@@ -35,6 +35,34 @@ _cmux_restore_scrollback_once() {
 }
 _cmux_restore_scrollback_once
 
+_cmux_resume_agent_once() {
+    local cmd="${CMUX_RESTORE_AGENT_COMMAND:-}"
+    [[ -n "$cmd" ]] || return 0
+    unset CMUX_RESTORE_AGENT_COMMAND
+
+    # Validate against an allowlist to prevent command injection from a
+    # tampered snapshot JSON file. Only known safe resume commands are accepted.
+    case "$cmd" in
+        "claude --continue"|"codex --resume"|"aider"|"goose session resume") ;;
+        *)
+            printf '\033[33mcmux:\033[0m Unknown agent resume command, skipping.\n'
+            return 0 ;;
+    esac
+
+    # Extract the binary name and verify it is installed.
+    local bin="${cmd%% *}"
+    if ! command -v "$bin" >/dev/null 2>&1; then
+        return 0
+    fi
+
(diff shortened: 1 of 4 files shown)


mrosnerr:
FYI — I've opened #2545 which takes a different approach to solving this problem:

**Key differences:**
- Uses Ghostty's `initialInput` to type the command after shell is ready (no shell integration timing issues)
- Single allowlist source in Swift (no duplicate logic in shell scripts)
- Detects foreground process at quit time via `ps` + process group matching (no caching, no stale state)
- No `eval` in shell — just types the command and presses Enter

Happy to discuss if you'd like to compare approaches!

PR #1206, opened 2026-03-12
Fix inactive split pane not rendering terminal output

### LINKED ISSUE #1205: Inactive split pane stops rendering terminal output
## Summary

When splitting a terminal pane (Cmd+D), the unfocused pane stops rendering output. Any running command (e.g. `while true; do date; sleep 1; done`) appears frozen until the pane regains focus.

## Steps to reproduce

1. Open a terminal pane and run `while true; do date; sleep 1; done`
2. Split the pane with Cmd+D
3. Click into the new (right) pane to move focus away
4. Observe the left pane — output stops updating

## Root cause

`setFocus(_ focused: Bool)` only reasserts the display ID when `focused == true`. During a split, the unfocused pane calls `ghostty_surface_set_focus(surface, false)` which can leave Ghostty's CVDisplayLink in a stuck state with no vsync callbacks. The display-ID reassertion that restarts the display link is gated behind `if focused`, so unfocused panes never recover.

## Fix

Remove the `if focused` guard so the display ID is reasserted on both focus and unfocus transitions, keeping the CVDisplayLink running for visible but unfocused split panes.

### LINKED ISSUE #2498: Unfocused split pane stops rendering terminal output
## Summary

When splitting a terminal into panes, the unfocused pane stops rendering terminal output (appears frozen).

## Steps to reproduce

1. Split a terminal pane
2. Focus one of the panes
3. Run a command in the unfocused pane — its output is not rendered

## Cause

In `setFocus(_:)`, the display ID reassignment is inside an `if focused` guard, so the CVDisplayLink for the unfocused pane gets stopped.

## User report

Reproduced on 0.62.0 (macOS, Apple Silicon) as reported in a comment on #1206.

## Related PR

Fixed in #1206 (awaiting review).

atani:
Fixes #2498

## Summary
- Reassert the display ID on both focus gain and loss transitions in `setFocus(_:)`
- Previously gated behind `if focused`, so unfocused split panes could end up with a stuck CVDisplayLink

## Root cause
When splitting a terminal pane, the unfocused side calls `ghostty_surface_set_focus(surface, false)`. The display-ID reassertion that restarts the CVDisplayLink was only performed on focus gain (`if focused`), leaving unfocused panes with no vsync callbacks and frozen output.

## Test plan
- [x] Split a terminal (Cmd+D) with a running command (`while true; do date; sleep 1; done`)
- [x] Move focus to the other pane
- [x] Confirm the unfocused pane continues rendering output
- [x] Verify no regressions when closing splits rapidly (Cmd+D → Ctrl+D)

Fixes #1205

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -4265,6 +4265,12 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
             let deltaMs = (now - lastScrollEventTime) * 1000
             Self.focusLog("resignFirstResponder: surface=\(terminalSurface?.id.uuidString ?? "nil") deltaSinceScrollMs=\(String(format: "%.2f", deltaMs))")
             ghostty_surface_set_focus(surface, false)
+
+            // Keep the display link alive for unfocused split panes so they
+            // continue rendering. Mirrors the reassertion in setFocus(_:).
+            if let displayID = window?.screen?.displayID, displayID != 0 {
+                ghostty_surface_set_display_id(surface, displayID)
+            }
         }
         return result
     }
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -3076,16 +3076,15 @@ final class TerminalSurface: Identifiable, ObservableObject {
         guard let surface = surface else { return }
         ghostty_surface_set_focus(surface, focused)
 
-        // If we focus a surface while it is being rapidly reparented (closing splits, etc),
-        // Ghostty's CVDisplayLink can end up started before the display id is valid, leaving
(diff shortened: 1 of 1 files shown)


atani:
@codex review

atani:
@lawrencecchen Just checking in on this one — it's been about a month.

I confirmed the root cause is still present on main. `setFocus(_:)` still gates `ghostty_surface_set_display_id` behind `if focused`. Unfocused split panes can end up with a stuck CVDisplayLink. Recent focus/rendering fixes (occlusion, surface recovery, drift repair) don't cover this path.

The fix is minimal — guard removal plus a mirrored reassertion in `resignFirstResponder`. The branch merges cleanly. Happy to rebase if needed.

No rush, just want to keep this from getting lost.

PR #1237, opened 2026-03-12
Fix memory leak in GhosttySurfaceScrollView deinit

atani:
## Summary

- Static debug dictionaries (`flashCounts`, `drawCounts`, `lastDrawTimes`, `presentCounts`, `dropOverlayShowCounts`, `lastPresentTimes`, `lastContentsKeys`) in `GhosttySurfaceScrollView` accumulate UUID-keyed entries that are never removed when surfaces deallocate, causing unbounded memory growth over long sessions
- Add `removeDebugCounters(for:)` and call it in `deinit` to clean up all 7 static dictionaries per surface
- Remove CA animations on `flashLayer` and `dropZoneOverlayView` in `deinit` to release layer backing stores promptly
- Detach `searchOverlayHostingView` in `deinit` so the SwiftUI hosting view hierarchy is freed synchronously

## Context

After 26+ hours of use, cmux NIGHTLY memory grows to ~1GB. `sample` output showed:
- Physical footprint: 990MB (peak 1.2GB)
- 5 renderer thread sets accumulated
- `CA::Layer::collect_animations_` recursing 10+ levels deep

The static `[UUID: ...]` dictionaries add entries every time a surface is created but never remove them on deallocation, which is the most direct contributor to memory growth in debug/nightly builds.

## Test plan

- [x] Verify existing `CmuxWebViewKeyEquivalentTests` pass (they use `resetFlashCounts()` and `flashCount(for:)` which are unchanged)
- [x] Open multiple tabs, close them, and verify `removeDebugCounters` is called in deinit logs
- [x] Monitor memory over extended sessions to confirm growth is reduced

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -5648,8 +5648,12 @@ final class GhosttySurfaceScrollView: NSView {
     private var lastDropZoneOverlayLogSignature: String?
     private var lastDragGeometryLogSignature: String?
     private var dragLayoutLogSequence: UInt64 = 0
+    /// Surface UUID captured at init time so deinit cleanup is unconditional
+    /// (the weak `terminalSurface` ref may already be nil by then).
+    private let debugStoredSurfaceId: UUID?
     private static let tabTransferPasteboardType = NSPasteboard.PasteboardType("com.splittabbar.tabtransfer")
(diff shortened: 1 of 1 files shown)


atani:
Debug-onlyの静的辞書（`#if DEBUG`ブロック内）であり、Releaseビルドには影響しません。Debugビルドでの蓄積もsurface1つあたり約400-500バイトで、実用上は無視できるレベルです。

優先度が低いためcloseします。必要になった際は再度openしてください。

PR #1355, opened 2026-03-13
fix: distinguish left/right Option key for macos-option-as-alt

### LINKED ISSUE #1349: macos-option-as-alt = left doesn't work: right Option key cannot produce special characters (@, ~, etc.)
## Bug Description

`macos-option-as-alt = left` in `~/.config/ghostty/config` does not work in cmux. Both left and right Option keys are treated as Alt/Meta, making it impossible to type special characters like `@`, `~`, `[`, `]`, `{`, `}`, `\`, `|` on non-US keyboard layouts that rely on the right Option key.

The same config works correctly in Ghostty. Only the left Option acts as Alt, and the right Option produces special characters as expected.

## Root Cause

In `Sources/GhosttyTerminalView.swift`, the `modsFromEvent` function never distinguishes left from right Option:

```swift
if event.modifierFlags.contains(.option)  { mods |= GHOSTTY_MODS_ALT.rawValue }
```

AppKit's `.option` is a device-independent flag that matches **both** physical Option keys. Because cmux only ever sets `GHOSTTY_MODS_ALT` (left), libghostty's `ghostty_surface_key_translation_mods` has no way to know the right Option key was pressed, so it cannot apply the `macos-option-as-alt = left` rule correctly.

libghostty defines `GHOSTTY_MODS_ALT_RIGHT` (bit 8) in `ghostty.h` for exactly this purpose, but cmux never sets it.

The same issue affects `consumedModsFromFlags`.

## Expected Behavior

With `macos-option-as-alt = left`:
- **Left Option** → acts as Alt/Meta (for terminal shortcuts)
- **Right Option** → acts as normal macOS Option (produces special characters like `@`, `~`)

## Suggested Fix

Check the raw device-dependent flags or `event.keyCode` to distinguish left (keyCode 58) from right (keyCode 61) Option:

```swift
if event.modifierFlags.contains(.option) {
    let isRight = event.modifierFlags.rawValue & 0x00000040 != 0
    mods |= isRight ? GHOSTTY_MODS_ALT_RIGHT.rawValue : GHOSTTY_MODS_ALT.rawValue
}
```

The same left/right distinction should also be applied for Shift, Control, and Command for full parity with Ghostty.

## Environment

- cmux (latest)
- macOS Sequoia (arm64)
- `macos-option-as-alt = left` in `~/.config/ghostty/config`

Sancerro:
## Problem

`macos-option-as-alt = left` does not work in cmux. Both left and right Option keys are treated as Alt, making it impossible to type special characters (`@`, `~`, `[`, `]`, etc.) on keyboard layouts that rely on the right Option key.

The same config works correctly in Ghostty.

Fixes #1349

## Root Cause

`modsFromEvent()` in `GhosttyTerminalView.swift` only ever sets `GHOSTTY_MODS_ALT`. It never sets `GHOSTTY_MODS_ALT_RIGHT`, so libghostty cannot distinguish left from right Option and cannot apply `macos-option-as-alt = left`.

Same issue in `consumedModsFromFlags()`.

## Fix

Uses the same **additive** pattern as upstream Ghostty (`Ghostty.Input.swift`): always set the generic modifier flag, then additionally set the sided flag when the right physical key is pressed, using named `NX_DEVICE*` constants:

```swift
if event.modifierFlags.contains(.option) { mods |= GHOSTTY_MODS_ALT.rawValue }
// ...
if raw & UInt(NX_DEVICERALTKEYMASK) != 0 { mods |= GHOSTTY_MODS_ALT_RIGHT.rawValue }
```

This avoids regressions in any code that assumes `GHOSTTY_MODS_ALT` is always set when Option is held, and no changes to the `translationMods` reconstruction are needed.

## Testing

Built and tested locally with `macos-option-as-alt = left`. Right Option now correctly produces special characters while left Option continues to act as Alt/Meta.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -4617,7 +4617,7 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
             case .control:
                 hasFlag = (translationModsGhostty.rawValue & GHOSTTY_MODS_CTRL.rawValue) != 0
             case .option:
-                hasFlag = (translationModsGhostty.rawValue & GHOSTTY_MODS_ALT.rawValue) != 0
+                hasFlag = (translationModsGhostty.rawValue & (GHOSTTY_MODS_ALT.rawValue | GHOSTTY_MODS_ALT_RIGHT.rawValue)) != 0
             case .command:
                 hasFlag = (translationModsGhostty.rawValue & GHOSTTY_MODS_SUPER.rawValue) != 0
             default:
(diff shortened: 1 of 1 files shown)


rrva:
Nice fix! One suggestion: the exclusive if/else approach (setting EITHER `ALT` or `ALT_RIGHT`) diverges from how Ghostty itself handles this. Ghostty's `ghosttyMods()` in `Ghostty.Input.swift:60-79` uses an **additive** pattern — it always sets the generic flag, then additionally sets the sided flag:

```swift
if flags.contains(.option) { mods |= GHOSTTY_MODS_ALT.rawValue }
// ...
if rawFlags & UInt(NX_DEVICERALTKEYMASK) != 0 { mods |= GHOSTTY_MODS_ALT_RIGHT.rawValue }
```

This means the `translationMods` switch changes (checking `ALT | ALT_RIGHT`) wouldn't be needed — the generic `ALT` flag is always present, so existing code paths that check it continue to work.

It also avoids risk of regressions in any other code that assumes `GHOSTTY_MODS_ALT` is set when Option is held, and you can use the named `NX_DEVICE*` constants instead of raw hex.

PR #1422, opened 2026-03-14
feat: double-click pane tab to toggle zoom

STRML:
## Summary
- Double-clicking a tab title in the **pane tab bar** (Bonsplit) now toggles pane zoom
- Uses the existing `onZoomToggle` callback already wired through `TabBarView`
- 3-line change in `vendor/bonsplit` — adds `.onTapGesture(count: 2)` before the existing single-tap handler

**Depends on:** https://github.com/manaflow-ai/bonsplit/pull/28

## Test plan
- [ ] Open a workspace with multiple splits (Cmd+D)
- [ ] Double-click a tab title in the pane tab bar → that pane should zoom to fill the content area
- [ ] Double-click the tab title again → splits should restore
- [ ] Single-click still selects tabs normally
- [ ] Middle-click still closes tabs
- [ ] Right-click context menu still works


STRML:
Closing - bonsplit main already does this.

`TapGesture(count: 2)` calling `onZoomToggle()` is at `Sources/Bonsplit/Internal/Views/TabItemView.swift:367` in 0501fed692, which is the pointer cmux main carries today. It went in as a `simultaneousGesture` rather than the `.onTapGesture(count: 2)` here, so it does not race the single-tap select, and there is a zoom indicator button next to it.

This PR is only a submodule pointer bump, and its bonsplit commit ad798c5545 is from March, so merging it now would move vendor/bonsplit backwards.

PR #1438, opened 2026-03-14
Fix Option+Delete not deleting previous word

### LINKED ISSUE #1424: Option + Delete does not delete previous word
### cmux version

cmux 0.62.1 (76) [ac98625eb]

### macOS version

macOS 26.0.1 (25A362)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

I could not test NIGHTLY

### Bug description

In cmux, Option + Arrow keys correctly move the cursor between words, like Ghostty. However, Option + Delete does not delete the previous word. The key combination appears to do nothing. This breaks the expected macOS word-editing behavior commonly supported in terminals (and Ghostty!)

### Expected behavior

Option + Delete should delete the previous word, consistent with standard macOS text editing behavior and other terminal emulators.

### Steps to reproduce

1. Open cmux
2. Start typing a command with multiple words
3. Use Option + Left Arrow to move between words (works)
4. Press Option + Delete

### Shell and environment

zsh
oh-my-zsh
no tmux running

```text

```

_No response_

### Additional context

_No response_

shouryamaanjain:
Fixes #1424

## Summary
- macOS maps Option+Delete to "∂" (U+2202), which was sent to Ghostty as literal text instead of letting the key encoder handle Alt+Backspace
- Option+Arrow works because arrow keys produce PUA characters (0xF700-0xF8FF) that `textForKeyEvent` already filters out — "∂" wasn't caught by any filter
- suppress text in `textForKeyEvent` when the unmodified key produces a control character (< 0x20 or DEL/0x7F) but Option transforms it into a different printable character

---
## Summary by cubic
Fixes Option+Delete on macOS so it deletes the previous word instead of inserting "∂". Suppresses printable text for Option+Backspace only so the key encoder handles Alt+Backspace.

- **Bug Fixes**
  - In `textForKeyEvent`, ignore text only when the unmodified key is Delete/Backspace (0x7F) and Option produces a different printable character (e.g., Option+Delete → "∂").
  - Keeps existing PUA filtering for arrow keys; Option+Arrow behavior is unchanged.

<sup>Written for commit cab1d9fcdc3cb79e0a20cf3560fa992f8aecb3ae. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard input handling for Option/Alt combinations that produce printable characters (e.g., Option+Delete → "∂"), preventing those characters from being sent as text so Alt+key sequences are correctly processed by the terminal.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -5248,6 +5248,15 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
             if scalar.value >= 0xF700 && scalar.value <= 0xF8FF {
                 return nil
             }
+
+            // When Option transforms a control/special key into a printable
+            // Unicode character (e.g. Option+Delete → "∂"), suppress the text
+            // so Ghostty's key encoder handles Alt+key instead.
+            if let unmodified = event.charactersIgnoringModifiers?.unicodeScalars.first,
+               (unmodified.value < 0x20 || unmodified.value == 0x7F),
+               scalar.value != unmodified.value {
+                return nil
+            }
         }
 
         return chars
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -5249,11 +5249,11 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
                 return nil
             }
 
-            // When Option transforms a control/special key into a printable
+            // When Option transforms Delete/Backspace into a printable
             // Unicode character (e.g. Option+Delete → "∂"), suppress the text
-            // so Ghostty's key encoder handles Alt+key instead.
+            // so Ghostty's key encoder handles Alt+Backspace instead.
             if let unmodified = event.charactersIgnoringModifiers?.unicodeScalars.first,
-               (unmodified.value < 0x20 || unmodified.value == 0x7F),
+               unmodified.value == 0x7F,
                scalar.value != unmodified.value {
                 return nil
             }


lawrencecchen:
Thanks for the fix! We investigated this and have a couple of questions before merging.

We tried reproducing on German keyboard layout but Option+Delete still correctly deletes the previous word without the fix. Could you help us narrow down what's different on your setup?

1. **What keyboard layout are you using?** (System Settings > Keyboard > Input Sources, the exact layout name)
2. **Do you have a custom `~/Library/KeyBindings/DefaultKeyBinding.dict`?**
3. **Can you reproduce on the latest nightly?** Install from https://cmux.dev/nightly and test there.

Also, we're considering narrowing the fix to backspace only (0x7F) rather than all control characters (< 0x20), to avoid potential side effects on Option+Return/Tab/Escape. The broader heuristic is more theoretically correct but harder to verify without reproducing.

pandec:
Hi! I ran into the same issue and after investigating the key event flow, I found that textForKeyEvent() isn't reached for this bug.  The "∂" text arrives via interpretKeyEvents → insertText → keyTextAccumulator before textForKeyEvent is ever called. It was working ok in pure terminal for me, but not in TUI apps. 
I submitted an alternative approach in #2246 that adds a fast path for Option+Delete before interpretKeyEvents, similar to the existing Ctrl fast path. This bypasses AppKit's text transformation entirely, so it works regardless of keyboard layout and macos-option-as-alt config. 
I cannot use macos-option-as-alt anyway, as I need option for my lovely polish lang abominations, like `żółć` (yes, that's a word... with equaly lovely meaning - `bile`).

PR #1522, opened 2026-03-16
fix: create offscreen NSWindow for background workspace PTY initialization

### LINKED ISSUE #1472: Background workspaces have dead PTYs — read-screen/send-key fail with 'Surface not ready'
## Bug Description

When workspaces are created programmatically via `cmux new-workspace` (or `workspace.create` V2 API), the terminal surfaces are listed correctly by `list-pane-surfaces` but have no functioning PTY. Commands that require a live `ghostty_surface_t` fail:

| Command | Result |
|---------|--------|
| `cmux send` | ✅ Succeeds (queued path), but text may not be delivered |
| `cmux read-screen` | ❌ `internal_error: ERROR: Terminal surface not found` |
| `cmux send-key` | ❌ `internal_error: Surface not ready` |
| `surface-health` | Shows `in_window: false` |

## Steps to Reproduce

```bash
# 1. Create workspace programmatically
WS=$(cmux new-workspace | awk '{print $2}')

# 2. List surfaces — looks normal
cmux list-pane-surfaces --workspace $WS
# → * surface:N  Terminal X  [selected]

# 3. Try to read screen — fails
cmux read-screen --workspace $WS --surface surface:N
# → Error: internal_error: ERROR: Terminal surface not found

# 4. Try to send key — fails
cmux send-key --workspace $WS --surface surface:N enter
# → Error: internal_error: Surface not ready

# 5. Check health — reveals the issue
cmux --json surface-health --workspace $WS
# → "in_window": false
```

## Root Cause

The PTY (`ghostty_surface_t`) is never created for background workspaces because:

1. **`workspace.create` is not in `focusIntentV2Methods`** (TerminalController.swift:89-104), so `socketCommandAllowsInAppFocusMutations()` returns `false`
2. This means `addTab(select: false, eagerLoadTerminal: true)` is called (TerminalController.swift:10685)
3. `eagerLoadTerminal` calls `requestBackgroundSurfaceStartIfNeeded()` (TabManager.swift:847)
4. But `requestBackgroundSurfaceStartIfNeeded()` (GhosttyTerminalView.swift:2727) requires `attachedView != nil` — which is `nil` because SwiftUI hasn't rendered the background workspace's view
5. Even if `attachedView` were set, `createSurface(for: view)` (GhosttyTerminalView.swift:2300) requires `view.window != nil` — background workspaces aren't attached to any `NSWindow`
6. **Result:** The surface is silently never created. No error is logged or returned.

The `send_text` handler (TerminalController.swift:4248-4252) has a fallback queue path for nil surfaces, but `read_text` (line 4414) and `send_key` (line 4296) do not — they require a live `ghostty_surface_t`.

## Workaround

Focus the workspace immediately after creation, then switch back:

```bash
WS=$(cmux new-workspace | awk '{print $2}')
cmux select-workspace --workspace $WS --focus
sleep 1
cmux select-workspace --workspace $CMUX_WORKSPACE_ID --focus
# Now read-screen/send-key work for $WS
```

## Suggested Fix

The eager background loading path needs to create the `ghostty_surface_t` without requiring an `NSWindow`. Options:

1. **Create a headless/offscreen ghostty surface** that doesn't need a view/window for PTY spawning — migrate it to a real view when the workspace is first displayed
2. **Briefly attach a hidden NSWindow** during background workspace creation to trigger `createSurface`
3. **Add `workspace.create` to `focusIntentV2Methods`** and briefly select + deselect the workspace (simplest, but causes visual flicker)

## Impact

This affects all programmatic/agent use of cmux. AI coding agents (Claude Code, Cursor, etc.) that create workspaces via the socket API get zombie terminals that can never execute commands. This is the #1 blocker for multi-agent terminal orchestration.

## Environment

- cmux 0.61.0 (73) [8caa5e9c9]
- macOS Darwin 25.3.0

EtanHey:
Fixes #1472

## Summary

- Fix background workspaces created via the socket API (`workspace.create` with `select: false`) having dead PTY surfaces that reject `send`, `read-screen`, and `send-key` commands.
- Add an offscreen `NSWindow` host so `ghostty_surface_new()` gets a valid Metal-backed window context even when the workspace has never been displayed.
- The offscreen window is released automatically when the view migrates to a real window or on teardown.

## Why

We built an [MCP server](https://github.com/EtanHey/cmuxlayer) on top of cmux for multi-agent AI orchestration — spawning Claude Code, Codex, and Gemini CLI sessions in separate workspaces programmatically. When creating workspaces via the socket API without selecting them, the surfaces are listed correctly but all commands fail with "Surface is not a terminal."

Root cause: `requestBackgroundSurfaceStartIfNeeded()` calls `attachToView`, which requires `surfaceView.window != nil`. Background workspaces have no `NSWindow`, so `createSurface` silently defers and the PTY is never forked.

## Testing

1. Create a background workspace: `cmux new-workspace` (without `--focus`)
2. Immediately run `cmux read-screen --surface <new-surface>`
3. **Before fix:** "Surface is not a terminal"
4. **After fix:** Returns terminal content (shell prompt)

Could not verify local build — `zig` is required for `GhosttyKit.xcframework`. CI should handle build verification.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -3412,6 +3412,10 @@ final class TerminalSurface: Identifiable, ObservableObject {
     private var pendingSocketInputBytes: Int = 0
     private let maxPendingSocketInputBytes = 1_048_576
     private var backgroundSurfaceStartQueued = false
+    /// Offscreen window used to host the view hierarchy during background surface creation.
+    /// Ghostty requires a live NSWindow for Metal layer setup and backing context.
+    /// Released when the view moves to a real window or on teardown.
+    private var offscreenHostWindow: NSWindow?
     private var surfaceCallbackContext: Unmanaged<GhosttySurfaceCallbackContext>?
     /// The desired focus state for the Ghostty C surface. May be set before the
     /// C surface exists (e.g. during layout restoration); `createSurface`
@@ -3844,6 +3848,7 @@ final class TerminalSurface: Identifiable, ObservableObject {
     func teardownSurface() {
         recordTeardownRequest(reason: "surface.teardown")
         markPortalLifecycleClosed(reason: "teardown")
+        offscreenHostWindow = nil
 
         let callbackContext = surfaceCallbackContext
         surfaceCallbackContext = nil
@@ -3960,6 +3965,10 @@ final class TerminalSurface: Identifiable, ObservableObject {
 #if DEBUG
             dlog("surface.attach.reuse surface=\(id.uuidString.prefix(5)) view=\(Unmanaged.passUnretained(view).toOpaque())")
 #endif
+            // Release the offscreen host window once the view has moved to a real window.
+            if let offscreen = offscreenHostWindow, view.window !== offscreen {
+                offscreenHostWindow = nil
+            }
             if let screen = view.window?.screen ?? NSScreen.main,
                let displayID = screen.displayID,
                displayID != 0,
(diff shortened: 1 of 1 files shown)


shuhei0866:
I have the same need — I'm building an orchestrator pattern where a conductor pane sends commands to other panes via `cmux send-text`, and hitting exactly the "Surface not ready" issue after `cmux break` / `new-workspace`.

Reviewed the diff and the approach looks solid:

- **Offscreen `NSWindow` at (-10000, -10000)**: clean way to satisfy Metal's requirement for a window-backed layer without visual side effects
- **Automatic cleanup**: releasing the offscreen window in both `attachToView` (when migrating to a real window) and `teardownSurface` — no leak risk
- **Minimal change**: only `GhosttyTerminalView.swift` touched, no API surface changes

One minor thought: the `surface != nil` check right after `addSubview` assumes `viewDidMoveToWindow` → `createSurface` fires synchronously. Worth confirming that's always the case in the Ghostty lifecycle, but from reading the existing code it looks like it is.

Would love to see this merged — it unblocks programmatic multi-workspace orchestration for both MCP servers and CLI-driven workflows. Happy to help test once a build is available.

EtanHey:
Rebased onto current main, resolved conflicts from portal sync changes. Also applied CodeRabbit's frame/autoresizing suggestion. Issue #1472 is still open — would appreciate a maintainer review when you get a chance! 🙏

EtanHey:
Rebased onto current main on April 9, 2026, and issue #1472 is still open. This still unblocks background workspace PTY initialization; would appreciate a maintainer review when convenient.

PR #1548, opened 2026-03-16
Fix terminal Cmd+= zoom on command-aware layouts

lawrencecchen:
## Summary
- preserve command-aware keyboard layout translation when deriving Ghostty shortcut codepoints
- fix terminal pane Cmd+= zoom in on layouts like Japanese where the command shortcut character differs from the unmodified key output
- add a regression test for the command-aware Cmd+= case

## Testing
- xcodebuild -project GhosttyTabs.xcodeproj -scheme cmux-unit -destination 'platform=macOS' -derivedDataPath /tmp/cmux-task-terminal-pane-cmd-equals-zoom-in-tests test -only-testing:cmuxTests/GhosttyUnshiftedCodepointTests

---
## Summary by cubic
Fixes Cmd+= zoom-in in the terminal on command-aware keyboard layouts (e.g., Japanese) by deriving shortcut codepoints using Command-aware layout translation. Adds a regression test to prevent future breakage.

- **Bug Fixes**
  - Compute unshifted codepoint with Command-aware layout first, then fall back; centralized in ghosttyUnshiftedCodepoint and used in `GhosttyNSView`.
  - Added unit test to ensure Cmd+= resolves to "=" when Command is held, even if the unmodified key outputs a different character.

<sup>Written for commit bef85712d8f4c3e60117c8e96055854b09ee26ed. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Tests**
  * Added tests to verify command key shortcut behavior
* **Refactor**
  * Reorganized internal keyboard event handling to improve code maintainability

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -10,6 +10,39 @@ import Bonsplit
 import IOSurface
 import UniformTypeIdentifiers
 
+func ghosttyUnshiftedCodepoint(
+    keyCode: UInt16,
+    modifierFlags: NSEvent.ModifierFlags,
+    charactersIgnoringModifiers: String?,
+    characters: String?,
+    layoutCharacterProvider: (UInt16, NSEvent.ModifierFlags) -> String? = KeyboardLayout.character(forKeyCode:modifierFlags:)
+) -> UInt32 {
+    let normalizedFlags = modifierFlags
+        .intersection(.deviceIndependentFlagsMask)
+        .intersection([.command])
+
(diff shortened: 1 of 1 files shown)


lawrencecchen:
Closing this because the reported bug is not reproduced on current `main`, and the original fix is not justified on current evidence.

The follow-up XCUITest work validates expected terminal zoom behavior in CI, but it does not prove a present regression that this PR fixes. This may already be fixed elsewhere or may be environment-specific.

If we get a concrete repro on `6c203b514` or a current build, we can reopen with verified failing behavior first.

PR #1576, opened 2026-03-17
Fix focus after pane-collapsing split closes

lawrencecchen:
## Summary
- add a regression test for closing the right column bottom-first in a 2x2 split
- preserve focus on the intended surviving panel when a last-tab close collapses a pane
- score fallback panes from normalized split-tree bounds so focus restore is stable before layout attaches

## Testing
- xcodebuild -project GhosttyTabs.xcodeproj -scheme cmux-unit -destination 'platform=macOS' -derivedDataPath /tmp/cmux-feat-split-close-focus-state test -only-testing:cmuxTests/WorkspacePanelGitBranchTests/testClosingRightColumnBottomFirstKeepsFocusOnBottomLeftPanel -only-testing:cmuxTests/WorkspacePanelGitBranchTests/testClosingFocusedSplitRestoresBranchForRemainingFocusedPanel

---
## Summary by cubic
Fixes focus loss after pane-collapsing closes. Focus now goes to the best neighboring panel (favoring same-row/column), stays on the current panel when closing an unfocused split, and survives close veto/retry.

- **Bug Fixes**
  - Record the next focused panel via `postCloseFocusPanelId` and apply it in `didCloseTab`; keep this target if the first close is vetoed and the close is retried.
  - Pick fallback panes from normalized split-tree bounds; prefer same-band neighbors, then score by overlap, distance, and center alignment for deterministic selection.
  - Preserve current focus when closing an unfocused split.
  - Add regression tests for side-by-side, stacked, L-shapes, and 2x2 grids, including right-column bottom-first and the close-veto retry path.

<sup>Written for commit 2cebb8705cf56734ee05bdb07822c87c430e319a. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Deterministic focus restoration after closing tabs/panels: focus reliably moves to the most sensible neighboring panel (direction-aware and respecting the selected panel), reducing unexpected focus jumps.

* **Tests**
  * Added comprehensive split/close focus tests covering side-by-side, stacked, column, and grid layouts, including retry/veto scenarios to ensure focus targets persist across retryed closes.


lawrencecchen:
Addressed cubic's retry-path focus comment in https://github.com/manaflow-ai/cmux/commit/50fe1f907df1e61a8c4f2d378d146f1f24380c55.

This keeps `postCloseFocusPanelId` intact across a vetoed `closeTab` attempt so the follow-up close still restores focus to the intended surviving panel, and it adds an executed regression test for that retry path.

PR #1639, opened 2026-03-18
Fix AZERTY Option+Delete word delete in Claude Code

### LINKED ISSUE #1635: AZERTY keyboards: Option+Delete (word delete) doesn't work in Claude Code
## Summary

On AZERTY keyboard layouts, Option+Delete (the macOS standard shortcut for deleting the previous word) does not work when running Claude Code inside cmux. This is a common keyboard shortcut that AZERTY users rely on heavily since their layout differs from QWERTY.

## Expected behavior

Pressing Option+Delete (⌥⌫) in a Claude Code session inside cmux should delete the previous word, matching standard macOS terminal behavior.

## Actual behavior

The keypress is either swallowed, produces no effect, or sends the wrong escape sequence. Users on AZERTY keyboards cannot use word-delete while editing prompts in Claude Code.

## Context

This likely affects other non-QWERTY keyboard layouts as well (QWERTZ, etc.), but the report is specifically from AZERTY users. The issue may be in how cmux translates Option+key combinations for non-US keyboard layouts — Option is used both as a modifier (for word-delete, word-jump, etc.) and as a character input key (AZERTY uses Option for common characters like `{`, `}`, `|`, `\`, etc.).

The terminal needs to distinguish between:
- **Option as Meta modifier** (Option+Delete → send `\e\x7f` for word-delete)
- **Option as character input** (Option+5 → `{` on AZERTY)

Investigate how cmux/Ghostty handles `NSEvent.modifierFlags` with `.option` for non-US input sources, and whether the `optionIsMeta` / `macos-option-as-alt` Ghostty config setting interacts with this correctly for AZERTY layouts.

## Related

This may be related to Ghostty's `macos-option-as-alt` configuration, which controls whether Option acts as Meta or as a character compose key. AZERTY users often need a hybrid behavior where Option+Delete sends Meta+Backspace but Option+letter still produces special characters.

austinywang:
## Summary
- treat DEL (`Option+Delete`) as a non-text control key in the terminal key bridge
- clear consumed text modifiers when a key event is forwarded without text so Option remains available as Meta for word-delete
- add a regression test covering synthetic `Option+Delete` through the real `keyDown` path

Closes #1635.

## Verification
- built and launched with `./scripts/reload.sh --tag fix-1635-azerty-delete`
- did not run local tests per repo policy

---
## Summary by cubic
Fixes Option+Delete on AZERTY by treating Delete (DEL) as a non-text key and not consuming Option, so Alt+Delete deletes a word again in the terminal. Adds a regression test that drives a synthetic Option+Delete through keyDown to ensure Alt is preserved and no DEL text is sent.

<sup>Written for commit e9db155943c1230e7314e84acde88070e381b6fb. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced keyboard event text extraction for improved input reliability and consistency
  * Improved control character detection to prevent accidental text transmission
  * Added defensive modifier state reset to ensure correct behavior in edge cases

* **Tests**
  * Added regression tests verifying Option+Delete correctly preserves the Alt modifier during word deletion

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -4817,7 +4817,7 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
         // Check if this event matches a Ghostty keybinding.
         let bindingFlags: ghostty_binding_flags_e? = {
             var keyEvent = ghosttyKeyEvent(for: event, surface: surface)
-            let text = event.characters ?? ""
+            let text = textForKeyEvent(event).flatMap { shouldSendText($0) ? $0 : nil } ?? ""
             var flags = ghostty_binding_flags_e(0)
             let isBinding = text.withCString { ptr in
                 keyEvent.text = ptr
@@ -5204,6 +5204,7 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
                     )
 #endif
                 } else {
+                    keyEvent.consumed_mods = GHOSTTY_MODS_NONE
                     keyEvent.text = nil
                     #if DEBUG
                     let ghosttySendStart = ProcessInfo.processInfo.systemUptime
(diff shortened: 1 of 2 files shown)


austinywang:
Superseded by #1640, which uses the branch on manaflow-ai/cmux instead of the fork branch.

PR #1731, opened 2026-03-18
fix(input): don't swallow unbound Cmd+Shift combinations

### LINKED ISSUE #1718: Unbound Cmd+Shift+<key> key combinations are silently swallowed, they never reach terminal/kitty protocol
### cmux version

0.62.2

### macOS version

macOS 26.3

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

I could not test NIGHTLY

### Bug description

Key combinations using `Cmd+Shift+<key>` that are **not bound** to any cmux or Ghostty action are silently consumed by the AppKit key-routing layer. They never reach the terminal application (e.g., Neovim) via the kitty keyboard protocol.

### Contrast with a bound key

- Press `Cmd+Shift+J` → Ghostty's default `write_screen_file:paste` fires (the file path is pasted). This confirms the key event reaches Ghostty — but only because it has an explicit binding.

### Works in Alacritty

The same Neovim mappings work correctly in Alacritty — `Cmd+Shift+K` is forwarded to Neovim and triggers the mapping:

```lua
vim.keymap.set('n', '<S-D-J>', '<C-o>', { desc = 'Back in jumplist' })
vim.keymap.set('n', '<S-D-K>', '<C-i>', { desc = 'Forward in jumplist' })
```

## Impact

Any terminal application relying on the kitty keyboard protocol to receive `super` (Cmd) modifier keys is affected. Neovim ≥ 0.10 supports `<D-...>` (Cmd) mappings via the kitty protocol, but they are unreachable in cmux.

### Expected behavior

**Expected:** A CSI escape sequence like `^[[107;13u` (kitty keyboard protocol encoding for super+shift+k)
**Actual:** Nothing is printed. The keystroke is lost.

### Steps to reproduce

1. Ensure no custom keybinding exists for `Cmd+Shift+K` (this is NOT a Ghostty or cmux default)
2. Open a terminal pane in cmux
3. Run `cat -v`
4. Press `Cmd+Shift+K`

### Shell and environment

_No response_

_No response_

_No response_

## Root cause

In `GhosttyTerminalView.swift`, `performKeyEquivalent` uses a two-pass timestamp mechanism for Cmd-modified keys (lines ~4596–4610):

```swift
// First pass: store timestamp, return false
if let lastPerformKeyEvent {
    self.lastPerformKeyEvent = nil
    if lastPerformKeyEvent == event.timestamp {
        equivalent = event.characters ?? ""
        break
    }
}
lastPerformKeyEvent = event.timestamp
return false  // ← event escapes into AppKit menu routing
```

The first pass returns `false`, expecting AppKit to redispatch the event for a second pass. However, when **no menu item** and **no Ghostty binding** matches, AppKit never redispatches — the event is dropped.

Events that DO match a Ghostty binding are caught earlier (line ~4544, `ghostty_surface_key_is_binding`) and routed through `keyDown`. The problem only affects **unbound** Cmd+Shift combinations.

BillionClaw:
Fixes #1718

Unbound Cmd+Shift+<key> combinations were being silently swallowed.
Now properly passed through to terminal applications via the kitty keyboard protocol.

## Bug Fixes

* Fixed unbound Cmd+Shift keyboard shortcuts being silently ignored in the terminal. These key combinations are now properly dispatched and will work as expected.

---
## Summary by cubic
Ensure unbound Cmd+Shift+<key> shortcuts pass through to terminal apps instead of being swallowed by AppKit. We now route them directly to keyDown in performKeyEquivalent so they are sent via the kitty keyboard protocol (fixes #1718).

<sup>Written for commit 65f4c8653bcdb91dd001b0ed3bda1a2b163c3913. Summary will update on new commits.</sup>

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -4886,8 +4886,14 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
                 }
             }
 
+            // First pass: instead of returning false and hoping for a redispatch
+            // (which AppKit may not do if no menu item matches), directly send
+            // unbound Cmd+Shift combinations through to keyDown so they reach the
+            // terminal application via the kitty keyboard protocol.
+            // This fixes issue #1718 where unbound Cmd+Shift+<key> was silently swallowed.
             lastPerformKeyEvent = event.timestamp
-            return false
+            keyDown(with: event)
+            return true
         }
 
         let finalEvent = NSEvent.keyEvent(


BillionClaw:
Closing per repository blocklist: maintainer threatened to ban. All submissions to this repo have been suspended.

PR #1788, opened 2026-03-19
Fix Claude remote-control wrapper passthrough in cmux

### LINKED ISSUE #1782: claude remote-control not working
### cmux version

 0.62.2

### macOS version

macOs Tahoe 26.3

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Bug description

Image

### Expected behavior

claude should load up as in normal terminal zsh or bash

### Steps to reproduce

claude rc 
claude remote-control

### Shell and environment

zsh

```text

```

_No response_

### Additional context

_No response_

austinywang:
## Summary
- add regression coverage for Claude wrapper passthrough when invoking `claude rc` / `claude remote-control`
- stop cmux from injecting `--session-id` / `--settings` into passthrough Claude subcommands after leading global options
- clear nested `CLAUDECODE` markers before passthrough execs so cmux-launched sessions stay isolated

## Verification
- built and launched `./scripts/reload.sh --tag issue-1782-claude-rc`
- verified the tagged app bundle includes the updated `Resources/bin/claude` wrapper
- verified the tagged bundle's wrapper reaches Claude's own remote-control path instead of failing with `Unknown argument: <uuid>`

Fixes #1782

---
## Summary by cubic
Hardens the `claude` wrapper in `cmux` to correctly passthrough `rc`/`remote-control` and other management commands without injecting session/settings flags. Improves option parsing so values aren’t mistaken for subcommands and clears `CLAUDECODE` to isolate sessions. Fixes #1782.

- **Bug Fixes**
  - Treat `rc`/`remote-control` and other management commands as passthrough, even after global options (including inline forms).
  - For passthrough, do not inject `--session-id` or `--settings`; always unset `CLAUDECODE`. Honor `--resume`/`-r` and `--from-pr` by skipping `--session-id`.
  - Add regression tests for passthrough after options and for edge cases where option values resemble subcommands.

<sup>Written for commit b7f0d16bcd7ab7a1e936f2880a15237a9b1af7fe. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * The CLI wrapper now more reliably detects management/configuration/update commands and skips injecting internal session/settings parameters (includes broader command set and resume-like flag handling).

* **Tests**
  * Added regression tests covering passthrough behavior for management commands (including cases with global options and resume-like flags) to ensure correct injection or skipping.

### DIFF
--- a/Resources/bin/claude
+++ b/Resources/bin/claude
@@ -18,6 +18,65 @@ find_real_claude() {
     return 1
 }
 
+# Return 0 for Claude global options that require a separate value.
+claude_option_requires_value() {
+    case "${1:-}" in
+        --add-dir|--agent|--agents|--allowedTools|--allowed-tools|--append-system-prompt|--betas|--debug-file|--disallowedTools|--disallowed-tools|--effort|--fallback-model|--file|--input-format|--json-schema|--max-budget-usd|--mcp-config|--model|--name|--output-format|--permission-mode|--plugin-dir|--session-id|--setting-sources|--settings|--system-prompt|--tools|-n)
+            return 0
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+}
+
+# Return 0 when the first Claude subcommand should bypass cmux hook/session
+# injection entirely. These commands manage Claude itself rather than launching
+# a normal interactive coding session, and some of them reject --session-id.
+claude_passthrough_subcommand() {
+    case "${1:-}" in
+        api-key|auth|config|doctor|help|install|mcp|plugin|plugins|rc|remote-control|setup-token|update|upgrade)
+            return 0
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+}
+
+claude_invocation_has_passthrough_subcommand() {
+    local arg
+    local pending_value=0
+
+    for arg in "$@"; do
+        if [[ "$pending_value" == "1" ]]; then
+            pending_value=0
+            continue
+        fi
+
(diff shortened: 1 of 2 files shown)


villeodell:
Hi there, a quick comment for consideration as you're working on this PR. I patched this locally with a different approach that attempts to be less tightly coupled to future parameter changes to claude code.

I replaced:

```bash
case "${1:-}" in
    mcp|config|api-key) exec "$REAL_CLAUDE" "$@" ;;
esac
```

with:

```bash
case "${1:-}" in
    ""|-*|*\ *) ;;  # no arg, flag, or multi-word string: interactive
    *) exec "$REAL_CLAUDE" "$@" ;;  # single bare word: subcommand
esac
```

This assumes interactive sessions start with either no arguments, a flag, or a multi-word quoted string (`claude`, `claude --resume`, `claude "fix the bug"`), and that subcommands are always a single bare word (`remote-control`, `mcp`, `config`). That pattern holds for claude code today. The only edge case that I could think of is a single unquoted word intended as a prompt (`claude hello`), which, at least for me, is unlikely in practice.

Note that I didn't dig too deep to validate and test, but I thought it might be helpful to consider the above or some other loosely coupled approach since the PR is still open.  Perhaps the claude code team reading https://github.com/anthropics/claude-code/issues/35894 may be influenced to formally adopt a predictable pattern to enable this.

PR #1821, opened 2026-03-19
feat: sidebar folder drop — open workspace or add terminal split

wonbywondev:
## Demo

https://github.com/user-attachments/assets/cf14e807-2a21-4b80-9e7f-6727096c458f

## Summary

- Dropping a folder from Finder onto the cmux sidebar opens a terminal in that directory
- If a workspace with the same path already exists, it is selected and a new terminal split is added inside it
- If no matching workspace exists, a new workspace is created at that path
- Browser-only workspaces are supported — the terminal appears as a split beside the browser, and the first click works immediately

## Behavior

| Drop target | Action |
|---|---|
| Sidebar + plain folder, workspace exists | Select workspace + add terminal split |
| Sidebar + plain folder, browser-only workspace | Add terminal split beside browser |
| Sidebar + plain folder, no match | Create new workspace |
| Sidebar + file | Silently ignored |
| Sidebar + package (`.app`, `.xcworkspace`, etc.) | Silently ignored |
| Terminal panel (any item) | Unchanged — path text pasted as before |
| Browser panel (any item) | Unchanged — browser handles the drop |

## Implementation

**`Sources/ContentView.swift`** (`FileDropOverlayView`)
- `performDragOperation`: when no terminal is under the cursor, reads folder URLs from the pasteboard and delegates to `onFolderDrop`
- `updateDragTarget`: returns `.copy` only when the payload contains at least one plain-directory URL (`hasDirectoryPath && !isPackage`). Packages are excluded via `resourceValues(forKeys: [.isDirectoryKey, .isPackageKey])` with fail-closed guard. Drops on terminal panels preserve the existing behavior.

**`Sources/TabManager.swift`** (`handleSidebarFolderDrop`)
- Filters to plain directories via `isDirectory + isPackage` resource keys (same exclusion logic as `updateDragTarget`)
- Normalises paths via `FinderServicePathResolver.orderedUniqueDirectories` (handles `/tmp → /private/tmp` and trailing slashes)
- Matches against `workspace.currentDirectory`; prefers `focusedPanelId`, falls back to any `TerminalPanel`, then any panel (for browser-only workspaces)
- Returns `false` when no terminal was created (drop not consumed)

**`Sources/Workspace.swift`** (`newTerminalSplit`)
- When the split source is a `BrowserPanel`, focuses the browser panel first (synchronously), then hands focus to the new terminal on the next runloop — after SwiftUI has repositioned the WKWebView portal to its new (smaller) rect. This prevents the first click on the terminal from being swallowed by the stale portal geometry.

## Focus fix detail

After splitting from a browser-only workspace, the WKWebView portal window covers the full workspace rect until SwiftUI layout shrinks it. Directly focusing the new terminal caused its first click to be intercepted by the stale portal.

Fix: `focusPanel(browser)` synchronously → `DispatchQueue.main.async { focusPanel(terminal) }`. SwiftUI updates the portal geometry between the two calls, so clicks land correctly on the terminal immediately.

Only `Workspace.swift` and `TabManager.swift` are changed for this fix — `BrowserPanel` and `BrowserWindowPortal` are untouched.

## Testing

```bash
./scripts/reload.sh --tag drag-folder-sidebar
tail -f /tmp/cmux-debug-drag-folder-sidebar.log | grep "sidebar.folderDrop"
```

1. Drag a plain folder onto the sidebar → new workspace opens at that path
2. Drag the same folder again → existing workspace selected, new split added (`sidebar.folderDrop.existing` in log)
3. Drag onto a browser-only workspace → terminal split opens beside the browser; first click works immediately
4. Drag a file → nothing happens, no copy cursor shown
5. Drag a `.app` bundle → nothing happens, no copy cursor shown
6. Drag a folder onto a terminal panel → path text pasted as before

## Related

Builds on #1571 (dock icon drag support).

## Checklist

- [x] Tested locally (all 6 scenarios above)
- [x] No new UI strings (no localization changes needed)
- [x] All code review bot comments resolved
- [x] Existing terminal and browser drop behavior unchanged

## Summary by CodeRabbit

Updated file-drop routing to forward dropped folder URLs to `TabManager.handleSidebarFolderDrop(_)`, changed copy-advertising to require a non-package directory payload, and adjusted terminal-split focus sequencing to defer focusing the new terminal when split originates from a `BrowserPanel`.

| File | Summary |
|---|---|
| `Sources/ContentView.swift` | When no terminal is under the drop point, read dropped URLs and call `onDrop?(urls)`; advertise `.copy` only if payload contains at least one non-package directory |
| `Sources/TabManager.swift` | Added `handleSidebarFolderDrop` — selects first plain directory, matches existing workspace by path, attempts a horizontal terminal split or creates a new workspace |
| `Sources/Workspace.swift` | In `newTerminalSplit`, special-case `BrowserPanel` sources: focus browser first, then focus new terminal on the next main-queue turn |

## Summary by cubic

Drag a plain folder from Finder onto the sidebar to open a terminal in that directory. Existing workspaces are reused with a split when possible; files and filesystem packages like `.app` are ignored.

- **New Features**
  - Matching workspace at path → select and add a terminal split; browser-only → split beside the browser; otherwise create a new workspace.
  - Drops on terminal or browser panels are unchanged; the sidebar handles folder drops only. The copy cursor is shown only for plain directories (non-packages).

- **Bug Fixes**
  - Allow sidebar folder drops by approving the prepare step when no terminal is under the cursor.
  - Restore first-click focus after splitting from a browser-only workspace by focusing the browser first, then the new terminal on the next runloop.
  - Return false when no terminal is created so the drop isn't consumed.
  - Switch workspace selection only after a successful split to avoid changing tabs on rejected drops.

<sup>Written for commit 53cc9fccd379a818d02d16d92963af02a7f36c51.</sup>

## Summary by CodeRabbit

* **New Features**
  * Folder drag-and-drop to the sidebar to open or create workspaces.

* **Bug Fixes**
  * Fallback handling for folder drops when no pane/terminal is under the cursor.
  * More reliable terminal split creation and focus when splitting from browser panels.
  * Improved drag-cursor feedback and detection for droppable folders.

### DIFF
--- a/Sources/ContentView.swift
+++ b/Sources/ContentView.swift
@@ -877,7 +877,11 @@ private func findFileDropOverlayView(in root: NSView?) -> FileDropOverlayView? {
 private func configureFileDropOverlay(_ overlay: FileDropOverlayView, tabManager: TabManager) {
     overlay.onDrop = { [weak tabManager] urls in
         MainActor.assumeIsolated {
-            guard let tabManager, let terminal = tabManager.selectedWorkspace?.focusedTerminalPanel else { return false }
+            guard let tabManager else { return false }
+            // Plain-directory drops onto the sidebar open or extend a workspace.
+            if tabManager.handleSidebarFolderDrop(urls) { return true }
+            // Otherwise fall back to pasting the path into the focused terminal.
+            guard let terminal = tabManager.selectedWorkspace?.focusedTerminalPanel else { return false }
             return terminal.hostedView.handleDroppedURLs(urls)
         }
     }
--- a/Sources/FileDropOverlayView.swift
+++ b/Sources/FileDropOverlayView.swift
@@ -305,7 +305,9 @@ final class FileDropOverlayView: NSView {
             preparedDragWebView = accepted ? webView : nil
             return accepted
         }
-        return hasPaneTarget
+        // Allow the folder-drop fallback path (e.g. sidebar) when no pane target is under
+        // the cursor — performDragOperation will route the URLs through `onDrop`.
+        return hasPaneTarget || onDrop != nil
     }
 
     override func performDragOperation(_ sender: any NSDraggingInfo) -> Bool {
@@ -394,7 +396,15 @@ final class FileDropOverlayView: NSView {
             return handled
         }
         activeDragWebView = nil
-        guard let terminal else { return false }
+        guard let terminal else {
+            // No terminal/browser/pane target under the drop point — delegate to the folder-drop
+            // handler. In practice the only non-pane area visible to this overlay is the sidebar.
+            let urls = sender.draggingPasteboard.readObjects(
+                forClasses: [NSURL.self],
+                options: [.urlReadingFileURLsOnly: true]
+            ) as? [URL] ?? []
+            return onDrop?(urls) ?? false
+        }
         return terminal.performDragOperation(sender)
     }
 
--- a/Sources/FileDropOverlayView.swift
+++ b/Sources/FileDropOverlayView.swift
@@ -62,6 +62,39 @@ final class FileDropOverlayView: NSView {
     var lastHitTestLogSignature: String?
     var lastDragRouteLogSignatureByPhase: [String: String] = [:]
 
+    /// Cached "pasteboard contains at least one plain directory" result for the current
+    /// drag session, keyed by pasteboard `changeCount`. Avoids re-reading the pasteboard
+    /// and running filesystem stat on every `draggingUpdated` event.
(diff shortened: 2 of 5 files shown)


wonbywondev:
Re: `newTerminalSplit(focus: true)` nitpick — keeping the call site implicit on purpose. The default is `true` and the surrounding `newTerminalSplit` call sites in this file (`createSplit`, command-palette new-terminal, drag-attach paths, etc.) all rely on the same default. Making this one call explicit would create a style mismatch without changing behavior. Same reasoning as in the previous review round.

wonbywondev:
@coderabbitai review

Issue #1418, opened 2026-03-14
new-surface: add --no-focus flag to prevent tab focus change


jasonkuhrt:
## Problem

`cmux new-surface` always focuses the newly created tab. When programmatically creating surfaces (e.g., dispatching work to parallel Claude Code sessions), this steals keyboard focus from the user's current tab.

This causes real UX issues when an agent creates a new surface in the background — the user's keystrokes suddenly go to the wrong tab, corrupting both their current work and the new surface's initialization.

## Proposed solution

Add a `--no-focus` or `--focus <true|false>` flag to `new-surface`:

```
cmux new-surface --focus false
```

The surface is created and added to the tab bar, but the user's current surface retains keyboard focus.

## Prior art

`move-surface` already has `--focus <true|false>`. This would be the same semantics applied at creation time.

## Workaround

Currently using `new-workspace --command` (workspaces don't auto-focus) or osascript keystroke simulation to immediately refocus the previous tab. Both are suboptimal.

joaqo:
Yep, have the same issue here.

Issue #140, opened 2026-02-20
CLI commands should not change focus state


lawrencecchen:
## Description

Running `cmux` CLI commands (e.g. `new-workspace`, `new-split`, `open-browser`, `select-workspace`, etc.) should never change which surface, pane, or workspace is currently focused.

Today, commands like `cmux new-workspace` or `cmux open-browser` can steal focus from the surface you're actively working in. This is disruptive — especially when CLI commands are being issued programmatically by agents or scripts running in a terminal pane.

The CLI should be a side-effect-free control plane: it creates/modifies resources but leaves the user's focus exactly where it was before the command ran.

PR #1942, opened 2026-03-22
cmux ssh: add --no-focus flag to suppress auto-focus on connect

maucher:
## Summary

Add `--no-focus` flag to `cmux ssh` so scripted/background SSH workspace creation does not steal the user's active workspace focus.

Without this flag, `workspace.select` is called immediately after `workspace.remote.configure` returns, switching the user away from their current workspace before SSH is even established. With `--no-focus`, the workspace is created and configured but the caller's workspace retains focus. The caller can then redirect to the new workspace later via `cmux select-workspace`.

## Usage

```
cmux ssh dev@my-host --no-focus
cmux ssh dev@my-host --name "gpu-box" --no-focus
```

## Consistency

Follows the same pattern as `break-pane --no-focus` and `join-pane --no-focus`.

## Changes

- `SSHCommandOptions`: add `noFocus: Bool` field
- `parseSSHCommandOptions`: parse `--no-focus` flag (also made `internal` for testability)
- `runSSH`: guard `workspace.select` call with `if !sshOptions.noFocus`
- Help text: document `--no-focus` flag
- Tests: fix existing `SSHCommandOptions` initializer call; add `testParseSSHCommandOptionsNoFocusFlag`

## Testing

cmux_ssh_no_focus

Tested locally by compiling the CLI directly and running against a live cmux socket:

```
/tmp/cmux-test-cli ssh devbox --name "no focus" --no-focus
# → OK workspace=workspace:N target=devbox state=connecting
# active workspace did not change
```

Also verified:
- `--no-focus` alongside other flags (`--name`, `--port`) parses correctly
- unknown flags still error as expected
- unit test `testParseSSHCommandOptionsNoFocusFlag` added to `CLIProcessRunnerTests`

## Issues

Addresses the `cmux ssh` case of #140 ("CLI commands should not change focus state") and is complementary to #1418 (`new-surface: add --no-focus flag`).

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -3510,7 +3510,7 @@ struct CMUXCLI {
         }
     }
 
-    private func parseSSHCommandOptions(_ commandArgs: [String], localSocketPath: String = "", remoteRelayPort: Int = 0) throws -> SSHCommandOptions {
+    func parseSSHCommandOptions(_ commandArgs: [String], localSocketPath: String = "", remoteRelayPort: Int = 0) throws -> SSHCommandOptions {
         var destination: String?
         var port: Int?
         var identityFile: String?
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -3294,6 +3294,7 @@ struct CMUXCLI {
         let port: Int?
         let identityFile: String?
         let workspaceName: String?
+        let noFocus: Bool
         let sshOptions: [String]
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Superseded by https://github.com/manaflow-ai/cmux/pull/2227 (cherry-picked onto main with conflict resolution). Thank you for the contribution!

PR #2033, opened 2026-03-24
Consume Cmd+number shortcuts when workspace index is out of bounds

### LINKED ISSUE #1970: Cmd+number workspace switch opens new window instead of switching tabs
## Bug description

Pressing `⌘+number` (Cmd+1~8) to switch workspaces opens a **new cmux window** instead of switching between tabs in the sidebar of the current window.

This appears to share the same root cause as #1929 (weak `NSWindow` reference becoming `nil`, causing the fallback path to create a new window), but the trigger is `⌘+number` workspace switching rather than `⌘+N`.

## Expected behavior

Pressing `⌘+number` should switch to the corresponding workspace tab within the **same window's sidebar**, not open a new cmux window.

## Steps to reproduce

1. Open cmux
2. Create multiple workspaces (tabs in sidebar)
3. Press `⌘+1`, `⌘+2`, etc. to switch between workspaces
4. Instead of switching tabs within the same window, a new cmux window opens

## Shell and environment

zsh

## Relevant logs or crash reports

_No crash — the app does not crash, it just opens a new window._

## Screenshots or screen recordings

_(will attach separately)_

## Additional context

- Likely same root cause as #1929: `MainWindowContext.window` weak reference becomes `nil`, causing `resolvedWindow(for:)` to fail and the handler to fall back to `openNewMainWindow()`.
- Also related to #1936 (request to customize numbered tab switching shortcuts).

anthhub:
## Summary
- When pressing Cmd+N (N=1-9) for a workspace index that doesn't exist, the shortcut event was not consumed and fell through to Ghostty's `goto_tab` binding, which could create a new window
- Fix: always consume the event when the digit shortcut matches, regardless of whether a valid workspace exists at that index

Fixes #1970

## Test plan
- [ ] Open cmux with 2 workspaces, press Cmd+5 — should do nothing (no new window)
- [ ] Press Cmd+1 / Cmd+2 — should switch workspaces normally
- [ ] Press Cmd+9 — should switch to last workspace

---
## Summary by cubic
Consume Cmd+number workspace shortcuts even when the index is out of bounds to prevent fallthrough to Ghostty’s goto_tab from creating a new window. Now Cmd+N with no matching workspace does nothing; valid numbers still switch (9 = last). Fixes #1970.

<sup>Written for commit 676374014c2506322ac090b4fca232c04381ead7. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of numbered workspace selection keyboard shortcuts (1–9). Shortcuts are now consistently recognized and consumed, preventing potential interference with other application functions.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -9481,18 +9481,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         }
 
         // Numeric shortcuts for specific workspaces (9 = last workspace)
-        if let manager = tabManager,
-           let digit = numberedShortcutDigit(
-               event: event,
-               shortcut: KeyboardShortcutSettings.shortcut(for: .selectWorkspaceByNumber)
-           ),
-           let targetIndex = WorkspaceShortcutMapper.workspaceIndex(forDigit: digit, workspaceCount: manager.tabs.count) {
+        // Always consume the event when the digit matches to prevent Ghostty's
+        // goto_tab fallback from creating a new window when the index is out of bounds.
+        if let digit = numberedShortcutDigit(
+            event: event,
+            shortcut: KeyboardShortcutSettings.shortcut(for: .selectWorkspaceByNumber)
+        ) {
+            if let manager = tabManager,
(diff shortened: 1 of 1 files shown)


lawrencecchen:

PR #2059, opened 2026-03-24
fix(browser): use native value setter for React compatibility

### LINKED ISSUE #2058: browser fill/type/select don't work on React forms
fill, type, and select set the DOM value but don't update React's internal state, so React-controlled forms stay broken (submit buttons disabled, validation fails, etc). I ran into this on a Next.js site.

The issue is that the injected JS does `el.value = x`, but React overrides the value setter on input instances. Direct assignment bypasses it, so React never sees the change.

The standard fix (used by Playwright and Puppeteer) is to call the native setter from the prototype:

```js
const nativeSetter = Object.getOwnPropertyDescriptor(
    HTMLInputElement.prototype, 'value'
).set;
nativeSetter.call(el, newValue);
el.dispatchEvent(new Event('input', { bubbles: true }));
```

I have a branch with the fix if a PR would be useful: joshuaswanson:fix/react-controlled-input-compat

joshuaswanson:
Fixes #2058

fill, type, and select use `el.value = x` which doesn't trigger state updates in React (or Vue/Angular). This uses the native setter from the prototype instead, same approach as Playwright and Puppeteer.

Falls back to direct assignment if the native setter isn't available.

---
## Summary by cubic
Use the native value setter so `browser.fill`, `browser.type`, and `browser.select` trigger React/Vue/Angular state updates. Supports cross‑realm elements and web components.

- **Bug Fixes**
  - Walk the prototype chain to call the native `value` setter; fallback to direct assignment. Works across realms and custom elements.
  - Dispatch `input` and `change` events after setting the value for proper framework updates.

<sup>Written for commit 0dbc2881b99675579be26f46d2ee907e84306b01. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input handling to use native browser mechanisms when updating values for inputs, textareas, and selects, ensuring value changes are applied more reliably.
  * Ensured `input` and `change` events are consistently dispatched after value updates; adjusted fill/type/select paths to compute and apply the final value before event dispatch.

### DIFF
--- a/Sources/TerminalController.swift
+++ b/Sources/TerminalController.swift
@@ -8076,18 +8076,20 @@ class TerminalController {
     }
 
     /// JavaScript snippet that sets an input element's value using the native
-    /// HTMLInputElement/HTMLTextAreaElement/HTMLSelectElement prototype setter.
-    /// Frameworks like React, Vue, and Angular override the value property on
-    /// instances, so a plain `el.value = x` assignment only updates the DOM
-    /// without notifying the framework's internal state. Calling the native
-    /// setter from the prototype bypasses the override and triggers the
-    /// framework's change-detection when followed by an `input` event.
+    /// prototype setter. Frameworks like React, Vue, and Angular override the
+    /// value property on instances, so a plain `el.value = x` assignment only
+    /// updates the DOM without notifying the framework's internal state.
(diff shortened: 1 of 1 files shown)


lawrencecchen:

PR #2075, opened 2026-03-25
fix: prevent Japanese IME confirmation Enter from executing command

rerun0510:
## Summary

- **Fix Japanese IME confirmation Enter from executing the command prematurely.**
- Root cause: commit b64fb301 (https://github.com/manaflow-ai/cmux/pull/1671) added logic to forward Return to Ghostty after IME composition commit, intended for Korean IME where Enter both confirms the syllable and executes the command in a single step.
- Japanese (and Chinese) IME use Enter only to confirm the conversion; a second Enter is needed to execute. The original code did not distinguish between input sources, causing Japanese IME confirmation to also trigger command execution.

### What changed

| File | Change |
|---|---|
| `Sources/GhosttyTerminalView.swift` | `shouldSendCommittedIMEConfirmKey` now checks `KeyboardLayout.id` and only sends the extra Return for Korean input sources |
| `Sources/KeyboardLayout.swift` | Added `debugInputSourceIdOverride` (DEBUG only) to allow tests to simulate specific input sources |
| `cmuxTests/CJKIMEInputTests.swift` | Korean IME test now sets the input source override so it passes regardless of the host's active input method |

### IME behavior difference

| IME | Enter behavior |
|---|---|
| Korean (한글) | Enter = confirm syllable + execute command (single step) |
| Japanese (日本語) | Enter = confirm conversion only; second Enter = execute command |
| Chinese (中文) | Enter = confirm conversion only; second Enter = execute command |

## Testing

- Verified Japanese IME: type text → convert → press Enter → conversion confirmed, command NOT executed. Press Enter again → command executes.
- Verified Korean IME unit test (`testReturnAfterKoreanCommitAlsoSendsReturnToSurface`) passes.
- Built and launched via `reload.sh --tag fix-jp-ime`.

## Demo Video

v0.62.2 (before the bug occurred)

https://github.com/user-attachments/assets/2dd4eee6-084c-4ede-b4e3-3ba8b7cf5a26

Before the fix (latest main branch)

https://github.com/user-attachments/assets/f4e2ee27-5640-4e4f-ad52-8ccf8131224f

After the fix

https://github.com/user-attachments/assets/49219a3c-924b-4e7b-8cd8-1e51c5843a25

---
## Summary by cubic
Fixes Enter with Japanese/Chinese IME so it only confirms conversion and does not execute the command; Korean IME behavior (confirm + execute in one Enter) remains unchanged.

- **Bug Fixes**
  - Restrict extra Return forwarding in `shouldSendCommittedIMEConfirmKey` to Korean input sources only, using a case-insensitive ID check.
  - Add `debugInputSourceIdOverride` (DEBUG-only) to `KeyboardLayout` and update the Korean IME unit test to use it.

<sup>Written for commit d74afefa627307018fd8b69bd19ad234204553a7. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Return/Enter key behavior for Korean input method editors so committed text and Enter press are delivered correctly; prevents incorrect extra Return for non-Korean IMEs (e.g., Japanese/Chinese).

* **Tests**
  * Updated tests to cover Korean IME Return behavior and prevent regressions.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -5765,7 +5765,7 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
         // Japanese/Chinese IME: Enter only confirms the conversion; a second Enter executes.
         // Only send the extra Return key for Korean input sources.
         guard let sourceId = KeyboardLayout.id else { return false }
-        return sourceId.contains("Korean") || sourceId.contains("korean")
+        return sourceId.range(of: "korean", options: .caseInsensitive) != nil
     }
 
     private func ghosttyKeyEvent(for event: NSEvent, surface: ghostty_surface_t) -> ghostty_input_key_s {


rerun0510:
@greptile-apps review

lawrencecchen:

PR #2108, opened 2026-03-25
fix(browser): keep IME Enter on composition path

### LINKED ISSUE #1814: Built-in browser: Bopomofo IME Enter key submits webpage instead of committing composition
### Problem or motivation

**Bug description**

When using macOS Bopomofo (注音) input method in the built-in browser, pressing Enter to confirm/commit an IME composition inside a webpage `<textarea>` (e.g., claude.ai's chat input) immediately submits the form, instead of only committing the composed character.

This does not happen in Chrome or Safari — in those browsers, Enter during active IME composition correctly commits the character without triggering any submit action.

**Steps to reproduce**
1. Open the built-in browser and navigate to claude.ai (or any page with a `<textarea>`)
2. Switch to macOS Bopomofo input method (注音)
3. Type any Bopomofo sequence (e.g., `ㄓㄨ`)
4. Press Enter to confirm the character

**Expected:** Character is committed (e.g., `注`), no form submission occurs
**Actual:** The message is submitted immediately

**Root cause hypothesis**

cmux may be forwarding a "raw" Enter `keydown` event to WKWebView pages before — or instead of — the IME `compositionend` event. Webapps like claude.ai check `event.isComposing` on `keydown` to guard against accidental submissions; if cmux doesn't correctly mark the event as `isComposing: true`, the webapp sees a plain Enter and submits.

This is distinct from the address bar fix in #789/#867 (which handled the omnibar `NSTextField`), and likely requires an equivalent `hasMarkedText()` guard in `CmuxWebView`'s key event dispatch path toward the WKWebView content.

**Environment**
- cmux version: [your version]
- macOS version: [your version]
- Input method: macOS Bopomofo (注音)
- Affected site: claude.ai (and likely any webapp using `event.isComposing` to gate Enter-to-submit)

### Proposed solution

cmux may be forwarding a "raw" Enter `keydown` event to WKWebView pages before — or instead of — the IME `compositionend` event. Webapps like claude.ai check `event.isComposing` on `keydown` to guard against accidental submissions; if cmux doesn't correctly mark the event as `isComposing: true`, the webapp sees a plain Enter and submits.

This is distinct from the address bar fix in #789/#867 (which handled the omnibar `NSTextField`), and likely requires an equivalent `hasMarkedText()` guard in `CmuxWebView`'s key event dispatch path toward the WKWebView content.

### Alternatives considered

_No response_

### Additional context

_No response_

austinywang:
## Summary

- keep browser Return and keypad Enter on AppKit's normal composition path while marked text is active, so IME confirmation does not submit web forms
- add regression coverage for browser Return and keypad Enter during marked-text composition
- Closes #1814

## Testing

- `./scripts/reload.sh --tag issue-1814-browser-ime-enter`
- did not run tests locally per repo policy

## Demo Video

- Video URL or attachment: not recorded

```text
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review
```

## Checklist

- [ ] I tested the change locally
- [ ] I updated docs/changelog if needed
- [ ] I requested bot reviews after my latest commit (copy/paste block above or equivalent)

---
## Summary by cubic
Keep Return and keypad Enter on the IME composition path while marked text is active, so confirming candidates doesn’t submit web forms in the embedded browser. Closes #1814.

- **Bug Fixes**
  - Skip forwarding Return/Enter to the browser when the first responder has marked text.
  - Detect marked text via `NSTextInputClient`/`NSTextView` and pass `firstResponderHasMarkedText` into Return/Enter routing.
  - Add regression tests for IME Return and keypad Enter, plus unit tests for the routing helper.

<sup>Written for commit 7d0dc4ed5f9a8431a77da9fa5f3821cd0334899c. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the Return/Enter key was being incorrectly intercepted by the browser during IME (Input Method Editor) text composition, causing disruptions to international keyboard input and text input methods.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -1504,12 +1504,31 @@ func browserOmnibarShouldSubmitOnReturn(flags: NSEvent.ModifierFlags) -> Bool {
     return normalizedFlags == [] || normalizedFlags == [.shift]
 }
 
+func browserResponderHasMarkedText(_ responder: NSResponder?) -> Bool {
+    guard let responder else { return false }
+
+    // During IME composition, Return/Enter belongs to the text system so the
+    // candidate list can commit or confirm the marked text.
+    if let textInputClient = responder as? NSTextInputClient {
+        return textInputClient.hasMarkedText()
+    }
+
+    if let textField = responder as? NSTextField,
+       let editor = textField.currentEditor() as? NSTextView {
+        return editor.hasMarkedText()
+    }
+
+    return false
+}
+
 func shouldDispatchBrowserReturnViaFirstResponderKeyDown(
     keyCode: UInt16,
     firstResponderIsBrowser: Bool,
+    firstResponderHasMarkedText: Bool = false,
     flags: NSEvent.ModifierFlags
 ) -> Bool {
     guard firstResponderIsBrowser else { return false }
+    guard !firstResponderHasMarkedText else { return false }
     guard keyCode == 36 || keyCode == 76 else { return false }
     // Keep browser Return forwarding narrow: only plain/Shift Return should be
     // treated as submit-intent. Command-modified Return is reserved for app shortcuts
@@ -12391,6 +12410,7 @@ private extension NSWindow {
         let firstResponderWebView = self.firstResponder.flatMap {
             Self.cmuxOwningWebView(for: $0, in: self, event: event)
         }
+        let firstResponderHasMarkedText = browserResponderHasMarkedText(self.firstResponder)
         if let ghosttyView = firstResponderGhosttyView {
             // If the IME is composing and the key has no Cmd modifier, don't intercept —
             // let it flow through normal AppKit event dispatch so the input method can
@@ -12434,6 +12454,7 @@ private extension NSWindow {
         if shouldDispatchBrowserReturnViaFirstResponderKeyDown(
             keyCode: event.keyCode,
             firstResponderIsBrowser: firstResponderWebView != nil,
+            firstResponderHasMarkedText: firstResponderHasMarkedText,
             flags: event.modifierFlags
         ) {
             // Forwarding keyDown can re-enter performKeyEquivalent in WebKit/AppKit internals.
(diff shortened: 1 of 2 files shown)


notoriouslab:
@austinywang 

首先感謝你們針對中文輸入法的改善努力，不過上次的修復還是有一點問題：

## Still reproducible after #2108

**Environment**
- macOS
- Input method: 注音 (Zhuyin / Bopomofo)
- Browser: CMUX built-in browser
- Tested on: claude.ai

### What still happens

After #2108 was merged, the issue persists. Pressing Enter after 注音 candidate selection still submits the web form instead of committing the composition.

### Why the fix likely still has a race condition

#2108 guards against dispatch by calling `hasMarkedText()` on the AppKit first responder inside `performKeyEquivalent`. However, by the time `performKeyEquivalent` is invoked for the Return key, WKWebView appears to have already committed and cleared the marked text — so `hasMarkedText()` returns `false`, the guard is bypassed, and Enter is forwarded to the browser as a form submission.

This is the same timing issue as the web-layer `isComposing` guard: the event ordering in WKWebView places Enter *after* marked text is cleared, making synchronous checks unreliable.

### Confirmed workaround (JS layer, tested)

Injecting the following via DevTools Console fully resolves the issue:

```javascript
(function() {
  let composing = false;
  document.addEventListener('compositionstart', () => composing = true, true);
  document.addEventListener('compositionend', () => {
    setTimeout(() => composing = false, 30);
  }, true);
  document.addEventListener('keydown', (e) => {
    if (e.key === 'Enter' && composing) {
      e.preventDefault();
      e.stopImmediatePropagation();
    }
  }, true);
})();
```

The key insight is the **30ms delay** on `compositionend` — this compensates for WKWebView clearing marked text before the keydown event is processed.

### Suggested fix direction

The same delay-based approach could be applied at the native layer: instead of checking `hasMarkedText()` synchronously at `performKeyEquivalent` time, maintain a short-lived `isComposing` flag that is set on `insertText:replacementRange:` / composition start and cleared with a brief delay after `setMarkedText:` receives an empty string. This would mirror the JS workaround's behavior without relying on a synchronous `hasMarkedText()` call that may already be stale.

PR #2146, opened 2026-03-25
fix(cli): suppress premature stop-hook notifications during auto-continue

### LINKED ISSUE #2077: Notifications fire prematurely on every Claude Code turn, not just task completion
## Bug

cmux sends a desktop notification on **every Claude Code turn completion** (each time Claude finishes a response and waits for input), not just when the overall task is complete. This causes spurious/premature notifications — the user sees "Completed" alerts while the agent is still mid-task.

### Root cause (investigation)

The `cmux claude-hook stop` handler ([`CLI/cmux.swift:10221-10278`](https://github.com/manaflow-ai/cmux/blob/main/CLI/cmux.swift#L10221-L10278)) fires a `notify_target` notification immediately when invoked. Claude Code's `stop` hook is a **per-turn** lifecycle event (Claude finished responding, waiting for next input), not a task-completion signal.

This means:
1. User sends a multi-step task to Claude
2. Claude completes first response → `stop` hook fires → notification sent → **premature**
3. Claude auto-continues or user sends follow-up → another `stop` → another notification
4. Repeat until task is actually done

There is no distinction between "Claude paused mid-task" and "Claude finished the entire task."

### Possible approaches

- **Debounce/delay notifications:** Wait N seconds after `stop` fires before sending the notification. If Claude resumes within the window (a new `session-start` or `notification` hook), cancel the pending notification.
- **Detect auto-continue:** If Claude Code is auto-continuing (tool use → response → tool use), suppress notifications for intermediate turns. Only notify on the final turn where Claude is actually waiting for human input.
- **Use the `notification` hook instead of `stop`:** Claude Code's `notification` hook is specifically for "needs user attention" events, which may be a better signal than `stop`.
- **Heuristic: check if the response ended with a question or tool result** vs. a final summary.

### Related issues

- https://github.com/manaflow-ai/cmux/issues/923 — Claude Code weird popup notifications (on workspace switch)
- https://github.com/manaflow-ai/cmux/issues/1529 — macOS notifications fire on Stop hook even when workspace is focused
- https://github.com/manaflow-ai/cmux/issues/1004 — Desktop notifications not suppressed when window is focused

### Key files

| Component | File | Lines |
|-----------|------|-------|
| Stop hook handler | `CLI/cmux.swift` | 10221-10278 |
| Notification storage/delivery | `TerminalNotificationStore.swift` | 893-942 |
| Session tracking | `CLI/cmux.swift` | 338-456 |

## Environment

- cmux nightly
- Claude Code with hooks enabled

anthhub:
## Problem

Closes #2077

The `claude-hook stop` handler fires on **every Claude Code turn**, not just when the task is fully complete. This caused spurious desktop notifications mid-task: each time Claude finished a response and paused (even briefly between tool calls), the user received a "Completed" notification.

## Root cause

`CLI/cmux.swift` stop handler calls `notify_target` unconditionally whenever `stop` fires. Claude Code's `stop` hook is a per-turn lifecycle event; there is no built-in way to distinguish "Claude paused between tool calls (auto-continue)" from "Claude finished the task and is waiting for the human."

## Fix: timestamp-based deduplication

Track the last `pre-tool-use` event time in the session record. In the stop handler, skip the notification if a tool was used within the past **2 seconds**:

- **Auto-continue turns:** `stop` fires immediately after Claude finishes a response, then `pre-tool-use` fires within ~500 ms as Claude resumes. When the *next* `stop` arrives, `lastToolUseAt` is fresh — notification is suppressed.
- **Genuine completions:** Claude produces a final response with no follow-up tool call. The 2-second window expires before any `pre-tool-use` updates the timestamp, so the notification fires normally.
- **First turn / no tool history:** `lastToolUseAt` is `nil` → notification fires as before (no regression).

### Changes

| File | Change |
|------|--------|
| `CLI/cmux.swift` | Add `lastToolUseAt: TimeInterval?` field to `ClaudeHookSessionRecord` |
| `CLI/cmux.swift` | Add `lastToolUseAt` parameter to `ClaudeHookSessionStore.upsert()` |
| `CLI/cmux.swift` | In `pre-tool-use` handler: record `Date().timeIntervalSince1970` on every tool call |
| `CLI/cmux.swift` | In `stop` handler: guard `notify_target` behind 2-second staleness check |

The `notification` hook path, OSC 777, and Codex notification paths are **not affected**.

## Test plan

- [ ] Start Claude Code with a multi-step task (file edits + shell commands). Verify **no** notification during intermediate turns.
- [ ] After the task finishes and Claude stops calling tools, verify the notification **does** fire (> 2 s elapsed).
- [ ] Single-turn question-answer (no tool calls, `lastToolUseAt` is nil). Verify notification fires normally.
- [ ] Trigger the `notification` hook (e.g. Claude uses `AskUserQuestion`). Verify notification is unaffected by this change.
- [ ] Rapid stop/restart within 2 s (e.g. user sends immediate follow-up). No spurious notification expected.

---
## Summary by cubic
Suppress premature "Completed" notifications during auto-continue by gating the `stop` hook with a per-turn `toolUsedThisTurn` flag and consuming/resetting it each turn so alerts only fire on true completion. Fixes #2077.

- **Bug Fixes**
  - Replace session-level timestamp with per-turn `toolUsedThisTurn`; set on `pre-tool-use`, reset on prompt submit and immediately after `stop`, and consume before the completion check to avoid stuck suppression when completion is nil.
  - In `stop`, send `notify_target` only when the flag is false; localize the "Claude Code" title in both the stop and `notification` handlers.

<sup>Written for commit ccf67f3e493ee15b3a7fd25a00c228e07a5fc167. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Improvements**
  * Per-turn session tracking of tool usage added to better control notification behavior.
  * Completion notifications are suppressed when a tool was used during the current turn, reducing premature or noisy alerts.
  * Tool-usage state is reset at the start of each new turn so normal notifications resume.
  * Notification title now uses a localized lookup with a sensible default for improved internationalization.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -328,9 +328,36 @@ private struct ClaudeHookSessionRecord: Codable {
     var lastBody: String?
     var startedAt: TimeInterval
     var updatedAt: TimeInterval
-    /// Timestamp of the most recent pre-tool-use event. Used to suppress
-    /// premature stop-hook notifications when Claude auto-continues mid-task.
-    var lastToolUseAt: TimeInterval?
+    /// Whether a tool was used during the current turn. Reset to false on prompt-submit
+    /// so that only tool uses within the same turn suppress stop-hook notifications.
+    /// Using a per-turn Bool rather than a session-level timestamp prevents stale
+    /// timestamps from a previous turn from incorrectly suppressing later notifications.
+    var toolUsedThisTurn: Bool
+
+    init(
+        sessionId: String, workspaceId: String, surfaceId: String,
+        cwd: String?, pid: Int?, lastSubtitle: String?, lastBody: String?,
+        startedAt: TimeInterval, updatedAt: TimeInterval, toolUsedThisTurn: Bool = false
+    ) {
+        self.sessionId = sessionId; self.workspaceId = workspaceId; self.surfaceId = surfaceId
+        self.cwd = cwd; self.pid = pid; self.lastSubtitle = lastSubtitle; self.lastBody = lastBody
+        self.startedAt = startedAt; self.updatedAt = updatedAt; self.toolUsedThisTurn = toolUsedThisTurn
+    }
+
+    init(from decoder: Decoder) throws {
+        let c = try decoder.container(keyedBy: CodingKeys.self)
+        sessionId = try c.decode(String.self, forKey: .sessionId)
+        workspaceId = try c.decode(String.self, forKey: .workspaceId)
+        surfaceId = try c.decode(String.self, forKey: .surfaceId)
+        cwd = try c.decodeIfPresent(String.self, forKey: .cwd)
+        pid = try c.decodeIfPresent(Int.self, forKey: .pid)
+        lastSubtitle = try c.decodeIfPresent(String.self, forKey: .lastSubtitle)
+        lastBody = try c.decodeIfPresent(String.self, forKey: .lastBody)
+        startedAt = try c.decode(TimeInterval.self, forKey: .startedAt)
+        updatedAt = try c.decode(TimeInterval.self, forKey: .updatedAt)
+        // Default to false for records persisted before this field was introduced.
+        toolUsedThisTurn = try c.decodeIfPresent(Bool.self, forKey: .toolUsedThisTurn) ?? false
+    }
 }
 
 private struct ClaudeHookSessionStoreFile: Codable {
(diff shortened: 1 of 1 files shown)


anthhub:
Quick triage request on this PR.

I still think the underlying stop-hook notification issue is worth fixing, but this branch likely wants one more polish pass before it is truly merge-ready. If maintainers are still interested in the approach here, I can come back, address the remaining review feedback, and refresh the branch.

If the team would rather not take this approach, I am also fine closing it instead of letting it linger.

Issue #2078, opened 2026-03-25
Memory leaks: NotificationCenter observers, per-surface queues, and profile stores never cleaned up


austinywang:
## Description

Code audit identified several memory leaks beyond the GPU/IOSurface accumulation tracked in https://github.com/manaflow-ai/cmux/issues/1435. These are application-level leaks that compound over long sessions, especially with many workspaces being created and destroyed.

Triggered by macOS reporting **99.24 GB** memory usage from cmux.

## Leak 1: GhosttyApp NotificationCenter observers never removed (CRITICAL)

**File:** `Sources/GhosttyTerminalView.swift` (~line 941, 1327-1336)

```swift
private var appObservers: [NSObjectProtocol] = []
// Two observers appended but never removed:
appObservers.append(NotificationCenter.default.addObserver(...))
appObservers.append(NotificationCenter.default.addObserver(...))
```

`GhosttyApp` has **no `deinit`** to clean up these observers. Every terminal surface creation adds observers that persist for the entire app lifetime, holding strong references and preventing deallocation.

**Fix:** Add `deinit` to `GhosttyApp` that iterates `appObservers` and calls `NotificationCenter.default.removeObserver`.

## Leak 2: AppDelegate observer imbalance — 39 adds / 14 removes

**File:** `Sources/AppDelegate.swift`

25+ orphaned `addObserver` calls with no matching `removeObserver`. Key examples:
- `DistributedNotificationCenter.addObserver(self, ...)` with strong `self`, never removed (~line 2245)
- `menuBarVisibilityObserver` created but never cleaned up (~line 6013)
- Multiple `NSWorkspace.shared` observers (~lines 3273, 3286, 3302)

`AppDelegate` has **no `deinit`**. While AppDelegate typically lives for the app lifetime, the retained observer closures and objects can hold strong references to other objects that should have been freed.

**Fix:** Audit all `addObserver` calls and add matching cleanup, or use block-based observers with `[weak self]`.

## Leak 3: TerminalController per-surface message queues never cleared (CRITICAL)

**File:** `Sources/TerminalController.swift` (~lines 196-197)

```swift
private var v2BrowserDialogQueueBySurface: [UUID: [V2BrowserPendingDialog]] = [:]
private var v2BrowserDownloadEventsBySurface: [UUID: [[String: Any]]] = [:]
```

[...]

Issue #2045, opened 2026-03-24
send/read-screen/send-key --surface surface:N errors 'not a terminal' for valid terminals


EmundoT:
## Summary

`cmux send --surface surface:N`, `cmux read-screen --surface surface:N`, and `cmux send-key --surface surface:N` all fail with `Error: invalid_params: Surface is not a terminal` when targeting **any** terminal surface by ref — even surfaces confirmed as `type=terminal` by `cmux tree` and `cmux surface-health`.

Using UUIDs directly works. Using `--workspace` works. Only surface ref (`surface:N`) targeting is broken.

## Environment

- cmux version: 0.62.2 (77) [6c203b514]
- macOS 15.5
- Apple Silicon (M4 Pro)
- Installation: Homebrew

## Reproduction Steps

```bash
# 1. Confirm terminal surfaces exist
cmux tree --all
# Shows: surface:1 [terminal], surface:29 [terminal], surface:55 [terminal], etc.

cmux surface-health
# surface:1  type=terminal in_window=true
# surface:29  type=terminal in_window=true

# 2. Try to send to ANY terminal surface by ref — all fail
cmux send --surface "surface:1" "echo hello"
# Error: invalid_params: Surface is not a terminal

cmux send --surface "surface:29" "echo hello"
# Error: invalid_params: Surface is not a terminal

cmux read-screen --surface "surface:29"
# Error: invalid_params: Surface is not a terminal

cmux send-key --surface "surface:29" Enter
# Error: invalid_params: Surface is not a terminal

cmux capture-pane --surface "surface:29"
# Error: invalid_params: Surface is not a terminal

# 3. Workaround: UUID works
cmux read-screen --surface "$CMUX_SURFACE_ID"
# => Returns screen content correctly

# 4. Workaround: --workspace works
cmux send --workspace "workspace:1" "echo hello"
# => OK surface:1 workspace:1
```

## Root Cause Analysis

`identify --surface "surface:29"` reveals the problem — the `--surface` flag is not resolved:

```json
{
  "caller": {
    "surface_ref": null,     // ← should be surface:29
    "surface_type": null,    // ← should be "terminal"
    "pane_ref": null
  },
  "focused": {
    "surface_ref": "surface:1",
    "surface_type": "terminal"
  }
}
```

[...]

PR #2150, opened 2026-03-25
fix(socket): return not_found when surface_id is provided but unresolvable

anthhub:
## Summary

- When `surface_id` is explicitly provided in a socket command but fails to resolve (stale ref, unknown ordinal, etc.), the four affected functions now return a `not_found` error instead of silently falling back to `ws.focusedPanelId`.
- When `surface_id` is **not** provided, the existing focused-pane fallback is preserved unchanged (backward compatible).
- Affected functions: `v2SurfaceSendText`, `v2SurfaceSendKey`, `v2SurfaceClearHistory`, `v2SurfaceReadText` (all in `Sources/TerminalController.swift`).

## Root Cause

`v2UUID(params, "surface_id")` calls `v2ResolveHandleRef()` internally. If the ref is unresolvable, it returns `nil`. The previous pattern `v2UUID(params, "surface_id") ?? ws.focusedPanelId` silently promoted this `nil` to the focused panel, causing two distinct failure modes:

- **#2042** — Commands succeed with exit code 0 but operate on the wrong surface (the focused pane), making automation that targets specific surfaces dangerous.
- **#2045** — When the fallback panel happens to be a browser surface, the subsequent `ws.terminalPanel(for:)` guard fails and returns `invalid_params: Surface is not a terminal`, making valid terminal surfaces appear broken when addressed by ref.

## Fix

```swift
// Before
let surfaceId = v2UUID(params, "surface_id") ?? ws.focusedPanelId

// After
let surfaceId: UUID?
if params["surface_id"] != nil {
    surfaceId = v2UUID(params, "surface_id")
    guard surfaceId != nil else {
        result = .err(code: "not_found", message: "Surface not found for the given surface_id", data: nil)
        return
    }
} else {
    surfaceId = ws.focusedPanelId
}
```

## Test Plan

- [ ] `cmux send --surface surface:9999 "hello"` → `ERROR: not_found: Surface not found for the given surface_id` (non-zero exit)
- [ ] `cmux send-key --surface surface:9999 Enter` → same error
- [ ] `cmux read-screen --surface surface:9999` → same error
- [ ] `cmux send "hello"` (no `--surface`) → operates on focused pane (backward compat preserved)
- [ ] `cmux send --surface surface:1 "hello"` with a valid ref → succeeds as before
- [ ] `cmux read-screen --surface surface:29` with a valid terminal ref → returns screen content

Fixes #2042, Fixes #2045

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #2151, opened 2026-03-25
fix: plug NotificationCenter observer accumulation and per-surface browser state leaks

### LINKED ISSUE #2042: send/send-key/read-screen silently fall back to focused pane when given a non-existent surface handle
### cmux version

0.62.2

### macOS version

macOS 15.5

### Mac chip

Apple Silicon (M4 Pro)

### Installation method

Homebrew

Not tested

### Bug description

When `cmux send`, `cmux send-key`, or `cmux read-screen` is called with a `--surface` handle that does not exist (e.g. `surface:9999`), the command succeeds silently (exit code 0) and operates on the **currently focused pane** instead of returning an error.

This is dangerous for automation: a script that sends a command to a stale/dead surface unknowingly sends it to whatever pane happens to be focused.

### Steps to Reproduce

```bash
# Confirm surface:9999 does not exist
cmux tree  # no surface:9999 in output

# Send text to non-existent surface
cmux send --surface surface:9999 "hello"
# => OK surface:20 workspace:8   (focused pane, not an error)
# => "hello" appears in the focused pane's input

# Same behavior with send-key
cmux send-key --surface surface:9999 "return"
# => OK surface:20 workspace:8

# Same behavior with read-screen
cmux read-screen --surface surface:9999
# => Returns the focused pane's screen content, exit code 0
```

### Expected behavior

An error should be returned (non-zero exit code + error message like `ERROR: Surface not found`) when the specified surface handle does not exist.

### Root cause (from source analysis)

The issue is a validation gap between the CLI and server layers:

1. **CLI** (`cmux.swift` ~L2910-2911): `normalizeSurfaceHandle()` recognizes `surface:9999` as a valid handle-ref format (`isHandleRef()` returns true) and passes it through to the server **without verifying the surface exists**.
   - Note: numeric-only inputs (e.g. `5`) *are* validated against `surface.list` and correctly error if not found (~L2913-2926).

2. **Server** (`TerminalController.swift` ~L5281): `v2SurfaceSendText()` resolves the surface ID with:
   ```swift
   let surfaceId = v2UUID(params, "surface_id") ?? ws.focusedPanelId
   ```
   When `v2UUID()` → `v2ResolveHandleRef()` returns `nil` for a non-existent handle, it silently falls back to `ws.focusedPanelId`.

3. The same `?? ws.focusedPanelId` fallback pattern exists in `v2SurfaceSendKey()` (~L5332), `v2SurfaceClearHistory()` (~L5366), and `v2SurfaceReadText()` (~L5417).

### Suggested fix

Either:
- **CLI side**: Validate handle-ref surfaces against `surface.list` before sending to the server (same as numeric inputs), or
- **Server side**: Return an error when `v2ResolveHandleRef()` returns `nil` for an explicitly provided `surface_id` parameter, instead of falling back to `ws.focusedPanelId`.

### Impact

This affects any automation that manages multiple panes by surface ID. If a pane is closed or its surface ID becomes stale, commands silently target the wrong pane. In our case (multi-agent orchestration with Claude Code), a Claude launch command was accidentally sent to the orchestrator's own pane instead of a dead agent pane.

anthhub:
## Summary

Partial fix for #2078 — addresses **Leak 1** and **Leak 3** only. Leak 2 (AppDelegate observer imbalance) and Leak 4 (BrowserProfileStore eviction) are left for a follow-up PR.

### Leak 1 — GhosttyApp NotificationCenter observers never removed

`GhosttyApp` registered two `NSApplication` activation observers via `NotificationCenter.default.addObserver(forName:object:queue:)` during `initializeGhostty()`, storing them in `appObservers`. The class had no `deinit`, so these block-based observers were never unregistered and the token objects were never released.

**Fix:** Add `deinit` to `GhosttyApp` that iterates `appObservers`, calls `NotificationCenter.default.removeObserver` on each token, then clears the array. While `GhosttyApp.shared` is a process-lifetime singleton, the deinit ensures observers are always balanced if the object is ever deallocated and makes the lifecycle intent explicit.

### Leak 3 — TerminalController per-surface browser dictionaries never cleared

Six `[UUID: …]` dictionaries in `TerminalController` accumulate state keyed by browser-surface ID:
- `v2BrowserFrameSelectorBySurface`
- `v2BrowserInitScriptsBySurface`
- `v2BrowserInitStylesBySurface`
- `v2BrowserDialogQueueBySurface`
- `v2BrowserDownloadEventsBySurface`
- `v2BrowserUnsupportedNetworkRequestsBySurface`

Entries were written when surfaces were created but never removed when surfaces were destroyed, causing unbounded growth over sessions with many workspace/surface creation cycles.

**Fix:** Add `cleanupBrowserSurface(surfaceId:)` to `TerminalController` that calls `removeValue(forKey:)` on all six dictionaries. Invoke it from `Workspace.splitTabBar(_:didCloseTab:fromPane:)` — the canonical surface-destruction callback — alongside the existing per-surface cleanup block (e.g. `panelDirectories`, `surfaceTTYNames`, etc.).

## Test plan

- [ ] Build and launch cmux; verify no regression in terminal or browser surface behavior
- [ ] Open and close many browser surfaces (repeated workspace create/destroy cycles); confirm memory footprint does not grow unboundedly
- [ ] Reload configuration repeatedly; confirm no observer accumulation (instrument with Allocations profiler, filter on `_NSObserverTarget`)
- [ ] Verify browser dialog, download-wait, and unsupported-request commands still function correctly after a surface-close/reopen cycle

---
## Summary by cubic
Fixes two memory leaks by unregistering `NotificationCenter` observers and clearing per-surface browser state on tab close. Also makes `surface_id` handling strict by returning `not_found` when provided but unresolvable to prevent misrouted commands.

- **Bug Fixes**
  - GhosttyApp: remove `NotificationCenter` observers in `deinit` to stop observer/token accumulation. (Leak 1)
  - TerminalController: add per-surface cleanup for six browser dictionaries and call it when a tab closes to prevent unbounded growth. (Leak 3)
  - v2SurfaceSendText / v2SurfaceSendKey / v2SurfaceClearHistory / v2SurfaceReadText: if `surface_id` is present but cannot be resolved, return `not_found` instead of falling back to the focused pane; no `surface_id` still falls back. Fixes #2042, #2045.
  - Partial fix for #2078 (addresses Leak 1 and Leak 3).

- **Migration**
  - Scripts/clients that pass `surface_id` must ensure it’s valid; stale or unknown IDs now return `not_found`.
  - To target the focused pane, omit `surface_id` instead of relying on implicit fallback.

<sup>Written for commit 53832d595577d341f7d3ec683e4570308becf49c. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed memory leaks from stale observers when closing the application and browser tabs.
  * Enhanced error handling for invalid browser panel references with clearer error messages.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -1032,6 +1032,13 @@ class GhosttyApp {
         initializeGhostty()
     }
 
+    deinit {
+        for observer in appObservers {
+            NotificationCenter.default.removeObserver(observer)
+        }
+        appObservers.removeAll()
+    }
+
     #if DEBUG
     private static let initLogPath = "/tmp/cmux-ghostty-init.log"
 
--- a/Sources/TerminalController.swift
+++ b/Sources/TerminalController.swift
@@ -216,6 +216,18 @@ class TerminalController {
         }
     }
 
+    /// Remove all per-surface browser state accumulated for the given surface ID.
+    /// Must be called on the main thread when a surface is destroyed.
+    @MainActor
+    func cleanupBrowserSurface(surfaceId: UUID) {
+        v2BrowserFrameSelectorBySurface.removeValue(forKey: surfaceId)
+        v2BrowserInitScriptsBySurface.removeValue(forKey: surfaceId)
+        v2BrowserInitStylesBySurface.removeValue(forKey: surfaceId)
+        v2BrowserDialogQueueBySurface.removeValue(forKey: surfaceId)
+        v2BrowserDownloadEventsBySurface.removeValue(forKey: surfaceId)
+        v2BrowserUnsupportedNetworkRequestsBySurface.removeValue(forKey: surfaceId)
+    }
+
     private nonisolated func withListenerState<T>(_ body: () -> T) -> T {
         listenerStateLock.lock()
         defer { listenerStateLock.unlock() }
@@ -5278,7 +5290,16 @@ class TerminalController {
                 result = .err(code: "not_found", message: "Workspace not found", data: nil)
                 return
             }
-            let surfaceId = v2UUID(params, "surface_id") ?? ws.focusedPanelId
+            let surfaceId: UUID?
+            if params["surface_id"] != nil {
+                surfaceId = v2UUID(params, "surface_id")
+                guard surfaceId != nil else {
+                    result = .err(code: "not_found", message: "Surface not found for the given surface_id", data: nil)
+                    return
+                }
+            } else {
+                surfaceId = ws.focusedPanelId
+            }
             guard let surfaceId else {
                 result = .err(code: "not_found", message: "No focused surface", data: nil)
                 return
@@ -5329,7 +5350,16 @@ class TerminalController {
                 result = .err(code: "not_found", message: "Workspace not found", data: nil)
                 return
             }
-            let surfaceId = v2UUID(params, "surface_id") ?? ws.focusedPanelId
+            let surfaceId: UUID?
+            if params["surface_id"] != nil {
+                surfaceId = v2UUID(params, "surface_id")
+                guard surfaceId != nil else {
+                    result = .err(code: "not_found", message: "Surface not found for the given surface_id", data: nil)
+                    return
+                }
+            } else {
+                surfaceId = ws.focusedPanelId
+            }
             guard let surfaceId else {
                 result = .err(code: "not_found", message: "No focused surface", data: nil)
                 return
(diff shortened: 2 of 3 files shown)


anthhub:
Closing this PR — the Leak 1 fix (deinit on a static singleton) is ineffective since GhosttyApp is `static let shared` and will never be deinitialized. The correct fix should clear observers at the start of `reloadConfiguration()` instead. The PR also inadvertently includes unrelated surface-ref changes from #2150. Will re-submit with a correct approach.

PR #2186, opened 2026-03-26
fix(quit): enforce Warn Before Quit when Cmd+Q arrives via app switcher

### LINKED ISSUE #2139: "Warn Before Quit" dialog not shown when quitting via Cmd+Q in Cmd+Tab app switcher
### cmux version

0.62.2

### macOS version

macOS 26.3.1

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Bug description

When "Warn Before Quit" is enabled in settings, pressing Cmd+Q while the macOS Cmd+Tab app switcher is open (i.e. cmux is highlighted but not focused in the foreground) quits cmux immediately without showing the confirmation dialog. The warning only appears when Cmd+Q is pressed while cmux is the active foreground app.

Other apps such as Google Chrome correctly intercept the Cmd+Q quit shortcut and display their warning dialog even when triggered from the Cmd+Tab switcher.

### Expected behavior

The "Warn Before Quit" confirmation dialog should appear regardless of whether cmux is the active foreground app or is being quit from the macOS Cmd+Tab app switcher, consistent with how apps like Chrome handle this.

### Steps to reproduce

1. Enable "Warn Before Quit" in cmux Settings
2. Have cmux running with active terminal sessions
3. Switch to another app so cmux is not in the foreground
4. Press Cmd+Tab to open the macOS app switcher, and navigate to cmux
5. While still holding Cmd, press Q to quit cmux
6. cmux quits immediately with no confirmation dialog

### Shell and environment

_No response_

```text

```

_No response_

### Additional context

_No response_

anthhub:
## Summary

- `applicationShouldTerminate` was unconditionally returning `.terminateNow`, bypassing the `QuitWarningSettings` check entirely
- When Cmd+Q is pressed in the Cmd+Tab app switcher, macOS calls `applicationShouldTerminate` directly without going through `handleCustomShortcut`, so the existing dialog in `handleQuitShortcutWarning` was never shown
- Fix: check `QuitWarningSettings.isEnabled()` in `applicationShouldTerminate`; if enabled, show the same confirmation alert (with suppression button) and reply asynchronously via `NSApp.reply(toApplicationShouldTerminate:)`
- Add `isQuitWarningConfirmed` flag to prevent a double dialog when the normal Cmd+Q path (`handleQuitShortcutWarning`) has already obtained user confirmation before calling `NSApp.terminate`

Fixes #2139

## Test plan

- [ ] Enable Settings → Warn Before Quit
- [ ] Press Cmd+Tab to open the app switcher, then press Cmd+Q while cmux is highlighted — confirm that the "Quit cmux?" alert appears and cancelling keeps the app running
- [ ] With warning enabled, press Cmd+Q normally (not via app switcher) — confirm only one alert appears (no double dialog)
- [ ] Tick "Don't warn again for Cmd+Q" in the app-switcher alert — confirm subsequent quits (both paths) skip the dialog
- [ ] Disable Warn Before Quit in Settings — confirm Cmd+Q via app switcher quits immediately without any alert

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed quit confirmation dialog appearing multiple times during app termination.
  * Improved stability when dismissing the quit warning prompt.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -2212,6 +2212,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
     private var lastTypingActivityAt: TimeInterval = 0
     private var didHandleExplicitOpenIntentAtStartup = false
     private var isTerminatingApp = false
+    // Set to true when the user has already confirmed quit via the warning dialog,
+    // so applicationShouldTerminate does not show a second alert.
+    private var isQuitWarningConfirmed = false
     private var didInstallLifecycleSnapshotObservers = false
     private var didDisableSuddenTermination = false
     private var commandPaletteVisibilityByWindowId: [UUID: Bool] = [:]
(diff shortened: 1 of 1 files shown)


lawrencecchen:

PR #2192, opened 2026-03-26
fix: clean up per-surface state on tab close to prevent unbounded memory growth

anthhub:
## Problem

`TerminalController` accumulates per-surface state that is never cleaned up when a surface is destroyed:

1. **6 browser panel dictionaries** (`v2BrowserFrameSelectorBySurface`, `v2BrowserInitScriptsBySurface`, `v2BrowserInitStylesBySurface`, `v2BrowserDialogQueueBySurface`, `v2BrowserDownloadEventsBySurface`, `v2BrowserUnsupportedNetworkRequestsBySurface`) — written during browser/socket interactions, never removed
2. **`v2BrowserElementRefs`** — element references keyed by surface ID, never pruned after surface close (stale refs can never be resolved anyway per the guard at L6899)
3. **`v2RefByUUID[.surface]` / `v2UUIDByRef[.surface]`** — handle mappings written on creation, only added-to by `v2RefreshKnownRefs()` but never removed

Over long sessions with many workspaces created and closed, these dictionaries grow without bound.

## Fix

**`Sources/TerminalController.swift`**: Add `cleanupSurfaceState(surfaceId:)` that clears all 8 categories of per-surface data in one call.

**`Sources/Workspace.swift`**: Call `cleanupSurfaceState` from `splitTabBar(_:didCloseTab:fromPane:)` alongside the existing panel dictionary cleanup.

## Note on Leak 1 from #2078

The reported observer accumulation in `GhosttyApp` does **not** exist — `appObservers.append` runs only inside `initializeGhostty()`, which is called once from `private init()`. Since `GhosttyApp` is `static let shared`, the observer count is always exactly 2.

## Test plan

- [ ] Run cmux for an extended session, repeatedly creating and closing workspaces with browser panels
- [ ] Verify via Instruments (Allocations) that the cleaned dictionaries do not grow after tab close
- [ ] Confirm normal browser panel functionality (navigation, script injection, dialogs, downloads) is unaffected

Partially addresses #2078

## Summary by CodeRabbit

* **Bug Fixes**
  * Tab and panel closing now consistently cleans up per-surface terminal and browser state, preventing leftover data from affecting later sessions.
  * Closing via detach/move preserves surface state so tabs moved between panes keep their existing session data.
  * Cleanup now runs at the correct teardown points for improved reliability and performance.

### DIFF
--- a/Sources/TerminalController.swift
+++ b/Sources/TerminalController.swift
@@ -201,16 +201,26 @@ class TerminalController {
     private var v2BrowserUnsupportedNetworkRequestsBySurface: [UUID: [[String: Any]]] = [:]
     private let v2BrowserUndefinedSentinel = V2BrowserUndefinedSentinel()
 
-    /// Remove all per-surface browser state for a destroyed surface.
+    /// Remove all per-surface state for a destroyed surface.
     /// Called from Workspace.splitTabBar(_:didCloseTab:fromPane:) to prevent unbounded dictionary growth.
+    ///
+    /// Cleans up:
+    /// - 6 browser panel dictionaries (frame selector, init scripts/styles, dialog queue, download events, unsupported network requests)
+    /// - v2BrowserElementRefs entries whose surfaceId matches (stale refs can never be resolved after surface close)
+    /// - v2RefByUUID[.surface] and v2UUIDByRef[.surface] entries for this surface (written on creation, never removed otherwise)
     @MainActor
-    func cleanupBrowserSurfaceState(surfaceId: UUID) {
+    func cleanupSurfaceState(surfaceId: UUID) {
         v2BrowserFrameSelectorBySurface.removeValue(forKey: surfaceId)
         v2BrowserInitScriptsBySurface.removeValue(forKey: surfaceId)
         v2BrowserInitStylesBySurface.removeValue(forKey: surfaceId)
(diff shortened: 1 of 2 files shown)


anthhub:
Gentle ping on this one — it fixes a memory leak where per-surface state grows unbounded as tabs are closed. Pretty small change if anyone has a few minutes to look.

anthhub:
Hi @austinywang @lawrencecchen 👋 Gentle ping on this one — it fixes unbounded memory growth from per-surface state not being cleaned up on tab close. Would appreciate a review when you get a chance!

anthhub:
Quick triage request on this PR.

This one still seems valuable to me because it targets the per-surface cleanup / memory-growth issue, but it is now conflict-stale. If maintainers still want this direction, I can rebase it and refresh the branch; if not, I am happy to close it rather than leave it hanging.

A short yes/no on whether this is still worth reviving would be enough.

anthhub:
Rebased this onto current main and resolved the Workspace.swift conflicts. I kept the later pending-input / restored-agent cleanup from main and preserved the original !isDetaching guard. I couldn’t do a full local build in this checkout because GhosttyKit.xcframework isn’t present here, but the branch should be back in reviewable shape.

PR #2235, opened 2026-03-27
fix: play system bell sound when bell-features is not configured

### LINKED ISSUE #2209: Terminal bell (\a) does not produce sound
## Bug

The terminal bell character (`\a` / BEL / `0x07`) does not produce any audible sound in cmux. The same bell works correctly in Terminal.app and iTerm2.

## Steps to reproduce

1. Open a terminal in cmux
2. Run `echo -e '\a'` or `printf '\a'`
3. No sound is heard

## Expected behavior

An audible bell sound should play (the system alert sound), matching the behavior of Terminal.app and iTerm2.

## Current behavior

No sound is produced. The bell character is silently ignored.

## Notes

- This may be a missing handler in the Ghostty terminal integration — Ghostty may emit a bell callback/event that cmux isn't handling
- Check if there's a `bell` callback in the Ghostty surface/terminal delegate that needs to be implemented
- The system alert sound can be played via `NSSound.beep()` or `AudioServicesPlayAlertSound(kSystemSoundID_UserAlert)`

anthhub:
Fixes #2209

Terminal bell (\a) did not produce any sound because Ghostty's default `bell-features` has `system` disabled. When a user hasn't explicitly configured `bell-features = system` in their Ghostty config, the `ringBell()` function checked the feature bits, found none set, and did nothing.

This adds a fallback: when no bell features are configured (features == 0), play the system beep sound by default. Users who explicitly set bell-features to a specific value are unaffected.

## Test plan
- Remove any `bell-features` line from Ghostty config
- Run `echo -e '\a'` in terminal → should hear system beep
- Set `bell-features = none` → should be silent (explicit opt-out)
- Set `bell-features = system` → should hear beep (explicit opt-in)

---
## Summary by cubic
Play the system bell by default when `bell-features` isn’t set, so the terminal bell (\a) makes a sound out of the box. Also fixes Ctrl shortcuts being dropped when a CJK IME is active.

- **Bug Fixes**
  - Bell: If `bell-features` is missing from the config, call `NSSound.beep()`. Explicit settings (including `bell-features = none`) are unchanged.
  - Input: Remove the `hasMarkedText()` check from the Ctrl key path so Ctrl+C/Z/D work even with IMEs that leave empty marked text.

<sup>Written for commit d09733164df7664059cbf6215346310bd296d5cc. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal bell behavior so the fallback system beep reliably when bell configuration is missing or unreadable.
  * Ensured feature-based bell options are only applied when valid configuration is present, preventing incorrect bell handling.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -1995,11 +1995,13 @@ class GhosttyApp {
         return String(cString: value)
     }
 
-    private func bellFeatures() -> CUnsignedInt {
-        guard let config else { return 0 }
+    private func bellFeatures() -> CUnsignedInt? {
+        guard let config else { return nil }
         var features: CUnsignedInt = 0
         let key = "bell-features"
-        _ = ghostty_config_get(config, &features, key, UInt(key.lengthOfBytes(using: .utf8)))
+        guard ghostty_config_get(config, &features, key, UInt(key.lengthOfBytes(using: .utf8))) else {
+            return nil
+        }
         return features
     }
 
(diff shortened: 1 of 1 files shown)


anthhub:
Closing this — after deeper investigation, `ghostty_config_get` always returns `true` for `bell-features` (it's a packed struct with a defined default, not an optional key). The nil fallback path would never be reached in practice.

The actual root cause is different: the default `bell-features` value has `system: false`, so the system sound bit (bit 0) is off by default. The fix should check if the system bit is unset and call `NSSound.beep()` in that case, rather than relying on the config getter's return value. Will resubmit with the correct approach.

PR #2216, opened 2026-03-26
refactor: physical keyCode as primary match path in matchShortcut

lawrencecchen:
## Summary

- Refactors `matchShortcut` to try physical ANSI keyCode matching first, aligning with Ghostty's `Binding.Set.getEvent()` which checks physical key before character
- Removes the special non-Latin guards added in https://github.com/manaflow-ai/cmux/pull/2202, since physical-first matching handles Russian, Korean, Arabic, etc. without special cases
- Preserves Dvorak/Colemak behavior: when a physical keyCode matches a letter shortcut but the event's character identity disagrees (e.g. physical I producing "c" on Dvorak), the physical match is rejected and character-based fallback takes over
- Non-letter shortcuts (digits, punctuation) always accept physical matches, handling AZERTY and non-US layouts correctly
- Adds new tests for Dvorak (physical C producing "j"), Colemak (physical F producing "e"), and Russian layout via physical keyCode

Closes https://github.com/manaflow-ai/cmux/issues/2215

## Test plan

- [ ] All existing shortcut routing tests pass in CI
- [ ] New Dvorak tests verify character-based matching when physical key disagrees
- [ ] New Colemak test verifies character-based matching for remapped keys
- [ ] New Russian physical keyCode test verifies non-Latin matching without special guards
- [ ] Manual test with Russian keyboard layout: Cmd+T, Cmd+N, Cmd+W should all work
- [ ] Manual test with Dvorak layout: Cmd+C (copy) should not trigger Cmd+I shortcut when pressing physical I

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard shortcut recognition for non-US keyboard layouts, including Dvorak, Colemak, and Russian layouts.

* **Tests**
  * Added comprehensive test coverage for keyboard shortcut routing with international keyboard layouts.

---
## Summary by cubic
Refactors shortcut matching to prioritize the physical ANSI `keyCode`, with a character-based fallback. This simplifies non‑Latin handling and makes shortcuts reliable on Dvorak, Colemak, AZERTY, and other layouts. Closes #2215.

- **Refactors**
  - Match via `event.keyCode` first (aligned with Ghostty’s physical‑first approach).
  - For letter shortcuts, require character agreement; otherwise fall back to character/layout matching.
  - For digits/punctuation, always accept the physical match.
  - Remove non‑Latin guards; add tests for Dvorak, Colemak, and Russian.

[...]

PR #2237, opened 2026-03-27
fix: prevent Cmd+Shift+U escape sequence leak in kitty protocol mode

### LINKED ISSUE #2198: Cmd+Shift+U outputs escape sequence when no notifications exist
## Description

The `Cmd+Shift+U` shortcut (Jump to Unread Notification) inserts a raw escape sequence `[12629;10u` into the terminal when there are no unread notifications. It should either do nothing or show a brief "no notifications" indicator.

## Steps to Reproduce

1. Open a cmux terminal
2. Ensure there are no unread notifications (no tabs with notification badges)
3. Press `Cmd+Shift+U`

**Expected:** Nothing happens, or a brief visual indicator shows "no unread notifications."
**Actual:** The string `[12629;10u` is inserted into the active terminal, which can corrupt commands or cause unexpected behavior.

## Analysis

The output `[12629;10u` looks like a CSI u (kitty keyboard protocol) escape sequence that is being passed through to the terminal instead of being consumed by the cmux keybinding handler. It seems the shortcut is only intercepted when a matching notification exists, and falls through to the terminal otherwise.

## Environment

- macOS Sequoia (Darwin 25.2.0)
- cmux (stable channel)

anthhub:
Fixes #2198

When no unread notifications exist, pressing Cmd+Shift+U would output the escape sequence `[12629;10u` into the terminal. This happened because the kitty keyboard protocol encodes key events differently, causing `matchShortcut` to fail character matching and letting the event pass through to the terminal.

## Root Cause

In `matchShortcut`, the layout-character match (`shortcutLayoutCharacterProvider`, which uses keyCode → keyboard layout translation) was placed *after* an early-return guard that blocked further matching for letter shortcuts when `charactersIgnoringModifiers` was non-empty ASCII but didn't match the shortcut key. In kitty protocol mode, `charactersIgnoringModifiers` carries a synthetic value, so the guard fired before the reliable keyCode-based translation had a chance to run.

## Fix

Move the `shortcutLayoutCharacterProvider` match *before* the guard. The keyCode → layout translation reliably identifies the physical key (e.g. keyCode 32 → "u") regardless of keyboard protocol mode. The guard is preserved after both character sources have been tried, so cross-layout collision protection for letter shortcuts remains intact.

## Test plan
- Press Cmd+Shift+U with no unread notifications → should do nothing (no escape sequence output)
- Press Cmd+Shift+U with unread notifications → should jump to latest unread
- Test with kitty keyboard protocol enabled (`echo -e '\e[>1u'`)
- Verify other letter shortcuts (Cmd+Shift+P, Cmd+N, etc.) still work correctly on non-US keyboard layouts

---
## Summary by cubic
Fixes #2198. Stops Cmd+Shift+U from printing a kitty keyboard protocol escape sequence when there are no unread notifications by matching the keyCode→layout character before the early guard.

- **Bug Fixes**
  - Move layout-character match ahead of the ASCII guard so letter shortcuts are recognized in kitty protocol mode; Cmd+Shift+U no longer passes through.
  - Keep the guard after both character sources to avoid cross-layout collisions; other shortcuts still work on non‑US layouts.

<sup>Written for commit fbb7d8f74e171ab124f116f192ccbfbd313044d7. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard shortcut matching for Cmd-based letter shortcuts to enhance character recognition and behavior across different keyboard layouts.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -10671,24 +10671,15 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
             return true
         }
 
-        // For command-based shortcuts, trust AppKit's layout-aware characters when present.
-        // Keep this strict for letter shortcuts to avoid physical-key collisions across layouts,
-        // while still allowing keyCode fallback for digit/punctuation shortcuts on non-US layouts.
-        // When a non-Latin input source is active (Russian, Korean, Chinese, Japanese, etc.),
-        // charactersIgnoringModifiers returns non-ASCII characters that can never match
-        // a Latin shortcut key — skip this guard and fall through to layout-based matching.
         let hasEventChars = !(eventCharsIgnoringModifiers?.isEmpty ?? true)
         let eventCharsAreASCII = eventCharsIgnoringModifiers?.allSatisfy(\.isASCII) ?? true
-        if hasEventChars,
-           eventCharsAreASCII,
-           flags.contains(.command),
-           !flags.contains(.control),
-           shouldRequireCharacterMatchForCommandShortcut(shortcutKey: shortcutKey) {
-            return false
-        }
 
         // Match using the current keyboard layout so Command shortcuts stay character-based
         // across layouts (QWERTY, Dvorak, etc.) instead of being tied to ANSI physical keys.
+        // This also provides a reliable fallback when the kitty keyboard protocol is active:
+        // in that mode, charactersIgnoringModifiers may be empty or carry a synthetic value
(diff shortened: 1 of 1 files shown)


anthhub:
Closing — this conflicts with PR #2216 (physical keyCode refactor by @lawrencecchen) which touches the same `matchShortcut` function with a larger rewrite. Will rebase on top of #2216 if the fix is still needed after that lands.

PR #2245, opened 2026-03-27
fix: skip hook injection for non-interactive claude invocations

### LINKED ISSUE #2206: Claude wrapper shim injects hooks into non-interactive claude invocations, creating phantom sessions that burn API credits
## Summary

cmux's `claude` PATH shim injects `--settings` with lifecycle hooks into every `claude` invocation inside a cmux terminal. This includes non-interactive, programmatic calls — such as auth token refresh — that were never meant to start a conversation. The hooks trigger a `"."` enqueue that Claude interprets as a user prompt, spawning a full response cycle with no human present.

Related: #2140 documents the shim behavior. This issue reports a concrete bug where the shim causes unattended API credit consumption.

## Root cause chain

1. A third-party tool (`opencode-claude-auth`) spawns `claude` to refresh an OAuth token
2. cmux's shim at `/Applications/cmux.app/Contents/Resources/bin/claude` intercepts the call because `CMUX_SURFACE_ID` is set in the terminal environment
3. The shim injects `--session-id` + `--settings` with SessionStart/Stop/PreToolUse hooks
4. The SessionStart hook fires `cmux claude-hook session-start`
5. This creates a `queue-operation` with `content: "."` in the JSONL session file
6. Claude receives `"."` as a user prompt → responds → triggers PreToolUse hooks → more responses → Stop hook
7. A full phantom conversation is created with ~10–27 messages, all from a single dot

## Impact

- **API credits burned unattended**: ~30 phantom sessions fired every ~30 minutes between 06:30–09:33 while I was asleep. Each triggers a full Claude response cycle.
- **Billing contamination**: Phantom sessions chain indefinitely — the ~30-minute hook interval bridges across the idle threshold, creating continuous multi-hour blocks of fake activity that run in the background even when the laptop lid is closed. In my case, 3+ hours of phantom billable time on a single day.

## Evidence

Session file `f7847c46-3a2a-4f62-9881-5e0d1fe30406.jsonl`:

```
 1. [queue-operation] enqueue    content: "."
 2. [queue-operation] dequeue
 3. [progress]        cmux claude-hook session-start
 4. [user]            (the "." arrives as a prompt)
 5. [assistant]       "Hi! How can I help you today?"
 6–25.               (tool use, hooks, responses)
26. [progress]        cmux claude-hook stop
```

436 JSONL files in the project directory, 434 contain cmux hook artifacts, totaling 2,034 phantom messages.

## Suggested fix

The shim already passes through `mcp`, `config`, and `api-key` subcommands (line 66). It should also bypass hook injection when:

- `claude` is invoked with non-interactive flags (e.g., `--print-auth-token`, `--output-format json`, `-p` with piped stdin)
- stdin is not a TTY (`[[ ! -t 0 ]]`)
- The invocation is clearly programmatic (no subcommand, piped input)

A TTY check would catch the most common case:

```bash
# Don't inject hooks for non-interactive invocations
if [[ ! -t 0 ]]; then
    exec "$REAL_CLAUDE" "$@"
fi
```

## Environment

- cmux 0.62.2 (77) [6c203b514]
- Claude Code (latest, via `~/.local/bin/claude`)
- opencode-claude-auth plugin (spawns `claude` for token refresh)
- macOS, Apple Silicon

anthhub:
## Problem

The `claude` PATH shim inside cmux terminals injects `--session-id` and `--settings` with lifecycle hooks into *every* `claude` invocation where `CMUX_SURFACE_ID` is set. This includes non-interactive, headless calls — such as OAuth token refresh via `opencode-claude-auth` — that have no interactive TTY and were never meant to start a conversation.

When the `SessionStart` hook fires, it enqueues `"."` in the JSONL session file. Claude receives it as a user prompt and generates a full response, triggering `PreToolUse`, `Stop`, and more hooks in a chain. The result is phantom session files with 10–27 messages each, generated every ~30 minutes, burning API credits unattended.

Fixes #2206

## Fix

Add a `[[ ! -t 0 ]]` guard after the subcommand passthrough block. Headless invocations have no stdin TTY, so this check cleanly detects piped/scripted calls without requiring exhaustive flag enumeration (`--print`, `--output-format`, `--print-auth-token`, etc.).

Interactive claude sessions inside cmux always have a TTY (the terminal provides stdin), so the check does not affect normal cmux usage.

## Test plan

- [ ] Non-interactive: `echo "test" | claude` inside a cmux terminal → no hook injection, passes through to real claude binary
- [ ] Interactive: `claude` inside a cmux terminal → hooks injected as before, session tracking works
- [ ] Subcommand passthrough still works: `claude mcp`, `claude config` → no hooks
- [ ] `cmux claude-hook` commands still fire for interactive sessions

---
## Summary by cubic
Skip hook injection for non-interactive `claude` runs inside `cmux`, preventing phantom sessions and wasted API credits (fixes #2206). Headless calls now pass through to the real binary without starting a session.

- **Bug Fixes**
  - Detect non-interactive calls via `[[ ! -t 0 ]]`, unset `CLAUDECODE`, and bypass hook injection.
  - Keep passthrough for `claude mcp`, `claude config`, etc.
  - Interactive sessions are unchanged; hooks still run and sessions track normally.

<sup>Written for commit 904baed483926b931ab15dcdc3b5c6617c36074a. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Non-interactive (piped/scripted) invocations now bypass session injection and lifecycle hooks, preserving original arguments and avoiding added session/settings flags—resulting in faster, more predictable behavior for scripted use.

### DIFF
--- a/Resources/bin/claude
+++ b/Resources/bin/claude
@@ -60,6 +60,13 @@ case "${1:-}" in
     mcp|config|api-key|rc|remote-control) exec "$REAL_CLAUDE" "$@" ;;
 esac
 
+# Pass through non-interactive (piped/scripted) invocations without injecting
+# hooks. Headless callers like auth-token refreshers have no TTY and should
+# not trigger session-start hooks that create phantom sessions burning API credits.
+if [[ ! -t 0 ]]; then
+    exec "$REAL_CLAUDE" "$@"
+fi
+
 # Unset CLAUDECODE to avoid "nested session" detection — cmux terminals are
 # independent sessions even when the parent shell was launched from Claude Code.
 unset CLAUDECODE
--- a/Resources/bin/claude
+++ b/Resources/bin/claude
@@ -64,6 +64,7 @@ esac
 # hooks. Headless callers like auth-token refreshers have no TTY and should
 # not trigger session-start hooks that create phantom sessions burning API credits.
 if [[ ! -t 0 ]]; then
+    unset CLAUDECODE
     exec "$REAL_CLAUDE" "$@"
 fi
 


mrm007:
@anthhub did you see I already have a PR open for this? #2213

> ## Why flag detection (not TTY check)
>
> The issue suggests `[[ ! -t 0 ]]` as a catch-all, but the existing test harness uses `subprocess.run(capture_output=True)` which makes stdin a pipe — a TTY check would break all existing tests without a significant test refactor. Flag detection is precise, testable, and directly addresses the reported bug (`claude -p .`).
> 
> A TTY check could be added later as defense-in-depth once the test harness supports pseudo-TTY simulation.

anthhub:
Small bump on this one.

From what I can see, there doesn’t seem to be any new actionable review feedback at the moment. It also looks like part of the remaining blocker may just be the Vercel/team auth side rather than the code itself, but please correct me if I’m missing something.

Happy to make follow-up changes if needed — just wanted to bring this back up in the queue. Thanks!

anthhub:
Closing this in favor of #2213.

That PR was opened earlier, follows the flag-detection approach discussed in this thread, and already includes the related test coverage. I’d rather reduce duplicate review surface here than keep two competing fixes open.

If maintainers specifically want the TTY-check approach instead, I’m happy to revisit it in a new focused PR.

PR #2259, opened 2026-03-27
Add web-browser entitlement for Keychain password autofill

### LINKED ISSUE #2258: Support macOS Keychain password autofill in browser panes
I noticed #1518 added the `com.apple.developer.web-browser` entitlement for Keychain autofill and #1727 had to revert it because it broke signing without a matching provisioning profile.

Have you looked into requesting the restricted entitlement from Apple through the provisioning portal? cmux does have real browser functionality, so it seems like it could qualify. Keychain autofill in browser panes would be a great usability improvement.

Curious if there's been any progress on this or if there's another approach being considered.

joshuaswanson:
Adds the `com.apple.developer.web-browser` entitlement so that macOS Keychain password autofill works in browser panes. The WKWebView configuration already uses the default persistent `WKWebsiteDataStore`, so the entitlement is the only missing piece.

Since cmux is distributed directly via Sparkle (not the App Store), this entitlement doesn't require Apple's approval process.

Closes #2258

---
## Summary by cubic
Enable macOS Keychain password autofill in browser panes by adding the `com.apple.developer.web-browser` entitlement. `WKWebView` already uses the default persistent store; distribution via Sparkle means no Apple approval needed.

<sup>Written for commit 681152d50c1580a0c6a9364b7cc95200e31151f1. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Chores**
  * Updated app permissions to enable web browser capabilities.

### DIFF
--- a/cmux.entitlements
+++ b/cmux.entitlements
@@ -14,5 +14,7 @@
 	<true/>
 	<key>com.apple.security.automation.apple-events</key>
 	<true/>
+	<key>com.apple.developer.web-browser</key>
+	<true/>
 </dict>
 </plist>


joshuaswanson:
Closing this — just found that #1518 already tried this and #1727 had to revert it because the restricted entitlement breaks nightly/release signing without a matching provisioning profile. Sorry for the noise.

PR #2453, opened 2026-04-01
Make sidebar PR lookups event-driven

### LINKED ISSUE #2414: GitHub PR lookups should be event-driven and avoid unnecessary gh queries
## Summary

The sidebar GitHub metadata path should do much less `gh` querying overall.

Right now the app already has good event-driven trigger points for git metadata refreshes:

- initial workspace setup
- directory change
- branch change
- branch cleared

Those are the moments when we actually learn something new locally. The problem is that we still also rely on periodic polling for the selected workspace and for tracked workspaces, which means we can keep re-running `gh` queries even when nothing meaningful has changed.

This matters because the goal is not just to handle rate limits better after the fact. The better fix is to avoid unnecessary GitHub queries in the first place.

Related:
- https://github.com/manaflow-ai/cmux/issues/2388
- https://github.com/manaflow-ai/cmux/pull/2405

## Current problem

Current behavior still appears to include:

- selected workspace poll every 5 seconds
- tracked workspace poll every 30 seconds
- `gh pr list` to resolve PR state for the current branch
- `gh pr checks` to resolve checks for open PRs

That means a long-running session with many workspaces can keep asking GitHub the same question repeatedly even when:

- the branch has not changed
- the directory has not changed
- the PR identity has not changed
- the last result is still fresh enough to reuse

## Desired direction

Prefer **event-driven refreshes** tied to actual local git state changes, and make timer-driven polling a narrow fallback instead of the default behavior.

## Suggested approach

1. Treat branch changes as the primary trigger for `gh pr list`.
   If the branch name is unchanged, do not keep re-running PR discovery every few seconds.

2. If a branch has no PR, remember that result and stop polling it aggressively.
   Re-check only when:
   - the branch changes
   - the directory/repo changes
   - the user explicitly refreshes
   - a much longer stale timeout expires

3. Separate PR discovery from PR checks refresh.
   `gh pr list` should mostly happen on branch transition.
   `gh pr checks` should only happen when there is already a known open PR.

4. Add freshness windows / TTLs.
   If a PR/checks result was fetched recently, reuse it instead of querying GitHub again immediately from another poll path.

5. Limit periodic polling to the cases that genuinely need it.
   For example:
   - maybe only the selected workspace
   - maybe only when there is a known open PR
   - maybe at a much slower cadence than branch detection

6. Track the last queried branch and repo per panel/workspace.
   If the local branch/repo identity is unchanged, avoid redundant GitHub lookups.

7. Add debug instrumentation for GitHub query volume.
   It should be easy to see:
   - how often `gh pr list` runs
   - how often `gh pr checks` runs
   - why each query was triggered
   - whether the result came from cache or a live query

8. Add tests or stress coverage that verify query volume stays bounded over long uptimes with many workspaces.

## Concrete goal

Make GitHub metadata refresh mostly driven by real local events like branch changes, not by constant periodic polling. The app should avoid asking GitHub the same question over and over when the branch/repo state has not changed.

lawrencecchen:
## Summary
- replace unconditional sidebar PR polling with per-panel cache files and force-signal files in the zsh/bash shell integrations
- cache both PR and no-PR results so idle workspaces reuse prior state, while branch changes and explicit git/gh activity can trigger immediate reprobes
- extend the existing shell integration regression harness to cover cached idle behavior and cached no-PR lookups

## Validation
- Did not run local tests per repo policy
- `zsh -n Resources/shell-integration/cmux-zsh-integration.zsh`
- `bash -n Resources/shell-integration/cmux-bash-integration.bash`
- `python3 -m py_compile tests/test_issue_1138_sidebar_pr_polling.py`

Closes #2414.

---
## Summary by cubic
Make sidebar PR lookups event-driven with per-panel caching, a reprobe signal, and a persistent idle poll loop to keep badges fresh while reducing `gh` calls. Restores predictable idle refresh cadence and reduces UI churn, addressing #2414.

- **New Features**
  - Keep the PR poll loop alive between prompts; wake early on branch changes, per-panel reprobe signals, and a `zsh` Git HEAD watcher.
  - Cache PR and no-PR state per panel and reuse it between prompts; optional debug logging to `/tmp/cmux-pr-debug.log`.

- **Bug Fixes**
  - Clear stale PR state and signal files on context changes and when stopping the loop; handle transient `gh` failures and timeouts without stopping the poller.
  - Fix `zsh` timer handling by falling back to `$SECONDS` when `EPOCHSECONDS` is unset; add `_cmux_now` in `bash` for consistent timing.

<sup>Written for commit a00b46ceaa5f5914a496bd88e24fa96257539bb6. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Per-panel PR state caching (including “no PR” results) for faster, more accurate PR badges.
  * Force-probe signal to refresh PR information on demand.
  * Optional debug logging for PR detection troubleshooting.

* **Improvements**
  * Polling optimized to wake early on demand and use a monotonic timestamp for prompt timing.
  * Prompt lifecycle now clears stale PR state when context (path/branch/HEAD) changes.

### DIFF
--- a/Resources/shell-integration/cmux-bash-integration.bash
+++ b/Resources/shell-integration/cmux-bash-integration.bash
@@ -62,8 +62,6 @@ _CMUX_PR_POLL_PWD="${_CMUX_PR_POLL_PWD:-}"
 _CMUX_PR_LAST_BRANCH="${_CMUX_PR_LAST_BRANCH:-}"
 _CMUX_PR_NO_PR_BRANCH="${_CMUX_PR_NO_PR_BRANCH:-}"
 _CMUX_PR_POLL_INTERVAL="${_CMUX_PR_POLL_INTERVAL:-45}"
-_CMUX_PR_RESULT_TTL="${_CMUX_PR_RESULT_TTL:-300}"
-_CMUX_PR_NO_PR_TTL="${_CMUX_PR_NO_PR_TTL:-300}"
 _CMUX_PR_FORCE="${_CMUX_PR_FORCE:-0}"
 _CMUX_PR_DEBUG="${_CMUX_PR_DEBUG:-0}"
 _CMUX_ASYNC_JOB_TIMEOUT="${_CMUX_ASYNC_JOB_TIMEOUT:-20}"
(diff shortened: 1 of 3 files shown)


avkrishnamurthy:
Im not sure this is fixed @austinywang  ? My gh graphql api rate limit is being saturated by cmux with 7 workspaces open, it uses ~2/second

Issue #2069, opened 2026-03-24
Markdown viewer: render images and mermaid code blocks


chriscoey:
The `cmux markdown open <file>` viewer handles text formatting well (headings, code blocks, lists, tables). Two things it doesn't currently render:

- **Images** — `![alt](path)` shows as text rather than rendering the image
- **Mermaid code blocks** — ` ```mermaid ` shows as plain code rather than a rendered diagram

Both are supported by VS Code's markdown preview and GitHub's markdown renderer. Adding them would make the cmux markdown viewer more useful as a side-by-side reference pane, especially for agent workflows that produce diagrams and screenshots.

Related: #1295

anthhub:
Images are probably the easier of the two. If the markdown viewer is SwiftUI-based, `AsyncImage` handles remote URLs natively, and for local paths you'd just load from disk and wrap in an `Image` view. The trickier part is resolving relative paths — the viewer needs to know the source file's directory to resolve `![alt](./diagram.png)` correctly.

Mermaid is harder. There's no native Swift/AppKit library for it — every implementation I've seen either shells out to a Node.js process or loads the mermaid.js bundle inside a `WKWebView`. The WKWebView approach is probably the right call: detect ` ```mermaid ` blocks, swap them out for a `WKWebView` that loads a minimal HTML page with the mermaid.js CDN (or bundled), and let it render. The main annoyance is sizing — you'd need JS→Swift messaging to communicate the rendered height so the webview doesn't clip the diagram.

If the viewer is already partially WebView-based (e.g. for HTML output), switching the whole renderer to WKWebView with a proper HTML/CSS pipeline might actually be simpler than mixing native SwiftUI views with embedded webviews for just the mermaid blocks.

PR #2463, opened 2026-04-01
feat: render mermaid diagrams in markdown panels

kangsan-rpls:
## Summary

- Mermaid fenced code blocks in markdown panels are now rendered as SVG diagrams instead of raw text
- Uses bundled mermaid.js (v11) inside WKWebView — no external dependencies, works offline
- Supports dark/light mode, horizontal scrolling for wide diagrams, dynamic resize handling, and inline error display for invalid syntax

Partially addresses https://github.com/manaflow-ai/cmux/issues/2069 (mermaid rendering; images not included in this PR).

## Before / After

| Before | After |
|--------|-------|
| before | after |

## How it works

1. MarkdownUI's `CodeBlockConfiguration.language` detects `mermaid` fenced blocks
2. `MermaidRenderView` (new) wraps a `WKWebView` that loads a minimal HTML page with the bundled `mermaid.min.js`
3. mermaid.js renders the code to SVG inside the WebView, then calls `bindFunctions` for tooltip/click handler interactivity
4. JS reports the rendered content height back to Swift via `WKScriptMessageHandler` + `@State` binding, so the WebView sizes correctly within the parent SwiftUI `ScrollView`
5. A `resize` event listener keeps the height in sync when the panel is resized
6. Non-mermaid code blocks are unchanged

## Changes

| File | Change |
|------|--------|
| `Sources/Panels/MermaidRenderView.swift` | **New** — WKWebView-based mermaid renderer |
| `Sources/Panels/MarkdownPanelView.swift` | Branch on `configuration.language == "mermaid"` in `.codeBlock` theme |
| `Resources/mermaid.min.js` | **New** — Bundled mermaid.js v11 for offline rendering |
| `Resources/Localizable.xcstrings` | Add `markdown.mermaid.syntaxError` key with 19 locale translations |
| `GhosttyTabs.xcodeproj/project.pbxproj` | Register new source + resource files |

## Design decisions

- **Bundled mermaid.js vs CDN**: Bundled for offline reliability. ~2.9 MB in the app bundle.
- **Inline WKWebView vs full renderer swap**: Keeps the existing MarkdownUI renderer and only swaps mermaid code blocks, minimizing blast radius. Complementary to #2424 (WKWebView markdown renderer).
- **vs PR #1396 (mmdc CLI)**: No Node.js / external tool dependency. Runs entirely in-process.
- **WeakScriptMessageHandler**: Prevents the retain cycle between `WKUserContentController` and the coordinator. Handlers are removed in `dismantleNSView`.
- **securityLevel: 'strict'**: Prevents raw HTML injection from diagram node labels. `</script>` is escaped to prevent XSS. Error messages use `textContent` instead of `innerHTML`.
- **Height management**: Uses `@State` binding (not Auto Layout constraints) so SwiftUI's `ScrollView` correctly tracks content height. Resize events re-report height to keep the binding fresh.

## Test plan

- [x] Flowchart, sequence, class diagrams render correctly
- [x] Dark mode theme applied
- [x] Wide diagrams scroll horizontally
- [x] Regular code blocks (python, swift, etc.) render unchanged
- [x] Live file reload works (MarkdownPanel's existing DispatchSource watcher)
- [x] Panel resize updates diagram height
- [ ] Invalid mermaid syntax shows inline error
- [ ] Diagram renders on first open without needing resize


kangsan-rpls:
Thanks for the thorough reviews! All feedback has been addressed in the latest push (fddb035c):

**@greptile P1 — `as? CGFloat` cast always fails:** Fixed. Now casting to `Double` first (which `NSNumber` bridges to), then converting to `CGFloat`.

**@greptile P2 — Unlocalized error string:** Fixed. Error prefix is now `String(localized: "markdown.mermaid.syntaxError", defaultValue: "Mermaid syntax error:")` and interpolated into JS.

**@coderabbitai — `innerHTML` XSS in error display:** Fixed. Switched to `textContent` via `document.createElement` + `appendChild` to prevent entity-encoded XSS.

**@coderabbitai — Missing `bindFunctions` call:** Fixed. Now destructuring `{ svg, bindFunctions }` from `mermaid.render()` and calling `bindFunctions?.(container)` after DOM insertion.

kangsan-rpls:
Addressed the two new findings in 0137d2c1:

**@coderabbitai — Stale height on resize:** Added `window.addEventListener('resize', scheduleHeightReport)` so the height binding updates when the panel is resized. Also extracted `reportHeight()` / `scheduleHeightReport()` helpers to reduce duplication.

**@coderabbitai — Missing localization key in xcstrings:** Added `markdown.mermaid.syntaxError` to `Resources/Localizable.xcstrings` with translations for all 19 supported locales (en, ja, zh-Hans, zh-Hant, ko, de, es, fr, it, da, pl, ru, bs, ar, nb, pt-BR, th, tr, uk).

Issue #2279, opened 2026-03-28
Terminal jitters/flickers when toggling between tabs


austinywang:
## Description

The terminal visibly jitters when switching between tabs. The rendering geometry appears to fire multiple updates in rapid succession during the tab switch, causing a visible flicker/jump before settling.

## Root cause hypothesis

Multiple geometry/resize callbacks are being triggered in quick succession when switching tabs — likely from libghostty surface resize notifications combined with cmux's own layout recalculations. This creates a cascade of redraws with intermediate geometry states that are briefly visible.

## Key facts

- **Does NOT happen in upstream Ghostty** — this is a cmux-induced bug, not a libghostty issue.
- The problem is on the cmux side: something in how cmux manages tab visibility, surface attachment, or layout is causing redundant or out-of-order geometry updates during the switch.

## Expected behavior

Tab switching should be instant with no visible geometry jitter — the new tab's content should appear in its final size/position immediately.

## Steps to reproduce

1. Open cmux with multiple tabs (terminal content in each)
2. Toggle between tabs (click sidebar or keyboard shortcut)
3. Observe the brief jitter/flicker as the terminal content resizes before settling

Issue #2224, opened 2026-03-26
Terminal goes blank until


jkitchin:
### cmux version

cmux 0.61.0 (73) [8caa5e9c9]

### macOS version

Tahoe 26.3.1 (25D2128)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Can you reproduce this on cmux NIGHTLY?

I could not test NIGHTLY

### Bug description

Occasionally (often when a question is asked) the terminal screen goes blank, i.e. all the text disappears. If I close the app with the red button, and then click on it again from the taskbar the text comes back. 

### Expected behavior

For that not to happen.

### Steps to reproduce

I don't know how to reproduce it, it just periodically happens. I notice this on three different macs.

### Shell and environment

i use bash.

### Relevant logs or crash reports

```text
there is no crash.
```

bencodezen:
I'm seeing something similar as well.

For my scenario, I'm working with Claude over a session, and when it's time for me to respond it goes completely blank. Resizing the window can trigger a refresh to show all the text, but it's finnicky. Not sure what the cause is just yet.

JLarky:
for me switching to cmux nightly solved the issue

diegoadmoai:
On cmux 0.64.17 (97), macOS (Apple Silicon). Still happening on current stable (survives past the "nightly fixed it" comment). New evidence narrows this to cmux's PTY read/ingest path and rules out the obvious alternatives.

**Symptom:** a fully-rendered line/reply goes blank right after it finishes streaming, leaving a gap. Intermittent, not forceable. Repro'd with `omp` in a cmux terminal surface; per #1156 also happens with Claude Code (not program-specific).

**Recovery matrix (the useful part):**
- `cmux refresh-surfaces` (re-snapshot the surface): does NOT recover it.
- Resizing the window/pane (SIGWINCH → inner program repaints viewport): does NOT recover it.
- Making the inner program **re-emit its committed transcript** (agent history navigation / full re-print): recovers it every time.

Only re-emission works; repainting and re-snapshotting don't. ⇒ the dropped bytes are **gone from cmux entirely** — not in the screen buffer, not in a snapshot — so they only come back when the program writes them again.

**This rules out:**
- *Paint/snapshot bug* — a resize would repaint from the buffer if the buffer had the row; it doesn't.
- *The inner program* — standalone Ghostty on the same machine never drops rows (#1156); only cmux's embedded terminal does.
[...]

Issue #914, opened 2026-03-05
Terminal pane goes blank when notification fires on new tab


bdbull:
## Bug Description
When a notification fires on any newly opened tab, the content of the active terminal pane goes blank.

## Steps to Reproduce
1. Open a new workspace tab
2. Trigger a notification (or wait for one to fire)
3. Active terminal pane content disappears

## Expected Behavior
Pane content remains visible; notification does not cause a redraw/blank.

## Actual Behavior
Pane goes blank — appears to be a missed redraw after the notification event.

bdbull:
https://github.com/user-attachments/assets/28658aca-ca75-41f8-afa0-ffca9ef57ed5

Watch when the notification happens.

austinywang:
Duplicate of #3026 (resolved by #3048) — closing. If this still happens on the latest build, please reopen.

PR #2484, opened 2026-04-01
Fix blank screen: remove occlusion cache that suppresses retries

### LINKED ISSUE #1156: Terminal screen goes blank after sending a message in Claude Code
## Description

When using Claude Code inside a cmux terminal, the screen disappears/goes blank every time a message is sent. The terminal content only reappears after clicking back into the terminal pane. This happens 100% of the time.

## Steps to Reproduce

1. Open a terminal surface in cmux
2. Launch `claude` (Claude Code CLI)
3. Type and send any message
4. **Observed**: Screen goes blank/disappears
5. **Expected**: Screen should remain visible with Claude Code's response rendering

## Environment

- **cmux**: latest (installed via DMG)
- **macOS**: Darwin 25.3.0
- **TERM**: `xterm-ghostty`
- **Surface health**: All surfaces report `in_window=true`, no issues
- **Layout**: Single pane, single workspace

## Additional Context

- **Works correctly in Ghostty** (same machine, same Claude Code version) — the issue is cmux-specific.
- Likely related to how cmux handles TUI redraws triggered by Claude Code's interface updates.

sefr-sefr:
## Summary

Fixes a bug where terminal surfaces permanently go blank after certain transitions (vim exit, tab switching, notification events).

### Root Cause

`setVisibleInUI` gates `ghostty_surface_set_occlusion()` behind a `lastRequestedPortalOcclusionVisible` cache. The cache is updated *before* the optional chain `terminalSurface?.setOcclusion(visible)` — so when `terminalSurface` is nil (transient state during surface creation), the cache records "sent" while Ghostty was never notified. All subsequent calls with the same value are suppressed by the guard. The renderer thread's `if (!self.flags.visible) return` in `drawFrame` drops all draw requests forever — permanent blank screen.

### Fix

Remove the cache entirely. `ghostty_surface_set_occlusion()` is cheap (lock-free mailbox push) and idempotent (renderer thread deduplicates via `if (self.flags.visible == v) break`). The cache optimized away calls that don't need optimization.

`setOcclusion` is now called unconditionally in `setVisibleInUI`. If `terminalSurface` is nil, the optional chain skips, but the next `setVisibleInUI` call retries since there is no cache to suppress it.

### What stays the same

- Notification posting and first responder management remain gated on actual visibility transitions (`wasVisible != visible`)
- No Ghostty fork changes
- No new timers, polling, or callbacks

## Test plan
- [ ] Build passes
- [ ] Multi-tab vim open/close no longer causes blank screen
- [ ] Rapid workspace switching doesn't cause blank surfaces
- [ ] No performance regression during normal terminal use

Fixes #1156
Relates to #2224, #914, #2279

---
## Summary by cubic
Fixes permanent blank screens after visibility changes by removing the occlusion cache and always sending occlusion updates to the renderer. Fixes #1156; relates to #2224, #914, #2279.

- **Bug Fixes**
  - Removed the `lastRequestedPortalOcclusionVisible` cache and now call `terminalSurface?.setOcclusion(visible)` unconditionally in `setVisibleInUI` to prevent suppressed retries when the surface is nil.
  - Notifications and first-responder logic still trigger only on real visibility changes; the renderer deduplicates occlusion updates, so no performance impact.

<sup>Written for commit 86a2a464bbeab9fe2c0e7dafc79cf408229c882f. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal surface visibility handling to ensure consistent occlusion updates during visibility transitions.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -7332,7 +7332,6 @@ final class GhosttySurfaceScrollView: NSView {
     private static let scrollToBottomThreshold: CGFloat = 5.0
     private var isActive = true
     private var lastFocusRefreshAt: CFTimeInterval = 0
-    private var lastRequestedPortalOcclusionVisible: Bool?
     private var activeDropZone: DropZone?
     private var pendingDropZone: DropZone?
     private var dropZoneOverlayAnimationGeneration: UInt64 = 0
@@ -8679,10 +8678,10 @@ final class GhosttySurfaceScrollView: NSView {
         let wasVisible = surfaceView.isVisibleInUI
         surfaceView.setVisibleInUI(visible)
         isHidden = !visible
-        if wasVisible != visible, lastRequestedPortalOcclusionVisible != visible {
-            lastRequestedPortalOcclusionVisible = visible
-            surfaceView.terminalSurface?.setOcclusion(visible)
-        }
+        // Unconditional: setOcclusion is a cheap mailbox push and the renderer
+        // deduplicates. Gating on a local cache risks permanent desync if
+        // terminalSurface is nil during a visibility transition.
+        surfaceView.terminalSurface?.setOcclusion(visible)
 #if DEBUG
         if wasVisible != visible {
             let transition = "\(wasVisible ? 1 : 0)->\(visible ? 1 : 0)"


sefr-sefr:
Related PRs: #2485 (sync occlusion on surface creation) and #2486 (reconcile occlusion in forceRefresh). Each PR is independent and can be reviewed/merged separately. This PR (cache removal) is the root cause fix. The other two close edge cases and add defense-in-depth.

sefr-sefr:

sefr-sefr:
Manually tested all three changes on a debug build. Verified: (1) no blank screen after vim exit with multiple tabs, (2) no jitter regression on rapid tab switching, (3) no blank surfaces on workspace switching. Reopening.

atani:
## Summary

Root-cause analysis is spot-on: the occlusion cache desync is the culprit for permanent blank screens. The fix approach (unconditional `setOcclusion`) is sound and minimal. Approve with a few suggestions below.

## Rebase needed (blocking)

The diff targets a base with `lastRequestedPortalOcclusionVisible`, but upstream `main` currently has neither the cache nor any `setOcclusion` call in `setVisibleInUI`. A rebase is needed — the good news is it simplifies the diff to just *adding* the unconditional call.

## Suggestions (non-blocking)

**Consolidate #2484 + #2485 + #2486 into a single PR?**
All three fix the same bug class (occlusion desync) in the same file. A single PR would make the layered defense strategy clear at a glance:
1. Primary path — `setVisibleInUI` (this PR)
2. Creation sync — `createSurface` (#2485)
3. Recovery — `forceRefresh` (#2486)

**Note the optional-chaining gap in the description.**
`surfaceView.terminalSurface?.setOcclusion(visible)` still silently no-ops when `terminalSurface` is nil. This is fine because #2485/#2486 provide recovery paths — worth calling out so reviewers see the full picture.

Great work tracking this down.

PR #2485, opened 2026-04-01
Sync occlusion state on surface creation

sefr-sefr:
## Summary

Adds `setOcclusion(view.isVisibleInUI)` in `createSurface()` immediately after the existing `ghostty_surface_set_focus` call. This mirrors the focus sync pattern already established in `createSurface`.

### Why

`createSurface` already syncs focus state to newly created Ghostty surfaces because "this surface may have been logically unfocused before the C surface existed." The same applies to occlusion: a surface created in a background tab or hidden workspace should start with the correct visibility state.

Without this, newly created surfaces rely on a subsequent `setVisibleInUI` call from SwiftUI to sync occlusion. If SwiftUI's view state is stable (no re-evaluation triggered), the sync may not happen, leaving the surface with Ghostty's default (visible=true) even when it should be occluded — or vice versa.

### Change

One `if let` block (~4 lines) added after `ghostty_surface_set_focus` in `createSurface()`. No other changes.

## Test plan
- [ ] Build passes
- [ ] Surfaces created in background tabs start correctly occluded
- [ ] Surfaces created in foreground tabs render immediately

Relates to #1156, #2224

---
## Summary by cubic
Syncs occlusion on surface creation by calling setOcclusion(view.isVisibleInUI) right after ghostty_surface_set_focus in createSurface(). This ensures new surfaces inherit the correct visibility (e.g., start occluded in background tabs) instead of defaulting to visible until SwiftUI triggers an update.

<sup>Written for commit 8e0fe840a96d2fa49a9dcf063e05543dabc49513. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Terminal surfaces now correctly respect their visibility state when first created, ensuring they display or hide appropriately in background tabs and hidden workspaces.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -3969,6 +3969,13 @@ final class TerminalSurface: Identifiable, ObservableObject {
         // Ghostty's default.
         ghostty_surface_set_focus(createdSurface, desiredFocusState)
 
+        // Sync occlusion state to the newly created C surface. Ghostty surfaces
+        // default to visible=true, but this surface may be in a background tab
+        // or hidden workspace. Mirrors the focus sync pattern above.
+        if let view = attachedView {
+            setOcclusion(view.isVisibleInUI)
+        }
+
         NotificationCenter.default.post(
             name: .terminalSurfaceDidBecomeReady,
             object: self,


sefr-sefr:
Related PRs: #2484 (root cause fix — cache removal) and #2486 (reconcile occlusion in forceRefresh). Each PR is independent. This PR ensures newly created surfaces start with correct occlusion state.

sefr-sefr:

sefr-sefr:
Manually tested on a debug build with all three changes applied. Reopening.

atani:
## Summary

Clean and well-placed fix. Syncing occlusion at surface creation mirrors the existing `ghostty_surface_set_focus` pattern and closes a real gap: surfaces created in background tabs previously defaulted to `visible=true` when they should start occluded. Approve with minor suggestions.

## Rebase needed (blocking)

The diff references `ghostty_surface_set_focus(createdSurface, desiredFocusState)` as context, which does not exist on current `main`. A rebase (or merging the prerequisite branch first) is needed.

## Suggestions (non-blocking)

**Self-contained comment.**
"Mirrors the focus sync pattern above" relies on code proximity that can shift. A small tweak makes it standalone:

```swift
// Ghostty surfaces default to visible=true, but this surface may be
// in a background tab or hidden workspace. Sync to match the view state.
```

**Consider consolidating with #2484 and #2486.**
All three address occlusion desync in the same file. A single PR would eliminate merge-order concerns and present the full fix clearly.

Nice catch on the creation-time edge case.

PR #2486, opened 2026-04-01
Defense-in-depth: reconcile occlusion state in forceRefresh

sefr-sefr:
## Summary

Adds `setOcclusion(view.isVisibleInUI)` to `TerminalSurface.forceRefresh()`, following the existing pattern of reconciling display ID and Metal layer size before forcing a surface redraw.

### Why

`forceRefresh` is the established recovery method for rendering issues — it is called on focus changes, geometry reconciliation, surface creation, user input, and via the `cmux refresh-surfaces` CLI command. It already re-asserts the display ID (to restart stuck CVDisplayLinks) and re-syncs the Metal layer size. Adding occlusion reconciliation makes it a recovery point for any occlusion state desync.

This is defense-in-depth: if any code path causes Ghostty's renderer to think a surface is occluded when it should be visible, the next `forceRefresh` call will correct it. `ghostty_surface_set_occlusion` is cheap (lock-free mailbox push) and idempotent (renderer thread deduplicates), so the redundant calls on already-synced surfaces have negligible cost.

### Change

One line added after the guard clauses in `forceRefresh()`. No other changes.

## Test plan
- [ ] Build passes
- [ ] Rapid tab switching shows no increase in jitter/flicker (regression check for #2279)
- [ ] `cmux refresh-surfaces` still works
- [ ] No performance regression during normal use

Relates to #1156, #2224, #914, #2279

---
## Summary by cubic
Add occlusion reconciliation to `TerminalSurface.forceRefresh()` by calling `setOcclusion(view.isVisibleInUI)` before forcing a redraw. This keeps renderer visibility in sync and makes each refresh a recovery point for occlusion desyncs (on focus changes, geometry updates, input, surface creation, and `cmux refresh-surfaces`) with negligible overhead.

<sup>Written for commit f4b47868c58a007b20ccd9cb85828c956f4bf8bc. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the terminal view's visibility state was not properly synchronized with the renderer during focus changes and window updates, ensuring consistent display behavior.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -4071,6 +4071,13 @@ final class TerminalSurface: Identifiable, ObservableObject {
         }
         guard let currentSurface = self.surface else { return }
 
+        // Reconcile occlusion: ensure Ghostty's visibility matches the view's.
+        // If a prior setOcclusion call was dropped (e.g. terminalSurface was nil),
+        // the renderer may think this surface is occluded. Re-asserting here
+        // provides a recovery path on focus changes, geometry updates, and
+        // manual cmux refresh-surfaces invocations.
+        setOcclusion(view.isVisibleInUI)
+
         // Re-read self.surface before each ghostty call to guard against the surface
         // being freed during wake-from-sleep geometry reconciliation (issue #432).
         // The surface can be invalidated between calls when AppKit layout triggers


sefr-sefr:
Related PRs: #2484 (root cause fix — cache removal) and #2485 (sync occlusion on creation). Each PR is independent. This PR adds defense-in-depth recovery for any occlusion desync.

sefr-sefr:

sefr-sefr:
Manually tested on a debug build with all three changes applied. No jitter regression from the additional setOcclusion calls. Reopening.

atani:
## Summary

Solid defense-in-depth addition. Making `forceRefresh` a recovery point for occlusion desync is a natural fit — it already reconciles display ID and Metal layer size, so adding occlusion follows the same pattern. Approve with minor suggestions.

## Rebase needed (blocking)

Line numbers and surrounding context differ from current `main`. A rebase is needed for a clean merge.

## Suggestions (non-blocking)

**Trim the comment.**
The current 5-line comment partially restates what the code does. The key insight is *why* this is here — the recovery aspect. Shorter version:

```swift
// Recovery: re-sync occlusion in case a prior call was dropped
// (e.g. terminalSurface was nil during a visibility transition).
setOcclusion(view.isVisibleInUI)
```

**Verify comment syntax.**
The patch shows `/ Reconcile occlusion:` (single slash) — if that matches the actual branch, it would be a syntax error. Likely just a patch formatting artifact, but worth double-checking.

**Consider consolidating with #2484 and #2485.**
Same file, same bug class — a single PR would make the layered fix strategy easy to review.

Great addition to the recovery toolkit.

PR #2838, opened 2026-04-12
fix: use physical key positions for Cmd shortcuts on alternative layouts

mithril-logic:
## Summary

- Cmd shortcuts on Dvorak, Colemak, and other alternative Latin layouts fire the wrong action because `ShortcutStroke.matches()` compares `event.charactersIgnoringModifiers` (which follows the active layout) against shortcut keys before checking the physical key code
- On Dvorak, physical W produces `,` — so Cmd+W matches Cmd+, (Settings) before reaching Cmd+W (Close Tab)
- Fix: for Command-modified shortcuts, match by ANSI key code (physical position) first, following the macOS convention that Cmd shortcuts bind to QWERTY positions regardless of layout

### Before

On Dvorak: Cmd+W (physical) → opens Settings (matches `cmd+,`)

### After

On Dvorak: Cmd+W (physical) → closes tab (matches `cmd+w` by keyCode 13 = kVK_ANSI_W)

## Details

The existing non-Latin layout fix (Russian, Korean, etc.) works because `charactersIgnoringModifiers` returns non-ASCII characters, triggering the keyCode fallback. Dvorak/Colemak produce ASCII characters that happen to match the *wrong* shortcut, so they never reach the fallback.

The fix adds a keyCode-first check for all Command shortcuts and simplifies the fallback chain:
1. Command shortcuts → match by ANSI key code (physical position)
2. Control shortcuts → match by ANSI key code (fallback, unchanged)
3. All others → match by character (unchanged)

## Test plan

- [ ] Dvorak: Cmd+W closes tab (not Settings), Cmd+, opens Settings
- [ ] Colemak: Cmd+W closes tab
- [ ] QWERTY: all shortcuts unchanged (keyCode and character agree)
- [ ] Russian/Korean: shortcuts still work (keyCode fallback still active)
- [ ] Existing unit tests pass

---
## Summary by cubic
Fixes Cmd shortcuts on alternative Latin layouts by matching via physical key positions (ANSI) and records Cmd/Control shortcuts by physical key. Prevents collisions like Cmd+W opening Settings on Dvorak; QWERTY and non‑Latin layouts unchanged.

- **Bug Fixes**
  - Cmd shortcuts now match by ANSI key code first (character match only if no known key code), resolving Dvorak/Colemak collisions like Cmd+W vs Cmd+,.
  - Recording stores the physical QWERTY key for Cmd and Control shortcuts; others still record by character. Control matching behavior is unchanged.

- **Refactors**
  - Added reverse keyCode→key mapping and removed `shouldRequireCharacterMatchForCommandShortcut` and `storedKey(from:)`; expanded tests for physical-key-first matching and recorder behavior.

<sup>Written for commit e9a17fcbc81ad2dabd634914220df48d70347fac. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Command-key shortcuts now match by physical key position (layout-independent) and use simplified fallback rules for more consistent behavior across layouts.
  * Shortcut recording stores the physical key for Command shortcuts instead of the layout-produced character.

* **Tests**
  * Added/updated unit tests (including Dvorak scenarios) to validate physical-key matching and prevent regressions.

### DIFF
--- a/Sources/KeyboardShortcutSettings.swift
+++ b/Sources/KeyboardShortcutSettings.swift
@@ -982,10 +982,14 @@ struct ShortcutStroke: Equatable {
     }
 
     static func from(event: NSEvent, requireModifier: Bool = true) -> ShortcutStroke? {
-        guard !isEscapeCancelEvent(event),
-              let key = storedKey(from: event) else { return nil }
-
+        guard !isEscapeCancelEvent(event) else { return nil }
         let flags = normalizedModifierFlags(from: event.modifierFlags)
+        let preferPhysicalKey = flags.contains(.command) || flags.contains(.control)
+        guard let key = storedKey(
+            keyCode: event.keyCode,
+            charactersIgnoringModifiers: event.charactersIgnoringModifiers,
+            preferPhysicalKey: preferPhysicalKey
+        ) else { return nil }
 
         let stroke = ShortcutStroke(
             key: key,
@@ -1034,6 +1038,16 @@ struct ShortcutStroke: Equatable {
             return keyCode == 36 || keyCode == 76
         }
 
+        // For Command-based shortcuts, match by physical key position (ANSI key
+        // code) first. macOS uses QWERTY positions for Cmd shortcuts regardless
+        // of the active keyboard layout. Without this, alternative Latin layouts
+        // (Dvorak, Colemak) produce false positives — e.g. physical W in Dvorak
+        // sends "," which incorrectly matches Cmd+, instead of Cmd+W.
+        if flags.contains(.command),
+           let expectedKeyCode = Self.keyCodeForShortcutKey(shortcutKey) {
+            return keyCode == expectedKeyCode
+        }
+
         if Self.shortcutCharacterMatches(
             eventCharacter: eventCharacter,
             shortcutKey: shortcutKey,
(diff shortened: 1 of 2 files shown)


mithril-logic:
Note: `shouldRequireCharacterMatchForCommandShortcut` (line 1168) is now dead code — both call sites were removed by this change. Happy to push a follow-up commit removing it, or leave it for you to handle — let me know.

mithril-logic:
Two more observations from review:

**1. Unit test coverage for `ShortcutStroke.matches()`**

There don't appear to be direct tests for the `matches()` method. A few cases that would lock down this fix:

- `ShortcutStroke(key: ",", command: true)` should NOT match keyCode 13 (physical W on Dvorak where `charactersIgnoringModifiers == ","`)
- `ShortcutStroke(key: "w", command: true)` should match keyCode 13 regardless of `charactersIgnoringModifiers`
- `ShortcutStroke(key: "w", control: true)` (no command) should still match by keyCode fallback

**2. Pre-existing: shortcut recorder stores layout character, not physical key**

On Dvorak, pressing physical W in the shortcut recorder stores `key: ","` (via `storedKey(from:)` falling through to `charactersIgnoringModifiers`). If a user records a custom Cmd shortcut on Dvorak, the stored key won't match the physical position after this fix. Not introduced here, but worth knowing about — the recorder may need the same keyCode-first treatment for consistency.

Issue #1492, opened 2026-03-16
Sidebar shows inconsistent Idle vs Needs Input for workspaces in the same state


austinywang:
## Description

The sidebar status indicator inconsistently shows "Idle" vs "Needs Input" for workspaces that are all in the same state. In the screenshot below, all three workspaces have a Claude Code session that is idle/waiting for input, but some show "🔵 Idle" and one shows "🔔 Needs input".

## Repro

1. Have multiple workspaces with Claude Code sessions that have finished their work and are waiting for input
2. Look at the sidebar status for each workspace
3. Some show "Idle" while others show "Needs input" — even though they're all in the same state

## Expected behavior

Workspaces in the same state should show the same status indicator. If a Claude Code session is waiting for user input, it should consistently show either "Needs input" or "Idle" — not a mix of both.

## Actual behavior

The status detection logic classifies identical states differently across workspaces. See screenshot:

- "New Markets" → Idle
- "General" → Idle  
- "legal" → Needs input

All three have a Claude Code session that finished and is waiting at the prompt.

## Related

- https://github.com/manaflow-ai/cmux/issues/1052 — sidebar badges don't refresh on state change
- https://github.com/manaflow-ai/cmux/issues/1238 — "Running" status not shown when agent is active

kyleawayan:
Maybe related comment I added in https://github.com/manaflow-ai/cmux/issues/1027#issuecomment-4173419253
> I think this also can be due to the notification hook handling in cmux, firing for all `Notification` hook types? https://code.claude.com/docs/en/hooks#notification -- I see `idle_prompt` is one, which I assume fires when you leave the Claude Code instance inactive. So it could go from Claude finishing its tasks -> Idle state in cmux -> then 5 minutes later, `idle_prompt` fires -> since cmux doesn't check the `Notification` hook type, "Needs input" shows up instead, even though theres no input needed most likely.

scarere:
It would also be useful to be able to manually mark sessions as idle.

Issue #1027, opened 2026-03-06
Sidebar 'Running' and 'Needs Input' status indicators are flaky and unreliable


austinywang:
## Bug

The sidebar status indicators ("Running" and "Needs Input") are unreliable:

1. **Shows "Running" when not actually running** — the indicator stays lit even after the agent has stopped
2. **Shows "Needs Input" when there's no input needed** — false positives
3. **"Needs Input" doesn't show up when it should** — the agent is actually waiting for input but the sidebar doesn't reflect it

The behavior is intermittent/flaky — sometimes it works correctly, sometimes it doesn't.

## Expected behavior

- "Running" should only appear when the agent process is actively executing
- "Needs Input" should reliably appear whenever the agent is blocked waiting for user input
- Both indicators should clear promptly when the state changes

## Actual behavior

The indicators appear to be out of sync with the actual terminal process state. The detection seems unreliable — sometimes correct, sometimes wrong in either direction.

kyleawayan:
I think this also can be due to the notification hook handling in cmux, firing for all `Notification` hook types? https://code.claude.com/docs/en/hooks#notification -- I see `idle_prompt` is one, which I assume fires when you leave the Claude Code instance inactive. So it could go from Claude finishing its tasks -> Idle state in cmux -> then 5 minutes later, `idle_prompt` fires -> since cmux doesn't check the `Notification` hook type, "Needs input" shows up instead, even though theres no input needed most likely.

scarere:
Also want to add to this that a common mistake is that when the main agent is waiting for a subagent to return with results (such as a background research or explore agent), cmux interprets this as the agent needing input or attention and sends a notification.

AuthorOfTheSurf:
This is pretty much the main feature that I need from cmux

I see `Claude is waiting for your input` and ` ⚡ Running` linger around indefinitely

Unfortunate because cmux is doing a better job than my other enhanced-terminals at notifications and notification sounds from Claude

I am also liking the vertical tab design quite a lot. Day 1 user here and appreciate all the good work!

Eibich:
> This is pretty much the main feature that I need from cmux
> 
> I see `Claude is waiting for your input` and ` ⚡ Running` linger around indefinitely
> 
> Unfortunate because cmux is doing a better job than my other enhanced-terminals at notifications and notification sounds from Claude
> 
> I am also liking the vertical tab design quite a lot. Day 1 user here and appreciate all the good work!

Some here.

[...]

PR #2845, opened 2026-04-12
Distinguish idle_prompt from real 'Needs input' in Claude hooks

### LINKED ISSUE #2576: Native Claude detection should distinguish 'Needs Input' vs 'Idle' states
## Description

When cmux natively detects a Claude Code session, the sidebar shows "Needs Input" with a blue notification ring when Claude is waiting for user input. This is great — but there's no distinction between two semantically different states:

1. **Needs Input** — Claude finished a task and is actively waiting for the user to type. This is attention-required — the user needs to respond for work to continue.
2. **Idle** — Claude finished and the session is just sitting there. No urgency. The user might have moved on to another workspace.

Currently both states appear the same in the sidebar. When running multiple Claude sessions across workspaces, it's hard to tell which ones need attention vs which are just parked.

## Proposed Behavior

| State | Sidebar Display | Ring/Glow |
|---|---|---|
| Running | ⚡ Running (current behavior) | None |
| Needs Input | 🔔 Needs Input (blue) | Blue ring/glow |
| Idle | ⏸ Idle (neutral) | None |

The transition would be:
- User submits prompt → **Running**
- Claude finishes responding → **Needs Input** (blue ring fires, clears when workspace is focused)
- After workspace is focused/acknowledged → **Idle** (neutral, no ring)

## Why This Matters

With `cmux claude-teams` spawning multiple agent sessions, or just having several Claude workspaces open, the sidebar becomes a dashboard. Being able to glance at it and know "workspace 3 needs me, workspaces 1 and 5 are just idle" is the difference between efficient multi-agent workflows and constantly checking each tab.

## Related Issues

- #2541 — shell integration not injected in claude-teams mode (no status at all)
- #2229 — no notifications in claude-teams mode

This issue is about the semantic model even when native detection *is* working (plain `claude` sessions).

## Environment

- cmux 0.63.1
- macOS 15
- Claude Code v2.1.92

Audiofool934:
## Summary

Claude Code's \`Notification\` hook fires for multiple reasons per the [official docs](https://code.claude.com/docs/en/hooks#notification):

- \`permission_prompt\` — Claude needs user approval
- \`elicitation_dialog\` — MCP server requests user input
- \`idle_prompt\` — session has been idle for a while (fires ~5 minutes of inactivity)
- \`auth_success\` — authentication completed

The existing hook handler hardcoded every fired \`Notification\` as \"Needs input\", so after Claude finished a task and sat idle, the \`idle_prompt\` event re-raised a false \"Needs input\" badge in the sidebar. This is the root cause reported in several issues.

This PR parses the authoritative \`notification_type\` field and routes each case correctly:
- \`permission_prompt\` / \`elicitation_dialog\` → \"Needs input\" (bell, blue)
- \`idle_prompt\` → \"Idle\" (pause, gray)
- \`auth_success\` → no status change
- Unknown types → fall back to keyword classification (existing behavior)

Also tightens the keyword fallback so bare \`idle_prompt\` mentions no longer bleed into the \"Waiting\" bucket.

## Test plan

Manual reproduction against the released 0.63.1 CLI:

\`\`\`bash
# idle_prompt — should NOT trigger \"Needs input\" but does on 0.63.1
echo '{\"hook_event_name\":\"Notification\",\"notification_type\":\"idle_prompt\",\"message\":\"Claude waited for your input\",\"session_id\":\"test\"}' \\
  | /Applications/cmux.app/Contents/Resources/bin/cmux claude-hook notification --workspace 1
# -> Before this PR: sidebar shows \"Needs input\" (bug)
# -> After this PR:  sidebar shows \"Idle\"

# permission_prompt — should still trigger \"Needs input\"
echo '{\"hook_event_name\":\"Notification\",\"notification_type\":\"permission_prompt\",\"message\":\"Permission needed\",\"session_id\":\"test\"}' \\
  | cmux claude-hook notification --workspace 1
# -> Sidebar shows \"Needs input\" (correct, unchanged)
\`\`\`

Verified locally by comparing the 0.63.1 installed CLI against a debug build with this patch against the same running app.

## Related issues

- Relates to #1027 (Running/Needs Input indicators flaky)
- Relates to #1492 (inconsistent Idle vs Needs Input)
- Closes #2576 (distinguish Needs Input vs Idle states)

---
## Summary by cubic
Fixes false “Needs input” badges by using `notification_type` in Claude `Notification` hooks. Sidebar shows “Idle” after inactivity, keeps the current pill on completion/auth, and only asks for input when action is required.

- **Bug Fixes**
  - Map `permission_prompt`/`elicitation_dialog` → “Needs input”; `idle_prompt` → “Idle”; `auth_success` and completion → no status change; preserve the existing pill and reuse a saved message body.
  - Defer the “needs input” placeholder until after classification so idle/auth/completion don’t get conflicting generic bodies.
  - Improve fallback: include `notification_type` in the signal and tighten keywords so idle text doesn’t trigger “Needs input”.

- **Refactors**
  - Rename status kind from `.none` to `.noStatus` for clarity.

<sup>Written for commit 80234f87ed2ffa1afb8b4504cfdfe9dc5b9df641. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Improvements**
  * Notification classification now prefers authoritative notification fields with a reliable keyword fallback, mapping notifications to "Needs input", "Idle", or default states.
  * Detected status is preserved when reusing saved session content.
  * Visuals updated: "Needs input" (bell), "Idle" (pause), or default.

* **Bug Fixes**
  * Sidebar status is only updated for actionable states, preventing unintended overwrites.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -12600,9 +12600,9 @@ struct CMUXCLI {
                     icon: "pause.circle",
                     color: "#8E8E93"
                 )
-            case .none:
-                // auth_success and other non-status events should not clobber
-                // the existing running/idle state.
+            case .noStatus:
+                // auth_success, completion, and other non-attention events
+                // should not clobber the existing running/idle pill state.
                 break
             }
             print(response)
@@ -13301,7 +13301,7 @@ struct CMUXCLI {
     enum ClaudeNotificationStatusKind {
         case needsInput  // permission_prompt, elicitation_dialog, unknown attention-class messages
         case idle        // idle_prompt — session parked, no action required
-        case none        // auth_success and other non-status events
+        case noStatus    // auth_success, completion, and other events that should not clobber the existing pill
     }
 
     private func summarizeClaudeHookNotification(parsedInput: ClaudeHookParsedInput) -> (subtitle: String, body: String, statusKind: ClaudeNotificationStatusKind) {
@@ -13316,10 +13316,13 @@ struct CMUXCLI {
         // The authoritative Notification hook type field per Claude Code docs.
         let notificationType = firstString(in: object, keys: ["notification_type"])
             ?? firstString(in: nested, keys: ["notification_type"])
+        // Include notification_type in signalParts so unknown-but-classifiable
+        // type values (e.g. future values not yet enumerated below) still
+        // influence the keyword-based fallback instead of being dropped.
         let signalParts = [
             firstString(in: object, keys: ["event", "event_name", "hook_event_name", "type", "kind"]),
-            firstString(in: object, keys: ["matcher", "reason"]),
-            firstString(in: nested, keys: ["type", "kind", "reason"])
+            firstString(in: object, keys: ["notification_type", "matcher", "reason"]),
+            firstString(in: nested, keys: ["notification_type", "type", "kind", "reason"])
         ]
         let messageCandidates = [
             firstString(in: object, keys: ["message", "body", "text", "prompt", "error", "description"]),
@@ -13354,7 +13357,7 @@ struct CMUXCLI {
             case "idle_prompt":
                 return ("Idle", message.isEmpty ? "Session idle" : message, .idle)
             case "auth_success":
-                return ("Auth", message.isEmpty ? "Authentication successful" : message, .none)
+                return ("Auth", message.isEmpty ? "Authentication successful" : message, .noStatus)
             default:
                 break  // fall through to keyword-based classification
             }
@@ -13368,7 +13371,10 @@ struct CMUXCLI {
             return ("Error", message.isEmpty ? "Claude reported an error" : message, .needsInput)
         }
         if lower.contains("complet") || lower.contains("finish") || lower.contains("done") || lower.contains("success") {
-            return ("Completed", message.isEmpty ? "Task completed" : message, .needsInput)
+            // Completion events should not clobber the current status with a
+            // blue "Needs input" bell; the Stop hook is responsible for moving
+            // the pill to Idle.
+            return ("Completed", message.isEmpty ? "Task completed" : message, .noStatus)
         }
         if lower.contains("idle_prompt") || lower.contains("idle prompt") {
             return ("Idle", message.isEmpty ? "Session idle" : message, .idle)
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -13328,7 +13328,12 @@ struct CMUXCLI {
             firstString(in: object, keys: ["message", "body", "text", "prompt", "error", "description"]),
             firstString(in: nested, keys: ["message", "body", "text", "prompt", "error", "description"])
         ]
-        let message = messageCandidates.compactMap { $0 }.first ?? "Claude needs your input"
+        // Keep the message empty when the payload doesn't supply one, so the
+        // per-type body defaults below ("Session idle", "Authentication
+        // successful", etc.) are chosen instead of the generic needs-input
+        // placeholder. Only fall back to the placeholder when the classifier
+        // actually produces a needs-input event with no body of its own.
+        let message = messageCandidates.compactMap { $0 }.first ?? ""
         let normalizedMessage = normalizedSingleLine(message)
         let signal = signalParts.compactMap { $0 }.joined(separator: " ")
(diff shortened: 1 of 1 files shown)


Audiofool934:
Thanks for the review. Addressed the empty-message edge case in 80234f8 — for empty payloads, the message now stays empty through classification so per-type defaults ('Session idle', 'Authentication successful', etc.) get used instead of the generic needs-input placeholder.

Re: the 'quote multi-word status label' comment — this is a non-issue in practice. `upsertSidebarMetadata` in `TerminalController.swift:14618` rejoins `parsed.positional[1...]` with spaces, so `set_status claude_code Needs input` is correctly reassembled on the server side. The 'Needs input' label has shipped unquoted in 0.63.1 and prior without issues.

epaynter:
Sharing findings from instrumenting this locally, in case useful for this PR.

The only built-in tools that block on the user *without* firing `Notification`/`PermissionRequest` are `AskUserQuestion` and `ExitPlanMode` — they fire just `PreToolUse` (before) and `PostToolUse` (after the answer), nothing during the block. Per the Claude Code hook docs that's the complete "silent blocking" set; permission-gated tools already surface correctly via `PermissionRequest`, and `idle_prompt` covers turn-end. So the gap is specifically those two tools showing "Running" while actually blocked.

One design note: bracketing the block with a custom hook (`PreToolUse`{AskUserQuestion|ExitPlanMode} → set, `PostToolUse`+`Stop` → clear) gets the *detection* right. But a `set-status` key with high `--priority` renders **stacked above** the internal `claude_code=Running` key rather than replacing it (orange "Needs input" over blue "Running" — both visible). So whatever the native fix does, it should *suppress/replace* the internal `claude_code` state during the block, not just add a higher-priority status, or the stale "Running" stays visible underneath.

Pawel-Kica:
Found a workaround that works right now on v0.64.13, no patch needed.

cmux already has a filter for this: `notifications.hooks` (https://cmux.com/docs/notifications). It runs before a notification shows and lets you kill the effects. It only sees the text, but that's enough to tell them apart:

- the annoying one: body `"Claude is waiting for your input"`
- permission: body `"Claude needs your permission"`
- turn done: subtitle `"Completed in <workspace>"`

So you match on the idle body and drop just that. Drop this in `~/.config/cmux/cmux.json` (you need `jq`):

```json
{
  "notifications": {
    "hooks": [
      {
        "id": "mute-idle-waiting",
        "command": "jq -c 'if ((.notification.body // \"\") | test(\"waiting for your input\"; \"i\")) then .effects.desktop=false | .effects.sound=false | .effects.paneFlash=false | .effects.markUnread=false | .effects.reorderWorkspace=false else . end'"
      }
    ]
  }
}
```

Hot-reloads on save. The idle banner, sound, flash, unread ring, and reorder all go away, but permission prompts and completions still come through. And it fails open, so if the hook breaks you just get the banner back.

Not a real fix since it matches on text not `notification_type`, but it unblocks you today.

pnientiedt:
👍 Confirming this is still needed and reproduces on current released versions — finished sessions get re-flagged as "Needs input" ~60s after they go idle.

**Environment**
- cmux `0.64.17 (97)` (latest release)
- Claude Code `2.1.207`
- macOS (Darwin 25.5.0), Apple Silicon

**Repro (from `~/.cmuxterm/events.jsonl`)** — every turn shows Stop correctly firing, then Claude's idle notification ~61s later flipping the badge back to blue "Needs input":
```
21:48:13  Stop          → session finished, goes Idle (correct)
21:49:14  Notification  → ~61s later → re-flagged "Needs input" ❌
21:53:10  Stop          → finished, Idle
21:54:11  Notification  → ~61s later → "Needs input" ❌
```

**Extra data point:** in the recorded `agent.hook.Notification` events, the payload has no discriminator at all — `notification_type: null`, `message: null`, `redacted: null`. So on top of routing `idle_prompt` → Idle, the handler needs a sane fallback for when neither `notification_type` nor `message` is present (right now that empty case falls through to "Needs input", which is the worst default). Worth confirming whether the field is actually being read from the hook stdin before classification.

Note this can't be worked around from the user side: cmux injects its Notification hook via `--settings` on the `claude` launch, so a user-defined `idle_prompt` matcher hook can't suppress it (as also noted in #3602).

PR #2847, opened 2026-04-12
Fix paste only pasting first character (#2805)

### LINKED ISSUE #2805: Copy stopped working with Open Code cli
### cmux version

0.63.2

### macOS version

26.4.1

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Bug description

Image
When i paste it , it only paste the first character 

Image

Its not a opencode issue , its a cmux issue , it was working okay 1-2 weeks ago until i updated cmux. 

The paste works well if you paste it say on a notepad app , or browser app , but it just cant paste into CMUX , i literally have to paste it on a notepad then copy and paste it back into cmux 

### Expected behavior

Copy to work well , to copy on opencode you actually drag on the text and it auto copies.

### Steps to reproduce

1. Install OpenCode
2. Try to copy something and paste it  WITHIN THE SAME APP

### Shell and environment

_No response_

```text

```

_No response_

### Additional context

_No response_

dezren39:
## Summary

- Fixes #2805: Cmd+V paste was only pasting the first character
- Root cause: Ghostty upstream made `paste_from_clipboard` a **performable** binding, which caused `performKeyEquivalent` to skip the menu path (Edit > Paste) and fall through to `keyDown`, double-firing the clipboard request
- Fix: remove the `!isPerformable` guard in the menu-first condition so performable bindings like paste still route through the menu, restoring pre-upgrade behavior

## Changes

**`Sources/GhosttyTerminalView.swift`** — one condition change in `performKeyEquivalent`:
- Removed `!isPerformable` from the menu-first gate so consumed bindings (including performable ones) try the menu before falling through to `keyDown`
- Added comments explaining why performable bindings need the menu path

## Testing

- Built tagged debug app (`cmux DEV fix-paste.app`)
- Verified Cmd+V now pastes full clipboard contents correctly

---
## Summary by cubic
Fixes Cmd+V only pasting the first character by sending performable paste through the Edit menu first to avoid double clipboard requests. Fixes #2805.

- **Bug Fixes**
  - In `performKeyEquivalent`, removed the `!isPerformable` check so Cmd+V routes through Edit > Paste; if the menu doesn’t handle it, it falls back to `keyDown`.

<sup>Written for commit 05ab5943118a77b05d46690ac10b01a258401435. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard binding behavior for terminal actions. Performable keybindings (such as paste_from_clipboard) now reliably execute through both menu routing and direct action dispatch, enhancing overall responsiveness.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -6398,12 +6398,19 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
             let isPerformable = (bindingFlags.rawValue & GHOSTTY_BINDING_FLAGS_PERFORMABLE.rawValue) != 0
 
             // If the binding is consumed and not meant for the menu, allow menu first.
-            if isConsumed && !isAll && !isPerformable && keySequence.isEmpty && keyTables.isEmpty {
+            // Performable bindings (e.g. paste_from_clipboard) also need the menu
+            // path so that Edit > Paste handles Cmd+V instead of keyDown double-
+            // firing the clipboard request through both interpretKeyEvents and
+            // ghostty_surface_key.
+            if isConsumed && !isAll && keySequence.isEmpty && keyTables.isEmpty {
                 if let menu = NSApp.mainMenu, menu.performKeyEquivalent(with: event) {
                     return true
                 }
             }
 
+            // For performable bindings where the menu didn't handle the event,
+            // fall through to keyDown so Ghostty can perform the action directly
+            // (e.g. paste when no menu item exists).
             keyDown(with: event)
             return true
         }


shaun0927:
One performance question on the `!isPerformable` guard removal at `Sources/GhosttyTerminalView.swift:6505`: with the guard gone, every consumed non-global Ghostty binding traverses the menu first (`NSMenu.performKeyEquivalent` walk + `validateUserInterfaceItem` probe per matching menu item). For `paste_from_clipboard` this is fine because it is amortised over user-paced paste, but is there a chance future PERFORMABLE Ghostty bindings would hit this on every keystroke? If so, narrowing the new path back to the specific bindings that need menu routing might be cheaper than auditing each new binding.

Asking because the typing-latency policy in CLAUDE.md is strict and this hunk widened the menu probing slightly past the documented motivation (paste-only).

PR #2855, opened 2026-04-13
Fix macOS modifier desync causing idle terminal input corruption

### LINKED ISSUE #2853: Terminal input corrupts to raw escape sequences after idle
## Bug

After leaving a terminal tab open and idle for a while, keyboard input becomes corrupted. Typing normal keys produces garbage output like raw escape sequences instead of the actual characters typed.

### Example output

```
zsh: command not found: 1:3u14
zsh: command not found: 1:3u
c9;1:3ue01;1:3u;1:3u
```

### Analysis

The `1:3u` fragments are kitty keyboard protocol CSI sequences (`CSI <keycode> ; <modifiers> u`). This suggests the terminal's keyboard protocol state is getting desynchronized after idle — Ghostty uses the kitty keyboard protocol, and something is causing the raw protocol bytes to leak through to the shell instead of being decoded.

Possible causes:
- A background process or keepalive mechanism is sending a sequence that corrupts the terminal state
- The kitty keyboard protocol mode is being toggled off/on incorrectly after the terminal resumes from idle
- A PTY read/write race condition after the terminal has been idle causes partial escape sequence reads
- The terminal emulator's state machine loses sync and starts passing raw CSI bytes to the shell

### Repro

1. Open a cmux terminal tab
2. Leave it idle for an extended period (exact duration unclear, but "a while")
3. Try typing normal commands
4. Observe corrupted input with raw escape sequence fragments

### Expected

Normal keyboard input regardless of idle duration.

austinywang:
## Summary
- fix macOS `flagsChanged` handling so modifier-only events emit correct press/release edges
- use explicit left and right device masks for Shift, Control, Option, and Command so releasing one side while the other stays held does not leave Ghostty desynced
- keep modifier-only `flagsChanged` events off AppKit character APIs to avoid crashes on non-text events
- add unit coverage for modifier edge detection used by the terminal bridge

## Testing
- ./scripts/reload.sh --tag issue-2853-terminal-input-corruption-idle
- ./scripts/reload.sh --tag issue-2853-terminal-input-corruption-idle --launch

Fixes #2853

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate detection of modifier key press vs. release for Shift, Control, Option, Command and Caps Lock.
  * Stops synthesizing modifier events when IME marked text is present, improving IME compatibility and preventing spurious modifier input.

* **Tests**
  * Added unit tests covering modifier key press/release behaviors for left/right keys and Caps Lock.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -49,6 +49,49 @@ private func cmuxTransparentWindowBaseColor() -> NSColor {
     // avoids visual artifacts that can happen with a fully clear window background.
     NSColor.white.withAlphaComponent(0.001)
 }
+
+// `flagsChanged` is used for both modifier presses and releases on macOS.
+// Returning the wrong edge leaves Ghostty with a phantom held modifier until
+// a later focus loss flushes release events into the PTY.
+func cmuxGhosttyModifierActionForFlagsChanged(
+    keyCode: UInt16,
+    modifierFlagsRawValue: UInt
+) -> ghostty_input_action_e? {
+    let flags = NSEvent.ModifierFlags(rawValue: modifierFlagsRawValue)
+    let modifierActive: Bool
+    switch keyCode {
+    case 0x39:
+        modifierActive = flags.contains(.capsLock)
+    case 0x38, 0x3C:
+        modifierActive = flags.contains(.shift)
+    case 0x3B, 0x3E:
+        modifierActive = flags.contains(.control)
+    case 0x3A, 0x3D:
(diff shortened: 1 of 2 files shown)


austinywang:
Addressed the review feedback on this branch:

- added explicit left-side device-mask handling so releasing left Shift/Control/Option/Command while the right-side peer remains held now emits a release instead of a phantom press
- expanded modifier-edge tests to cover the left/right crossover case plus right Control and right Option
- kept `flagsChanged` on a modifier-only event path so it no longer probes AppKit character APIs for non-text events

Latest branch tip: `151075f7`.

PR #2877, opened 2026-04-14
fix: prefer UTF-8 plain text in pasteboard to avoid Mac OS Roman loss

### LINKED ISSUE #2818: Cmd+V pastes question marks instead of Cyrillic (non-Latin) text
## Bug

Pasting Cyrillic (Russian) text via Cmd+V replaces all non-Latin characters with `?`. The clipboard contains valid UTF-8 — confirmed via `pbpaste` and hex dump.

**Version:** cmux 0.63.2 (79) [179b16ce6]
**OS:** macOS, locale `en_GB.UTF-8`
**Worked before:** yes, broke after a recent update

## Reproduction

1. Copy Russian text from Telegram Desktop (or any Qt app)
2. Cmd+V into cmux terminal
3. All Cyrillic characters appear as `?`

`pbpaste` in the same terminal shows correct UTF-8 text.

## Root cause

Telegram (Qt) places these types on the pasteboard, in this order:

```
com.apple.traditional-mac-plain-text  (9 bytes)   ← Mac OS Roman, NO Cyrillic support
CorePasteboardFlavorType 0x54455854   (9 bytes)
public.utf16-plain-text               (18 bytes)
CorePasteboardFlavorType 0x75747874   (18 bytes)
public.utf8-plain-text                (18 bytes)  ← correct UTF-8
NSStringPboardType                    (18 bytes)
public.text                           (18 bytes)
com.trolltech.anymime.text--plain     (18 bytes)
```

Hex dump proof:

```
traditional-mac-plain-text: 0x3F 0x3F 0x3F 0x3F 0x3F 0x3F 0x3F 0x3F 0x3F  →  "?????????"
public.utf8-plain-text:     0xD1 0x81 0xD0 0xBB 0xD0 0xB5 ...              →  "следующий"
```

`plainTextContents()` in `GhosttyPasteboardHelper` iterates `pasteboard.types` **in order** and returns the first match. `com.apple.traditional-mac-plain-text` passes `isPlainTextType()` (because `UTType` conforms to `.plainText`), and `pasteboard.string(forType:)` returns a non-empty string of `?` — so it's returned before `public.utf8-plain-text` is ever checked.

## Suggested fix

In `plainTextContents(from:)`, check `utf8PlainTextType` first before iterating:

```swift
private static func plainTextContents(from pasteboard: NSPasteboard) -> String? {
    // Prefer UTF-8 to avoid lossy encodings like Mac OS Roman,
    // which replaces non-Latin characters (e.g. Cyrillic) with '?'
    if let value = pasteboard.string(forType: utf8PlainTextType), !value.isEmpty {
        return value
    }

    for type in pasteboard.types ?? [] {
        guard type != utf8PlainTextType else { continue }
        guard isPlainTextType(type) else { continue }
        guard let value = pasteboard.string(forType: type), !value.isEmpty else { continue }
        return value
    }
    return nil
}
```

This affects all non-Latin scripts (Cyrillic, CJK, Arabic, etc.) when copied from apps that place `com.apple.traditional-mac-plain-text` before `public.utf8-plain-text` on the pasteboard.

dasanworld:
## Summary

Fixes #2818 — Cmd+V replaces non-Latin characters (Korean, Chinese, Japanese, Cyrillic, Arabic, …) with `?` when pasting from Qt-based apps such as Telegram Desktop.

## Root Cause

`GhosttyPasteboardHelper.plainTextContents(from:)` iterated `pasteboard.types` in declaration order and returned the first value that satisfied `isPlainTextType(_:)`.

Qt on macOS registers pasteboard types in this order:

1. `com.apple.traditional-mac-plain-text` ← **Mac OS Roman encoding** (no support for non-Latin scripts)
2. `public.utf8-plain-text`
3. `public.utf16-external-plain-text`

The helper therefore read the Mac OS Roman value, converting every non-Latin character to `?`.

## Fix

Explicitly probe `public.utf8-plain-text` and `.string` (UTF-8 on modern macOS) **first** before falling back to the general loop. The fallback loop is preserved so apps that publish only a non-preferred plain-text UTType (e.g. `public.plain-text`, `NSFilenamesPboardType`) keep working — the existing `testAlternatePlainTextUTIExtractsPlainText` case still passes.

```swift
private static func plainTextContents(from pasteboard: NSPasteboard) -> String? {
    let allTypes = pasteboard.types ?? []

    for preferred in [utf8PlainTextType, NSPasteboard.PasteboardType.string] {
        guard allTypes.contains(preferred) else { continue }
        guard let value = pasteboard.string(forType: preferred), !value.isEmpty else { continue }
        return value
    }

    for type in allTypes {
        if type == utf8PlainTextType || type == .string { continue }
        guard isPlainTextType(type) else { continue }
        guard let value = pasteboard.string(forType: type), !value.isEmpty else { continue }
        return value
    }

    return nil
}
```

## Tests

Two new regression cases in `GhosttyPasteboardHelperTests`:

- **`testPrefersUTF8PlainTextOverLegacyMacRomanType`** — builds a pasteboard that declares `com.apple.traditional-mac-plain-text` **before** `public.utf8-plain-text` (mirroring Qt), writes a mangled ASCII value to the legacy slot and the correct Korean string `삼성전자 거래량 미충족` to the UTF-8 slot, and asserts `cmuxPasteboardStringContentsForTesting` returns the Korean original.
- **`testFallsBackWhenOnlyNonPreferredPlainTextTypePresent`** — declares only the legacy Mac OS Roman type and asserts the helper still returns the string (fallback-loop coverage, no regression for legacy-only pasteboards).

All existing tests in the file are unchanged.

## Out of scope

This PR only fixes the read path. The write path (`writeString`) already uses `.string`, which is UTF-8 on modern macOS, so no change is needed there. The Ghostty submodule is not touched.

## Test plan

- [ ] CI passes (new regression tests + existing suite)
- [ ] Manually verify: copy Korean text from Telegram Desktop → paste into a cmux pane → characters arrive intact
- [ ] Manually verify: copy from Safari / VS Code / Notes (non-Qt apps) → paste still works (they don't publish the legacy type, fallback path unchanged)

---
## Summary by cubic
Prefer UTF‑8 plain text from the macOS pasteboard to avoid Mac OS Roman corruption when pasting from Qt-based apps (e.g., Telegram Desktop). Non‑Latin characters now paste correctly. Fixes #2818.

- **Bug Fixes**
  - Update `GhosttyPasteboardHelper.plainTextContents(from:)` to read `public.utf8-plain-text` and `.string` first, then fall back to other plain-text types.
  - Keep fallback so legacy-only types still work.
  - Add regression tests for preferring UTF‑8 over `com.apple.traditional-mac-plain-text` and for legacy-only fallback.

<sup>Written for commit 4d49a7bae5781271a33b4d5e4849c9c4e07b6909. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced text paste functionality to correctly prioritize UTF-8 plain text formats, improving how text is handled when multiple text types are available on the clipboard.

* **Tests**
  * Added regression tests for clipboard plain-text format selection, including coverage for scenarios with both modern UTF-8 and legacy text format types.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -264,7 +264,22 @@ enum GhosttyPasteboardHelper {
     }
 
     private static func plainTextContents(from pasteboard: NSPasteboard) -> String? {
-        for type in pasteboard.types ?? [] {
+        let allTypes = pasteboard.types ?? []
+
+        // Prefer UTF-8 plain text whenever available. Some apps — notably
+        // Qt-based ones like Telegram Desktop — register
+        // `com.apple.traditional-mac-plain-text` (Mac OS Roman, which cannot
+        // represent non-Latin scripts) *before* the UTF-8 variants. Iterating
+        // `pasteboard.types` in order then returns a lossy value where every
+        // non-Latin character becomes "?". Fixes #2818.
+        for preferred in [utf8PlainTextType, NSPasteboard.PasteboardType.string] {
+            guard allTypes.contains(preferred) else { continue }
+            guard let value = pasteboard.string(forType: preferred), !value.isEmpty else { continue }
+            return value
+        }
+
+        for type in allTypes {
+            if type == utf8PlainTextType || type == .string { continue }
             guard isPlainTextType(type) else { continue }
             guard let value = pasteboard.string(forType: type), !value.isEmpty else { continue }
             return value
--- a/cmuxTests/TerminalAndGhosttyTests.swift
+++ b/cmuxTests/TerminalAndGhosttyTests.swift
@@ -49,6 +49,49 @@ final class GhosttyPasteboardHelperTests: XCTestCase {
         )
     }
 
+    /// Regression test for https://github.com/manaflow-ai/cmux/issues/2818 —
+    /// Qt-based apps (Telegram Desktop, etc.) register the legacy
+    /// `com.apple.traditional-mac-plain-text` type (Mac OS Roman encoding,
+    /// no CJK/Cyrillic/Arabic support) *before* UTF-8. Iterating the
+    /// pasteboard types in order used to return the lossy legacy value,
+    /// mangling every non-Latin character into "?". The helper must
+    /// prefer UTF-8 whenever it is also present on the pasteboard.
+    func testPrefersUTF8PlainTextOverLegacyMacRomanType() {
+        let pasteboard = NSPasteboard(name: .init("cmux-test-utf8-priority-\(UUID().uuidString)"))
+        pasteboard.clearContents()
+
+        let koreanText = "삼성전자 거래량 미충족"
+        let legacyType = NSPasteboard.PasteboardType("com.apple.traditional-mac-plain-text")
+        let utf8Type = NSPasteboard.PasteboardType("public.utf8-plain-text")
+
+        // Order matters: declare legacy FIRST to mirror Qt's behaviour.
+        pasteboard.declareTypes([legacyType, utf8Type], owner: nil)
+        pasteboard.setString("?? ??? ???", forType: legacyType)
+        pasteboard.setString(koreanText, forType: utf8Type)
+
+        XCTAssertEqual(
+            cmuxPasteboardStringContentsForTesting(pasteboard),
+            koreanText
+        )
+    }
+
+    /// Fallback-loop coverage: when *only* a legacy / unknown plain-text
+    /// type is present and no UTF-8 variant exists, the helper should still
+    /// return whatever string the pasteboard does expose (best-effort).
+    func testFallsBackWhenOnlyNonPreferredPlainTextTypePresent() {
+        let pasteboard = NSPasteboard(name: .init("cmux-test-only-legacy-\(UUID().uuidString)"))
+        pasteboard.clearContents()
+
+        let legacyType = NSPasteboard.PasteboardType("com.apple.traditional-mac-plain-text")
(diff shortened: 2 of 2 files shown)


dasanworld:
### Standalone verification on live `NSPasteboard` (macOS 15.3)

Could not run `xcodebuild test` on my machine (no Xcode / Zig toolchain for the ghostty submodule), so I extracted the buggy and fixed implementations into a 120-line `swift` script that drives a real `NSPasteboard`. Six scenarios, all pass:

```
✅ Qt-style: legacy→utf8 (Korean)
   buggy : ?? ?,???? MA??0.2% ???0.9? ???:???
   fixed : 삼성전자 9,290원 MA거리0.2% 거래량0.9배 미충족:거래량

✅ Qt-style: legacy→utf8 (Russian/Cyrillic)
   buggy : ??????
   fixed : Москва — столица России

✅ Qt-style: legacy→utf8 (Japanese)
   buggy : ???????
   fixed : こんにちは世界

✅ Normal: utf8 only (ASCII)           — no regression
✅ Alternate UTType only (fallback)    — existing test still green
✅ Legacy only (fallback)              — best-effort path preserved
```

The bug reproduces 1:1 with Qt's pasteboard ordering (`com.apple.traditional-mac-plain-text` declared before `public.utf8-plain-text`), and the patched function recovers every sample string intact. The fallback loop retains behavior for pasteboards that don't publish a UTF-8 variant.

Happy to rerun against any additional scenario / edge case you'd like covered.

PR #2946, opened 2026-04-17
[codex] Restore terminal redraw after notifications and workspace selection

### LINKED ISSUE #2555: new workspace terminal doesn't render until tab switch — first metal frame deferred by main thread congestion
## bug

when creating a new workspace (cmd+t), the terminal area stays blank — no cursor, no prompt, nothing renders. the shell is actually running (you can type blind and it executes), but the terminal surface is invisible. it stays this way for 10+ seconds or indefinitely until you manually switch to another workspace and switch back.

this has been driving me crazy for weeks. i thought it was my shell config, my ghostty settings, or the git probe delays — spent hours profiling every component. turns out the shell starts in ~150ms. the terminal just doesn't draw.

## steps to reproduce

1. run cmux for a few hours with 5+ workspaces open (sidebar visible, branch/pr details showing)
2. press cmd+t to create a new workspace
3. observe: terminal area is blank, no prompt visible
4. switch to a different workspace (cmd+1 or click sidebar)
5. switch back to the new workspace
6. terminal now renders instantly — prompt is there, shell was running the whole time

harder to reproduce on a fresh cmux launch with low cpu. becomes very consistent after a few hours of use when idle cpu is 30%+.

## expected behavior

the terminal should render its first frame immediately when a new workspace is created and selected.

## actual behavior

the ghostty metal surface gets created but never receives its first `display()` call. the `WindowTerminalPortal` doesn't synchronize the hosted view's geometry until a second layout pass is triggered (e.g. by switching workspaces). the shell and pty are fully functional — just invisible.

## root cause analysis

i traced this through the source code. here's what happens:

`TabManager.addWorkspace()` does three things in rapid succession on the main thread:
1. `tabs = updatedTabs` — mutates `@Published`, fires `objectWillChange`
2. `selectedTabId = newWorkspace.id` — another `@Published` mutation
3. `NotificationCenter.post(.ghosttyDidFocusTab)` — triggers 6+ `.onReceive` handlers in ContentView

each of these cascades through SwiftUI view invalidations. `Workspace` has 30+ `@Published` properties, and the sidebar's `sidebarObservationPublisher` merges 15+ combine streams. the initial layout pass for the new terminal view gets deferred because the run loop is saturated processing these cascades.

the `HostContainerView.onGeometryChanged` → `WindowTerminalPortal.synchronizeForAnchor()` path only fires when the view is re-presented (i.e. switching away and back), which is why that workaround works.

after hours of uptime, the constant `@Published` mutations from git/pr polling (every 5s for selected workspace, every 30s for all) keep the main thread at 30-40% idle cpu, making the deferral worse.

## workaround

```
cmux set-hook after-new-workspace 'cmux refresh-surfaces'
cmux set-hook after-new-split 'cmux refresh-surfaces'
```

this forces `synchronizeLayoutHierarchy()` after workspace creation, which triggers the first metal frame. completely fixes the issue — new tabs render instantly.

## suggested fix

the hook shouldn't be necessary. a few options:

1. call `synchronizeLayoutHierarchy()` at the end of `addWorkspace()` after the run loop settles (e.g. `DispatchQueue.main.async` one cycle later)
2. schedule `CAMetalLayer.setNeedsDisplay()` on the new surface's layer after insertion
3. debounce the `@Published` mutations during workspace creation so the main thread isn't starved during the initial layout pass

## environment

- cmux 0.63.1 (78), stable channel
- macOS 26.3.1 (Tahoe)
- apple m5
- typical session: 5-10 workspaces, mix of local and ssh tabs, sidebar visible with branch/pr details

## related issues

- #230 — app hanging 2s+ from swiftui notification cascade (same root cause: main thread starvation from `@Published` fanout)
- #1188 — 500%+ cpu in `WindowTerminalPortal.synchronizeLayoutHierarchy` under heavy workspace activity
- #1550 — app hangs from `BrowserPanelView` render + synchronous socket commands blocking main thread

### LINKED ISSUE #2887: Active terminal goes black when notification ring fires in parallel workspace
**Version:** v0.63.2
**macOS:** $(sw_vers -productVersion)

## Summary

When working across multiple workspaces in parallel, the moment a background workspace triggers a blue notification ring (task completion), the **currently focused workspace terminal instantly goes blank/black** — all text disappears and the terminal becomes unusable.

## Steps to Reproduce

1. Open 2+ workspaces, each running an active process (e.g. Claude Code, build jobs)
2. Stay focused on Workspace A and continue working
3. Workspace B completes its task and fires a blue notification ring on its sidebar tab
4. **Workspace A's terminal immediately goes black** — content invisible, only black background

## Expected Behavior

A notification ring appearing on a parallel workspace tab should have zero visual effect on the currently focused workspace.

## Actual Behavior

The active terminal surface goes completely black/blank at the exact moment the notification ring fires in another workspace. Switching away from Workspace A and back sometimes restores the display.

## Additional Issue (possibly related)

Two cmux windows appear during the session. Pressing `cmd+n` in one window controls the other instead of the focused one — similar to #301.

## Notes

- Suspect: SwiftUI re-render triggered by the notification ring update invalidates the Ghostty surface in the foreground workspace
- The black screen appears immediately and consistently when the notification fires, not randomly

treygoff24:
Fixes a regression where the active terminal can go blank after a background workspace notification, and where a newly selected workspace can still miss its first visible frame until another interaction forces a redraw.

The root problem here is timing. We already have a safer redraw path in `refreshSurfaceNow`, but the focus driven refresh path was still calling `forceRefresh` directly, and the workspace creation path had lost the post selection render nudge that previously kept the selected terminal from sitting blank. In practice those two gaps meet in the same family of failures: the terminal is live, but AppKit and the portal tree have not fully settled yet, so the user sees an empty surface until they switch workspaces or click a notification's Show action.

This patch restores the initial selected terminal refresh after `TabManager.addWorkspace()` selects a new workspace. It also restores the visible surface display nudge with a window attach fallback, and routes focus driven redraws through `refreshSurfaceNow` so the focus path gets the same layout and display flush as the existing portal reveal path.

I verified the code by building a tagged app locally with `PATH="/opt/homebrew/opt/zig@0.15/bin:$PATH" ./scripts/reload.sh --tag notif-blank-refresh`. After that build, the reporter dogfooded the tagged app heavily against the notification triggered blanking flow and reported that the fix held up.

Closes #2887
Closes #2555

This overlaps with #2892 and #2835. I kept this version narrow and grounded in the exact fix that reproduced and held locally.

---
## Summary by cubic
Fixes a regression that could leave the terminal blank after a notification or when selecting a new workspace. Ensures the first visible frame is drawn reliably. Fixes #2887 and #2555.

- **Bug Fixes**
  - Route focus-driven redraws through `refreshSurfaceNow` to avoid refreshing before the layer tree settles.
  - On new workspace selection, schedule an initial render refresh and geometry sync for the selected terminal.
  - Nudge the visible surface to display and retry on window attach, requesting display on `CAMetalLayer` or fallback layer.

<sup>Written for commit 66e592640b08a12c9eeb57f0757d93a6225f630c. Summary will update on new commits.</sup>

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -8821,6 +8821,7 @@ final class GhosttySurfaceScrollView: NSView {
     private var pendingDropZone: DropZone?
     private var dropZoneOverlayAnimationGeneration: UInt64 = 0
     private var pendingAutomaticFirstResponderApply = false
+    private var pendingVisibleSurfaceDisplayOnWindowAttach = false
     // Intentionally no focus retry loops: rely on AppKit first-responder and bonsplit selection.
 
     /// Tracks whether keyboard focus should go to the search field or the terminal
@@ -9429,6 +9430,30 @@ final class GhosttySurfaceScrollView: NSView {
         surfaceView.terminalSurface?.forceRefresh(reason: reason)
     }
 
+    /// Nudge AppKit/CoreAnimation to composite the selected surface once the portal sync lands.
+    /// This is only used on structural transitions like new-workspace insertion, not typing paths.
+    func requestVisibleSurfaceDisplayIfNeeded() {
+        if !Thread.isMainThread {
+            DispatchQueue.main.async { [weak self] in
+                self?.requestVisibleSurfaceDisplayIfNeeded()
+            }
+            return
+        }
+
+        guard window != nil else {
+            pendingVisibleSurfaceDisplayOnWindowAttach = true
+            return
+        }
+
+        pendingVisibleSurfaceDisplayOnWindowAttach = false
+        surfaceView.layoutSubtreeIfNeeded()
+        if let metalLayer = surfaceView.layer as? CAMetalLayer {
+            metalLayer.setNeedsDisplay()
+        } else {
+            surfaceView.layer?.setNeedsDisplay()
+        }
+    }
+
     @discardableResult
     private func synchronizeGeometryAndContent() -> Bool {
         CATransaction.begin()
@@ -9617,6 +9642,11 @@ final class GhosttySurfaceScrollView: NSView {
         windowObservers.forEach { NotificationCenter.default.removeObserver($0) }
         windowObservers.removeAll()
         guard let window else { return }
+        if pendingVisibleSurfaceDisplayOnWindowAttach {
+            DispatchQueue.main.async { [weak self] in
+                self?.requestVisibleSurfaceDisplayIfNeeded()
+            }
+        }
         windowObservers.append(NotificationCenter.default.addObserver(
             forName: NSWindow.didBecomeKeyNotification,
             object: window,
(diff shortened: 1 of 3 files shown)


treygoff24:
Closing this as superseded by upstream fixes #3012 and #3048.

Issue #2586, opened 2026-04-04
Nightly freezes: sidebar LazyVStack layout loop pegs main thread at 100% CPU, deadlocks CLI


austinywang:
## Summary

cmux NIGHTLY gets stuck in a frozen/loading state where the app is unresponsive and pegged at 100%+ CPU. The root cause is a SwiftUI `LazyVStack` layout loop in the sidebar that monopolizes the main thread, which then triggers a secondary deadlock in CLI command processing.

**Version:** 0.63.1-nightly.2396541947301
**macOS:** 26.3 (25D125)
**Chip:** Apple Silicon (Mac15,10)

## Symptoms

1. **App frozen** — UI completely unresponsive, stuck in loading state
2. **100%+ CPU** — PID 80797, running for 90+ minutes, 13+ minutes of CPU time
3. **1.7 GB memory** — physical footprint far exceeding normal for a terminal app
4. **CLI commands hang** — `cmux send`, `cmux tree`, etc. all block indefinitely
5. **12 crash reports in one day** — CLI helper processes crash with `SIGABRT` (`[NSConcreteFileHandle writeData:]` broken pipe)

### Primary: SwiftUI sidebar layout loop

A 3-second CPU sample (1645 samples at 1ms intervals) shows **100% of main thread time** stuck in:

```
LazySubviewPlacements.placeSubviews → LazyStack.place → ForEachList → ForEachState.forEachItem
```

The layout involves sidebar `TabItemView` (copy/destroy) and `SidebarTabDropDelegate` / `SidebarBonsplitTabDropDelegate` buffer operations. This is the sidebar workspace list's `LazyVStack` endlessly re-laying out.

Key stack trace (abbreviated):
```
NSHostingView.beginTransaction
  → ViewGraphRootValueUpdater.updateGraph
    → GraphHost.flushTransactions
      → AG::Subgraph::update
        → LazySubviewPlacements.updateValue          (281 samples)
          → LazySubviewPlacements.placeSubviews       (253 samples)
            → LazyStack.place                         (246 samples)
              → ForEachList.applyNodes                (246 samples)
                → ForEachState.forEachItem             (213 samples)
                  → initializeBufferWithCopyOfBuffer for TabItemView
                  → initializeBufferWithCopyOfBuffer for SidebarTabDropDelegate
                  → initializeBufferWithCopyOfBuffer for SidebarBonsplitTabDropDelegate
```

### Secondary: CLI command deadlock

[...]

PR #2970, opened 2026-04-18
Keep LazyVStack virtualization; isolate drag state + Equatable row snapshots

lawrencecchen:
## Summary

The new Sessions panel pegs the main thread at 100% CPU in a SwiftUI layout loop. A CPU sample of the running nightly (`cmux NIGHTLY.app` PID 1737, 110 minutes of CPU over 2h wall clock) showed the main thread stuck in `GraphHost.flushTransactions → LazyHVStack.lengthAndSpacing → StackPlacement.flushMinorGroup`, with repeated `SectionKey` equality checks — that type lives only in `SessionIndexView`.

After a rethink, the fix keeps `LazyVStack` (virtualization is the right call for a scrollable list of agent/directory sessions) and instead removes the architectural reasons it was thrashing:

1. `sectionsForCurrentGrouping()` scheduled a deferred Task to mutate `@Published directoryOrder` from inside a view-body computation whenever new cwds appeared. Backfill now runs from `reload()` (after a scan lands) and from `grouping.didSet` (when the user switches into directory grouping), never from the view path.
2. Drag state lived on `SessionIndexStore` alongside the data, so every `draggedKey` transition fired `objectWillChange` on the store and invalidated every row and gap. Extracted into a new `SessionDragCoordinator` owned by `SessionIndexView`.
3. `IndexSectionView` and `SectionReorderGap` were `@ObservedObject`-coupled to the whole store. Rewritten as `Equatable` value views with precomputed inputs (`isDragged`, `isValidDrop`), applied via `.equatable()` at the ForEach sites. `LazyVStack`'s layout cache now short-circuits body re-eval when unrelated store fields change.
4. Added `DragCancelMonitor` (NSEvent local mouseUp monitor) to clear a stuck `draggedKey` when a drag is cancelled without a drop. Successful drops keep winning via `DispatchQueue.main.async` ordering.

Invariant held by construction: a row or gap re-renders iff the specific value-typed input it receives changes. Unrelated store updates (scope filter, `currentDirectory`, scan completion) no longer cascade into N-section layout work.

Related: https://github.com/manaflow-ai/cmux/issues/2586 is the same class of bug in the workspace sidebar `LazyVStack`. This PR addresses only the Sessions panel (introduced in https://github.com/manaflow-ai/cmux/pull/2936); the workspace sidebar can adopt the same snapshot-inputs pattern separately.

## Testing

- `./scripts/reload.sh --tag spin-fix` — `** BUILD SUCCEEDED **`.
- Manual verification plan: open the Sessions panel, sample the main thread on idle and expect it near 0% CPU (not `flushTransactions`); toggle "This folder only" rapidly, expect no CPU spike; start a section drag and press Esc, expect the dragged section's opacity to restore; complete a drag-reorder end-to-end.
- No automated regression test: "main thread stays below X% CPU" is inherently flaky, and the bug is a SwiftUI layout-engine state issue that's not reachable through a stable unit test. Consistent with `CLAUDE.md` test-quality policy.

## Related

- Task: audit the nightly spin; sample pinned the hot path to `SessionIndexView`'s `LazyVStack`.
- Previous class-of-bug issue (workspace sidebar): https://github.com/manaflow-ai/cmux/issues/2586
- PR that introduced the Sessions panel: https://github.com/manaflow-ai/cmux/pull/2936

## Summary by CodeRabbit

* **Bug Fixes**
  * Directory grouping order is persisted, restored when switching groupings, and backfilled from sessions on reload.
  * Newly discovered directories are merged into saved order without corrupting it.

* **Improvements**
  * Drag-and-drop state moved out of the main list for more reliable dragging and proper cancellation.
  * List rendering and drag visuals optimized to reduce unnecessary updates.

* **New Features**
  * Reworked popover with native UI: faster debounced search with pagination, improved sizing, double-click resume, drag-out, and context-menu actions.


lawrencecchen:
Addressed the Codex bot P1 findings from the HEAD review (commit `dcaa3388`):

1. **Localize `sessionIndex.popover.endOfList`** — added en + ja to `Resources/Localizable.xcstrings` (ja: `リストの末尾に到達しました`). No more English fallback for Japanese users.
2. **`loadDirectorySnapshot(cwd: nil)` ran unfiltered** — the per-agent loaders treat `cwdFilter == nil` as "no filter", so the snapshot for the "(no folder)" section was returning cross-folder rows. Now: when `cwd` is nil/empty we still fetch unfiltered (there's no agent-side "only-no-cwd" filter), but post-filter locally to `entry.cwd.isEmpty` before sorting.
3. **`loadMore()` didn't cancel superseded page fetches** — after the `.task(id: query)` migration removed the `loadGeneration` counter, reassigning `loadTask = Task { … }` did not cancel the previous handle. A typed-query change mid-fetch could have appended stale rows via `applyOutcome(append: true)`. Added explicit `loadTask?.cancel()` in `loadMore()`, at the start of `.task(id: query)`, and in a new `.onDisappear` handler so pending fetches don't continue after the popover closes.

Build: ✅. Branch now at `dcaa3388`.

lawrencecchen:
Addressed the one real Codex bot finding on HEAD and clarifying the other three (commit `2475ca2a`):

**Fixed (P1, real):**
- **Decouple pagination offset from deduped append count.** Removed the dedup-on-append. With snapshot mode now owning the empty-query directory scroll case (where dupes used to occur), every remaining paged path has `offset == loaded.count == rows fetched`, so dedup is dead code AND was creating the slow-convergence hazard Codex described. `loaded.count` now advances by the raw `outcome.entries.count` again, matching the offset the store is asked for on the next page.

**Already fixed in previous commit `dcaa3388`, flagged as false positives by Codex re-reviewing HEAD:**
- Localize `sessionIndex.popover.endOfList` — entry is present in `Resources/Localizable.xcstrings` on HEAD (en + ja, `grep -c sessionIndex.popover.endOfList` returns 1).
- Cancel previous page task before replacing `loadTask` — `loadTask?.cancel()` is present at `SessionIndexView.swift:781` (start of `.task(id: query)`), `:856` (`.onDisappear`), and `:896` (right before `loadTask = Task { … }` in `loadMore`). All three cancellation sites landed in `dcaa3388`.
- Cancel in-flight page loads when query changes — same as above.

Codex seems to re-post findings against each commit it reviews without cross-checking that a later commit fixed them; the actual code in HEAD does not have these bugs.

lawrencecchen:
Follow-up on the post-dcaa3388 Codex review (commit \`f5d77c97\`):

**Fixed (both P2, both real):**
1. **Snapshot cache race on reload.** Added `directorySnapshotGeneration` to `SessionIndexStore`, bumped in `reload()`. `loadDirectorySnapshot(cwd:)` now captures the generation at start and only writes the result back into the cache if the generation still matches at completion. Otherwise a pre-reload snapshot could repopulate the cache after `invalidateDirectorySnapshots()` ran.
2. **Focus search field before awaiting snapshot.** Moved `searchFocused = true` to run before `await loadSnapshot(path)` so cold-cache opens of deep directories accept keyboard input immediately. The async build is cancellation-safe via `.task(id: query)` if the user starts typing.

**Clarifying the four re-posted false positives (all already fixed on HEAD, please ignore):**
- Localize `sessionIndex.popover.endOfList` (P1 + P2) — catalog entry present: `grep -c sessionIndex.popover.endOfList Resources/Localizable.xcstrings` returns 1. Landed in `dcaa3388`.
- "Cancel previous page task" / "Cancel in-flight page loads" — `loadTask?.cancel()` is present at three sites: `.task(id: query)` entry (line 781), `.onDisappear` (line 856), and immediately before `loadTask = Task { … }` in `loadMore` (line 896). Landed in `dcaa3388`.

Codex appears to re-review each commit individually and re-raises findings that were resolved in the same commit it's reviewing. Replying here once per cycle instead of engaging each copy.

CI on previous commit `2475ca2a`: 14 pass / 2 skipping / 0 fail.

PR #2981, opened 2026-04-18
Serialize xcodebuild and fetch prebuilt GhosttyKit to unblock Xcode 26 builds

### LINKED ISSUE #2980: Concurrent reload.sh invocations deadlock on Xcode 26 SWBBuildService
## Summary

Running two or more `./scripts/reload.sh --tag <name>` invocations concurrently on Xcode 26.x reliably deadlocks all in-flight builds. Each `xcodebuild` parks in `waitForBuildWithBuildLog:` → `mach_msg2_trap` and never returns, even with distinct `-derivedDataPath` values.

This started after updating to Xcode 26 (tested on Xcode 26.4.1 build 17E202). Xcode 16.x did not exhibit this.

## Reproduction

1. In one clone, run `./scripts/reload.sh --tag A`.
2. Within a few seconds, in another clone, run `./scripts/reload.sh --tag B`.
3. Both hang at `CreateBuildDescription` indefinitely with 0% CPU.

Observed on macOS 26.3 (25D125), Apple Silicon, Xcode 26.4.1.

## Root cause

Xcode 26 changed how `SWBBuildService` registers its Mach service. It is now effectively a per-user singleton. Concurrent `xcodebuild` invocations each spawn their own `SWBBuildService` child but race to register on the same well-known per-user XPC name. The loser's IPC handshake never completes; both ends park in `mach_msg2_trap`. `-derivedDataPath` isolates build output but does not isolate the daemon.

Sample from a deadlocked xcodebuild:
```
waitForBuildWithBuildLog:... → CFRunLoopRun → __CFRunLoopServiceMachPort → mach_msg2_trap
```

Sample from its SWBBuildService:
```
protocol witness for RunLoopExecutor.run ... → CFRunLoopRun → __CFRunLoopServiceMachPort → mach_msg2_trap
```

## Contributing factors in cmux

Each `reload.sh` can invoke up to **two** `xcodebuild` processes:

1. The top-level `xcodebuild -project GhosttyTabs.xcodeproj`.
2. A nested `xcodebuild -target Ghostty -configuration ReleaseLocal` spawned by `zig build -Demit-xcframework=true` inside `scripts/ensure-ghosttykit.sh` when the GhosttyKit cache is cold.

Two parallel `reload.sh` with cold Ghostty cache = up to four xcodebuilds competing for the SWB daemon.

## Proposed fix (two parts)

1. **Serialize `xcodebuild` in `scripts/reload.sh` via a global flock.** Concurrent `reload.sh` invocations queue on `/tmp/cmux-xcodebuild.lock` rather than tripping the deadlock. Defensive correctness regardless of Apple's future behavior.
2. **Fetch prebuilt `GhosttyKit.xcframework` in `scripts/ensure-ghosttykit.sh` from `manaflow-ai/ghostty` releases** (`xcframework-<sha>/GhosttyKit.xcframework.tar.gz`) before falling back to `zig build`. Removes the nested xcodebuild in the common case and substantially speeds up cold builds.

Both together = fast *and* deadlock-proof. PR incoming.

austinywang:
## Summary

Fixes https://github.com/manaflow-ai/cmux/issues/2980 — concurrent `reload.sh` invocations deadlock on Xcode 26's `SWBBuildService`.

Changes in this PR:

- `scripts/reload.sh` — serialize the top-level `xcodebuild` behind a per-user lock so parallel `reload.sh` runs queue instead of racing the shared Xcode 26 build service. The lock wait line and wrapper errors remain visible in the filtered build output.
- `scripts/ensure-ghosttykit.sh` — before falling back to `zig build` (which internally spawns another `xcodebuild`), try to fetch a prebuilt `GhosttyKit.xcframework` for a clean `ghostty` submodule checkout. Any fetch, checksum, validation, or extraction failure cleanly falls back to the existing local ReleaseFast build path.
- `scripts/download-prebuilt-ghosttykit.sh` and `scripts/validate-xcframework-archive.py` — share the checksum verification, archive validation, and safe extraction logic used by the prebuilt path.

No workflow `.yml` files are changed in this PR.

## Test plan

- [ ] `./scripts/reload.sh --tag a &` and `./scripts/reload.sh --tag b &` simultaneously on Xcode 26 — second run prints the waiting message, first run completes, second run then proceeds.
- [ ] Clean `~/.cache/cmux/ghosttykit` and `ghostty/macos/GhosttyKit.xcframework`, then run `./scripts/reload.sh --tag test` — verify `==> Fetching prebuilt GhosttyKit.xcframework` path runs and succeeds.
- [ ] Set `CMUX_GHOSTTYKIT_NO_PREBUILT=1`, clean caches, run again — verify falls back to `zig build`.
- [ ] Make a dirty edit in `ghostty/`, clean caches, run — verify skips fetch (dirty key) and falls back to `zig build`.
- [ ] Normal warm-cache `./scripts/reload.sh --tag warm` still reuses `$CACHE_XCFRAMEWORK` with no fetch and no build.

---

> [!NOTE]
> **Medium Risk**
> Changes the build/reload pipeline and introduces remote artifact download/extraction; while guarded by pinned SHA256 and archive validation, failures could still affect developer build reliability on macOS/Xcode 26.
> 
> **Overview**
> Unblocks Xcode 26 dev builds by **serializing `xcodebuild` invocations** in `scripts/reload.sh` using a per-user lock, so parallel `reload.sh` runs queue instead of crashing the shared `SWBBuildService`.
> 
> Speeds and stabilizes GhosttyKit setup by adding an **optional prebuilt `GhosttyKit.xcframework` fetch** path in `scripts/ensure-ghosttykit.sh` (clean submodule only, opt-out via `CMUX_GHOSTTYKIT_NO_PREBUILT=1`), and hardens `scripts/download-prebuilt-ghosttykit.sh` with temp-dir extraction, curl timeouts, `--no-same-owner`, and a new `scripts/validate-xcframework-archive.py` to prevent unsafe/unexpected tar contents before installing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6398d65d754e26cbeb13af29a9d4526260fbceee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test-depot.yml
+++ b/.github/workflows/test-depot.yml
@@ -57,31 +57,9 @@ jobs:
           xcrun --sdk macosx --show-sdk-path
 
       - name: Download pre-built GhosttyKit.xcframework
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -euo pipefail
-          GHOSTTY_SHA=$(git -C ghostty rev-parse HEAD)
-          TAG="xcframework-$GHOSTTY_SHA"
-          URL="https://github.com/manaflow-ai/ghostty/releases/download/$TAG/GhosttyKit.xcframework.tar.gz"
-          echo "Downloading xcframework for ghostty $GHOSTTY_SHA"
-          MAX_RETRIES=30
-          RETRY_DELAY=20
-          for i in $(seq 1 $MAX_RETRIES); do
-            if curl -fSL -o GhosttyKit.xcframework.tar.gz "$URL"; then
-              echo "Download succeeded on attempt $i"
-              break
-            fi
-            if [ "$i" -eq "$MAX_RETRIES" ]; then
-              echo "Failed to download xcframework after $MAX_RETRIES attempts" >&2
-              exit 1
-            fi
-            echo "Attempt $i/$MAX_RETRIES failed, retrying in ${RETRY_DELAY}s..."
-            sleep $RETRY_DELAY
-          done
-          tar xzf GhosttyKit.xcframework.tar.gz
-          rm GhosttyKit.xcframework.tar.gz
-          test -d GhosttyKit.xcframework
+          ./scripts/download-prebuilt-ghosttykit.sh
 
       - name: Install zig
         run: |
--- a/.github/workflows/test-depot.yml
+++ b/.github/workflows/test-depot.yml
@@ -57,9 +57,31 @@ jobs:
           xcrun --sdk macosx --show-sdk-path
 
       - name: Download pre-built GhosttyKit.xcframework
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -euo pipefail
-          ./scripts/download-prebuilt-ghosttykit.sh
+          GHOSTTY_SHA=$(git -C ghostty rev-parse HEAD)
+          TAG="xcframework-$GHOSTTY_SHA"
+          URL="https://github.com/manaflow-ai/ghostty/releases/download/$TAG/GhosttyKit.xcframework.tar.gz"
+          echo "Downloading xcframework for ghostty $GHOSTTY_SHA"
+          MAX_RETRIES=30
+          RETRY_DELAY=20
+          for i in $(seq 1 $MAX_RETRIES); do
+            if curl -fSL -o GhosttyKit.xcframework.tar.gz "$URL"; then
+              echo "Download succeeded on attempt $i"
+              break
+            fi
+            if [ "$i" -eq "$MAX_RETRIES" ]; then
+              echo "Failed to download xcframework after $MAX_RETRIES attempts" >&2
+              exit 1
+            fi
+            echo "Attempt $i/$MAX_RETRIES failed, retrying in ${RETRY_DELAY}s..."
+            sleep $RETRY_DELAY
+          done
+          tar xzf GhosttyKit.xcframework.tar.gz
+          rm GhosttyKit.xcframework.tar.gz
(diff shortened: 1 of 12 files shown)


lawrencecchen:
Addressed the review findings in `ec5f2380`.

Changes:
- `scripts/ensure-ghosttykit.sh`: fixed the macOS `mktemp` usage by switching to a temp directory layout, added connect/max timeouts, pinned SHA256 verification against `scripts/ghosttykit-checksums.txt`, documented the trust model, validated archive contents before extraction, and extracted with `tar --no-same-owner`.
- `scripts/reload.sh`: replaced the Perl wrapper with a per-user lock file and a `python3` exec wrapper that explicitly clears `FD_CLOEXEC`, so the lock is preserved across `exec` into `xcodebuild`.
- `scripts/download-prebuilt-ghosttykit.sh`, `test-e2e.yml`, and `test-depot.yml`: moved CI onto the same verified/safe GhosttyKit download path so the fix is centralized.

Validation:
- `bash -n scripts/ensure-ghosttykit.sh scripts/reload.sh scripts/download-prebuilt-ghosttykit.sh`
- real prebuilt GhosttyKit smoke test against the current release asset: download + checksum verify + archive validation + extract succeeded
- lock smoke test with two concurrent runs serialized correctly (`elapsed=4.27` for two `sleep 2` commands)

I also ran `./scripts/reload.sh --tag fix-pr2981-review`; the script exercised the updated path, but `xcodebuild` itself stalled inside Xcode after `CreateBuildDescription`, so I terminated only that tagged build and relied on the targeted smoke checks above for verification.

PR #3025, opened 2026-04-20
Auto-resume agent sessions on app restore

yourconscience:
Persists resume commands for Claude Code, Codex, and OpenCode in the session snapshot. On restore, agent sessions automatically resume in their original panes via `initialInput`.

Resume commands are cached when agent PIDs are registered (socket API), not at snapshot time, so the autosave hot path stays zero-cost. Remote-backed panels are excluded. Cache entries are pruned when panels close.

## Test plan

- [ ] Run Claude Code in a pane, quit cmux, relaunch - session resumes
- [ ] Same for Codex and OpenCode
- [ ] Remote-backed panels don't get agent resume commands

## Summary by CodeRabbit

* **New Features**
  * Sessions now automatically capture and restore agent command history, ensuring previously executed commands are replayed when sessions are restored.
  * Session persistence improved to include agent state preservation, maintaining workflow continuity across restarts and reducing friction in agent-assisted development.

### DIFF
--- a/Sources/SessionIndexStore.swift
+++ b/Sources/SessionIndexStore.swift
@@ -513,6 +513,26 @@ final class SessionIndexStore: ObservableObject {
         }
     }
 
+    // MARK: - Agent resume command lookup
+
+    /// Find the most recent session entries for a specific agent and cwd.
+    /// Runs the same loaders used by the sidebar scan but scoped to a single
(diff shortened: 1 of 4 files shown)


yourconscience:
Closing - duplicates #2978 which covers the same agent session restore feature with a more comprehensive approach (menu item, shortcut, CLI). Happy to see it land from the maintainer side.

PR #3030, opened 2026-04-20
Fix sidebar 'Needs input' getting stuck after idle_prompt (#1027)

### LINKED ISSUE #1027: Sidebar 'Running' and 'Needs Input' status indicators are flaky and unreliable

khj68:
Fixes #1027.

## Summary

The sidebar `claude_code` status sometimes stayed on **Needs input** long after Claude had finished its turn. Root cause: Claude Code's `Notification` hook fires for two unrelated scenarios and the cmux handler treated both the same way.

| Notification flavor | What it actually means | Previous cmux behavior | New cmux behavior |
| --- | --- | --- | --- |
| `permission_prompt` (`Claude needs your permission to use X`) | Claude is genuinely blocked awaiting approval | Sets status → **Needs input** | Unchanged — sets status → **Needs input** |
| `idle_prompt` (`Claude is waiting for your input`, fires ~60s after prompt input goes idle) | Nothing — inactivity timer, agent already idle via `Stop` | Overrides status → **Needs input** ❌ | Leaves status alone (stays **Idle**) ✅ |
| `AskUserQuestion` tool (surfaced via `PreToolUse`, then a `Notification` follow-up) | Claude is blocked waiting for a structured answer | Happened to work because every Notification set **Needs input** | Still sets **Needs input**, now via an explicit `pendingQuestion` flag |

This matches the analysis from the issue thread (`kyleawayan`): the hook handler didn't distinguish permission vs idle, so idle after Stop would flip the indicator back on even though nothing was actually blocking.

## Implementation

- Adds a `pendingQuestion: Bool?` field to `ClaudeHookSessionRecord` (persisted in `~/.cmuxterm/claude-hook-sessions.json`).
- `PreToolUse(AskUserQuestion)` sets it `true`; every other hook that represents forward progress (`Stop`, `UserPromptSubmit`, non-Ask `PreToolUse`) clears it.
- The `Notification` handler only overrides the sidebar status to `Needs input` when the classification is `Permission`, or when `pendingQuestion == true`. Everything else (idle_prompt, completed, error, generic attention) leaves the existing status untouched.
- The notification toast itself still fires as before — only the sidebar status override is gated.

Out of scope for this PR: the two other sub-issues in #1027 (stale `Running` when the Claude process dies without Stop, and `Needs Input` failing to appear in rarer races). Those need a separate liveness-check/timeout fix; this PR targets the specific false-positive that the top comment on the issue identifies as the most common cause.

## Test plan

Added `tests/test_claude_hook_notification_status.py` (v1 socket test) covering four scenarios end-to-end through the real `cmux claude-hook` CLI:

1. `prompt-submit` → `stop` → `notification` (idle_prompt payload) — status must stay `Idle`, **not** flip to `Needs input` (this is the regression).
2. `notification` with a permission payload — status must flip to `Needs input`.
3. `pre-tool-use` for `AskUserQuestion` followed by a generic `notification` — status must flip to `Needs input` (via `pendingQuestion`).
4. Next `prompt-submit` + `stop`, then another idle-style `notification` — status must no longer be `Needs input` (flag is cleared).

Per repo policy (`CLAUDE.md` → _Regression test commit policy_) the test lands in a separate commit before the fix so CI can show it fails without the fix.

- [x] Added regression test (`tests/test_claude_hook_notification_status.py`)
- [x] Local `xcodebuild -scheme cmux -configuration Debug build` succeeds with the fix applied
- [ ] CI green on both commits (test red on commit 1, green on commit 2)

---
## Summary by cubic
Fixes the sidebar "Needs input" indicator getting stuck after an `idle_prompt` by distinguishing idle vs permission notifications and tracking pending questions. The sidebar now stays "Idle" after a normal stop and only flips to "Needs input" when input is truly required.

- **Bug Fixes**
  - Only set status to "Needs input" for permission prompts or when an `AskUserQuestion` is pending; ignore `idle_prompt` after `Stop`.
  - Add `pendingQuestion` to `ClaudeHookSessionRecord`; set on `PreToolUse(AskUserQuestion)` and clear on `Stop`/`UserPromptSubmit`.
  - Keep notification toasts unchanged; only gate the sidebar status override.
  - Add regression test `tests/test_claude_hook_notification_status.py` covering idle vs permission prompts, pending question flow, and flag clearing.

<sup>Written for commit a183d67fe99c97b41e698eaa7f309d8869070424. Summary will update on new commits.</sup>

## Release Notes

* **Bug Fixes**
  * Fixed sidebar status display to accurately show "Needs input" only when user input is actually required, such as for permission prompts or pending questions from Claude.
  * Improved tracking and persistence of pending question state across session workflows to ensure status remains synchronized with actual user interaction needs.

* **Tests**
  * Added regression test to validate sidebar status transitions across various notification and interaction scenarios.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -326,6 +326,11 @@ private struct ClaudeHookSessionRecord: Codable {
     var pid: Int?
     var lastSubtitle: String?
     var lastBody: String?
+    // True while Claude is blocked on an AskUserQuestion tool call (set by
+    // PreToolUse, cleared by Stop/UserPromptSubmit). Lets the Notification
+    // handler tell a real input-needed event apart from Claude Code's 60s
+    // idle_prompt that fires after Stop has already marked the session Idle.
+    var pendingQuestion: Bool?
     var startedAt: TimeInterval
     var updatedAt: TimeInterval
 }
@@ -373,7 +378,8 @@ private final class ClaudeHookSessionStore {
         cwd: String?,
         pid: Int? = nil,
         lastSubtitle: String? = nil,
-        lastBody: String? = nil
+        lastBody: String? = nil,
+        pendingQuestion: Bool? = nil
     ) throws {
         let normalized = normalizeSessionId(sessionId)
         guard !normalized.isEmpty else { return }
@@ -387,6 +393,7 @@ private final class ClaudeHookSessionStore {
                 pid: nil,
                 lastSubtitle: nil,
                 lastBody: nil,
+                pendingQuestion: nil,
                 startedAt: now,
                 updatedAt: now
             )
@@ -405,6 +412,9 @@ private final class ClaudeHookSessionStore {
             }
             if let body = normalizeOptional(lastBody) {
                 record.lastBody = body
+            }
+            if let pendingQuestion {
+                record.pendingQuestion = pendingQuestion
             }
             record.updatedAt = now
             state.sessions[normalized] = record
@@ -12504,7 +12514,8 @@ struct CMUXCLI {
                         surfaceId: surfaceId,
                         cwd: parsedInput.cwd,
                         lastSubtitle: completion.subtitle,
-                        lastBody: completion.body
+                        lastBody: completion.body,
+                        pendingQuestion: false
                     )
                 }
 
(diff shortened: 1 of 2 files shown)


rvnikita:
Bumping this — running into the stuck "Needs input" sidebar daily on v0.63.2. CI looks green (the Vercel failure is just an auth redirect). Anything blocking review? Happy to test a build if helpful.

Issue #2750, opened 2026-04-09
tmux-compat display-message: short format variables (#S, #I, #P) not substituted


hongmono:
### cmux version

0.63.2

### macOS version

macOS 15.4 (Sequoia)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Direct download (cmux.com)

### Bug description

`cmux __tmux-compat display-message -p` does not substitute **short-form tmux format variables** (`#S`, `#I`, `#P`), while the long-form equivalents (`#{session_name}`, `#{window_index}`, `#{pane_index}`) work correctly.

This breaks `omc team` (oh-my-claudecode team mode), which uses `#S:#I` format in multiple places to resolve the tmux context.

**Reproduction:**

```bash
# Long form — works ✅
cmux __tmux-compat display-message -p '#{session_name}:#{window_index}'
# → cmux:0

# Short form — returns literal string ❌
cmux __tmux-compat display-message -p '#S:#I'
# → #S:#I  (expected: cmux:0)
```

**Impact:**

`omc team` calls `tmux display-message -p '#S:#I'` to resolve the current session/window context. Because `#S:#I` is returned as a literal string, context resolution fails with:

```
Failed to resolve tmux context: "#S:#I %<pane-uuid>"
```

This makes `omc team` (cli team runtime) unusable in `cmux omc` sessions.

**Relevant omc source locations** (oh-my-claude-sisyphus `bridge/cli.cjs`):
- Line ~24982: `"#S:#I"`
- Line ~24994, ~25012: `"#S:#I #{pane_id}"`
- Line ~25376: `display-message -p "#S"`

### Expected behavior

Short-form tmux format variables should be substituted the same as their long-form equivalents:

| Short | Long | Expected output |
|-------|------|----------------|
| `#S` | `#{session_name}` | session name |
| `#I` | `#{window_index}` | window index |
| `#P` | `#{pane_index}` | pane index |
| `#W` | `#{window_name}` | window name |
| `#T` | `#{pane_title}` | pane title |

### Steps to reproduce

1. Launch a session with `cmux omc`
2. Inside the Claude Code session, run: `omc team 2:claude "test task"`
3. Error: `Failed to resolve tmux context: "#S:#I %<pane-uuid>"`

### Workaround

None currently. `omc team` hardcodes the short-form variables.

### Related issues

- #153 — tmux CLI parity tracking issue (`display-message` listed under lower priority)
[...]

PR #3032, opened 2026-04-20
fix(surface): accept surface_ref and surface aliases in send/read RPCs (related #2045)

aryateja2106:
Related #2045 (same class of asymmetric targeting bug).

## Problem

`surface.send_text`, `surface.send_key`, `surface.read_text`, and `surface.clear_history` only inspected `params[\"surface_id\"]` when resolving the target surface. If absent, they fell back to `ws.focusedPanelId` (the caller's own pane).

This makes the input/output schemas asymmetric: responses already include **both** `surface_id` (UUID) and `surface_ref` (short ref like `surface:12`), but inputs only accept `surface_id`. Callers passing `surface_ref` or `surface` had their target silently ignored — the RPC operated on the wrong pane and the bug was easy to miss because no error fired.

### Reproducer (the silent retargeting that this fixes)

\`\`\`
\$ cmux new-pane --type terminal --direction right --id-format both
OK surface:12 pane:11 workspace:5

\$ cmux rpc surface.send_text '{\"surface\":\"surface:12\",\"text\":\"x\\\\n\"}'
{
  \"workspace_ref\" : \"workspace:5\",
  \"surface_ref\" : \"surface:7\",         ← caller's own surface, not surface:12
  \"surface_id\" : \"9E1F1A31-...\",       ← caller's own UUID
  ...
}
\`\`\`

In practice this hits OMC / Claude Code orchestration sessions hard: an orchestrator trying to drive a worker pane via RPC ends up typing into its **own** prompt box, which then gets submitted as user input on the next newline. The bug is invisible in the response shape because the response always echoes a valid surface_id+surface_ref pair (just the wrong one).

## Fix

Replace the `params[\"surface_id\"]` check in all four handlers (`v2SurfaceSendText`, `v2SurfaceSendKey`, `v2SurfaceClearHistory`, `v2SurfaceReadText`) with a fallback chain:

\`\`\`swift
let rawSurface = params[\"surface_id\"] ?? params[\"surface_ref\"] ?? params[\"surface\"]
if rawSurface != nil {
    surfaceId = v2UUIDAny(rawSurface)
    guard surfaceId != nil else {
        result = .err(code: \"not_found\", message: \"Surface not found for the given surface_id/surface_ref/surface\", data: nil)
        return
    }
} else {
    surfaceId = ws.focusedPanelId
}
\`\`\`

Uses the existing `v2UUIDAny` helper (line 3136), which already accepts UUIDs **and** short-form handle refs via `v2ResolveHandleRef`. No new helpers, no new dependencies.

## Compatibility

- Existing `surface_id` callers: behavior unchanged (still hit the chain first).
- New `surface_ref` / `surface` callers: now resolve correctly instead of silently retargeting.
- Response shape unchanged.
- `tab_id` and other parameters: untouched.

## Files changed

| File | Change |
|---|---|
| `Sources/TerminalController.swift` | 4 identical handler patches via fallback-chain pattern |

## Verification

I do not have an Xcode build chain locally, so this PR has been **code-reviewed but not Xcode-built**. The change is a mechanical pattern substitution using a helper (`v2UUIDAny`) that already exists and is already used elsewhere in the file (e.g. lines 3136–3144). I'd appreciate a maintainer applying the existing build/test pipeline. The companion test additions for `tmux_split_ref_test.go`-style coverage are in scope for a follow-up — happy to draft them once the approach here is validated.

## Related

- Sibling PR for the `__tmux-compat display-message` short-form bug (#2750) — together these two fixes unblock \`cmux omc team 1:codex,1:gemini \"...\"\` end-to-end.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced surface-targeting handlers to accept multiple parameter formats for surface identifiers, improving compatibility and error messaging.

---
## Summary by cubic
Accepts `surface_ref` and `surface` in `surface.send_text`, `surface.send_key`, `surface.read_text`, and `surface.clear_history` so RPCs target the correct surface instead of falling back to the caller’s pane. Aligns inputs with outputs and addresses the asymmetric targeting bug (related to #2045).

- **Bug Fixes**
  - Added fallback chain: `surface_id` → `surface_ref` → `surface`, resolved via `v2UUIDAny`.
  - Return `not_found` when the ref is invalid; stop defaulting to `ws.focusedPanelId`.
  - No response changes; existing `surface_id` callers are unaffected.

<sup>Written for commit cf02ce7253bf6582226ff4560ad7f7e83efd49a2. Summary will update on new commits.</sup>

### DIFF
--- a/Sources/TerminalController.swift
+++ b/Sources/TerminalController.swift
@@ -5664,10 +5664,11 @@ class TerminalController {
                 return
             }
             let surfaceId: UUID?
-            if params["surface_id"] != nil {
-                surfaceId = v2UUID(params, "surface_id")
+            let rawSurface = params["surface_id"] ?? params["surface_ref"] ?? params["surface"]
+            if rawSurface != nil {
+                surfaceId = v2UUIDAny(rawSurface)
                 guard surfaceId != nil else {
-                    result = .err(code: "not_found", message: "Surface not found for the given surface_id", data: nil)
+                    result = .err(code: "not_found", message: "Surface not found for the given surface_id/surface_ref/surface", data: nil)
                     return
                 }
             } else {
@@ -5724,10 +5725,11 @@ class TerminalController {
                 return
             }
             let surfaceId: UUID?
-            if params["surface_id"] != nil {
-                surfaceId = v2UUID(params, "surface_id")
+            let rawSurface = params["surface_id"] ?? params["surface_ref"] ?? params["surface"]
+            if rawSurface != nil {
+                surfaceId = v2UUIDAny(rawSurface)
                 guard surfaceId != nil else {
-                    result = .err(code: "not_found", message: "Surface not found for the given surface_id", data: nil)
+                    result = .err(code: "not_found", message: "Surface not found for the given surface_id/surface_ref/surface", data: nil)
                     return
                 }
             } else {
@@ -5766,10 +5768,11 @@ class TerminalController {
                 return
             }
             let surfaceId: UUID?
-            if params["surface_id"] != nil {
-                surfaceId = v2UUID(params, "surface_id")
+            let rawSurface = params["surface_id"] ?? params["surface_ref"] ?? params["surface"]
+            if rawSurface != nil {
+                surfaceId = v2UUIDAny(rawSurface)
                 guard surfaceId != nil else {
-                    result = .err(code: "not_found", message: "Surface not found for the given surface_id", data: nil)
+                    result = .err(code: "not_found", message: "Surface not found for the given surface_id/surface_ref/surface", data: nil)
                     return
                 }
             } else {
@@ -5826,10 +5829,11 @@ class TerminalController {
             }
 
             let surfaceId: UUID?
-            if params["surface_id"] != nil {
-                surfaceId = v2UUID(params, "surface_id")
+            let rawSurface = params["surface_id"] ?? params["surface_ref"] ?? params["surface"]
+            if rawSurface != nil {
+                surfaceId = v2UUIDAny(rawSurface)
                 guard surfaceId != nil else {
-                    result = .err(code: "not_found", message: "Surface not found for the given surface_id", data: nil)
+                    result = .err(code: "not_found", message: "Surface not found for the given surface_id/surface_ref/surface", data: nil)
                     return
                 }
             } else {


EtanHey:
Reviewed in the context of a session-mining sweep on 2026-04-23 where we surfaced the exact symptom this PR fixes: `cmux send --surface surface:N` being silently retargeted to the caller's own pane when the ref wasn't a UUID.

The patch looks right to me. A few observations from the code-review side:

- **Pattern is clean.** Four identical handler patches against `Sources/TerminalController.swift` (`v2SurfaceSendText`, `v2SurfaceSendKey`, `v2SurfaceClearHistory`, `v2SurfaceReadText`). Each replaces the `params["surface_id"] != nil` check with a `surface_id ?? surface_ref ?? surface` fallback chain and routes through the existing `v2UUIDAny` helper (already used elsewhere in the file, e.g. ~line 3136). No new helpers, no new dependencies — the blast radius is contained.
- **Error message improvement is a win on its own.** `"Surface not found for the given surface_id/surface_ref/surface"` is self-describing; callers hitting the new error path will know all three accepted keys.
- **Backwards-compat is solid.** Existing `surface_id` callers still hit the chain first and see the same behavior. Response shape unchanged. No change to `tab_id` handling.
- **Symmetry with the response shape.** Closing the input/output asymmetry (response always returns both `surface_id` and `surface_ref`, inputs now accept both plus the short `surface` key) is the architectural payoff — it eliminates a whole class of silent-retargeting bugs, not just this one.

**What I'd like to see before merge** (nothing blocking, just flagging):

1. **Xcode build verification.** Author notes they don't have a local Xcode toolchain; a maintainer run would close the loop. The change is mechanical so confidence is high, but Swift's type system could catch something subtle (e.g. `v2UUIDAny` signature specifics on optional handling that differ from `v2UUID`'s two-arg form).
2. **Test coverage.** The PR body mentions `tmux_split_ref_test.go`-style coverage as a follow-up. For a fix to an asymmetric-targeting bug where the old behavior was *silent* (not an error), a regression test is high-value — without one, a future refactor could re-introduce the fallback-to-focusedPanelId default and CI would stay green. Happy to stub the test shape on a follow-up PR if that helps.
3. **Consider also covering read paths.** `v2SurfaceReadText` is in scope here (good). If there are related read/inspect RPCs (`surface.info`, enumerate, etc.) with the same `params["surface_id"]`-only check, worth grep'ing before merge so we don't leave half the surface area silently wrong.

From the cmuxlayer side (the TypeScript MCP wrapper that sits above this), this unblocks `send_to` / `send_to_agent` routing to work correctly across workspaces without the wrapper having to pre-resolve refs to UUIDs. Strong +1 on shipping.

— Context: [session-mining 2026-04-23](https://github.com/EtanHey/cmuxlayer/pull/86) surfaced three bugs in this class; this PR covers Bug 1 directly.

Issue #770, opened 2026-03-03
Split pane click activates wrong window (distinct from #616 sporadic window creation)


gx-hidemi-ito:
## Summary

Clicking a non-focused split pane within the **same workspace** causes the app to switch to a **different window** instead of simply activating the clicked pane. In some cases, the split operation itself creates an entirely new window.

**This is a distinct issue from #616** (sporadic new window creation from in-app clicks). While #616 covers _spontaneous new window creation_, this issue describes a different symptom: **focus-driven window switching** during normal split pane interaction, and **intermittent new window creation on split**.

### Related but distinct issues

| Issue | Symptom | How this differs |
|-------|---------|------------------|
| #616 | Sporadic new window creation from clicks | This issue is about **switching to an existing window** on pane click, not always creating a new one |
| #301 | Workspace switch spawns new empty window | Triggered by workspace switch, not by clicking a split pane within a workspace |
| #708 | `+` opens workspace in wrong window | Trigger is the `+` button, not split pane focus |
| #394 (fixed) | Actions routing to wrong window | Fixed keyboard shortcut routing, but **mouse click focus path is different** |

---

## Steps to Reproduce

1. Open cmux with a single window
2. Create a workspace and split a pane (Cmd+D for vertical or Cmd+Shift+D for horizontal)
3. Click on the non-focused split pane to activate it

**Observed behavior (intermittent, two variants):**
- **Variant A**: A _different existing window_ becomes key/active instead of the clicked pane gaining focus within the same window
- **Variant B**: The split operation itself creates a _new unwanted window_, and subsequent pane clicks switch between these windows

**Expected behavior:** Clicking a split pane should only change focus within the same window. No new windows should be created during split operations.

---

## Root Cause Analysis

After tracing the source code, I identified the following mechanisms:

### Primary: `ensureFocus()` → `makeKeyAndOrderFront` → `setActiveMainWindow` chain

When a split pane is clicked, the following call chain executes:

```
mouseDown (GhosttyTerminalView.swift:~3434)
  → window.makeFirstResponder(self)
    → becomeFirstResponder (GhosttyTerminalView.swift:~2811)
      → onFocus() callback
        → Workspace.focusPanel() (Workspace.swift:~2774)
          → TerminalPanel.focus() (TerminalPanel.swift:~105)
            → ensureFocus() (GhosttyTerminalView.swift:~4768)
```

Inside `ensureFocus()` (~line 4768):
```swift
if !window.isKeyWindow {
[...]

PR #3176, opened 2026-04-27
fix: resolve event window's tabManager for numeric workspace/surface shortcuts

jaehoonkim-x:
## Summary

When two or more windows are open, pressing `Cmd+number` (select workspace) or `Ctrl+number` (select surface) could activate a tab in the **wrong window**.

The shortcuts used `AppDelegate.tabManager` — a cached weak reference that tracks the last-active window — instead of the window the key event actually came from. If the cached pointer drifted to window B while the user was working in window A, only shortcuts whose index existed in window B would misbehave; higher-indexed shortcuts silently fell through (`selectTab(at:)` has an out-of-range guard), making the bug appear to affect only specific numbers.

This is the keyboard-shortcut analogue of #770 (split-pane click activates wrong window).

## What changed

- `selectWorkspaceByNumber` and `selectSurfaceByNumber` shortcut handlers now resolve the `TabManager` from the key event's window using the existing `contextForMainWindow(mainWindowForShortcutEvent(event))` pattern — the same approach already used by sidebar and other shortcut handlers.
- Falls back to the cached `tabManager` only when no window context can be resolved from the event (preserves existing behaviour for single-window setups).

## Verification

- [x] `xcodebuild -scheme cmux -configuration Debug` passes (no new errors or warnings)
- [ ] Open two windows, each with multiple workspaces → `Cmd+1`…`Cmd+N` stays within the focused window
- [ ] Surface shortcuts (`Ctrl+1`…`Ctrl+N`) likewise stay within the focused window
- [ ] Single-window behaviour unchanged

## Test note

A unit test for this path requires AppKit multi-window infrastructure (two real `NSWindow` instances with live `TabManager` objects wired to `AppDelegate.mainWindowContexts`). No existing test harness covers this, and a fake structural test would violate the project's test quality policy. Manual verification with two windows is the practical check here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric shortcuts for workspace and surface selection now consistently target the correct window context, including the “9 = last” behavior.
  * Improved shortcut routing so numeric selections work reliably across different window situations.

---
## Summary by cubic
Fixes Cmd+number (workspace) and Ctrl+number (surface) so they act in the focused window when multiple windows are open.

- **Bug Fixes**
  - Resolve the `TabManager` from the shortcut event’s window for `selectWorkspaceByNumber` and `selectSurfaceByNumber`, falling back to the cached manager only when no window context is found.

<sup>Written for commit f1379eba02381e8d6503b92c6a5c9c94e524c512. Summary will update on new commits.</sup>

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -10477,7 +10477,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         // Always consume the event when the digit matches to prevent Ghostty's
         // goto_tab fallback from creating a new window when the index is out of bounds.
         if let digit = numberedConfiguredShortcutDigit(event: event, action: .selectWorkspaceByNumber) {
-            if let manager = tabManager,
+            let manager = contextForMainWindow(mainWindowForShortcutEvent(event))?.tabManager ?? tabManager
+            if let manager,
                let targetIndex = WorkspaceShortcutMapper.workspaceIndex(forDigit: digit, workspaceCount: manager.tabs.count) {
 #if DEBUG
                 cmuxDebugLog(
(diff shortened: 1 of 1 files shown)


jaehoonkim-x:
Closing to reopen with a clean commit history.

Issue #1835, opened 2026-03-20
Feature Request: Font settings in Settings UI


flor3z-github:
## Current

Changing the terminal font requires manually editing `~/.config/ghostty/config`. Users who don't already use Ghostty have to learn its config format just to pick a font. cmux Settings already has controls for sidebar colors, notification sounds, and workspace colors — font settings are a natural gap.

## What to build

- Add **Font family** (monospace font picker) and **Font size** (numeric stepper) to Settings
- Read current values from Ghostty config on launch, write back on change
- No new config layer — `~/.config/ghostty/config` stays the single source of truth
- Auto-create config file/directory if missing
- Trigger config reload after write (same path as Cmd+Shift+,)

## Implementation notes

- **UI**: Add a "Terminal Font" section following the existing `SettingsCardRow` pattern in `Sources/cmuxApp.swift:5712`. Place it near the Sidebar Appearance section (`Sources/cmuxApp.swift:4859`). Use `NSFontManager` or a filtered font list for monospace families.

```
 ┌───────────────────────────────────────────────────────────────────┐
 │  Font                                         [Monaco ▾] [12 ⬍]   │
 │  Choose the font used in the terminal.                            │
 └───────────────────────────────────────────────────────────────────┘
```

- **Read**: Extend the Ghostty config parsing at `Sources/GhosttyConfig.swift:225` (`parse()` method — already parses `font-family` / `font-size`) to expose current values to the Settings view.
- **Write**: Line-based read-modify-write on `~/.config/ghostty/config` (insert if absent, replace if present). Ghostty config is `key = value` per line, no nested structure. Config file paths defined at `Sources/GhosttyConfig.swift:196`.
- **Reload**: Call `GhosttyApp.shared.reloadConfiguration(source:)` — same path used by Cmd+Shift+, handler at `Sources/AppDelegate.swift:9151` and the reload implementation at `Sources/GhosttyTerminalView.swift:1561`. Invalidate config cache via `GhosttyConfig.invalidateLoadCache()` (`Sources/GhosttyConfig.swift:77`).

## References

- Ghostty config format: `font-family = SF Mono`, `font-size = 13`
- Settings UI patterns: `Sources/cmuxApp.swift:4307-5329` (App/Notifications/Workspace Colors/Sidebar sections)
- Config parsing: `Sources/GhosttyConfig.swift:225-312` (`parse()` method)
[...]

PR #3302, opened 2026-04-29
Add sidebar.titleFontSize setting for workspace title

legenduck:
## Summary

Adds a `sidebar.titleFontSize` key to `~/.config/cmux/settings.json` that controls the point size of the **workspace title text** in each sidebar tab. Range 9–24pt, default 12.5pt (current hardcoded value).

Refs #1835 — that issue asks for general font settings in the Settings UI; this is a smaller first step that exposes the most-requested knob via the JSON settings file. Settings UI surface for it can come later.

## Why this scope

I limited the setting to the **workspace title only** (the prominent `Text(workspaceSnapshot.title)` row at `Sources/ContentView.swift:12690`). Secondary detail rows (branch / cwd / ports / log / notification message) keep their fixed sizes so the existing sidebar tab layout and width policy stay untouched, which keeps the PR small and reduces regression risk in `SidebarResize*UITests`.

If reviewers prefer a single `sidebar.fontSize` that scales every text row in the tab proportionally, happy to follow up — but that touches ~15 `.font(.system(size: …))` call sites and may need width-policy adjustments, so I figured a focused PR was a safer first cut.

## Implementation

- `web/data/cmux-settings.schema.json` — new `sidebar.titleFontSize` property (number, 9–24, default 12.5)
- `Sources/KeyboardShortcutSettingsFileStore.swift` — whitelist entry, parser that silently clamps out-of-range numeric values (matching the `sidebarAppearance.tintOpacity` precedent in the same function), and an entry in `defaultTemplate()` so newly-generated `settings.json` files surface the key alongside the other sidebar entries
- `Sources/TabManager.swift` — new `SidebarTabTitleFontSettings` enum next to the other sidebar settings namespaces, with `userDefaultsKey`, `defaultSize`, `minSize`, `maxSize`, and a `resolvedSize(defaults:)` helper
- `Sources/ContentView.swift` — adds `titleFontSize: Double` to `SidebarTabItemSettingsSnapshot`, populates it in `init`, and uses it in `TabItemView`'s title `.font(...)` modifier in place of the hardcoded `12.5`

Hot reload works through the existing path: file watcher → `managedUserDefaults` → `UserDefaults.didChangeNotification` → `SidebarTabItemSettingsStore.refreshSnapshot()` → SwiftUI re-render. `TabItemView`'s `Equatable` conformance compares `settings`, so the new `Double` field flows through the snapshot diff automatically (no `==` rewrite needed).

## Test plan

- [ ] Manual: write `{ "sidebar": { "titleFontSize": 18 } }` to `~/.config/cmux/settings.json`, observe sidebar tab titles render at 18pt
- [ ] Manual: change the value while the app is running, verify hot reload re-renders without restart
- [ ] Manual: out-of-range values (e.g. `5`, `100`) clamp silently to 9–24
- [ ] Manual: a fresh launch with no `settings.json` writes a template that includes the new `titleFontSize` line under `sidebar`
- [ ] No regression in `cmuxTests/SidebarWidthPolicyTests.swift` — title font is still inside the same `HStack` with `layoutPriority(1)` and width policy is unchanged

## Notes

- Did not add a parser unit test because there's no existing precedent for testing `parseSidebarSection` end-to-end and CLAUDE.md's test-quality policy discourages source-shape tests; happy to add a runtime-behavior test if reviewers point me at the right harness.
- Did not touch `CHANGELOG.md` since the release flow (`/release`) generates the entry from commits.

### DIFF
--- a/Sources/ContentView.swift
+++ b/Sources/ContentView.swift
@@ -9467,6 +9467,7 @@ private struct SidebarTabItemSettingsSnapshot: Equatable {
     let selectionColorHex: String?
     let notificationBadgeColorHex: String?
     let visibleAuxiliaryDetails: SidebarWorkspaceAuxiliaryDetailVisibility
+    let titleFontSize: Double
 
     init(defaults: UserDefaults = .standard) {
         sidebarShortcutHintXOffset = Self.double(
@@ -9523,6 +9524,7 @@ private struct SidebarTabItemSettingsSnapshot: Equatable {
         activeTabIndicatorStyle = SidebarActiveTabIndicatorSettings.current(defaults: defaults)
         selectionColorHex = defaults.string(forKey: "sidebarSelectionColorHex")
         notificationBadgeColorHex = defaults.string(forKey: "sidebarNotificationBadgeColorHex")
+        titleFontSize = SidebarTabTitleFontSettings.resolvedSize(defaults: defaults)
     }
 
     private static func bool(
@@ -12688,7 +12690,7 @@ private struct TabItemView: View, Equatable {
                 }
 
                 Text(workspaceSnapshot.title)
-                    .font(.system(size: 12.5, weight: titleFontWeight))
+                    .font(.system(size: settings.titleFontSize, weight: titleFontWeight))
                     .foregroundColor(activePrimaryTextColor)
                     .lineLimit(1)
                     .truncationMode(.tail)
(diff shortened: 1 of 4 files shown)


legenduck:
On the `logInvalid` for non-numeric `titleFontSize` — the bots are pulling in opposite directions here, so flagging this for a human call.

The first review (Greptile, on `caed93b`) flagged the original `else if logInvalid` branch as inconsistent with the silent-clamp behavior on out-of-range numbers, and pointed at the closest existing precedent in the same function — `sidebarAppearance.tintOpacity` at L675–678 — which is fully silent on both invalid types and out-of-range values. I dropped the branch in `881dae0` to match that precedent.

CodeRabbit's nit on this commit asks to put the branch back so non-numeric input emits a diagnostic. Reasonable concern on its own, but it's exactly the inconsistency Greptile flagged.

Two coherent end states, happy with either:

1. **Keep current** — silent on both invalid type and clamp, matching `tintOpacity` (status quo of this PR).
2. **Log on both** — restore the `else if logInvalid` branch *and* add a log when `clamped != value`, so any malformed `titleFontSize` value is surfaced. This is the fully-consistent superset of both bots' suggestions.

Leaning (1) since `tintOpacity` is the most analogous existing parser, but happy to switch to (2) on request.

PR #3443, opened 2026-05-02
Show Toggle Pane Zoom for browser panes

### LINKED ISSUE #3439: Toggle Pane Zoom is hidden for browser panes in command palette
## Summary

`Toggle Pane Zoom` is incorrectly hidden in command palette when the focused pane is a browser pane.

## Repro

1. Create a split workspace with:
   - one terminal pane
   - one browser pane
2. Focus the terminal pane
3. Open command palette and search `Toggle Pane Zoom`
4. Command is visible
5. Focus the browser pane
6. Open command palette and search `Toggle Pane Zoom`
7. Command is missing

## Expected

`Toggle Pane Zoom` should be available whenever:
- a pane is focused
- the workspace has splits

It should not depend on the focused pane being terminal-only.

## Actual

The command is gated behind terminal-pane context, so it disappears for browser panes even though the action is still valid.

## Likely Cause

The command-palette visibility check is too narrow.

Current behavior appears equivalent to:

- `panelIsTerminal == true`
- `workspaceHasSplits == true`

But the action itself works at the workspace/pane level, so the condition should be closer to:

- `hasFocusedPanel == true`
- `workspaceHasSplits == true`

## Environment

- cmux v0.63.2
- macOS Tahoe 26.x
- Apple Silicon

honghua:
## Summary

Fix command palette visibility for `Toggle Pane Zoom` so it is available when a browser pane is focused.

## Problem

The command was previously gated behind terminal-only context:

- `panelIsTerminal == true`
- `workspaceHasSplits == true`

That meant the command disappeared in browser-pane context even though the underlying action still works at the workspace/pane level.

## Fix

Relax the visibility condition to:

- focused panel exists
- workspace has splits

Also switch the subtitle to the workspace-level label so the command reads naturally outside terminal-only context.

## Testing

- Added `ToggleSplitZoomCommandPaletteVisibilityTests`
- Ran:
  - `./scripts/test-unit.sh -only-testing:cmuxTests/ToggleSplitZoomCommandPaletteVisibilityTests test`

Closes #3439

---
## Summary by cubic
Make “Toggle Pane Zoom” visible in the Command Palette when a browser pane is focused, not just terminals. Prevents the command from disappearing even though the action works at the workspace level.

- **Bug Fixes**
  - Show when any panel is focused and the workspace has splits (`shouldShowToggleSplitZoomCommandInPalette`).
  - Use the workspace-level subtitle via `workspaceSubtitle` for consistent wording.
  - Add unit tests for the new visibility logic.

<sup>Written for commit 0fcbfff09807922eef9a3cb8b9f2925799760128. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * "Toggle Split Zoom" now appears in the command palette whenever a panel is focused and the workspace has split panes, not limited to terminal panels.
  * Command subtitle now updates dynamically to reflect the current workspace context.

### DIFF
--- a/Sources/ContentView.swift
+++ b/Sources/ContentView.swift
@@ -1613,6 +1613,13 @@ private final class SelectedWorkspaceDirectoryObserver: ObservableObject {
     }
 }
 
+func shouldShowToggleSplitZoomCommandInPalette(
+    hasFocusedPanel: Bool,
+    workspaceHasSplits: Bool
+) -> Bool {
+    hasFocusedPanel && workspaceHasSplits
+}
+
 struct ContentView: View {
     @ObservedObject var updateViewModel: UpdateViewModel
     let windowId: UUID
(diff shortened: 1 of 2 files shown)


honghua:
Followed up the review feedback in `0fcbfff0`:
- renamed the helper test to match the actual boolean inputs
- added a doc comment to the extracted helper

At this point the remaining red checks are the two Vercel deploy statuses, which are failing on team authorization for this fork PR rather than on a code/build issue.

PR #3606, opened 2026-05-06
fix: close all targeted tabs in pane right-click 'Close Other Tabs' (#3605)

### LINKED ISSUE #3605: REGRESSION: close other tab is broken
### cmux version

0.64.2

### macOS version

26.4.1

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Direct download (DMG)

I could not test NIGHTLY

### Bug description

if i click on a tab and then "close other tabs" only one tab gets closed.

### Expected behavior

all other tabs in the workspace should close

### Steps to reproduce

1.open cmux
2. open a workspace
3. open multiple tabs
4. click "close other tabs" 
5. confirm closing

### Shell and environment

_No response_

```text

```

_No response_

### Additional context

_No response_

psh4607:
## Summary

Right-click **Close Other Tabs** / **Close Tabs to Right** / **Close Tabs to Left** on a panel tab regressed: when every tab in the pane wants confirm-on-close (e.g. each runs a shell), only **one** of the targeted tabs ever closes. The user sees their confirmation, one tab disappears, and the rest silently stay open. Closes #3605.

The keyboard shortcut path (`Cmd+Option+T`, [`TabManager.closeOtherTabsInFocusedPaneWithConfirmation`](Sources/TabManager.swift)) was not affected.

## Root cause

`Workspace.closeTabs(_:)` iterated `tabIds` and called `bonsplitController.closeTab(tabId)` directly without setting `forceCloseTabIds`. The first call kicks off `Workspace.shouldCloseTab`'s async per-panel `beginSheetModal` confirmation, which sets `TabManager.isCloseConfirmationInFlight = true`.

PR #2989 (\`a5a1cafb\` — *Fix close confirmation bypass when spamming close*) added an early-return on that flag at the top of `shouldCloseTab` to fix close-spam. As a side effect, every subsequent call in the loop returned `false` synchronously and was never retried, so only the first tab closed after the user confirmed it.

## Fix

`Workspace.closeTabs(_:)` now matches the working keyboard-shortcut path:

1. Bail early if a confirmation is already on-screen, preserving the PR #2989 anti-spam guard for unrelated flows.
2. Filter pinned panels up front.
3. If any candidate panel needs confirmation, show **one** batched dialog listing every tab title — using the existing localized `dialog.closeOtherTabs.title` / `.message.one` / `.message.other` strings.
4. After the user confirms (or if no panel needs confirmation), insert each candidate into `forceCloseTabIds` and call `closeTab`. If `bonsplitController.closeTab` returns false, remove the stale entry so we never leak `forceCloseTabIds` bypass state.

This keeps the keyboard-shortcut path and the right-click context menu *semantically* aligned without unifying anchor calculation: the keyboard shortcut still closes everything except the focused/selected tab, while the context menu still closes relative to the right-clicked tab.

## Two-commit structure

Per AGENTS.md *Regression test commit policy*:

1. **\`9bfe2078\`** — Adds the failing regression test only. CI should go red.
2. **\`6362f8ff\`** — Adds the fix. CI should go green.

The new test (`cmuxTests/WorkspaceCloseTabsContextMenuTests.swift`) drives `splitTabBar(_:didRequestTabContextAction:for:inPane:)` directly for `.closeOthers` and `.closeToRight`, with three terminal panels in the same pane all marked \`setNeedsConfirmCloseOverrideForTesting(true)\`, and a `confirmCloseHandler` injected into the `TabManager`. Without the fix only the first panel ever leaves the pane (and only after async drain); with the fix all targeted panels close after a single batched confirmation.

## Manual repro

1. Open a workspace, open 3+ tabs in the same pane (each with a running shell).
2. Right-click any tab → *Close Other Tabs*.
3. Confirm.

Before this PR: only one of the other tabs closes.
After this PR: all other tabs close.

## Notes

- I'm running on a borrowed work tree where local Zig is `0.16.0` so `xcodebuild -scheme cmux-unit` can't link the Ghostty CLI helper here. Swift compilation of the changed files is clean (LSP) and the change is intentionally minimal Swift; I'm relying on CI to run `cmuxTests` end-to-end.
- The fix benefits all three context-menu actions (`.closeOthers`, `.closeToLeft`, `.closeToRight`) because they share `closeTabs(_:)`.

---
## Summary by cubic
Fixes a regression where right‑clicking “Close Other Tabs” or “Close Tabs to Left/Right” only closed one tab when all tabs required confirm-on-close. Now a single, batched confirmation closes all targeted tabs, and both keyboard and context‑menu paths show the same localized dialog.

- **Bug Fixes**
  - Updated `Workspace.closeTabs(_:)`: bail if a confirmation is in flight, skip pinned tabs, show one batched dialog using `dialog.closeOtherTabs.*`, then force-close each tab with cleanup on failure.
  - Applies to `.closeOthers`, `.closeToLeft`, and `.closeToRight`; keyboard shortcut logic is unchanged, and `TabManager.closeOtherTabsInFocusedPaneWithConfirmation` now uses the same localized strings.
  - Added `WorkspaceCloseTabsContextMenuTests` covering all three actions and cancellation to prevent regressions.

<sup>Written for commit 721a17257f40e3501cba6ae6f69f2f3fb7ca9d4d. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added user confirmation prompts when closing multiple tabs from context menus to reduce accidental closures.

* **Refactor**
  * Dialog text for tab-closing confirmations replaced with localized strings to support translations and pluralization.

* **Tests**
  * Added regression tests covering context-menu tab-closing actions, confirmation flows, and cancellation scenarios.

### DIFF
--- a/GhosttyTabs.xcodeproj/project.pbxproj
+++ b/GhosttyTabs.xcodeproj/project.pbxproj
@@ -292,6 +292,7 @@
 		B900002FA1B2C3D4E5F60719 /* CMUXCLI+ThemeSupport.swift in Sources */ = {isa = PBXBuildFile; fileRef = B900002DA1B2C3D4E5F60719 /* CMUXCLI+ThemeSupport.swift */; };
 		B9000035A1B2C3D4E5F60719 /* CMUXCLI+DocsSettings.swift in Sources */ = {isa = PBXBuildFile; fileRef = B9000031A1B2C3D4E5F60719 /* CMUXCLI+DocsSettings.swift */; };
 		D7AB00000000000000000011 /* WorkspaceAdjacentPaneMoveTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D7AB00000000000000000012 /* WorkspaceAdjacentPaneMoveTests.swift */; };
+		D7AB36050000000000000001 /* WorkspaceCloseTabsContextMenuTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D7AB36050000000000000002 /* WorkspaceCloseTabsContextMenuTests.swift */; };
 		B9000033A1B2C3D4E5F60719 /* CMUXCLI+TopRendering.swift in Sources */ = {isa = PBXBuildFile; fileRef = B9000032A1B2C3D4E5F60719 /* CMUXCLI+TopRendering.swift */; };
 		B9000012A1B2C3D4E5F60719 /* AutomationSocketUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B9000011A1B2C3D4E5F60719 /* AutomationSocketUITests.swift */; };
 		FEED0000000000000000F009 /* FeedSidebarUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = FEED0000000000000000F008 /* FeedSidebarUITests.swift */; };
@@ -722,6 +723,7 @@
 		B900002DA1B2C3D4E5F60719 /* CMUXCLI+ThemeSupport.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "CMUXCLI+ThemeSupport.swift"; sourceTree = "<group>"; };
 		B9000031A1B2C3D4E5F60719 /* CMUXCLI+DocsSettings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "CMUXCLI+DocsSettings.swift"; sourceTree = "<group>"; };
 		D7AB00000000000000000012 /* WorkspaceAdjacentPaneMoveTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WorkspaceAdjacentPaneMoveTests.swift; sourceTree = "<group>"; };
+		D7AB36050000000000000002 /* WorkspaceCloseTabsContextMenuTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WorkspaceCloseTabsContextMenuTests.swift; sourceTree = "<group>"; };
 		B9000032A1B2C3D4E5F60719 /* CMUXCLI+TopRendering.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "CMUXCLI+TopRendering.swift"; sourceTree = "<group>"; };
 		B9000011A1B2C3D4E5F60719 /* AutomationSocketUITests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AutomationSocketUITests.swift; sourceTree = "<group>"; };
 		FEED0000000000000000F008 /* FeedSidebarUITests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FeedSidebarUITests.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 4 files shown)


psh4607:
Pushed `721a1725` to address review feedback.

## Changes in `721a1725`

1. **Added `testCloseToLeftClosesAllTabsLeftOfAnchorWhenAllNeedConfirmation`** — the third context-menu action is now exercised by the regression suite (covers Copilot's docstring/coverage comment, Greptile P2, and CodeRabbit's nitpick).
2. **Unified the dialog copy between the keyboard-shortcut path and the context-menu path.** `TabManager.closeOtherTabsInFocusedPaneWithConfirmation` now uses the same `dialog.closeOtherTabs.title` / `dialog.closeOtherTabs.message.one` / `.message.other` localized keys that `Workspace.closeTabs` already uses, so both entry points emit identical copy in every supported locale (closes the string-drift half of Greptile P1).

## Reviewer concerns I'm intentionally not addressing in this PR

### Greptile P1 (full architecture refactor)

> The close-confirmation contract (in-flight flag, `forceCloseTabIds` management, dialog construction) is now split between `TabManager.closeOtherTabsInFocusedPaneWithConfirmation` and `Workspace.closeTabs`.

Agreed that the *long-term* fix is a single `TabManager.confirmAndClose(tabIds:in:)` helper that both entry points route through. I'd prefer to keep that out of this PR though:

- AGENTS.md *Bugfix rule*: "fix MINIMALLY. NEVER refactor while fixing." This PR is a regression fix and the two-commit policy is meant to keep it auditable.
- The keyboard path and the context-menu path have genuinely different anchor calculation (focused/selected tab vs. right-clicked tab), pinned-pane filtering rules, and pane scoping. Extracting a clean shared helper without subtly changing semantics deserves its own PR.
- The string-drift half of the concern — which is what would have made the next regression silently slip through — is fixed in `721a1725`.

If a maintainer wants the unification done in this PR I'm happy to do it; otherwise I'll file a follow-up issue once this lands.

### Copilot inline: `confirmManager == nil` silent no-op

The `guard let confirmManager else { return }` only fires when **both** `owningTabManager` *and* `AppDelegate.shared?.tabManager` (and `tabManagerFor(tabId:)`) are nil. In a real app this can't happen — every `Workspace` is attached to a `TabManager` before it's wired into Bonsplit, which is the only thing that can deliver a `splitTabBar(_:didRequestTabContextAction:…)` callback.

The unit tests in `WorkspaceCloseTabsContextMenuTests` use `TabManager()` (whose default workspace has `owningTabManager` set), so they hit the normal manager path too. A truly-detached `Workspace` instance with no TabManager and no AppDelegate would not have a window to host an `NSAlert` either, so silently dropping the close is the safer default than force-closing without confirmation.

Happy to add an inline `assertionFailure` in DEBUG if reviewers want a louder signal, but I don't think a real fallback is justified.

psh4607:
@coderabbitai full review

The original `CHANGES_REQUESTED` review was filed against `6362f8ff` for the missing `.closeToLeft` test case. `721a1725` adds that test plus addresses Greptile's string-drift concern. All 5 review threads are now resolved. Please re-evaluate the latest commit.

psh4607:
Closing in favor of #3628, which uses a cleaner abstraction (`CloseOtherTabsConfirmationPrompt` struct + shared `requestCloseTab` helper) and respects the swift-file-length-budget by splitting into a separate file.

Leaving this PR as reference — the root-cause analysis in the description (tracing the regression to commit `a5a1cafb` / PR #2989, \"Fix close confirmation bypass when spamming close\") is more detailed than #3628's, and the cancel-path / no-confirm-path test cases here aren't in #3628 yet. I'll open a follow-up PR with just those two tests after #3628 lands — see review comment there.

PR #3621, opened 2026-05-06
Surface auth sign-in errors in Settings (#3617)

### LINKED ISSUE #3617: Settings → Sign In… opens auth sheet that closes immediately (no page rendered)
https://github.com/user-attachments/assets/06bbf633-a22c-4218-aa7b-7ce63e957dc4

## Summary

Clicking **Settings → Account → Sign In…** on the Release DMG opens an `ASWebAuthenticationSession` sheet that flashes briefly and closes immediately without rendering any page (no Stack Auth sign-in form, no error). The app remains in the "Not signed in" state and no user-visible error is shown.

The feature was added in #3027 (merged 2026-04-20) and shipped in v0.64.0 (2026-05-05) but is **not mentioned anywhere in `CHANGELOG.md`** — `grep -in "sign\|account\|login" CHANGELOG.md` only matches unrelated entries (browser-pane Passkey support, Settings sidebar shell, Claude auth env, Google sign-in in browser panes). I could not find any prior issue or discussion describing this symptom.

## Environment

- cmux: **0.64.3 (build 83)** [`aea6cfcde`]
- macOS: **26.4.1** (build 25E253)
- Build source: Release DMG from GitHub releases
- Default browser: **not Safari** (other browser used as default)
- Pre-existing cmux.com session in Safari WebKit cookies: **none** (never signed in to https://cmux.com prior to clicking Sign In…)

## Steps to reproduce

1. Install the Release DMG (latest from https://github.com/manaflow-ai/cmux/releases).
2. Launch cmux → open **Settings → Account**.
3. Click **Sign In…**.

## Expected

`ASWebAuthenticationSession` sheet opens and renders `https://cmux.com/handler/sign-in?after_auth_return_to=…`, i.e. the Stack Auth sign-in form. (`/handler/sign-in` returns HTTP/2 200 with the Stack Auth React payload — verified via `curl`, so the server is healthy.)

## Actual

The sheet briefly opens and closes immediately. No page is rendered. No error is shown to the user. App still says "Not signed in".

## Diagnostic notes

`AuthManager.beginSignIn` swallows all errors via `NSLog` and never surfaces them to the user — see https://github.com/manaflow-ai/cmux/blob/main/Sources/Auth/AuthManager.swift#L200-L217 :

```swift
} completionHandler: { [weak self] callbackURL, error in
    Task { @MainActor [weak self] in
        ...
        if let error {
            NSLog("auth.webauth failed: %@", "\(error)")  // invisible to the user
            return
        }
        ...
    }
}
```

Localization keys `settings.account.error.invalidCallback` ("로그인 콜백이 잘못되었습니다.") and `settings.account.error.missingAccessToken` already exist in `Resources/Localizable.xcstrings` but are never displayed in any view; the `AuthSettingsRow` UI in `Sources/cmuxApp.swift` only reads `authManager.isAuthenticated` and `authManager.currentUser` and has no error surface.

I was not able to capture the underlying `NSError` from Console.app on this attempt (the sheet closed before I could switch to Console). To anyone repro-ing: open Console.app first, filter on `process:cmux` plus subsystem `com.apple.AuthenticationServices`, **then** click Sign In… — the failure event is logged after the sheet dismisses.

## Suspected root causes (ranked)

The "no page rendered" detail is the smoking gun — it rules out anything that depends on a page actually loading (manual click on "Return to cmux", token parsing in `AuthCallbackRouter`, etc.). Plausible mechanisms:

1. **Most likely** — `ASWebAuthenticationSessionError.canceledLogin` (code 1) firing before any page renders. Triggers:
   - macOS consent dialog ("cmux.app wants to use cmux.com to sign in. This will allow the app and website to share information about you.") was dismissed (clicked Cancel, lost focus, or stolen by another window).
   - Sandbox / TCC policy blocks the private `com.apple.WebAuthentication` capability for the signed Release bundle.
   - First-request network failure (proxy / VPN / captive portal blocking `cmux.com`).

2. **Possible** — Double-fire of `beginSignIn()` from a SwiftUI body re-evaluation (e.g. settings search re-mounting `AuthSettingsRow`) cancels the in-flight session via `webAuthSession?.cancel()` at https://github.com/manaflow-ai/cmux/blob/main/Sources/Auth/AuthManager.swift#L188-L191 . There is no debounce / in-flight guard in `buttonAction()` (https://github.com/manaflow-ai/cmux/blob/main/Sources/cmuxApp.swift#L7583-L7591).

3. **Less likely on Release** — `AuthManager.shared` / `webAuthSession` reference dropped due to lifecycle churn.

I ruled out web-side auto-redirect bugs:

- `grep useEffect|window.location|router.push|http-equiv|<meta refresh` on `web/app/handler/after-sign-in/` returns zero matches; `OpenNativeClient` is a static page with a manual `<a href>` link only.
- `curl https://cmux.com/handler/sign-in?...` returns HTTP/2 200 with the Stack Auth React payload.
- `curl https://cmux.com/handler/after-sign-in?...` (no auth cookie) returns HTTP/2 200 with `redirect("/")` (homepage, not a `cmux://` deep link), so the after-sign-in page cannot accidentally close the sheet for an unauthenticated user.

## Suggested fix

Make `AuthManager.beginSignIn`'s error path visible to the user instead of silently `NSLog`-ing it. Concretely:

1. Add `@Published var lastSignInError: AuthManagerError?` to `AuthManager`.
2. Set it on the failure path of the completion handler (and on `handleCallbackURL` throws).
3. Render the localized error message under the **Sign In…** button in `AuthSettingsRow` (`Sources/cmuxApp.swift`), reusing the existing `settings.account.error.*` localization keys.

This is universally good — it fixes the next user too regardless of which of the three root causes above is the actual one in any given case, and turns an invisible failure into a debuggable one.

A separate UX nit: `web/app/handler/after-sign-in/OpenNativeClient.tsx` relies on the user manually clicking "Return to cmux"; the page has no `useEffect`/`window.location` auto-redirect to the deep link. Worth either auto-redirecting or making the manual click more obvious.

## Related

- #3027 — Bring minimal macOS auth scaffolding to main (merged) — adds the flow.
- #3044 — Raise all cmux windows on auth callback, Settings on top (open) — fixes a different post-callback UX bug, but touches the same `AppDelegate.application(_:open:)` area.

## Willing to PR

Yes — happy to send a PR for the visible-error UX fix above.

psh4607:
## Summary

Closes #3617.

`AuthManager.beginSignIn` and `handleCallbackURL` previously swallowed every failure via `NSLog` only. A user clicking **Settings → Sign In…** who hit any of `ASWebAuthenticationSessionError` (canceled consent, sandbox/TCC, network unreachable…), `handleCallbackURL` throwing `.invalidCallback` because the after-sign-in handler emitted a malformed deep link, or `session.start()` returning false saw the auth sheet flash and close with no visible error and no way to debug. The app stayed at "Not signed in" silently.

This PR publishes the most recent failure as `AuthManager.lastSignInError` and renders it in red below the Account row subtitle in `AuthSettingsRow`, so the failure mode that motivated #3617 is no longer invisible — regardless of which underlying root cause hits in any given environment.

## What changed

- `Sources/Auth/AuthManager.swift`
  - `AuthManagerError` is now `Equatable` and gains a `signInSessionFailed` case for non-`AuthManagerError` failures (network, sandbox, etc.).
  - New `@Published private(set) var lastSignInError: AuthManagerError?` published so SwiftUI can render it.
  - `beginSignIn` clears `lastSignInError` on entry, sets `.signInSessionFailed` on `session.start()` failure, and routes the completion handler's `error` through a new `recordSignInSessionError(_:)` helper that suppresses the error for `ASWebAuthenticationSessionError.canceledLogin` (explicit user cancel) so the UX stays quiet when the user cancels themselves.
  - `handleCallbackURL` sets `lastSignInError = .invalidCallback` before the early throw and otherwise sets it to the typed `AuthManagerError` (or `.signInSessionFailed` fallback) on any throw from `tokenStore.seed` / `refreshSession`. Clears it on success.
- `Sources/cmuxApp.swift` — `AuthSettingsRow` reads `authManager.lastSignInError?.errorDescription` and renders it in red under the existing subtitle. No layout change when no error is set.
- `Resources/Localizable.xcstrings` — adds `settings.account.error.signInSessionFailed` with `en` / `ja` / `uk` / `ko` translations, matching the existing `settings.account.error.*` key pattern.

This deliberately doesn't try to fix the underlying root cause(s) hypothesized in #3617 (`canceledLogin` from a dismissed consent dialog, double-fire from view-body re-evaluation, sandbox denial, network failure). It makes those root causes diagnosable so the *next* PR can target the actual cause once a Console.app log is captured.

## Test plan

`AGENTS.md` regression test policy is two commits:

1. `e0cd032a` — adds the failing regression test `cmuxTests/AuthManagerSignInErrorVisibilityTests.swift`. The test references `AuthManager.lastSignInError`, which doesn't exist yet on this commit, so CI compile fails — proving the test catches the bug.
2. `0cbea3d8` — ships the fix. The test now compiles and asserts `lastSignInError == .invalidCallback` after `handleCallbackURL` throws on a payload-less `cmux://auth-callback` URL.

The CI signal across the two commits is the proof: red on commit 1, green on commit 2.

Manual verification (not run locally per `AGENTS.md` "Never run tests locally"):

- [ ] Settings → Sign In… on a build where the sheet flashes-and-closes now shows `Sign-in didn't complete. Please try again.` under the subtitle.
- [ ] Settings → Sign In…, dismiss the macOS consent dialog explicitly. No error is shown (we suppress `canceledLogin`).
- [ ] Successful sign-in clears any prior error message.
- [ ] Sign Out leaves no stale error message.

## Out of scope (follow-up issues)

- Actual root-cause fix for the sheet flashing-and-closing on Release v0.64.3 — needs a captured `auth.webauth failed: …` line from Console.app, then targeted fix per the canceledLogin / double-fire / sandbox / network branches in #3617.
- `web/app/handler/after-sign-in/OpenNativeClient.tsx` UX nit — auto-redirect to the deep link instead of relying on a manual click. Different reviewer (web), different PR.
- Double-fire guard on `AuthSettingsRow.buttonAction()` (`webAuthSession != nil` / `isLoading == true`). Different concern, separate PR.

---
## Summary by cubic
Surfaced sign-in errors in Settings so users see why auth failed instead of a silent close. Fixes #3617 by publishing the last sign-in error, showing it under Account, and clearing it on success and sign-out.

- **Bug Fixes**
  - AuthManager now publishes `lastSignInError`; `AuthManagerError` is `Equatable` with new `.signInSessionFailed`. Error is set on `session.start()` failure, invalid callback, and other throws; suppressed for `ASWebAuthenticationSessionError.canceledLogin`; cleared on successful flows and on sign-out to avoid stale messages.
  - Settings shows the error message in red under the Account subtitle; no change when there’s no error.
  - Added `settings.account.error.signInSessionFailed` with en/ja/uk/ko translations.
  - Tests: added regression for invalid callback populating the error and another asserting sign-out clears it; tightened error matching.

- **Refactors**
  - Replaced three `NSLog` calls in the sign-in flow with `authLog` (DEBUG-guarded).

<sup>Written for commit d48988d3375fd79a1f8fb98e5c03b36de85df5d3. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Users now see clear sign-in error messages shown in the account/auth settings row when authentication fails.

* **Localization**
  * Added translations for the new sign-in error message in English, Japanese, Ukrainian, and Korean.

* **Tests**
  * Added regression tests to verify sign-in error visibility and lifecycle.

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -110152,6 +110152,35 @@
         }
       }
     },
+    "settings.account.error.signInSessionFailed": {
+      "extractionState": "manual",
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Sign-in didn't complete. Please try again."
+          }
+        },
+        "ja": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "サインインを完了できませんでした。もう一度お試しください。"
+          }
+        },
+        "uk": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Не вдалося завершити вхід. Спробуйте ще раз."
+          }
+        },
+        "ko": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "로그인을 완료하지 못했습니다. 다시 시도해 주세요."
+          }
+        }
+      }
+    },
     "remote.state.connected.vmNoProxy": {
       "extractionState": "manual",
       "localizations": {
--- a/Sources/Auth/AuthManager.swift
+++ b/Sources/Auth/AuthManager.swift
@@ -31,10 +31,11 @@ private final class AuthPresentationContext: NSObject, ASWebAuthenticationPresen
     }
 }
 
-enum AuthManagerError: LocalizedError {
+enum AuthManagerError: LocalizedError, Equatable {
     case invalidCallback
     case missingAccessToken
     case missingRefreshToken
+    case signInSessionFailed
 
     var errorDescription: String? {
         switch self {
@@ -52,6 +53,11 @@ enum AuthManagerError: LocalizedError {
             return String(
                 localized: "settings.account.error.missingRefreshToken",
                 defaultValue: "Account refresh token is unavailable."
+            )
+        case .signInSessionFailed:
+            return String(
+                localized: "settings.account.error.signInSessionFailed",
+                defaultValue: "Sign-in didn't complete. Please try again."
             )
         }
     }
@@ -128,6 +134,7 @@ final class AuthManager: ObservableObject {
     @Published private(set) var isLoading = false
     @Published private(set) var isRestoringSession = false
     @Published private(set) var didCompleteBrowserSignIn = false
+    @Published private(set) var lastSignInError: AuthManagerError?
     @Published var selectedTeamID: String? {
         didSet {
             guard selectedTeamID != oldValue else { return }
@@ -189,6 +196,7 @@ final class AuthManager: ObservableObject {
         loginPollTask?.cancel()
         webAuthSession?.cancel()
         webAuthSession = nil
+        lastSignInError = nil
         isLoading = true
 
         let signInURL = AuthEnvironment.signInURL()
@@ -206,6 +214,7 @@ final class AuthManager: ObservableObject {
                 }
                 if let error {
                     NSLog("auth.webauth failed: %@", "\(error)")
+                    self.recordSignInSessionError(error)
                     return
                 }
                 guard let callbackURL else { return }
@@ -223,8 +232,19 @@ final class AuthManager: ObservableObject {
             webAuthSession = session
         } else {
             NSLog("auth.webauth: session.start() returned false")
+            lastSignInError = .signInSessionFailed
             isLoading = false
         }
+    }
+
+    private func recordSignInSessionError(_ error: Error) {
+        // canceledLogin = explicit user cancel; surfacing an error there is
+        // hostile. Other codes (network, sandbox, etc.) surface to fix #3617.
+        if let asError = error as? ASWebAuthenticationSessionError,
+           asError.code == .canceledLogin {
+            return
+        }
+        lastSignInError = .signInSessionFailed
     }
 
     /// Starts the ASWebAuthenticationSession popup and awaits the user's
@@ -395,18 +415,25 @@ final class AuthManager: ObservableObject {
 
     func handleCallbackURL(_ url: URL) async throws {
         guard let payload = AuthCallbackRouter.callbackPayload(from: url) else {
+            lastSignInError = .invalidCallback
             throw AuthManagerError.invalidCallback
         }
 
         isLoading = true
         defer { isLoading = false }
 
-        await tokenStore.seed(
-            accessToken: payload.accessToken,
-            refreshToken: payload.refreshToken
-        )
-        try await refreshSession()
-        didCompleteBrowserSignIn = true
+        do {
+            await tokenStore.seed(
+                accessToken: payload.accessToken,
+                refreshToken: payload.refreshToken
+            )
+            try await refreshSession()
+            didCompleteBrowserSignIn = true
+            lastSignInError = nil
+        } catch {
+            lastSignInError = (error as? AuthManagerError) ?? .signInSessionFailed
+            throw error
+        }
     }
 
     func seedTokensFromCLI(refreshToken: String, accessToken: String?) async {
--- a/Sources/Auth/AuthManager.swift
+++ b/Sources/Auth/AuthManager.swift
@@ -213,7 +213,7 @@ final class AuthManager: ObservableObject {
                     self.webAuthSession = nil
                 }
                 if let error {
-                    NSLog("auth.webauth failed: %@", "\(error)")
+                    Self.authLog("webauth failed: \(error)")
                     self.recordSignInSessionError(error)
                     return
                 }
@@ -221,7 +221,7 @@ final class AuthManager: ObservableObject {
                 do {
                     try await self.handleCallbackURL(callbackURL)
                 } catch {
-                    NSLog("auth.webauth callback failed: %@", "\(error)")
+                    Self.authLog("webauth callback failed: \(error)")
                 }
             }
         }
@@ -231,7 +231,7 @@ final class AuthManager: ObservableObject {
         if session.start() {
             webAuthSession = session
         } else {
-            NSLog("auth.webauth: session.start() returned false")
+            Self.authLog("webauth: session.start() returned false")
             lastSignInError = .signInSessionFailed
             isLoading = false
         }
@@ -460,6 +460,7 @@ final class AuthManager: ObservableObject {
         await tokenStore.setTokens(accessToken: resolvedAccess, refreshToken: refreshToken)
         do {
             try await refreshSession()
+            lastSignInError = nil
             authLog("seedTokensFromCLI: success user=\(currentUser?.primaryEmail ?? "nil")")
         } catch {
             authLog("seedTokensFromCLI: refreshSession failed: \(error)")
@@ -532,6 +533,7 @@ final class AuthManager: ObservableObject {
         isAuthenticated = true
         selectedTeamID = Self.resolveTeamID(selectedTeamID: selectedTeamID, teams: result.teams)
         didCompleteBrowserSignIn = true
+        lastSignInError = nil
         authLog("applySignInResult: user=\(result.email ?? "nil") teams=\(result.teams.count) teamID=\(selectedTeamID ?? "nil")")
     }
 
@@ -597,6 +599,7 @@ final class AuthManager: ObservableObject {
         selectedTeamID = Self.resolveTeamID(selectedTeamID: selectedTeamID, teams: teams)
         authLog("signInWithCredential: success user=\(user.primaryEmail ?? "nil") teams=\(teams.count) teamID=\(selectedTeamID ?? "nil")")
         didCompleteBrowserSignIn = true
+        lastSignInError = nil
     }
 
     func signOut() async {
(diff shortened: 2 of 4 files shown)


psh4607:
Locally tested the visibility fix on a `reload.sh --tag auth-error-vis` Debug build. **It works** — the Account row now renders the localized error under the Sign In… button when sign-in fails. Screen recording confirms the new red text appears as expected.

**Diagnostic result from this build**: the surfaced error is `AuthManagerError.invalidCallback` (`"로그인 콜백이 잘못되었습니다."` in ko-KR locale), not `.signInSessionFailed`.

That narrows #3617's root cause significantly:

- ASWebAuthenticationSession **did** receive a `cmux-dev://auth-callback` redirect (so it isn't `canceledLogin`, sandbox/TCC, or first-request network failure as I'd hypothesized in the issue body).
- But the URL was missing valid `stack_refresh` / `stack_access` query items — `AuthCallbackRouter.callbackPayload` returned nil, which makes `handleCallbackURL` throw `.invalidCallback`.

Likely culprit is in [`web/app/handler/after-sign-in/page.tsx`](https://github.com/manaflow-ai/cmux/blob/main/web/app/handler/after-sign-in/page.tsx#L102-L121): the gate emits `OpenNativeClient` when `refreshToken && accessCookie` are truthy, but `accessCookie` can be a JSON-encoded `[refreshToken, ""]` (empty access token) when `user.createSession()` succeeds for `refreshToken` but `getTokens()` returns an empty `accessToken`. The Swift-side `AuthCallbackRouter.decodeAccessToken` then returns nil for that empty string and we land on `.invalidCallback`.

Will follow up with a separate PR targeting `web/app/handler/after-sign-in/page.tsx` to either tighten the gate (require non-empty `accessToken` before emitting the deep link) or fall through to the sign-in form retry instead. This visibility PR keeps its scope.

psh4607:
Pushed `d48988d3` addressing review feedback:

| Reviewer | Issue | Resolution |
|---|---|---|
| Cubic `Cmux Swift Logging` | 3 `NSLog` in `beginSignIn` (L216/224/234) without `#if DEBUG` guards | Replaced with the existing `authLog` helper which is already `#if DEBUG`-guarded; same call sites, no Release-mode logs. |
| Greptile P1 + Cubic `Cmux Architecture Rethink` + CodeRabbit | `lastSignInError` not cleared in `clearSessionState` / `applySignInResult` / `signInWithCredential` / `seedTokensFromCLI` success paths → stale red error could render after sign-out or alongside an authenticated row | Added `lastSignInError = nil` to all four. `signOut` is covered transitively via `clearSessionState`. |
| CodeRabbit test rigor | `catch is AuthManagerError` swallows any case | Tightened to `catch AuthManagerError.invalidCallback`; an `XCTFail` is raised for any other `AuthManagerError` variant. |
| CodeRabbit (new) | No coverage for the stale-error scenario | Added `testSignOutClearsStaleLastSignInError`: primes `lastSignInError = .invalidCallback`, calls `signOut()`, asserts `lastSignInError` is `nil`. |

Skipped:
- CodeRabbit's `deinit { }` SwiftLint suggestion: repo has no `.swiftlint.yml`, the `required_deinit` rule isn't enforced, and other test files don't carry an explicit `deinit`. Adding it would be inconsistent without ack from maintainers.
- Cubic `Docstring Coverage` warning: the project's existing style does not require docstrings on every method (most existing tests/methods don't have them). Adding 80% coverage just for this PR would be inconsistent.

Verified locally with `reload.sh --tag auth-error-vis`: LSP clean on all changed files, Debug build compiles, two-commit invariant preserved (1st commit's failing test still fails to compile without the fix; 2nd commit's fix + 3rd commit's review-fix both pass).

Diff size: +34/-4 across `Sources/Auth/AuthManager.swift` (+10/-3) and `cmuxTests/AuthManagerSignInErrorVisibilityTests.swift` (+24/-1).

psh4607:
Closing — superseded by #3624, which lands the same visibility fix in a tighter shape: extracts `AuthSignInError` to a dedicated `AuthManagerErrors.swift`, moves `AuthSettingsRow` to `Sources/Settings/AuthSettingsRow.swift` (satisfies the CI Swift file-length budget — something my PR didn't do), adds a duplicate-sign-in guard for in-flight sessions (covers hypothesis #2 "double-fire" from #3617), and registers the v0.64.0 Settings → Sign In addition + this fix in `CHANGELOG.md` (also missing here).

The diagnostic data captured during local verification on this branch is still useful — particularly that the actual surfaced error on Debug builds is `AuthManagerError.invalidCallback` (not `.signInSessionFailed`), which narrows #3617's root cause to `web/app/handler/after-sign-in/page.tsx` emitting a deep link with a malformed `[refreshToken, ""]` access cookie when `user.createSession()`'s `getTokens()` returns an empty `accessToken`. I'll mirror that finding into a comment on #3624 and open a separate web-only PR targeting the after-sign-in handler so the actual sign-in completes once visibility lands.

PR #3634, opened 2026-05-06
Require valid accessToken for native app handoff to prevent auth crashes

psh4607:
## Summary

Prevent the native auth handoff from emitting a malformed callback when Stack returns a refresh token but an empty access token.

The current route-handler implementation now requires both token slots to be present and non-empty before redirecting to the native callback. If the access-token slot is empty, the request falls back to the web root instead of sending an invalid payload to the macOS app.

## Current implementation

- `web/app/handler/after-sign-in/handler.ts` — requires `accessToken` in both native-handoff branches.
- `web/tests/after-sign-in-route.test.ts` — covers the `[refreshToken, ""]` cookie shape and verifies the route redirects to `/`.

The earlier `page.tsx` / `native-handoff.ts` implementation was replaced by the route-handler architecture on `main`; the conflict refresh preserved the behavior in the current files instead of restoring those deleted files.

## Verification

- Final diff is limited to the current route handler and its behavioral route test.
- GitHub reports the refreshed head as mergeable.
- Tests are left to CI per repository policy.

This remains defense-in-depth alongside #3650.

Refs #3617


psh4607:
## Bot review triage (174afc6)

### Applied
- **Copilot — `NATIVE_SCHEME` defined twice (drift risk)**: confirmed valid. Removed `const NATIVE_SCHEME` from `page.tsx` and now import a single `DEFAULT_NATIVE_RETURN_TO` from `native-handoff.ts`. `buildNativeHref` and the fallback path both use the shared constant.
- **Greptile — `.length > 0` is redundant after the truthy check**: confirmed valid. `Boolean(args.refreshToken && args.accessToken)` already excludes empty strings, so the explicit `> 0` check was noise. Simplified.

### Rejected (with reasoning)
- **Greptile — `NATIVE_SCHEME` is dead code in `page.tsx`**: only partially correct in the original commit. The constant was still referenced by `buildNativeHref` for the production fallback URL, so it was *not* dead at the time the comment was filed. Resolved indirectly by the Copilot fix above (constant moved into the shared module).
- **CodeRabbit — fallback `buildNativeHref(null, ...)` hardcodes production scheme**: filed against the previous commit. After the latest fix, the fallback explicitly uses `DEFAULT_NATIVE_RETURN_TO` (= `cmux://auth-callback`), which is the correct scheme for the only client that today reaches `/handler/after-sign-in` *without* an explicit `native_app_return_to`. The Debug `cmux DEV` app always sets `native_app_return_to=cmux-dev://auth-callback` via `AuthEnvironment.signInURL()`, so it never falls through to this branch. Pulling the scheme from a cookie or the User-Agent here would add an attack surface for picking an arbitrary URL scheme without buying anything for the actual reproducible bug, so I'm intentionally not expanding the fallback to "scheme inference".

psh4607:
## Local integration evidence

Even without real Stack Auth keys we can prove the fix removes the exact malformed deep link the production app was choking on, by injecting the bug-shape cookie directly:

### Repro shape
`stack-access` cookie value (URL-encoded JSON): `["fake-refresh-token-abc",""]`

This is the same shape `web/app/handler/after-sign-in/page.tsx` was producing in production after `stackServerApp.getUser().createSession().getTokens()` returned an empty `accessToken`.

### Pre-fix run (temporarily reverted `shouldEmitNativeHandoff` to the old `Boolean(args.refreshToken)`):
```
$ curl -s -b 'stack-access=%5B%22fake-refresh-token-abc%22%2C%22%22%5D' \
    'http://localhost:3779/handler/after-sign-in?native_app_return_to=cmux-dev%3A%2F%2Fauth-callback' \
  | grep -oE 'cmux(-dev)?://auth-callback\?[^"]*'
cmux-dev://auth-callback?stack_refresh=fake-refresh-token-abc&stack_access=%5B%22fake-refresh-token-abc%22%2C%22%22%5D
```
→ exactly the `[refresh, ""]` payload `AuthCallbackRouter.decodeAccessToken` rejects with `.invalidCallback`. This matches the two `webauth callback failed: invalidCallback` lines logged in `/tmp/cmux-auth-debug.log` against production.

### Post-fix run (current branch):
```
$ curl -s -b 'stack-access=%5B%22fake-refresh-token-abc%22%2C%22%22%5D' \
    'http://localhost:3779/handler/after-sign-in?native_app_return_to=cmux-dev%3A%2F%2Fauth-callback' \
  | grep -cE 'auth-callback\?stack_'
0
```
→ no deep link emitted. Handler falls through to `redirect("/")` and the response is the home page (≈520 KB of normal HTML, no `OpenNativeClient` script).

### Happy-path sanity (current branch):
```
$ curl -s -b 'stack-access=%5B%22fake-refresh%22%2C%22fake-access%22%5D' \
    'http://localhost:3779/handler/after-sign-in?native_app_return_to=cmux-dev%3A%2F%2Fauth-callback' \
  | grep -oE 'cmux(-dev)?://auth-callback\?stack_refresh=[^"&\\]*' | head -n 1
cmux-dev://auth-callback?stack_refresh=fake-refresh
```
→ valid deep link still emitted when both tokens are present.

This validates the diagnosed root cause and the boolean gate at the handler level. The remaining gap is exercising the live `ASWebAuthenticationSession` round trip, which still requires real Stack keys (see PR body for "What is NOT verified end-to-end").

psh4607:
## dev-agents review: Reviewing code

Attempt: #101
Trigger: self-review
Head: fc5e52d41d57
Started: 2026-05-23T12:16:09.096Z
Updated: 2026-05-23T12:16:57.228Z

Progress:
- [x] Trigger accepted
- [x] Review job queued
- [x] Preparing worktree
- [x] Running build
- [~] Reviewing code
- [ ] Preparing GitHub review
- [ ] Submitting GitHub review
- [ ] Completed

psh4607 (commented):
Comment 남깁니다.

### P3

- **DEFAULT_NATIVE_RETURN_TO가 export되었지만 어디에서도 사용되지 않음** (`web/app/handler/after-sign-in/native-handoff.ts:3`)
  `DEFAULT_NATIVE_RETURN_TO`는 이 PR에서 새로 추가되었지만 `page.tsx`와 테스트 파일 어디에서도 import하지 않습니다. fallback 경로는 `nativeAuthCallbackForReturnTo()`가 동적으로 생성하고 있어 이 상수는 dead code입니다. 외부 소비자를 위해 의도적으로 남긴 것이 아니라면 제거하는 것이 좋겠습니다.

---

> ℹ️ **Read-only 리뷰 모드**
>
> 이 repo owner는 `UNWATCHED_REVIEW_TRUSTED_OWNERS`에 포함되어 있지 않아 dependency install/build/test를 실행하지 않았습니다. diff 기반 코드 리뷰만 진행했습니다.

---

> 🔄 **수동 재리뷰**: PR 우측 사이드바 **Reviewers** 섹션의 **Re-request review** 버튼(아래 빨간 박스)을 누르면 이 봇이 다시 리뷰합니다.

Re-request review 버튼

psh4607:
@lawrencecchen @austinywang Conflicts are resolved and this PR is mergeable. Could you review and merge? The remaining `CHANGES_REQUESTED` is stale from code no longer in the diff.

PR #3857, opened 2026-05-11
Gate minimal-mode chrome and sidebar on focusPaneOnFirstClick

### LINKED ISSUE #3856: focusPaneOnFirstClick is ignored by minimal-mode toolbar and the workspace sidebar
### Summary

With `app.focusPaneOnFirstClick: false` (in JSON settings) and the corresponding off state in App Settings, cmux is supposed to suppress first-click activation of pane content when the window is inactive. Currently the flag is honoured for pane bodies but bypassed in two visible places, the most surprising being the workspace sidebar.

### Repro (v0.64.3, also reproduces on Nightly)

1. Set `app.focusPaneOnFirstClick` to `false` (default behaviour for me, but explicit in `~/.config/cmux/settings.json`)
2. With `minimalMode: true`, give another app focus so cmux is inactive
3. Click any of: the sidebar toggle, notification bell, or `+` button in the cmux titlebar (left of the tab bar)
4. Or: from workspace 1, click directly on a different workspace row in the sidebar while cmux is inactive

### Expected

The first click activates cmux and focuses the clicked surface, but does not fire the underlying action. A second click is required to actually toggle the sidebar, open notifications, create a tab, or switch workspaces.

### Actual

- The titlebar buttons (toggle sidebar / bell / `+`) fire on first click while cmux is inactive
- Clicking another workspace row in the sidebar both activates cmux **and** immediately switches to that workspace, which is the most surprising case because it acts on a destructive intent, not just focus

### Root cause

`acceptsFirstMouse` is gated on `PaneFirstClickFocusSettings.isEnabled()` in three places:

- `Sources/Panels/CmuxWebView.swift`
- `Sources/Panels/MarkdownPanelView.swift`
- An inner part of `Sources/Panels/FilePreviewPanel.swift`

But the following overrides hardcode `true`, ignoring the setting:

- `Sources/Update/MinimalModeSidebarControls.swift:145-147` (`MinimalModeSidebarControlActionView.acceptsFirstMouse`)
- `Sources/Panels/FilePreviewPanel.swift:1050-1054` (`FilePreviewPDFChromeHostingView.acceptsFirstMouse`)

And the workspace sidebar is a SwiftUI surface inside `NSHostingView`. The hit test resolves to SwiftUI's internal `NSHostingPlatformView`, whose `acceptsFirstMouse` defaults to `true` for SwiftUI `Button`s, so neither the outer host nor any cmux-owned override applies. This is the case behind the workspace-switch-on-first-click symptom.

### Suggested fix

Two parts:

1. Gate the two ungated `acceptsFirstMouse` overrides on `PaneFirstClickFocusSettings.isEnabled()`, matching the existing pattern. Add tests in `cmuxTests/InactivePaneFirstClickFocusTests.swift` (same shape as the existing terminal/browser/markdown cases). I am happy to send a PR for this part.

2. The SwiftUI workspace sidebar needs a different approach since SwiftUI bypasses host-view gating. One option is a custom `NSHostingView` subclass (e.g. `FirstMouseGatedHostingView`) that overrides `hitTest(_:)` to short-circuit to itself when the window is not key and the setting is disabled, then have its `acceptsFirstMouse` return the setting's value. Would prefer maintainer input on whether to broaden the scope this way, or instead clarify the schema description to scope the flag explicitly to pane content and add a separate top-level toggle for chrome.

### Settings reference

Schema description (`web/data/cmux.schema.json`):

> "When cmux is inactive, the first click can activate and focus the clicked pane."

The wording says "pane", which arguably puts toolbar buttons and sidebar rows out of scope, but the natural reading when a user toggles it off is "nothing in cmux activates on first click while inactive". Aligning the two is the deciding question.

### Environment

- cmux: v0.64.3 (also reproduces on Nightly)
- macOS: Tahoe 26.3, Apple Silicon

rursache:
Closes #3856.

## Summary

With `app.focusPaneOnFirstClick: false`, clicks on cmux chrome while the window was inactive still fired actions on first click. The setting was already honoured for pane content (terminal, web view, markdown observer) but bypassed for chrome and sidebar via five separate code paths that I traced with a temporary instrumented build:

- `TitlebarControlButton` SwiftUI `Button`s in the titlebar (bell, sidebar toggle, `+`)
- The `NSEvent.addLocalMonitor` in `WindowDecorationsController` that resolves minimal-mode chrome clicks via the hit-region registry and calls `performMinimalModeSidebarControlAction` directly (this one bypasses every view-level check, including any `acceptsFirstMouse` override)
- The per-window send-event hook in the same controller
- The sidebar workspace row's `.onTapGesture`, which switched workspaces on first click — the most surprising symptom

## Approach

Adds `PaneFirstClickGate` next to `PaneFirstClickFocusSettings` in `WorkspaceRuntimeSettings.swift`. It observes `NSApplication.didBecomeActiveNotification` and exposes `shouldSwallowFirstClick(now:)`, which returns `true` for 200 ms after activation when the setting is disabled. The four action sites consult the gate before firing.

The grace-window approach is necessary because by the time these SwiftUI actions and event-monitor callbacks run, `NSApp.isActive` is already `true` — there's no other reliable per-click signal. 200 ms is short enough that a deliberate second click is well past the window and long enough that the activation click reliably falls inside it.

## Test plan

- `xcodebuild -project GhosttyTabs.xcodeproj -scheme cmux-unit -configuration Debug -destination 'platform=macOS' -only-testing:cmuxTests/InactivePaneFirstClickFocusTests test` — all 9 tests pass (6 existing + 3 new gate-behaviour cases)
- Manual on a tagged Debug build with `focusPaneOnFirstClick: false` and `minimalMode: true`, cmux unfocused:
  - Bell / sidebar toggle / `+` button: first click only activates cmux, second click fires the action
  - Click workspace 2 from workspace 1: first click only activates cmux, second click switches workspace
  - With `focusPaneOnFirstClick: true`: all four still fire on first click as before

## Summary by CodeRabbit

* **New Features**
  * First-click protection: the app now ignores the first click for a short grace period (~0.2s) after the app becomes active to prevent accidental actions. This applies to sidebar rows, titlebar controls, and minimal-mode sidebar chrome, and can be toggled in settings.
* **Tests**
  * Added automated tests validating the timing and behavior of the first-click gating.

[Review Change Stack](https://app.coderabbit.ai/change-stack/manaflow-ai/cmux/pull/3857)

### DIFF
--- a/Sources/App/WorkspaceRuntimeSettings.swift
+++ b/Sources/App/WorkspaceRuntimeSettings.swift
@@ -1,3 +1,4 @@
+import AppKit
 import Darwin
 import Foundation
 
@@ -89,6 +90,37 @@ enum PaneFirstClickFocusSettings {
 
     static func isEnabled(defaults: UserDefaults = .standard) -> Bool {
         defaults.object(forKey: enabledKey) as? Bool ?? defaultEnabled
+    }
+}
+
+@MainActor
+enum PaneFirstClickGate {
+    private static let graceInterval: TimeInterval = 0.2
+    private static var lastBecameActiveAt: TimeInterval = 0
+    private static var installed = false
+
+    static func install() {
+        guard !installed else { return }
+        installed = true
+        NotificationCenter.default.addObserver(
+            forName: NSApplication.didBecomeActiveNotification,
+            object: nil,
+            queue: .main
+        ) { _ in
+            MainActor.assumeIsolated {
+                lastBecameActiveAt = ProcessInfo.processInfo.systemUptime
+            }
+        }
+    }
+
+    static func shouldSwallowFirstClick(now: TimeInterval = ProcessInfo.processInfo.systemUptime) -> Bool {
+        if PaneFirstClickFocusSettings.isEnabled() { return false }
+        let elapsed = now - lastBecameActiveAt
+        return elapsed >= 0 && elapsed < graceInterval
+    }
+
+    static func markActivatedForTesting(at time: TimeInterval = ProcessInfo.processInfo.systemUptime) {
+        lastBecameActiveAt = time
     }
 }
 
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -996,6 +996,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         let isRunningUnderXCTest = isRunningUnderXCTest(env)
         let telemetryEnabled = TelemetrySettings.enabledForCurrentLaunch
         AppIconLaunchState.markDidFinishLaunching()
+        PaneFirstClickGate.install()
         if isRunningUnderXCTest {
             NSApp.setActivationPolicy(.regular)
         } else {
--- a/Sources/ContentView.swift
+++ b/Sources/ContentView.swift
@@ -12612,6 +12612,7 @@ private struct TabItemView: View, Equatable {
             lastSidebarSelectionIndex: $lastSidebarSelectionIndex
         ))
         .onTapGesture {
+            if PaneFirstClickGate.shouldSwallowFirstClick() { return }
             updateSelection()
         }
         .safeHelp(workspaceSnapshot.title)
--- a/Sources/Update/UpdateTitlebarAccessory.swift
+++ b/Sources/Update/UpdateTitlebarAccessory.swift
@@ -357,7 +357,11 @@ struct TitlebarControlButton<Content: View>: View {
     @State private var isHovering = false
 
     var body: some View {
-        let baseButton = Button(action: action) {
+        let gatedAction: () -> Void = {
+            if PaneFirstClickGate.shouldSwallowFirstClick() { return }
+            action()
+        }
+        let baseButton = Button(action: gatedAction) {
             content()
                 .frame(width: config.buttonSize, height: config.buttonSize)
                 .contentShape(Rectangle())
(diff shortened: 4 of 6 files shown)


rursache:
Addressed CodeRabbit's actor-isolation note in the latest force-push: annotated `PaneFirstClickGate` with `@MainActor` (so all static state and members are main-actor isolated per SE-0316), used `MainActor.assumeIsolated { ... }` inside the `NotificationCenter` observer body and at the two `WindowDecorationsController` call sites that live in non-isolated closures (`NSEvent.addLocalMonitorForEvents` handler and the non-`@MainActor` `handleMinimalModeSidebarChromeMouseDown` helper). The other call sites (AppDelegate, SwiftUI `Button` action, `.onTapGesture`, XCTest `@MainActor` class) are already main-actor isolated so they call straight through. Unit tests still green.

rursache:
Added a near-boundary assertion (`now + 0.19`) alongside the existing `now + 0.05` case in `testPaneFirstClickGateSwallowsWithinGraceWhenSettingDisabled`, per CodeRabbit's suggestion. Both points are inside the 200 ms grace and should swallow; the existing `testPaneFirstClickGateAllowsAfterGrace` test (using `now + 1.0`) already covers the after-grace side. The earlier `@MainActor` isolation note is addressed in the previous push and the pre-merge "Cmux Swift Concurrency" check now passes.

rursache:
closing because it's fixed in https://github.com/manaflow-ai/cmux/pull/3881

PR #4138, opened 2026-05-14
Add clean copy as default terminal copy behavior

### LINKED ISSUE #3096: Copying a soft-wrapped long command inserts hard newlines at wrap points, breaking paste into other terminals
### Summary
When the terminal contains a single-line command that's visually wrapped across multiple rows (because it was wider than the viewport), copying that selection puts `\n` at each wrap point onto the clipboard. Pasting into another terminal (or any field that auto-submits on newline) executes partial fragments instead of the intended full command.

### Repro
1. Size a cmux pane narrow (e.g. 80 cols).
2. Print a single long line:
   ~~~
   printf 'curl -X POST "https://example.com/api/v1/ep?a=1&b=2&c=3&d=4&e=5&f=6" -H "Authorization: Bearer xxx" -d @/tmp/body.json\n'
   ~~~
3. Drag-select the whole line (it spans multiple visual rows).
4. Cmd+C, then paste into Terminal.app / iTerm2 / another cmux pane.

**Expected:** one contiguous string, no embedded `\n`.
**Actual:** `\n` at every visual wrap point; shell executes each fragment as a separate command.

### Why it hurts
This is the dominant failure mode when copying AI-generated shell commands — they almost always wrap. Every cross-terminal copy-paste of a long command silently breaks.

### Env
- cmux 0.62.2 (77) [6c203b514], macOS

### Triage note
cmux depends on `migueldeicaza/SwiftTerm` (`Package.swift`). SwiftTerm's `Terminal._getSelectedLines` (in `Sources/SwiftTerm/Terminal.swift`) already has an `isWrapped` check meant to suppress hard newlines at soft-wrap boundaries. So the bug is likely one of:
1. cmux has its own selection→clipboard path that doesn't honor `isWrapped`, or
2. lines aren't being marked `isWrapped=true` when cmux's writer wraps them.

### Related
- #2203 (umbrella "selection/copy behavior doesn't match iTerm" — this is one concrete symptom)
- ghostty-org/ghostty#11293 — same symptom class upstream, auto-closed by vouching bot without a fix. Not inherited code (cmux uses SwiftTerm, not libghostty), but useful reference.

Note: a selection fix was mentioned as landing in nightly in #2203 — worth checking whether that fix covers this specific soft-wrap-newline case.

bbowser:
## Summary

Fixes #3096

- **Cmd+C now produces clean text** — strips common leading whitespace, removes terminal decoration characters (●, ◆, ▶, etc.), re-joins hard-wrapped paragraph lines, and joins URL lines without inserting spaces
- **Preserves structure** — code fences, table rows, headings, blockquotes, bullet points maintain their formatting
- **Cmd+Shift+C** and right-click → "Copy Raw Text" provide the old verbatim copy behavior as a fallback

## Test plan

- [ ] Select wrapped paragraph text in terminal, Cmd+C, paste — lines should be joined without extra indentation
- [ ] Select a long URL that wraps across lines, Cmd+C, paste — URL should be one line with no inserted spaces
- [ ] Select bullet list, Cmd+C, paste — each bullet stays on its own line, wrapped continuations rejoin
- [ ] Select text starting with ● or similar decoration, Cmd+C, paste — decoration char stripped
- [ ] Select text with code fences, Cmd+C, paste — code block content preserved verbatim
- [ ] Select text with `> ` blockquotes, Cmd+C, paste — blockquotes stay on own line
- [ ] Cmd+Shift+C still produces raw/verbatim copy
- [ ] Right-click context menu shows "Copy" and "Copy Raw Text"
- [ ] Verify in Settings → Keyboard Shortcuts that "Copy Raw Text" appears and is configurable

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -86962,6 +86962,40 @@
         }
       }
     },
+    "shortcut.copyRawText.label": {
+      "extractionState": "manual",
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Copy Raw Text"
+          }
+        },
+        "ja": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "生テキストをコピー"
+          }
+        }
+      }
+    },
+    "terminalContextMenu.copyRawText": {
+      "extractionState": "manual",
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Copy Raw Text"
+          }
+        },
+        "ja": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "生テキストをコピー"
+          }
+        }
+      }
+    },
     "sidebar.closeWorkspace.tooltip": {
       "extractionState": "manual",
       "localizations": {
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -11642,6 +11642,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
             return handled
         }
 
+        if matchConfiguredShortcut(event: event, action: .copyRawText) {
+            let handled = tabManager?.copyRawTextFromFocusedTerminal() ?? false
+            return handled
+        }
+
         // Workspace navigation: Cmd+Ctrl+] / Cmd+Ctrl+[
         if matchConfiguredShortcut(event: event, action: .nextSidebarTab) {
 #if DEBUG
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -7087,6 +7087,7 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
         }
 
         var result: [String] = []
+        var inCodeFence = false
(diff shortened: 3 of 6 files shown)


bbowser:
@coderabbitai review

bbowser:
## Addressing Greptile pre-merge suggestions

**File/Package Boundaries** — `GhosttyTerminalView.swift` is already ~13k lines; the ~60 lines of text-processing logic added here follow the existing pattern of colocating clipboard operations in the same file. Happy to extract into a separate module if maintainers prefer, but keeping it out of scope for this PR.

**Architecture (dual raw-copy paths)** — The context menu action calls `performBindingAction` directly on the view (which has access), while the keyboard shortcut routes through `TerminalSurface`. Both paths produce identical results. Can unify in a follow-up if desired.

**Docstring Coverage** — This repo's conventions default to no comments (per CLAUDE.md). The threshold doesn't apply here.

bbowser:
Re: bullet + prose merge edge case — this scenario (a bullet immediately followed by non-list prose with no blank line) is uncommon in real terminal output. Fixing it would break the more common case of wrapped bullet continuations rejoining correctly. Noting for a follow-up if it comes up in practice.

PR #4179, opened 2026-05-14
Restore Claude session after /clear by capturing rotated session id

elitecoder:
## Summary

Fix a gap in cmux's Claude session restore: a Claude session that has been `/clear`-ed cannot be restored after a cmux relaunch. The wrong (pre-clear) session resumes — so the user's recent post-clear context is silently lost.

## Repro

1. Open a Claude pane in cmux (or any tool that uses the bundled Claude wrapper).
2. Have a non-trivial conversation in session **A**. Wait for autosave to capture the snapshot (`session-com.cmuxterm.app.json` → panel → `terminal.agent.sessionId = A`).
3. Run `/clear` inside Claude. Claude rotates to session **B** but does **not** kill its process and does **not** emit `SessionEnd` for **A** — only `SessionStart(source=clear, session_id=B)`.
4. Quit cmux **before** typing the first post-clear turn (so the on-disk transcript for **B** has not been written yet).
5. Relaunch cmux → `cmux restore-session`.
6. Observed: cmux runs `claude --resume A` — the very session the user just abandoned.
7. Expected: cmux runs `claude --resume B`, or at minimum does not silently roll back to **A**.

## Root cause

In `runClaudeHook` → `case "session-start"` (`CLI/cmux.swift`), the rotated record is upserted with `isRestorable=false`:

```swift
try? sessionStore.upsert(
    sessionId: sessionId,
    ...
    isRestorable: false,                 // ← also for /clear and replacement
    markActive: shouldPromoteActiveSession,
    turnId: parsedInput.turnId
)
```

`RestorableAgentSessionIndex.load()` (the input to autosave) then filters out the rotated record because `isRestorable == false` *and* its transcript file does not yet exist on disk. The pre-clear record (whose transcript still exists) survives the filter and wins on `updatedAt`. Autosave bakes the abandoned `sessionId` into the session snapshot, and `restore-session` resumes that.

In addition, even after the user types and the rotated record becomes restorable, the stale pre-clear record lingers in `~/.cmuxterm/claude-hook-sessions.json` until the 7-day prune. A late pre-clear `Stop` / `Notification` refresh on the old record can bump its `updatedAt` past the rotated one and re-introduce the same wrong-resume.

## Fix

Two targeted changes in `runClaudeHook` `case "session-start"`:

1. **Mark the rotated record restorable immediately** when SessionStart establishes a new active boundary (`/clear` or replacement of a stopped owner — i.e. `shouldPromoteActiveSession == true`). Autosave now captures the rotated id before any transcript exists, so killing cmux mid-rotation no longer rolls back to the pre-clear session.
2. **Prune pre-clear sibling records** scoped to the same `(workspace, surface)` of the rotation. The abandoned record can no longer out-rank the rotated one through ordering flukes, and the hook store stops accumulating per-panel cruft.

The new active-session pointer (`activeSessionsByWorkspace[ws] = rotated`) is preserved across the prune, so the existing "stale Stop from old session must not clobber clear status" guard (`testClaudeStopFromPreviousSessionDoesNotClobberClearRunningStatus`) still holds.

Pruning runs from a separate `withLockedState` after the upsert completes — sequential, not nested — so the existing `flock` discipline is unchanged.

## Test plan

Three new integration tests in `cmuxTests/CLINotifyProcessIntegrationRegressionTests`:

- `testClaudeClearSessionStartMarksRotatedSessionRestorable` — proves the rotated record is `isRestorable=true` immediately after the clear hook, before any transcript exists.
- `testClaudeClearSessionStartPrunesPreClearSiblingOnSamePanel` — proves the pre-clear record is removed from `state.sessions` after the rotation, leaving the rotated record as the sole candidate.
- `testClaudeClearSessionStartKeepsSiblingsOnDifferentPanels` — proves pruning is scoped: a session bound to a different `surfaceId` survives the rotation untouched.

Existing regression tests still pass:

- `testClaudeClearSessionStartMarksWorkspaceRunning`
- `testClaudeSessionStartRecordIsNotRestorableUntilPrompt` — non-`/clear` `source=startup` continues to write `isRestorable=false`, because `shouldPromoteActiveSession` is false in that path.
- `testClaudeStopFromPreviousSessionDoesNotClobberClearRunningStatus` — late stale Stop on the pre-clear session is still rejected by `isCurrent` (active pointer = rotated, not pre-clear).
- `testClaudePromptSubmitFromNewSessionCanReplaceStoppedSession`
- `testClaudeSessionEndChecksConsumedWorkspaceBeforeClearingVisibleState`
- `testClaudeSessionEndDoesNotConsumeSameSessionStaleTurn`

Local: `xcodebuild -scheme cmux-unit … test -only-testing:cmuxTests/CLINotifyProcessIntegrationRegressionTests/<each>` → all 9 pass.

## Out of scope

- Behavior on `compact` and `resume` SessionStart sources is unchanged (those keep the existing `isRestorable=false` until the first prompt).
- The pre-existing 7-day `pruneExpired` cutoff is unchanged.
- No changes to the Claude wrapper script (`Resources/bin/claude`) or to the hooks JSON it injects.

## Author note

I'm a happy cmux dogfooder — opening this draft so a maintainer can sanity-check the boundary semantics around `shouldPromoteActiveSession`. If `replacement-of-stopped-owner` should NOT be marked restorable until first prompt (to keep that path symmetric with startup/resume), I'm happy to narrow the change to `isClearSessionStart` only.

---
## Summary by cubic
Fixes wrong resume after `/clear` by immediately marking the rotated Claude session restorable and pruning only same-panel siblings so autosave restores the right session. Tightens pruning to require both `workspaceId` and `surfaceId`, and keeps replacement-of-stopped sessions immediately restorable.

- **Bug Fixes**
  - `SessionStart` that promotes a new active boundary (`/clear` or replacement of a stopped owner) now sets `isRestorable=true` so autosave captures the rotated `sessionId` before a transcript exists.
  - On `/clear`, exhaustively prunes same-panel siblings in the hook store (scoped by `(workspaceId, surfaceId)`), with a strict guard that requires both IDs to avoid widening the sweep; replacement-of-stopped-owner does not prune.
  - Tests cover immediate restorability, same-panel pruning (including multi-sibling), cross-panel safety, replacement marking as restorable, and non-pruning on replacement.

- **Refactors**
  - In `removePanelSiblings`, collect-then-remove to avoid mutating during iteration; no behavior change.

<sup>Written for commit 94db2d92e00e90d8f821eaa934911ad5593254b1. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improve session handling during panel clear/rotation: pre-clear sibling sessions on the same panel are pruned, and rotated/replacement sessions are persisted as restorable so autosave/restoration selects the correct session.
* **Tests**
  * Add regression tests covering session-start, prompt-submit, stop, clear flows, sibling-pruning behavior, and restorable-session marking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/manaflow-ai/cmux/pull/4179)

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -755,9 +755,13 @@ private final class ClaudeHookSessionStore {
         surfaceId: String
     ) throws -> [String] {
         let keep = normalizeSessionId(keepSessionId)
-        let normalizedWorkspace = normalizeOptional(workspaceId)
-        let normalizedSurface = normalizeOptional(surfaceId)
-        guard !keep.isEmpty, normalizedWorkspace != nil || normalizedSurface != nil else {
+        // Require both workspace and surface: pruning is a panel-scoped
+        // operation, and proceeding with only one would silently widen the
+        // sweep (e.g. an empty surfaceId would prune every session in the
+        // workspace).
+        guard !keep.isEmpty,
+              let workspace = normalizeOptional(workspaceId),
+              let surface = normalizeOptional(surfaceId) else {
             return []
         }
         return try withLockedState { state in
@@ -766,13 +770,9 @@ private final class ClaudeHookSessionStore {
             // checks. Compute the set, then remove in a separate pass.
             var removed: [String] = []
             for (sid, record) in state.sessions {
-                guard sid != keep else { continue }
-                let recordWorkspace = normalizeOptional(record.workspaceId)
-                let recordSurface = normalizeOptional(record.surfaceId)
-                if let normalizedSurface, recordSurface != normalizedSurface {
-                    continue
-                }
-                if let normalizedWorkspace, recordWorkspace != normalizedWorkspace {
+                guard sid != keep,
+                      normalizeOptional(record.workspaceId) == workspace,
+                      normalizeOptional(record.surfaceId) == surface else {
                     continue
                 }
                 removed.append(sid)
@@ -783,10 +783,9 @@ private final class ClaudeHookSessionStore {
             // If the active-session pointer still targets a removed record,
             // clear it so later Stop/SessionEnd events cannot revive state
             // that no longer has a matching session record.
-            if let normalizedWorkspace,
-               let active = state.activeSessionsByWorkspace[normalizedWorkspace],
+            if let active = state.activeSessionsByWorkspace[workspace],
                removed.contains(active.sessionId) {
-                state.activeSessionsByWorkspace.removeValue(forKey: normalizedWorkspace)
+                state.activeSessionsByWorkspace.removeValue(forKey: workspace)
             }
             return removed
         }
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -761,6 +761,9 @@ private final class ClaudeHookSessionStore {
             return []
         }
         return try withLockedState { state in
+            // Collect victims first; mutating `state.sessions` while iterating
+            // it is undefined behavior in Swift and can trap under runtime
+            // checks. Compute the set, then remove in a separate pass.
             var removed: [String] = []
             for (sid, record) in state.sessions {
                 guard sid != keep else { continue }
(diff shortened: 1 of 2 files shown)


elitecoder:
Addressed reviewer feedback in 3fff076 (collect-then-remove to avoid dictionary mutation during iteration) and 94db2d9 (tighten guard to require both workspaceId and surfaceId, simplify loop, reuse workspace binding for active-pointer cleanup). All 12 Claude-hook regression tests still pass.

@greptileai please re-review

@coderabbitai please re-review

PR #4243, opened 2026-05-16
browser: expose webview lifecycle state in top

lidge-jun:
## Summary
- Track browser WKWebView lifecycle state as new_tab, live_visible, live_hidden, or closing.
- Record browser portal/view visibility timestamps and reasons without resetting hidden duration on duplicate hide events.
- Include lifecycle data in cmux top browser webview payloads for memory diagnostics.
- Reset lifecycle metadata when a browser panel is reused for a fresh workspace context.
- Add focused BrowserPanel lifecycle unit coverage.

## Why
Hidden browser tabs/workspaces currently keep their WKWebView and WebContent process alive, but cmux top does not distinguish visible webviews from hidden retained ones. This PR adds the observability needed before adding discard/reclaim policy in follow-up PRs.

## Review follow-up
- Addressed duplicate hide events resetting hidden_duration_ms.
- Deferred portal update lifecycle recording out of NSViewRepresentable render/update.
- Reset lifecycle metadata during workspace context reset.
- Removed test double-close.

## Validation
- `git diff --check`
- `swiftc -parse Sources/Panels/BrowserPanel.swift Sources/Panels/BrowserPanelView.swift Sources/TerminalController.swift cmuxTests/GhosttyConfigTests.swift`

Not run: `./scripts/test-unit.sh -only-testing:cmuxTests/BrowserPanelWebViewLifecycleTests` because this machine has only Command Line Tools selected and no Xcode.app install; xcodebuild exits with: tool xcodebuild requires Xcode.

## Summary by CodeRabbit

* **New Features**
  * Browser panel now persists WebView lifecycle states (new, visible, hidden, closing) with timestamped visibility transitions and computed hidden duration.
  * Lifecycle state and payload are included in workspace/browser surface telemetry.

* **Behavior**
  * Panel view and portal interactions now report UI visibility changes; hiding the portal records a visibility transition.

* **Tests**
  * Added tests verifying lifecycle transitions, payload contents, and closing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/manaflow-ai/cmux/pull/4243?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/Panels/BrowserPanel.swift
+++ b/Sources/Panels/BrowserPanel.swift
@@ -4572,6 +4572,7 @@ extension BrowserPanel {
 
     func resetForWorkspaceContextChange(reason: String) {
         guard needsWorkspaceContextReset else {
+            resetWebViewLifecycleMetadata()
 #if DEBUG
             cmuxDebugLog(
                 "browser.contextReset.skip panel=\(id.uuidString.prefix(5)) " +
--- a/Sources/Panels/BrowserPanel.swift
+++ b/Sources/Panels/BrowserPanel.swift
@@ -2069,6 +2069,13 @@ private enum BrowserInsecureHTTPNavigationIntent {
     case newTab
 }
 
+enum BrowserWebViewLifecycleState: String {
+    case newTab = "new_tab"
+    case liveVisible = "live_visible"
+    case liveHidden = "live_hidden"
+    case closing
+}
+
 /// Observable state for browser find-in-page. Mirrors `TerminalSurface.SearchState`.
 @MainActor
 final class BrowserSearchState: ObservableObject {
@@ -2487,8 +2494,22 @@ final class BrowserPanel: Panel, ObservableObject {
 
     /// Whether the browser panel should render its WKWebView in the content area.
     /// New browser tabs stay in an empty "new tab" state until first navigation.
-    @Published private(set) var shouldRenderWebView: Bool = false
+    @Published private(set) var shouldRenderWebView: Bool = false {
+        didSet {
+            if oldValue != shouldRenderWebView {
+                refreshWebViewLifecycleState()
+            }
+        }
+    }
     private var restoredSessionShouldRenderWebView: Bool?
+
+    @Published private(set) var webViewLifecycleState: BrowserWebViewLifecycleState = .newTab
+    private(set) var webViewLastVisibleAt: Date?
+    private(set) var webViewLastHiddenAt: Date?
+    private(set) var webViewLastVisibilityChangeAt: Date?
+    private(set) var webViewLastVisibilityChangeReason: String?
+    private var isWebViewVisibleInUI: Bool = false
+    private var isClosingWebViewLifecycle: Bool = false
 
     /// True when the browser is showing the internal empty new-tab page (no WKWebView attached yet).
     var isShowingNewTabPage: Bool {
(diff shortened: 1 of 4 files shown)


lidge-jun:
@coderabbitai resolve

austinywang (changes requested):
Summary

This is pointed in a useful direction: putting browser WebContent visibility into `system.top` is the right diagnostic surface, and the payload changes are mostly additive. I do not see a new WKWebView retain cycle, delegate-cycle, or observer teardown problem in the diff. I do see a correctness hole in the lifecycle state machine: attached-DevTools/local-inline hosting can keep reporting a hidden retained webview as `live_visible`, which is exactly the class of case this telemetry is supposed to distinguish.

Strengths

- The top payload change preserves the existing `webviews` array and `browser_web_content_pid` fields, so existing JSON consumers should not break on missing old keys.
- `BrowserPanel.close()` now marks the lifecycle as `closing` before delegate/KVO teardown, and workspace-context reset clears the new metadata.
- The model-level duplicate-hide behavior does avoid resetting `last_hidden_at`; the direct unit test covers that helper-level invariant.
- The new code does not add NotificationCenter observers, KVO observers, WK delegates, or strong self-capturing WKWebView closures.

Concerns

- `Sources/Panels/BrowserPanelView.swift:6482`: `updateUsingLocalInlineHosting` never records lifecycle visibility, and that makes the state wrong when an attached-DevTools browser is hidden. The visible path is gated into local-inline mode at `Sources/Panels/BrowserPanelView.swift:1328`, then `updateUsingLocalInlineHosting` clears/suppresses portal ownership around `Sources/Panels/BrowserPanelView.swift:6494`. When the same panel later loses pane ownership or visibility, `updateNSView` falls back to the window-portal path, but the hidden lifecycle update at `Sources/Panels/BrowserPanelView.swift:6738` only runs if a portal host was accepted or released. After local-inline hosting there may be no active portal lease to release, so `noteWebViewVisibility(false, ...)` is skipped and `cmux top` keeps showing `live_visible` for a hidden retained WKWebView. That hides the exact memory-diagnostic case this PR is trying to expose. The fix should make effective webview visibility a single transition independent of whether the current host is window-portal or local-inline.

- `Sources/Panels/BrowserPanel.swift:2701`: lifecycle ownership is split across `BrowserPanelView.onAppear` (`Sources/Panels/BrowserPanelView.swift:717`), SwiftUI `isVisibleInUI` changes (`Sources/Panels/BrowserPanelView.swift:780`), portal acceptance/release (`Sources/Panels/BrowserPanelView.swift:6738`), and workspace/tab cleanup (`Sources/Panels/BrowserPanel.swift:6335`). Architecturally this is still bolted onto several ad-hoc lifecycle signals rather than derived from one owner of "effective browser visibility." The local-inline miss above is a symptom of that split. I would expect one `@MainActor` owner on `BrowserPanel` to receive explicit lifecycle events or snapshots from the view/portal layer and derive `new_tab/live_visible/live_hidden/closing` in one place, with portal/local-inline as implementation details rather than separate truth sources.

- `cmuxTests/GhosttyConfigTests.swift:1175`: the new tests call `BrowserPanel.noteWebViewVisibility` directly, so they prove the helper behavior but not the real call paths that can disagree: selected-tab changes, workspace retirement, portal host replacement, local-inline DevTools hosting, and `system.top` serialization. Add at least one behavior-level test or small runtime seam that drives an effective visibility transition through the same path used by `BrowserPanelView`/portal ownership, plus a top-payload assertion. Otherwise the current tests would pass while the attached-DevTools hidden-state bug above remains.

Suggestions

- If the goal is human memory diagnostics from `cmux top`, consider also rendering the lifecycle in the default tree label in `CLI/cmux.swift:12486`; right now it is only visible to JSON consumers.
- Consider using truncation instead of rounding for `hidden_duration_ms` in `Sources/Panels/BrowserPanel.swift:2773` if downstream tooling treats it as elapsed time rather than a displayed approximation.

Verdict

Request changes. The payload shape is fine, but the lifecycle source of truth is not durable yet, and at least one real browser hosting mode can report hidden retained WebViews as visible.

Issue #4256, opened 2026-05-16
Resume launch command uses panel's current cwd, but session may have been created in a different cwd → session-not-found after restart


treerobin06:
### cmux version

0.64.6 (86), commit 2f96c15c2

### macOS version

macOS 26.4.1 (build 25E253)

### Mac chip

Apple Silicon (M1/M2/M3/M4)  

### Installation method

Direct download (DMG)

### Can you reproduce this on cmux NIGHTLY?

I could not test NIGHTLY

### Bug description

When cmux restores a previously-running Claude Code panel (after Cmd-Q or app
restart), the resume command it generates uses the panel's **current** working
directory as the `cd` target:

```
cd '/some/project/path' && '.../claude' --resume '<sid>' ...
```

However, Claude Code stores each session under a path-encoded directory
based on the cwd **at the time of first launch**, not the current cwd:

```
~/.claude/projects/<dash-encoded-first-launch-cwd>/<sid>.jsonl
```

If the user `cd`-s during the session (or Claude's Bash tool does), the
resume location stops matching the storage location. Claude can't find the
session and either silently starts a new one or errors. Workaround: manually
edit the `cd` target back to where Claude was originally launched (often
`$HOME`).

### Expected behavior

The resume command should `cd` to the cwd at which the session was first
created, regardless of where the panel's `$PWD` has drifted to during the
session. Resume should reliably reopen the existing session.

### Steps to reproduce

1. Open a cmux panel in `$HOME`. Run `claude` from there.
   - Claude creates `~/.claude/projects/-Users-<user>/<sid>.jsonl`.
2. Inside that session, do work that involves a project subdirectory — for
   instance, ask Claude to operate on files under
   `~/Desktop/some-project/`. Claude's tooling drifts the panel's `$PWD`
   into that subdirectory.
3. cmux's zsh shell integration (`_CMUX_PWD_LAST_PWD` precmd hook in
   `Resources/shell-integration/cmux-zsh-integration.zsh`) tracks `$PWD`
   live and updates `panel.directory` in cmux's serialized session state.
4. Cmd-Q cmux. cmux serializes `panel.directory = /Users/<user>/Desktop/some-project`.
5. Reopen cmux. The restored panel's title shows e.g.:
   ```
   cd '/Users/<user>/Desktop/some-project' && '.../claude' '--resume' '<sid>'
   ```
6. Claude launches with cwd `/Users/<user>/Desktop/some-project`, looks for
   the session in
[...]

PR #4401, opened 2026-05-19
Stop exporting CLAUDE_CONFIG_DIR=$HOME/.claude in the Resume-in-new-tab launch command

### LINKED ISSUE #4255: Resume launch command exports CLAUDE_CONFIG_DIR=$HOME/.claude (the default), triggers Claude CLI custom-config-dir branch and forces re-authentication
### cmux version

0.64.6 (86), commit 2f96c15c2

### macOS version

macOS 26.4.1 (build 25E253)

### Mac chip

Apple Silicon (M1/M2/M3/M4)  

### Installation method

Direct download (DMG)

I could not test NIGHTLY

### Bug description

When cmux generates the "resume session in terminal" command for an existing
Claude Code session, the command always includes:

```
env CLAUDE_CONFIG_DIR=$HOME/.claude \
    CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV=1 \
    CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV_KEYS=CLAUDE_CONFIG_DIR \
    claude --resume <session-id> ...
```

Even though `CLAUDE_CONFIG_DIR=$HOME/.claude` is the Claude CLI's default
value, **the act of explicitly exporting the variable** changes Claude's
behavior. The Claude CLI distinguishes "variable not set" from "variable set
to default value" (likely a `!!process.env.CLAUDE_CONFIG_DIR` check, not a
value comparison). When set, Claude takes a "custom config dir" code path
that fails to find the existing keychain credential / oauth account, and
greets the user with "Welcome to Claude Code, please sign in" — even though
the user is already signed in and the same `claude --resume` command works
fine without the env prefix.

The two `CMUX_PRESERVE_*` env vars only have meaning inside the cmux-bundled
`Resources/bin/claude` wrapper (they gate `should_preserve_claude_auth_selection_key`).
When exported to the user's external shell, they are inert — the wrapper
already runs its unset pass before exec'ing the real claude binary.

### Expected behavior

The resume command should let the user resume an existing Claude session
without triggering re-authentication. Either:

1. Omit `CLAUDE_CONFIG_DIR=...` from the `env` prefix when the value equals
   `$HOME/.claude` (the Claude CLI default). Only include it for true
   multi-account / custom-config-dir setups.
2. Omit the `CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV*` pair as well —
   they're wrapper-internal-only.

### Steps to reproduce

1. Sign in to Claude Code normally and run a session inside a cmux panel
   long enough for it to be persisted.
2. Use the cmux GUI feature that emits a copyable command for resuming the
   session in an external terminal. Observe that the command starts with
   `env CLAUDE_CONFIG_DIR=$HOME/.claude CMUX_PRESERVE_*=... claude --resume <id> ...`.
3. Paste and run it in any shell (a fresh cmux panel works too).
4. Claude prints "Welcome to Claude Code, please sign in" instead of resuming.

Five-variant bisection from the same cmux panel:

| # | Command | Result |
|---|---|---|
| A | `/Users/$USER/.local/bin/claude --resume <id>` (bypass cmux wrapper) | OK |
| B | `claude --resume <id>` (through cmux wrapper, no env prefix) | OK |
| C | B + `--model claude-opus-4-7` | OK |
| D | C + `--permission-mode bypassPermissions` | OK |
| E | `env CLAUDE_CONFIG_DIR=$HOME/.claude CMUX_PRESERVE_*=... claude --resume <id> --model ... --permission-mode bypassPermissions` | **FAIL** — re-auth prompt |

So the failure is fully attributable to the `env CLAUDE_CONFIG_DIR=...`
prefix (cmd E vs D is the only difference).

### Shell and environment

- zsh, no `ANTHROPIC_*` env vars exported globally (`zsh -ic 'env | grep ANTHROPIC'` returns empty).
- Standard OAuth login: keychain entry `Claude Code-credentials` with `acct = $USER`.
- Claude CLI binary contains a tellingly named warning string:
  ```
  subprocess CLAUDE_CONFIG_DIR likely differs from parent (custom spawnClaudeCodeProcess / container?)
  ```
  consistent with the "presence check, not value check" hypothesis.

N/A — no crash. Claude proceeds normally to its onboarding flow.

### Additional context

**Suggested fix locations in the cmux source tree** (based on grep over the
public repo):

- `Packages/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentLaunchEnvironmentPolicy.swift`
- `Sources/TerminalStartupEnvironment.swift`
- `Sources/RestorableAgentSession.swift`

**Belt-and-suspenders patch in `Resources/bin/claude`** (the bash wrapper):
end of `normalize_claude_config_dir()`, after the legacy-path rewrite block:

```bash
if [[ "$CLAUDE_CONFIG_DIR" == "$HOME/.claude" ]]; then
    unset CLAUDE_CONFIG_DIR
fi
```

This would mask the bug even if the GUI-emitted launch command continues to
export the redundant variable, at the small cost of changing behavior for
the (probably empty) set of users who deliberately set
`CLAUDE_CONFIG_DIR=$HOME/.claude` to force the custom-config-dir code path.

**Related** — see companion issue about `cd` target in the same resume
command being inferred from the panel's current `$PWD` rather than the cwd
at which Claude was first launched (which is where `~/.claude/projects/<dash-encoded-cwd>/<sid>.jsonl`
lives). Both stem from cmux's GUI-emitted launch command making assumptions
that don't survive in an external shell.

mvanhorn:
## Summary

There are two independent fix sites; the primary one (1) is sufficient on its own, and (2) is a defense-in-depth wrapper-side patch that masks future regressions.

1. **Primary fix — stop emitting the redundant env prefix.** Find where the Vault resume launch command is assembled and only include the `env CLAUDE_CONFIG_DIR=...` prefix when the value to be exported is **not** equal to `$HOME/.claude` (i.e. when the user really is on a non-default custom config dir). Candidate sites the reporter flagged via grep:
   - `Packages/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentLaunchEnvironmentPolicy.swift` — likely owns the env-var allowlist policy.
   - `Sources/TerminalStartupEnvironment.swift` — likely owns the terminal-side startup env composition.
   - `Sources/RestorableAgentSession.swift` — likely owns the resume command string assembly (matches the grep hit pattern from #4256, which is the companion issue at the same code path).

   Also drop the two `CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV*` variables from the emitted command — they are wrapper-internal and the wrapper unsets them anyway. Keep them inside the wrapper code path; just stop exporting them to the user's external shell.

2. **Defense-in-depth in the wrapper (`Resources/bin/claude`).** At the end of `normalize_claude_config_dir()`, after the legacy-path rewrite block, add:

   ```bash
   if [[ "$CLAUDE_CONFIG_DIR" == "$HOME/.claude" ]]; then
       unset CLAUDE_CONFIG_DIR
   fi
   ```

   This means even if a future code path regresses and re-exports the default value, the wrapper unsets it before exec'ing claude. Tiny behavior change for the (probably empty) set of users who deliberately set `CLAUDE_CONFIG_DIR=$HOME/.claude` to force the custom-config-dir code path — documented in the PR.

3. **Do not change** the non-default `CLAUDE_CONFIG_DIR` path. Users with a genuinely custom config dir (multi-account workflows, container setups) must continue to see it exported. The fix is conditional on equality with `$HOME/.claude`.

## Why this matters

The GUI command emitted by cmux's "Vault by folder → Resume in new tab" feature always begins with:

```
env CLAUDE_CONFIG_DIR=$HOME/.claude \
    CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV=1 \
    CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV_KEYS=CLAUDE_CONFIG_DIR \
    claude --resume <id> ...
```

`$HOME/.claude` is the Claude CLI's default for `CLAUDE_CONFIG_DIR`. But because Claude CLI distinguishes "variable not set" from "variable set to the default value" (via something equivalent to `!!process.env.CLAUDE_CONFIG_DIR` rather than a value comparison), the act of exporting the var takes Claude down its "custom config dir" branch, which fails to read the existing keychain credential and forces the user to re-authenticate every Vault resume.

Reporter (treerobin06) bisected this with five command variants on cmux 0.64.6 / macOS 26.4.1 / Apple Silicon. Removing the `env CLAUDE_CONFIG_DIR=...` prefix is the only variable that flips behavior from "re-auth prompt" to "session resumes cleanly". The two `CMUX_PRESERVE_*` env vars are inert outside the cmux-bundled `Resources/bin/claude` wrapper (the wrapper already unsets them before exec'ing the real claude binary), so they are dead weight in the emitted command.

## Testing

1. **Vault resume default OAuth**: sign in normally, run a session, Cmd-Q, reopen cmux, Vault by folder → Resume in new tab. Generated command must not contain `CLAUDE_CONFIG_DIR=$HOME/.claude` and must not contain `CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV*`. Pasting the command into a fresh shell resumes the session without re-auth prompt.
2. **Vault resume with custom CLAUDE_CONFIG_DIR**: launch cmux with `CLAUDE_CONFIG_DIR=/tmp/alt-claude-config claude` once; create a session; Cmd-Q; reopen and use Resume in new tab. The generated command MUST still export `CLAUDE_CONFIG_DIR=/tmp/alt-claude-config` because the value is non-default.
3. **Wrapper unit**: `Resources/bin/claude` invoked with `CLAUDE_CONFIG_DIR=$HOME/.claude` env set must unset the variable before exec; with `CLAUDE_CONFIG_DIR=/some/other/path` must preserve it.
4. **Five-variant bisection regression**: reproduce reporter's table A-E; only variant E (the original buggy emission) should be impossible to produce; A-D continue to work; the new "default-omitted" variant is a sixth row that also works.
5. **No regression on companion #4256 cd-target path**: the `cd '/path' && ...` portion of the same emitted command remains unchanged here; #4256 is a separate plan.

Fixes #4255

AI was used for assistance.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/4401"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

---
## Summary by cubic
Prevents forced re-auth when using “Resume in new tab” by normalizing the environment: if `CLAUDE_CONFIG_DIR` equals `$HOME/.claude`, the wrapper unsets it so the CLI treats it as default. Fixes #4255.

- **Bug Fixes**
  - In `Resources/bin/claude`, added and invoked `normalize_claude_config_dir` to unset `CLAUDE_CONFIG_DIR` only when it equals `$HOME/.claude`; non-default paths are preserved.

<sup>Written for commit cf098c2a4eb104642141e41ddab0bf291d5f6f20. Summary will update on new commits. <a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/4401?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Chores**
  * Enhanced Claude configuration directory environment normalization in the cmux wrapper to improve consistency when default paths are used.

[Review Change Stack](https://app.coderabbit.ai/change-stack/manaflow-ai/cmux/pull/4401?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Resources/bin/claude
+++ b/Resources/bin/claude
@@ -35,12 +35,22 @@ cmux_socket_available() {
         "$cmux_bin" --socket "$socket" ping >/dev/null 2>&1
 }
 
+normalize_claude_config_dir() {
+    # Issue #4255: Claude treats CLAUDE_CONFIG_DIR being set to its default
+    # path differently from the variable being absent, so normalize that case.
+    if [[ "${CLAUDE_CONFIG_DIR:-}" == "$HOME/.claude" ]]; then
+        unset CLAUDE_CONFIG_DIR
+    fi
+}
+
 # Pass through if not in a cmux terminal, hooks are disabled, or the cmux
 # socket is unavailable (stale env / app not running).
 IN_CMUX=0
 if [[ -n "$CMUX_SURFACE_ID" ]]; then
     IN_CMUX=1
 fi
+
+normalize_claude_config_dir
 
 if [[ "$IN_CMUX" == "0" || "$CMUX_CLAUDE_HOOKS_DISABLED" == "1" ]] || ! cmux_socket_available; then
     # In cmux-launched shells, preserve old behavior and always clear nested


mvanhorn:
Greptile is right -- the PR description listed both fix sites but the diff only ships the wrapper-side defense-in-depth one. I went looking for the 3 Swift files I named in the issue body (AgentLaunchEnvironmentPolicy / TerminalStartupEnvironment / RestorableAgentSession) and grep does'nt find `CLAUDE_CONFIG_DIR` anywhere in the Swift sources -- the code that assembles the resume command lives somewhere I can't immediately locate. The wrapper-side normalization unconditionally unsets `CLAUDE_CONFIG_DIR=$HOME/.claude` before exec, so the bug is masked for the typical user path (claude resolves to the cmux wrapper). Any execution path that bypasses the wrapper still hits the re-auth, so the root-cause fix in the launch-command assembler is still worth doing as a follow-up if a maintainer can point at the right file.

Issue #4515, opened 2026-05-22
cmux hooks setup: surface more visibly to user (current entrypoint is buried)


lawrencecchen:
`cmux hooks setup` is the recommended path for installing per-agent hooks (`codex`, `claude`, `gemini`, `opencode`, etc.) into the user's agent config. Today the entrypoint is buried — discoverable only via `cmux --help` deep in the hooks section, and there's no visible nudge when the user starts a fresh agent in cmux that lacks hooks.

### Asks

1. **First-run UX**: when an agent starts in a cmux workspace and the manager detects no cmux hooks installed for that agent (no `.claude/settings.json` cmux hooks block, no `~/.codex/hooks.json` entry, etc.), surface a one-time inline toast / sidebar pill / status-line message: "cmux hooks not installed for codex — install via `cmux hooks setup --agent codex`".
2. **Settings UI entry**: add a `Hooks` section to `cmux.app` settings with one-click install/uninstall per agent (codex / claude / gemini / opencode / amp / pi). Mirror the existing keyboard-shortcut settings page UX.
3. **CLI improvements**:
   - `cmux hooks status` (already exists in some form) — surface to top-level help.
   - `cmux hooks setup` without an agent flag should print a list of agents found in PATH and offer to install for all detected ones in one shot.
4. **Docs**: link the hooks setup flow from the cmux app's first-run welcome and from the public docs landing page; today it's only in `cmux --help`.

### Why

The hooks are the difference between cmux understanding agent state (running / waiting / blocked / asking) and treating an agent terminal as opaque. Users who don't run the setup miss most of the manager / sidebar / notification value. Making it visible removes the "why is my sidebar not showing agent status?" support load.

PR #4682, opened 2026-05-24
Wait for shell prompt before typing 'cmux welcome' (fixes #1900)

### LINKED ISSUE #1900: new-workspace --command uses send_text instead of existing initial_command API, breaking vi-mode shells
## Description

`cmux new-workspace --command "..."` delivers the command via `surface.send_text` (keystroke simulation) instead of using the `initial_command` parameter that `workspace.create` already supports.

This causes the command to be interpreted by the shell's key binding handler rather than executed as text. On fish with `fish_vi_key_bindings`, the first characters of every command are consumed as vi normal-mode editing commands (e.g. `echo hello` becomes `o hello` because `e`=end-of-word, `c`=change, `h`=motion).

The fix exists in the codebase already. The V2 `workspace.create` handler reads `initial_command` from params (TerminalController.swift ~L3319), passes it through `TabManager.addWorkspace` → `Workspace.init` → `TerminalPanel` → `GhosttyTerminalView` → `surfaceConfig.command` → Ghostty's C API. The terminal process starts as the command directly, bypassing the shell entirely.

The CLI at `CLI/cmux.swift` ~L1670 creates the workspace without passing `initial_command`, then separately calls `surface.send_text` with the command text. The `Thread.sleep(0.5)` delay that was previously used to wait for shell init has been removed, but the send_text approach remains.

### Steps to Reproduce

```bash
# With fish + fish_vi_key_bindings as default shell
cmux new-workspace --command "echo hello > /tmp/test.txt"
# Expected: /tmp/test.txt contains "hello"
# Actual: fish shows "o hello > /tmp/test.txt" and errors with "Unknown command: o"
```

### Suggested Fix

In `CLI/cmux.swift`, the `new-workspace` handler should pass `commandOpt` as `params["initial_command"]` to the V2 `workspace.create` call, instead of using a separate `surface.send_text` call:

```swift
// Current (~L1670):
var params: [String: Any] = [:]
if let cwdOpt { params["cwd"] = resolvePath(cwdOpt) }
let response = try client.sendV2(method: "workspace.create", params: params)
// ... then send_text

// Suggested:
var params: [String: Any] = [:]
if let cwdOpt { params["cwd"] = resolvePath(cwdOpt) }
if let commandText = commandOpt { params["initial_command"] = commandText }
let response = try client.sendV2(method: "workspace.create", params: params)
// no send_text needed
```

### Environment

- cmux 0.62.2
- Fish 3.x with `fish_vi_key_bindings`
- macOS

juliankmazo:
## Summary

The first-launch welcome banner is injected by typing `cmux welcome\n` into the new workspace's PTY 0.5s after the Ghostty surface becomes ready. That blind delay races shell init. With oh-my-zsh's auto-update prompt reading stdin during sourcing, the `c` of `cmux welcome` is consumed by the `Would you like to update? [Y/n]` prompt, and the rest of the line lands on the real shell prompt as `mux welcome`, leaving every fresh oh-my-zsh user with:

```
[oh-my-zsh] Would you like to update? [Y/n]
[oh-my-zsh] You can update manually by running `omz update`
☁  ~  mux welcome
zsh: command not found: mux
```

This is the canonical instance of the bigger pattern called out in #1900: cmux types commands into a shell that isn't actually at a prompt yet.

## Approach

`Workspace` owns the wait-for-prompt state machine. cmux's shell integration already drives `panelShellActivityStates[panelId]` through `.promptIdle` / `.commandRunning` via `Workspace.updatePanelShellActivityState`; this PR wires that into a single owner:

- `Workspace.updatePanelShellActivityState` posts `.panelShellActivityStateDidChange` on every real state change (deduplicated; userInfo carries workspaceId, panelId, state).
- `Workspace.sendTextOnNextPromptIdle(_:beforeSend:)` is the single coordination point. It checks the synchronous fast path, otherwise registers a `[weak self]` observer that fires `beforeSend` + `sendText` exactly once when `.promptIdle` is first observed for the focused panel.
- Outstanding observer tokens are tracked on `Workspace` and swept in `deinit`, so workspaces without shell integration installed (or workspaces closed before integration fires) do not leak their observer block.
- `AppDelegate.sendWelcomeCommandWhenReady` is a one-line delegate to the workspace.
- `TabManager`'s auto-welcome branch calls the workspace directly. The old `TabManager.sendWelcomeWhenReady` (with its blind `asyncAfter(0.5)`) is deleted.

**No timing fallback.** If cmux shell integration is not installed the panel never reaches `.promptIdle` and the welcome banner is skipped. That's the architectural invariant the project's `swift-blocking-runtime.md` rule asks for — any time-based fallback would still race shell init, which is the bug we're fixing. Users without shell integration get a clean prompt instead of `mux welcome` / `command not found: mux`; the better fix for them is to surface `cmux hooks setup` more prominently (#4515) rather than to keep typing into a shell whose state we can't verify.

## Net change

```
 Sources/AppDelegate.swift                       |  18 ++---
 Sources/TabManager.swift                        |  92 +---------------------
 Sources/TerminalController.swift                |   9 +++
 Sources/Workspace.swift                         |  74 ++++++++++++++++++++
 cmuxTests/PanelShellActivityNotificationTests.swift |  126 ++++++++++++++
 cmux.xcodeproj/project.pbxproj                  |   4 +
 6 files changed, 219 insertions(+), 104 deletions(-)
```

More code is deleted than added. No new `DispatchQueue.asyncAfter`, no new `NSLog`, no duplicate coordination logic.

## Regression test

Per `CLAUDE.md`'s two-commit regression policy, this PR is structured as:

1. **Add failing test** — `PanelShellActivityNotificationTests` asserting the new notification is posted with the expected userInfo and not posted for duplicate state. CI should go red on this commit alone (no post site exists).
2. **Apply the fix** — production change in `Workspace.swift`, `AppDelegate.swift`, `TabManager.swift`. CI should go green.

Later commits in the branch address reviewer feedback (single-owner refactor, observer leak fix, synchronous test delivery to remove flake risk).

Tests cover:
- Notification fires on real state transitions with correct userInfo (`testStateChangePostsNotificationWithWorkspaceAndPanelIds`).
- Notification does NOT fire on duplicate state (`testDuplicateStateDoesNotPostNotification`).
- `sendTextOnNextPromptIdle` waits, fires exactly once on first `.promptIdle`, and is one-shot through later state bounces (`testSendTextOnNextPromptIdleFiresExactlyOnceAtFirstPromptIdle`).
- The pending observer does not retain its Workspace when `.promptIdle` never fires (`testSendTextOnNextPromptIdleDoesNotLeakWorkspaceWhenPromptIdleNeverFires`).

`./scripts/lint-pbxproj-test-wiring.sh` passes locally.

## Test plan

- [ ] CI: `cmuxTests` runs `PanelShellActivityNotificationTests` and the rest of the suite passes
- [ ] Manual: launch a fresh-install cmux with cmux shell integration set up (`cmux hooks setup`). Banner appears under the prompt, no `command not found: mux`, even with oh-my-zsh's auto-update prompt waiting on stdin
- [ ] Manual: launch with `~/.zshrc` that does not run oh-my-zsh and no shell integration — welcome banner is skipped (intentional; better than typing into an unverified prompt)
- [ ] Manual: open many workspaces in succession without shell integration — no Workspace retain cycles or growing observer count

## Notes for reviewers

- The single-owner refactor (`Workspace.sendTextOnNextPromptIdle`) was done in response to Greptile's P1 (duplicate coordination logic) and CodeRabbit's architectural-rethink pre-merge check.
- The decision to skip the welcome for users without shell integration is deliberate: the project rules prohibit timing-based race repair on latency-sensitive paths, and any timing fallback would still hit the same bug it's meant to fix. If a non-timing fallback is preferred (e.g. write the banner content to the terminal as output instead of input, bypassing the PTY input race entirely), happy to follow up — that needs new plumbing through `ghostty_surface_*`.
- CodeRabbit's `@concurrent` finding is a false positive: `workspacePullRequestAuthHeaderValue` was added by Austin Wang in `e39f5e02c8` (April 2026) and is unmodified by this PR.

## Summary by CodeRabbit

* **New Features**
  * Notify when a panel’s shell activity state changes and add a one-shot "send text on next prompt idle" delivery to the focused terminal.

* **Bug Fixes**
  * Consolidated welcome sending to the workspace-level mechanism: welcome text is deferred until the terminal is ready and the onboarding-shown flag is set when scheduled.

* **Tests**
  * Added tests covering notifications, prompt-idle delivery semantics, one-shot behavior, idempotency, and leak regression.

[Review Change Stack](https://app.coderabbit.ai/change-stack/manaflow-ai/cmux/pull/4682?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -7695,7 +7695,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
     }
 
     func sendWelcomeCommandWhenReady(to workspace: Workspace, markShownOnSend: Bool = false) {
-        sendTextWhenReady("cmux welcome\n", to: workspace) {
+        // The welcome command is typed into the user's shell, so it can collide with
+        // anything that consumes stdin during shell init (oh-my-zsh's auto-update prompt
+        // is the most common culprit). Wait for the shell to report it is sitting at
+        // an actual prompt (cmux shell integration emits OSC sequences that drive
+        // `panelShellActivityStates[panelId] == .promptIdle`) before sending. If shell
+        // integration isn't installed we still send eventually via a fallback timeout
+        // so the welcome banner doesn't silently disappear.
+        sendTextWhenReady(
+            "cmux welcome\n",
+            to: workspace,
+            waitForShellPromptIdle: true
+        ) {
             if markShownOnSend {
                 UserDefaults.standard.set(true, forKey: WelcomeSettings.shownKey)
             }
@@ -8290,9 +8301,21 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         _ text: String,
         to tab: Tab,
         preferredPanelId: UUID? = nil,
+        waitForShellPromptIdle: Bool = false,
         beforeSend: (() -> Void)? = nil
     ) {
         let isReactGrabPasteback = preferredPanelId != nil
+
+        // When the caller asked us to wait for an actual interactive shell prompt
+        // (welcome injection is the canonical case), require both the surface to be
+        // ready and the shell-integration state for the target panel to be
+        // `.promptIdle`. Without this, typed bytes race with anything that
+        // consumes stdin during shell init (oh-my-zsh's auto-update prompt
+        // famously eats the first character: https://github.com/manaflow-ai/cmux/issues/1900).
+        func shellPromptReady(panel: TerminalPanel) -> Bool {
+            guard waitForShellPromptIdle else { return true }
+            return tab.panelShellActivityStates[panel.id] == .promptIdle
+        }
 #if DEBUG
         let initialTargetPanel = Self.resolveTerminalPanelForTextSend(
             in: tab,
@@ -8324,7 +8347,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
             in: tab,
             preferredPanelId: preferredPanelId
         ),
-           terminalPanel.surface.surface != nil {
+           terminalPanel.surface.surface != nil,
+           shellPromptReady(panel: terminalPanel) {
 #if DEBUG
             if isReactGrabPasteback {
                 cmuxDebugLog(
@@ -8352,6 +8376,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         var readyObserver: NSObjectProtocol?
         var focusObserver: NSObjectProtocol?
         var firstResponderObserver: NSObjectProtocol?
+        var shellActivityObserver: NSObjectProtocol?
         var panelsCancellable: AnyCancellable?
 
         func cleanupObservers() {
@@ -8364,7 +8389,26 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
             if let firstResponderObserver {
                 NotificationCenter.default.removeObserver(firstResponderObserver)
             }
+            if let shellActivityObserver {
+                NotificationCenter.default.removeObserver(shellActivityObserver)
+            }
             panelsCancellable?.cancel()
+        }
+
+        func send(panel terminalPanel: TerminalPanel, mode: String) {
+            resolved = true
+            cleanupObservers()
+            beforeSend?()
+            terminalPanel.sendText(text)
+#if DEBUG
+            if isReactGrabPasteback {
+                cmuxDebugLog(
+                    "reactGrab.pasteback h2.send.sent " +
+                    "workspace=\(Self.debugShortId(tab.id)) " +
+                    "target=\(Self.debugShortId(terminalPanel.id)) mode=\(mode) len=\(text.count)"
+                )
+            }
+#endif
         }
 
         func finishIfReady() {
@@ -8386,20 +8430,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
 #endif
             guard !resolved,
                   let terminalPanel,
-                  terminalPanel.surface.surface != nil else { return }
-            resolved = true
-            cleanupObservers()
-            beforeSend?()
-            terminalPanel.sendText(text)
-#if DEBUG
-            if isReactGrabPasteback {
-                cmuxDebugLog(
-                    "reactGrab.pasteback h2.send.sent " +
-                    "workspace=\(Self.debugShortId(tab.id)) " +
-                    "target=\(Self.debugShortId(terminalPanel.id)) mode=delayed len=\(text.count)"
-                )
-            }
-#endif
+                  terminalPanel.surface.surface != nil,
+                  shellPromptReady(panel: terminalPanel) else { return }
+            send(panel: terminalPanel, mode: "delayed")
         }
 
         panelsCancellable = tab.$panels
@@ -8484,7 +8517,28 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
             }
             finishIfReady()
         }
-        DispatchQueue.main.asyncAfter(deadline: .now() + 3.0) {
+        if waitForShellPromptIdle {
+            shellActivityObserver = NotificationCenter.default.addObserver(
+                forName: .panelShellActivityStateDidChange,
+                object: nil,
+                queue: .main
+            ) { note in
+                guard let workspaceId = note.userInfo?[PanelShellActivityNotificationKey.workspaceId] as? UUID,
+                      workspaceId == tab.id else { return }
+                if let preferredPanelId,
+                   let surfaceId = note.userInfo?[PanelShellActivityNotificationKey.panelId] as? UUID,
+                   surfaceId != preferredPanelId {
+                    return
+                }
+                finishIfReady()
+            }
+        }
+        // When we're waiting on shell integration to report a real prompt, give
+        // shells with heavy init (oh-my-zsh, starship, plugin managers) more time
+        // before falling back, and SEND on timeout so the welcome banner isn't
+        // silently dropped for users without cmux shell integration installed.
+        let fallbackDeadline: TimeInterval = waitForShellPromptIdle ? 8.0 : 3.0
+        DispatchQueue.main.asyncAfter(deadline: .now() + fallbackDeadline) {
             if !resolved {
 #if DEBUG
                 if isReactGrabPasteback {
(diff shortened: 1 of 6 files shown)


Sources/AppDelegate.swift
```diff
@@ -8403,8 +8457,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
...
 #endif
-                cleanupObservers()
-                NSLog("Command send: surface not ready after 3.0s")
+                if waitForShellPromptIdle,
+                   let terminalPanel = Self.resolveTerminalPanelForTextSend(
+                       in: tab,
+                       preferredPanelId: preferredPanelId
+                   ),
+                   terminalPanel.surface.surface != nil {
+                    send(panel: terminalPanel, mode: "timeoutFallback")
```
juliankmazo: This fallback can recreate the same stdin race for users who do have shell integration but are still blocked in a shell-init prompt after 8s. In that case `.promptIdle` has not arrived because `precmd` has not run yet, but we still type `cmux welcome\n` into whatever is reading stdin, including the oh-my-zsh update prompt. That means the PR does not fully fix the shell-integration case; it just moves the blind timer from 0.5s to 8s. The fallback needs to distinguish “no shell integration signal is possible” from “shell integration exists but the prompt is not idle yet”, or avoid sending while a non-idle state is known.

cmuxTests/PanelShellActivityNotificationTests.swift
```diff
@@ -0,0 +1,76 @@
...
+            notificationCount += 1
+        }
+        defer { NotificationCenter.default.removeObserver(observer) }
+
+        // No state change — must not post.
+        workspace.updatePanelShellActivityState(panelId: panelId, state: .promptIdle)
+        XCTAssertEqual(notificationCount, 0)
+
+        // Real transition — must post exactly once.
+        workspace.updatePanelShellActivityState(panelId: panelId, state: .commandRunning)
```
juliankmazo, line 86: This assertion is likely racy. The observer was registered with `queue: .main`, so delivery is queued asynchronously; asserting `notificationCount == 1` immediately after posting can fail on CI even if the production post is correct. Existing notification-style tests in this repo usually use `queue: nil` for synchronous assertions or an expectation. Please make this deterministic before relying on this as the red/green regression proof.

Sources/TabManager.swift
```diff
@@ -2620,32 +2620,43 @@ class TabManager: ObservableObject {
 
     @MainActor
     private func sendWelcomeWhenReady(to workspace: Workspace) {
```
juliankmazo: This duplicates the same observer/resolved/fallback state machine now living in `AppDelegate.sendTextWhenReady(waitForShellPromptIdle:)`. Per the shared behavior policy, the prompt-idle wait should have one implementation that both entrypoints use. The two copies already differ in details and will be easy to regress independently when the timeout/state behavior changes.

juliankmazo (commented):
Blocking from my side, even though GitHub will only let me post this as a comment on my own PR. The notification signal is a good direction, but the current fallback can still type into an active shell-init prompt, and the new regression test has a likely async assertion race.

juliankmazo:
Thanks — both addressed in the follow-up commits:

- **Fallback can still type into active shell-init prompt** — the 8s `DispatchQueue.asyncAfter` fallback is gone. The project's `swift-blocking-runtime.md` rule banned it on this latency-sensitive path, but you're right that the more important reason is correctness: any time-based fallback still races the omz prompt. Now if shell integration isn't installed, `.promptIdle` never fires and the welcome banner is skipped on purpose. PR description updated to call this out clearly instead of promising a fallback that doesn't exist. Commit: `348d8329d`.

- **Async assertion race in the regression test** — the observer is now registered with `queue: nil` (synchronous delivery on the poster's thread) and uses `MainActor.assumeIsolated` to call back into MainActor logic without a `Task { @MainActor in ... }` hop. The test asserts immediately after each `updatePanelShellActivityState` call, no runloop spin. A comment in the test pins this invariant so a future refactor that reintroduces an async hop is caught. Commit: `876055c5`.

Also fixed an observer leak Greptile flagged on the same pass: tokens are tracked on `Workspace` and swept in `deinit`, with `[weak self]` in the closure as a secondary safety net. New test `testSendTextOnNextPromptIdleDoesNotLeakWorkspaceWhenPromptIdleNeverFires` covers it.

juliankmazo:
Thanks for the assertive review. Addressing each of the four findings:

**1. Test uses `RunLoop.current.run(until: ...)` → flake risk.** Fixed in [`876055c5`](https://github.com/manaflow-ai/cmux/pull/4682/commits/876055c5). The observer is now registered with `queue: nil` (synchronous delivery on the posting thread) and uses `MainActor.assumeIsolated` so notifications drive the send without a `Task { @MainActor in ... }` hop. The test asserts immediately after each `updatePanelShellActivityState` with no runloop spins. A comment in the test pins this invariant.

**2. `WelcomeSettings.shownKey` only flipped on send → infinite retries when shell integration is missing.** Fixed in [`936851143`](https://github.com/manaflow-ai/cmux/pull/4682/commits/936851143). The auto-welcome path now flips `shownKey` synchronously the first time it *attempts* to type the welcome, not inside `beforeSend`. The welcome banner is one-shot onboarding; one attempt per fresh install is the correct semantic, regardless of whether the banner actually rendered. Also dropped the now-dead `markShownOnSend` parameter on `AppDelegate.sendWelcomeCommandWhenReady`.

**3. `TabManager` welcome waits forever without shell integration.** Deliberately not adding a timing fallback in this PR. The project's `swift-blocking-runtime.md` rule prohibits new `DispatchQueue.asyncAfter` on latency-sensitive paths (and was the basis for one of the pre-merge errors on the previous revision). More importantly, any timing fallback would still race the exact bug we're fixing — oh-my-zsh's auto-update prompt blocks stdin indefinitely, so no fixed timeout is safe. With finding #2 fixed, the cost is "users without `cmux hooks setup` don't see the auto-banner once" rather than "permanent retry loop." A proper UX nudge to run `cmux hooks setup` belongs with #4515 ("surface `cmux hooks setup` more visibly to user") which is the actual root cause for this user segment.

   If reviewers want the auto-banner to land for users without shell integration, the right plumbing is to write the welcome content to the terminal as output (bypassing the PTY input race entirely) rather than typing — but that needs a new Ghostty surface API and is out of scope for the bug fix.

**4. Observer leak in `sendTextOnNextPromptIdle`.** Fixed in [`876055c5`](https://github.com/manaflow-ai/cmux/pull/4682/commits/876055c5). The closure captures `self` weakly, tokens are tracked on `Workspace.pendingPromptIdleObservers`, and the existing `Workspace.deinit` sweeps any outstanding observer. New regression test `testSendTextOnNextPromptIdleDoesNotLeakWorkspaceWhenPromptIdleNeverFires` uses `weak var` + `autoreleasepool` to pin this.

Note: the `@concurrent` finding from the pre-merge checks is a false positive — `workspacePullRequestAuthHeaderValue` was added by Austin Wang in `e39f5e02c8` (April 2026) and is unchanged in this PR.

juliankmazo:
@coderabbitai review

juliankmazo:
@coderabbitai review

PR #4720, opened 2026-05-25
Fix file explorer not refreshing on changes in nested subdirectories

### LINKED ISSUE #4649: File explorer doesn't auto-refresh on changes inside nested (non-root) subdirectories
## Summary

The Files sidebar does not auto-refresh when a file or folder is created (or deleted/renamed) inside a **nested** subdirectory — i.e. anywhere below a *direct* child of the explorer root. The new entry exists on disk and is visible in Finder, but never appears in cmux's tree until the app is restarted or some unrelated change happens at the root level.

## Steps to reproduce

1. Open the Files sidebar on a local folder, e.g. root = `~/Documents/Work`.
2. Expand a subfolder, e.g. `~/Documents/Work/project-a/` (so its children are loaded/cached).
3. From outside cmux (Finder, a terminal, or an AI agent running in a pane), create a new folder/file **inside that already-expanded subfolder**, e.g. `mkdir ~/Documents/Work/project-a/build && touch ~/Documents/Work/project-a/build/index.html`.
4. Look at the Files sidebar.

**Expected:** `build/` appears under `project-a/` (it's on disk; Finder shows it).
**Actual:** It never shows up. Collapsing and re-expanding `project-a/` does **not** help. Only restarting cmux, switching the root, or making a change to a *direct child of the root* makes it appear.

## Root cause

The only auto-refresh mechanism is a single, **non-recursive** kqueue vnode watch on the **root directory only**:

- `Sources/FileExplorerStore.swift:749` `updateDirectoryWatcher()` calls `directoryWatcher?.watch(path: rootPath)` — only `rootPath` is ever watched.
- `Sources/FileExplorerStore.swift:1129` `FileExplorerDirectoryWatcher.watch(path:)` does `open(path, O_EVTONLY)` + `DispatchSource.makeFileSystemObjectSource(... eventMask: [.write, .link, .rename, .delete])`.

A kqueue vnode source on a directory fd fires only when the **immediate contents of that one directory** change. Creating `project-a/build/` modifies the mtime of `project-a/`, **not** of the watched `rootPath`, so no event fires and `reload()` is never called.

Compounding it: once a directory is expanded, its listing is cached in `node.children`. `expand()` (`FileExplorerStore.swift:801`) only reads from disk when `children == nil`, and `collapse()` does **not** clear the cache — so collapse+re-expand reuses the stale listing. The only thing that re-lists nested expanded dirs from disk is a full `reload()` (which does recurse over expanded paths — `FileExplorerStore.swift:934-944`), and `reload()` is only triggered by `setRootPath`, the root watcher firing, or SSH hydrate.

## Why it's easy to miss

In active coding workflows the condition is hit constantly (an agent scaffolds `src/feature/…`, a tool writes into a nested dir), but it usually **self-heals within seconds** because normal activity keeps tripping the root watcher — git branch switches, `npm/bun install`, saving/editing a root-level file (`package.json`, lockfiles, `README`), etc. all touch a *direct child of the root* and trigger a full `reload()` that recursively re-lists expanded subtrees.

It becomes **permanently visible** when all file activity is concentrated deep in one subtree and nothing ever touches the root — e.g. using an agent to generate a self-contained nested project/site folder, or "one folder per project / research task" usage. There it never self-heals; restart is the only way.

This fits the broader "external state changed, UI doesn't auto-update" family already tracked here (see related).

## Proposed fixes (in rough order of merge-friendliness / risk)

1. **Watch the set of currently-expanded directories**, not just the root (one kqueue source per expanded dir, added on expand / removed on collapse). fd cost is bounded by visible/expanded dirs (small), and it never watches `node_modules`/`.git` unless the user explicitly expands them. This precisely fixes the symptom for exactly what the user can see. *(Preferred.)*
2. **Recursive `FSEventStream`** over the root with an ignore list (`.git`, `node_modules`, build output) and debounced `reload()`. More complete (collapsed dirs also become fresh), but heavier and riskier on large repos — event storms → repeated `reload()` + `git status --porcelain`, which is the CPU-spin class this codebase is careful about. Would need solid exclusion logic.
3. **Low-risk safety net**, regardless of the above: clear `node.children` in `collapse()` so collapse+re-expand re-reads disk, and/or add an explicit **Refresh** action (context menu / shortcut) that calls `store.reload()`. Small and deterministic; gives users a manual escape hatch today.

Happy to implement (1) — with red/green regression coverage per the contributing rules — if a maintainer can confirm the preferred approach before I open a PR.

## Related

- #666 — sidebar doesn't update branch name on agent checkout
- #4531 — Antigravity sessions don't appear in Vault until session exits
- #2121 — sidebar updates but terminal content doesn't refresh on workspace switch
- #3742 — Files sidebar flickers when closing a pane
- #4644 — File Explorer: no New Folder / New File / Rename / Delete actions

## Environment

- cmux `0.64.9` (stable), macOS 26.5 (25F71)
- Local (non-SSH) Files root
- Confirmed by reading current source; behavior present in latest `main`/nightly.

zackzeng00:
## Summary

Fixes #4649.

The Files sidebar only auto-refreshed from a single, non-recursive kqueue vnode watch on the explorer **root**. A vnode source only reports changes to the *immediate* contents of the one directory it watches, so creating/deleting/renaming an entry inside an already-expanded **nested** subdirectory never tripped the watch — the change stayed invisible until a restart, a root switch, or some unrelated activity happened to touch a *direct child of the root* and trigger a full reload.

This implements the issue's preferred fix (option 1): **watch every expanded directory, not just the root.**

## Changes

- `FileExplorerDirectoryWatcher` now manages one vnode source per path via `watch(paths:)`, diffing the requested set against live sources so existing watches are kept (in-flight events aren't dropped) and only deltas are added/removed. Each source owns and closes its own fd in its cancel handler.
- `updateDirectoryWatcher()` computes `{root} ∪ expandedPaths`, filtered to the current root via the existing containment helper so stale paths from a previous root don't leak file descriptors.
- `expand()` / `collapse()` resync the watch set, so the watched directories track exactly what the user can see.

`reload()` already re-lists every expanded subtree, so a single watcher firing is enough to refresh the visible tree. The fd count stays bounded by the directories the user has actually expanded, and `node_modules`/`.git` are only ever watched if the user explicitly expands them.

## Test plan

Two-commit red/green structure per the contributing rules:

1. **Commit 1** adds `testCreatingEntryInNestedExpandedDirectoryRefreshesTree`, which drives a real `LocalFileExplorerProvider` against a temp directory, expands a nested subdirectory, then creates a folder *inside* it and asserts it appears in the tree without a restart. This fails against the root-only watcher.
2. **Commit 2** adds the fix; the test passes.

The test exercises the kqueue watch end-to-end through the store's public API.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/4720"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782283829&installation_id=133855650&pr_number=4720&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F4720&signature=840e86767ea1248b6dc18c482d16149997c718092f4beb86f948710ddd9e2cd3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Watch the root and all visible expanded directories so nested folder changes refresh immediately. Prevents missing updates in expanded subfolders until a restart.

- **Bug Fixes**
  - Switched to per-path vnode watching via `watch(paths:)` in `FileExplorerDirectoryWatcher`, diffing adds/removes and closing fds on cancel.
  - `updateDirectoryWatcher()` now watches `{root} ∪ visibleExpandedPaths`: only paths under the current root whose full ancestor chain is expanded; `expand()`/`collapse()` resync the set so fd usage matches the visible tree.
  - Guarded debounce scheduling with a lock to avoid races between the watch queue and main thread.
  - Tests: regression for nested-dir refresh and a watch-set visibility test via `watchedDirectoryPathsForTesting()`.

<sup>Written for commit afc6dc274690740789e283c970cc242cf8f7147e. Summary will update on new commits. <a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/4720?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **New Features**
  * File explorer now monitors changes in real-time across all expanded directories and nested subdirectories, improving responsiveness to filesystem updates.

* **Bug Fixes**
  * Directory watcher now stays in sync with expanded/collapsed folders so new files and folders appear reliably without restarting.

* **Tests**
  * Added regression tests validating nested expansion updates and that the watched set matches visible expanded directories.

[Review Change Stack](https://app.coderabbit.ai/change-stack/manaflow-ai/cmux/pull/4720?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/FileExplorerStore.swift
+++ b/Sources/FileExplorerStore.swift
@@ -761,17 +761,36 @@ final class FileExplorerStore: ObservableObject {
     }
 
     /// Directories the local watcher should observe: the root plus every
-    /// currently expanded directory under it. A kqueue vnode source only fires
-    /// for the immediate contents of the one directory it watches, so a
-    /// root-only watch never notices entries created inside nested expanded
-    /// subdirectories (issue #4649). Paths from a previous root are filtered out
-    /// so the fd count stays bounded by what the user can actually see.
+    /// expanded directory that is currently *visible* in the tree. A kqueue
+    /// vnode source only fires for the immediate contents of the one directory
+    /// it watches, so a root-only watch never notices entries created inside
+    /// nested expanded subdirectories (issue #4649). The fd count stays bounded
+    /// by what the user can actually see: directories hidden under a collapsed
+    /// ancestor (and anything from a previous root) are excluded.
     private func watchedDirectoryPaths() -> Set<String> {
         var paths: Set<String> = [rootPath]
-        for path in expandedPaths where Self.path(path, isContainedIn: rootPath) {
+        for path in expandedPaths where isVisibleExpandedDirectory(path) {
             paths.insert(path)
         }
         return paths
+    }
+
+    /// Whether an expanded directory is actually rendered right now: it must
+    /// live under the current root and every directory between it and the root
+    /// must also be expanded. `collapse()` only forgets the collapsed node
+    /// itself (so re-expanding restores descendants), which is why a descendant
+    /// can remain in `expandedPaths` while hidden — such paths must not be
+    /// watched.
+    private func isVisibleExpandedDirectory(_ path: String) -> Bool {
+        guard Self.path(path, isContainedIn: rootPath) else { return false }
+        var ancestor = (path as NSString).deletingLastPathComponent
+        while ancestor != rootPath, Self.path(ancestor, isContainedIn: rootPath) {
+            if !expandedPaths.contains(ancestor) { return false }
+            let parent = (ancestor as NSString).deletingLastPathComponent
+            if parent == ancestor { break }
+            ancestor = parent
+        }
+        return true
     }
 
     private func setProvider(_ newProvider: FileExplorerProvider?, reloadIfAvailable: Bool = true) {
@@ -788,6 +807,11 @@ final class FileExplorerStore: ObservableObject {
     #if DEBUG
     func setProviderForTesting(_ newProvider: FileExplorerProvider?, reloadIfAvailable: Bool = true) {
         setProvider(newProvider, reloadIfAvailable: reloadIfAvailable)
+    }
+
+    /// Exposes the local watcher's target directory set for tests.
+    func watchedDirectoryPathsForTesting() -> Set<String> {
+        watchedDirectoryPaths()
     }
     #endif
 
@@ -1138,14 +1162,16 @@ final class FileExplorerStore: ObservableObject {
 /// created inside nested expanded subdirectories (issue #4649).
 ///
 /// Debounces events to avoid rapid-fire reloads during bulk operations (e.g.,
-/// git checkout). `watch(paths:)`/`stop()` are only called from the main actor;
-/// the per-source event handlers run on `watchQueue` and only touch the
-/// debounce work item.
+/// git checkout). `watch(paths:)`/`stop()` are only called from the main actor
+/// (so the `watches` map needs no extra synchronization), while the per-source
+/// event handlers run on `watchQueue`; `debounceWorkItem` is therefore touched
+/// from both and guarded by `debounceLock`.
 final class FileExplorerDirectoryWatcher {
     /// path -> live vnode source. Each source owns its file descriptor and
     /// closes it from its cancel handler, so cancelling is enough to release it.
     private var watches: [String: DispatchSourceFileSystemObject] = [:]
     private let watchQueue = DispatchQueue(label: "com.cmux.fileExplorerWatcher", qos: .utility)
+    private let debounceLock = NSLock()
     private var debounceWorkItem: DispatchWorkItem?
     private let onChange: () -> Void
 
@@ -1186,8 +1212,10 @@ final class FileExplorerDirectoryWatcher {
     }
 
     func stop() {
+        debounceLock.lock()
         debounceWorkItem?.cancel()
         debounceWorkItem = nil
+        debounceLock.unlock()
         for source in watches.values {
             source.cancel()
         }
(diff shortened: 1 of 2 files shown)


zackzeng00:
Thanks for the review — addressed both points in afc6dc2:

**1. fd accumulation on collapse.** Correct. `collapse()` deliberately forgets only the collapsed node (so re-expanding restores descendants' expansion state), which left hidden descendants in `expandedPaths` and therefore watched. Rather than recursively pruning `expandedPaths` (which would break that restore behavior), `watchedDirectoryPaths()` now keeps a path only when its entire ancestor chain up to the root is expanded — i.e. only directories actually rendered are watched. Re-expanding the parent re-watches them. Added `testWatchSetTracksVisibleExpandedDirectoriesOnly` to lock this in.

**2. `debounceWorkItem` race.** `debounceWorkItem` is now guarded by an `NSLock`, since it's written from `watchQueue` (event handler) and cancelled from the main actor (`stop()`). I did *not* add `@MainActor` to the watcher as suggested: its event handlers (`scheduleReload`) run on `watchQueue` by design, so the type can't be main-actor isolated. The `watches` map remains main-actor-only and needs no lock; I documented that contract on the class.

PR #5058, opened 2026-05-31
Extract CmuxSocketControl package (modular refactor, wave 2 PR A)

azooz2003-bit:
Wave-2 leaf of the modular refactor. Grows the former `CMUXSocketPathDomain` package into `CmuxSocketControl`, the single source of truth for control-socket configuration and auth. Behavior is unchanged.

## What moved into the package
- `SocketControlMode` (cases, file permissions, auth requirement).
- `SocketControlSettings` path/mode resolution + env overrides (pure static functions; all deps injected).
- Socket-path marker recording (`SocketControlSettings+SocketPathMarkers`), alongside the existing `SocketPathMarkerFiles` / `SocketPathVariant`.

## Password store redesign
`SocketControlPasswordStore` goes from a namespace-enum of statics to an injectable `Sendable` struct: environment, password-file URL, and the legacy keychain load/delete ops are constructor-injected, so it is fully testable against a temp file and a fake keychain. The API stays synchronous, so `TerminalController`'s `nonisolated` socket-auth path is unchanged. The lazy keychain-fallback cache uses `OSAllocatedUnfairLock` (justified inline: the sync auth path has no async seam to await an actor).

## App target
- Display strings stay in the app (`SocketControlMode+Display`) since `String(localized:)` resolves the app's catalog, keeping the package UI-agnostic.
- `TerminalController` injects the password store via `init`; `cmuxApp` and the settings file store use injected/constructed instances instead of statics.
- 8 consumer files gain `import CmuxSocketControl`.

## Tests
- 21 Swift Testing tests in the package (mode, settings resolution, password load/verify/save/migrate, marker files).
- App-target tests that used the moved types now import the package; the old static-API password test is migrated to the injected struct API.

Package renamed `CMUXSocketPathDomain` -> `CmuxSocketControl` in the Xcode project and workspace.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/5058"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782850689&installation_id=133855650&pr_number=5058&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F5058&signature=ae2d0a95c57345c9d90d90d4b6d3acc926ceb9abeec71d0392a0b1d04871dd52"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Extracted socket-control config and auth into a new `CmuxSocketControl` package and converged the socket password onto one secure file owned by `CmuxSettings`. Behavior is unchanged; plaintext in `cmux.json` is migrated out, and consumers import the package with an injectable, testable password store.

- **New Features**
  - Added `SecretFileStore` and `SecretFileKey`; `automation.socketPassword` now lives in a `0600` file and binds via `SecretValueModel` in Settings.
  - Added `PlaintextSecretMigration` to lift any legacy plaintext password into the secure file and scrub the config; idempotent and JSONC‑tolerant.

- **Refactors**
  - `CmuxSocketControl` now owns control-socket types and marker recording; `SocketControlMode` gains `socketFilePermissions` and `requiresPasswordAuth`.
  - `SocketControlPasswordStore` is an injectable `Sendable` struct with `FileManager` injection and constant‑time verify; removed lock/cache; `TerminalController`, `cmuxApp`, and settings file store construct/inject it.
  - Hardened paths and secrets: sanitize `SocketPathVariant` slugs; validate `SecretFileKey.fileName`; fix `SecretFileStore.hasValue` to read the file directly.
  - Migration robustness: make secret save throwing to avoid data loss on failure; string‑aware trailing‑comma stripping; leave unparseable files intact.
  - CI runs `swift test` for `CmuxFoundation`, `CmuxSettings`, `CmuxSettingsUI`, and `CmuxSocketControl`; fixed a cmuxTests build error by using the correct `String.write` label.

<sup>Written for commit 516d11431c1ba3cea067ae68f8a792400a13e3b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/5058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Configure socket passwords via environment or app storage with verification and migration from legacy keychain.
  * Stable per-flavor socket path resolution (stable/nightly/staging/dev) with predictable marker locations.
  * Mode display metadata for clearer Settings UI (names, descriptions, ordering).

* **Bug Fixes**
  * Reliable one‑time legacy password migration with retry on failure and change notifications.

* **Chores**
  * Socket control logic moved into a dedicated module and tests added/updated.


azooz2003-bit:
Addressed the review feedback in `061cad6` (CI compile fix) and `516d114` (review feedback). Mapping each thread:

**Fixed**
- **Lock in `SocketControlPasswordStore` (CodeRabbit / Greptile public reset helper):** removed `OSAllocatedUnfairLock` and the lazy cache entirely. The keychain is the lowest-priority source and is only read on demand when neither env nor file holds a password, so the store keeps no shared mutable state and stays trivially `Sendable`. I did not take the eager-load-in-init suggestion: it would read the keychain on every construction and break the "not consulted unless allowed" contract (and the test that asserts it). The public `resetLazyKeychainFallbackCacheForTests` is gone with the cache.
- **Inject `FileManager` (CodeRabbit):** injected via init (`nonisolated(unsafe)`, with a thread-safety justification comment), replacing `FileManager.default`.
- **`.newlines` vs `.whitespacesAndNewlines` (CodeRabbit / cubic):** both `savePassword` and `normalized` now trim `.whitespacesAndNewlines`; added coverage for surrounding-space trimming and whitespace-only clearing.
- **Non-constant-time compare (Greptile):** `verify` now uses a constant-time comparison.
- **`SecretFileKey.fileName` validation (cubic):** `init` rejects path separators and `..`.
- **`SocketPathVariant` slug sanitization (CodeRabbit):** slug is reduced to a filesystem-safe token before building marker/`/tmp` paths.
- **`SecretFileStore.hasValue` (cubic):** reads the stored file directly, ignoring `defaultValue`, with a regression test.
- **`PlaintextSecretMigration.saveSecret` failure (cubic):** `saveSecret` is now throwing; on failure the migration leaves `cmux.json` intact (`saveFailedLeftIntact`) so the only copy of the secret is never lost. Call site propagates the error.
- **Trailing-comma regex (cubic):** replaced the global regex with a string-aware pass so a `,}`/`,]` inside a string value is preserved; added a regression test.
- **`SecretValueModel` redundant `MainActor.run` (cubic):** removed; the `Task` inherits the `@MainActor` method's isolation.
- **Inject `passwordStore` into `CmuxSettingsFileStore` (CodeRabbit / cubic):** added as an injected init parameter.
- **`NSLock` in package test helper (CodeRabbit):** `Counter` is now lock-free under a `.serialized` suite.
- **XCUITest launch-bypass coverage (CodeRabbit):** added the `CMUX_UI_TEST_` case.

**Not changed (with reasons)**
- **`cmuxApp.swift` `hasSocketPasswordConfigured` synchronous read (CodeRabbit):** this lives in the legacy `SettingsView` chain in `cmuxApp.swift` that is already dead (the live settings window uses the `SettingsWindowRoot` package path) and is scheduled for deletion in a separate follow-up PR, so I'm not modifying it here.
- **Missing `km`/`zh-Hant` socketControl translations (CodeRabbit):** `zh-Hant` is already present for all ten `socketControl.*` keys. `km` is sparsely populated catalog-wide (~230 of ~2578 keys) and its absence here matches the rest of the catalog and the en+ja policy, so this is not a regression.

Also fixed a pre-existing compile error this surfaced: `cmuxTests/SocketControlPasswordStoreTests.swift` used `write(to:atomic:encoding:)` instead of `atomically:`, which broke the `tests` job's `cmuxTests` build (the `cmux` scheme doesn't compile that target). That was the entire `tests` failure.

All four package suites pass locally (`swift test`, 80 tests, matching the CI `--skip userDefaultsStorageKeysAreUnique` gate) and the full `cmux` app scheme builds clean.

PR #5113, opened 2026-06-01
Add @LiveSetting: @State + store stream, reactive in the AppKit-hosted main window (all key kinds)

azooz2003-bit:
Resolves the root cause behind #5094/#5110: settings read in the main window weren't reactive because `ContentView` is hosted in an AppKit `NSHostingView` that doesn't inherit the App scene's environment, and `@Setting` reads through an `@Observable` model whose async update SwiftUI didn't turn into a redraw in that host.

### `@LiveSetting`
A unified `CmuxSettingsUI` property wrapper that holds the value in its **own `@State`** and forwards the store's `values(for:)` `AsyncStream` into it via a small `SettingReadDriver`. Re-rendering rides on `@State` invalidation, which is host-reliable inside `NSHostingView`.

- **One mechanism, any store.** The driver only needs an `AsyncStream<Value>`, so the same wrapper covers UserDefaults- and JSON-backed keys (`JSONConfigStore` exposes the same `values(for:)`, backed by a file watcher). Secret is the same pattern (follow-up).
- **Catalog stays the single source of truth.** You pass a catalog key path; writes go back through the same store.
- Replaces the interim `@AppStorage`-backed `@DefaultsSetting` from earlier on this branch.

### Also required
Re-adds the `SettingsRuntime` injection into the AppKit-hosted `ContentView` root (threaded via `AppDelegate.configure(...)` from `cmuxApp`). Any store-stream-based read in the main window needs it. The env key is optional, so a nil runtime degrades to the seeded catalog default.

### Verification
- Proven re-render: the sidebar footer body logged `false` (seeded default) then flipped to `true` when the driver pushed the stored value into `@State` — an async store update re-rendering the `NSHostingView`.
- Live Settings toggle confirmed: Settings → Beta Features → Extensions shows/hides the puzzle button and extension sidebar with no relaunch.
- Empirically verified on the **Defaults** (Bool) path. JSON path is structurally identical (compiles, same `values(for:)` stream) but not yet toggled end-to-end; secret keys not yet wired into `@LiveSetting`.

## Summary by CodeRabbit

* **Improvements**
  * Settings now update live across the UI; the Extensions toggle and other feature flags are reactive and immediate.
  * App initialization now provides the settings runtime to the main window so live setting access is consistent.

* **New Features**
  * Added support for secret-file-backed (secure string) settings that flow into the UI like other settings.


azooz2003-bit:
Triage of review-bot findings (verified against current HEAD):

**Addressed**
- **DocC docs** (CodeRabbit): added full DocC to `@LiveSetting`'s `wrappedValue`, `projectedValue`, `update()`, and `- Parameter` callouts on the inits.
- **Optimistic-write rollback** (CodeRabbit, `LiveSetting`): removed the optimistic `value = newValue`; the observation stream is now the sole writer (mirrors `DefaultsValueModel`), so a failed JSON/secret write leaves the displayed value at the last committed one — no stranded unsaved value.
- **Build under strict concurrency**: constrained `SettingReadDriver`'s `Value: Sendable` (the original CI break).

**No longer applicable** (these reviewed commit `3f62a344`/`f27e360b`, since superseded)
- **Suite-isolated `UserDefaults(suiteName:)` per init / `@AppStorage` KVO mismatch / silent `.standard` fallback** (greptile P1, CodeRabbit, cubic): these targeted the interim `@DefaultsSetting.swift`, which was **replaced** by the closure-based `@LiveSetting`. It no longer creates per-init `UserDefaults` instances or uses `@AppStorage`; reads go through the app's `SettingsRuntime` stores via `values(for:)`, the same path `@Setting`/`DefaultsValueModel` use.
- **Two parallel observation paths** (greptile P2): same — no `@AppStorage` KVO path anymore; the single `values(for:)` stream (with `bufferingNewest(1)`) is the only observation source.
- **Force-cast at `LiveSetting.swift:101`** (CodeRabbit Minor): removed — the secret store's `AsyncStream<String>` is now consumed directly as `AsyncStream<Value>` (the `where Value == String` init captures the stream), so there's no cast or wrapper stream.

**Verified-safe / by-design**
- **`deinit` accesses `@MainActor` `task`** (greptile P1): builds clean under the package's strict concurrency (CI green); `task` is a `Sendable` `Task`, and `Task.cancel()` is safe from any context.
- **`isEnabled` reads `UserDefaults.standard`** (CodeRabbit): this is the synchronous AppKit/static path (toggle menu, command-palette, browser opener); the settings store is `actor`-isolated/async-only with no synchronous read. It resolves the catalog key (`betaFeatures.extensions`, default suite) against `UserDefaults`, which is the same suite+key the store persists to. SwiftUI reads use `@LiveSetting`.

PR #5115, opened 2026-06-01
Fix #3799: paint OSC 11 overrides on the surface host layer

### LINKED ISSUE #3799: Per-pane OSC 11 background tints lost on inactive panes since #3179 (window-root backdrop only paints active surface)
### cmux version

0.64.3 (also reproduces on nightly 0.64.3-nightly.2559808846301)

### macOS version

15.x (Darwin 25.3.0)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Direct download (DMG)

### Bug description

OSC 11 (`\e]11;<color>\a`) sent from a TUI or wrapper script to set the terminal background color is dropped for any pane that isn't currently focused. The active pane visibly tints; inactive split panes render with the default terminal background as if no OSC 11 was ever sent. This  worked through 0.63.2 and broke in late-April nightlies.

Regression introduced by **commit `259de58` / PR #3179** ("Fix transparent titlebar backdrop matching"), which added an unconditional:

```swift
// Sources/Workspace.swift
nonisolated static func usesWindowRootTerminalBackdrop() -> Bool {
  true
}
```
That makes `applySurfaceBackground()` in Sources/GhosttyTerminalView.swift route every per-surface fill through
the single window-root backdrop:
```swift
let usesHostLayerFill = renderingMode.usesWindowHostBackdrop &&
  !sharesWindowBackdrop &&
  !usesBonsplitPaneBackdrop
let color = usesHostLayerFill ? effectiveBackgroundColor() : .clear
```
`sharesWindowBackdrop` is now always true, so `usesHostLayerFill` is always false, so the per-surface host layer is always painted `.clear`. The window backdrop is fed by `applyWindowBackgroundIfActive()`, which short-circuits unless the surface is the active one — so inactive panes have nowhere to render their OSC 11 override.

### Suggested fix (one line)

Force the host layer to fill whenever a surface has an explicit OSC override, while leaving default backgrounds owned by the shared window backdrop:
```diff
       let usesHostLayerFill = renderingMode.usesWindowHostBackdrop &&
-            !sharesWindowBackdrop &&
-            !usesBonsplitPaneBackdrop
+            (backgroundColor != nil || (!sharesWindowBackdrop && !usesBonsplitPaneBackdrop))
```

Verified locally — split panes with different OSC 11 tints both render their colors regardless of focus. Default-background panes are unaffected (still owned by the window backdrop, so #3179's titlebar-matching behavior is preserved).

Caveat I haven't tested: with background-opacity < 1, the override path may double-composite (host layer + window backdrop both painting the translucent color). My config is opaque. A maintainer with context on #3179 may prefer a different gate that also handles translucent backgrounds cleanly — flagging here rather than opening a PR for that reason.

### Shell and environment

- TUI: any (reproduced with bare printf and with Claude Code / Codex CLI wrapper scripts)
- Ghostty config: defaults (no background-opacity override)
- TERM=xterm-256color, COLORTERM=truecolor

### Related

- #3179 — backdrop refactor that introduced the regression
- #3655 — adjacent symptom (TUI-painted cell backgrounds don't blend); same code area, different mechanism

### Expected behavior

Each pane keeps its own OSC 11 background tint regardless of which pane is focused. Through 0.63.2 this worked: a TUI or wrapper script could emit `\e]11;<color>\a` once and the pane would stay tinted across focus changes.

### Steps to reproduce

1. Create a workspace with two split panes.
2. In pane A, run:
```bash
 printf '\e]11;#D2EEF9\a'   # soft blue
```
3. Switch focus to pane B and run:
```bash
 printf '\e]11;#E2D2F0\a'   # pale lavender
```
4. Alternate focus between panes.

### Shell and environment

zsh 5.9 with powerlevel10k prompt; zsh-autosuggestions and zsh-syntax-highlighting loaded. No tmux. Reproduces with bare zsh too (the OSC 11 path is independent of prompt).

```text

```

Expected behavior (screenshot from a local build with the suggested one-line fix  applied) — four panes, four distinct OSC 11 tints, all visible regardless of focus

Image

### Additional context

_No response_

adifsgaid:
#3886 fixed this but got reverted in #5106 about an hour after it landed, so #3799 is back on main. Still repros on v0.64.11 and nightly 264455e with an opaque config: a single full-window pane and split panes both ignore `\e]11;...\a`.

This drops the Core Image cutout that #3886 used and takes the gate from the original report instead. When a surface has an explicit OSC 11 override (`backgroundColor != nil`), it paints the surface host layer itself instead of going through the shared window-root backdrop. That backdrop only paints the active surface, which is why the tint disappeared on every other pane.

Surfaces with no override are untouched (still owned by the shared backdrop), so the titlebar matching from #3179 stays as-is.

Only opaque overrides take ownership: a translucent override would composite over the shared backdrop on the active pane and read too dark, so those keep using the shared backdrop and behave exactly as before. The gate is the one @ct-mahendra verified locally in the report.

## Summary by CodeRabbit

* **Bug Fixes**
  * Opaque per-pane custom background colors are now correctly rendered and preserved, ensuring panes with explicit opaque background overrides display as expected and no longer blend unexpectedly with the shared backdrop.
  * Panes without opaque overrides retain the previous backdrop behavior.

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -7697,9 +7697,13 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
         // The window root backdrop is the single surface fill for solid-color
         // terminal backgrounds. Painting the terminal host layer too would
         // double-composite it against the surrounding chrome.
+        //
+        // An explicit OSC 11 override is pane-local, so it owns the host layer
+        // directly. The shared backdrop only paints the active surface, so routing
+        // overrides through it drops the tint on every other pane (#3799). Panes
+        // without an override stay on the shared backdrop, preserving #3179.
         let usesHostLayerFill = renderingMode.usesWindowHostBackdrop &&
-            !sharesWindowBackdrop &&
-            !usesBonsplitPaneBackdrop
+            (backgroundColor != nil || (!sharesWindowBackdrop && !usesBonsplitPaneBackdrop))
         let color = usesHostLayerFill
             ? effectiveBackgroundColor()
             : .clear
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -7702,8 +7702,14 @@ class GhosttyNSView: NSView, NSUserInterfaceValidations {
         // directly. The shared backdrop only paints the active surface, so routing
         // overrides through it drops the tint on every other pane (#3799). Panes
         // without an override stay on the shared backdrop, preserving #3179.
+        //
+        // Only opaque overrides take ownership: a translucent fill would composite
+        // over the shared backdrop on the active pane and read too dark, so those
+        // keep using the shared backdrop.
+        let opaqueOverride = backgroundColor != nil &&
+            GhosttyApp.shared.defaultBackgroundOpacity >= 1
         let usesHostLayerFill = renderingMode.usesWindowHostBackdrop &&
-            (backgroundColor != nil || (!sharesWindowBackdrop && !usesBonsplitPaneBackdrop))
+            (opaqueOverride || (!sharesWindowBackdrop && !usesBonsplitPaneBackdrop))
         let color = usesHostLayerFill
             ? effectiveBackgroundColor()
             : .clear


adifsgaid:
Pushed a follow-up for the translucent case Greptile flagged: opaque overrides take the host layer, translucent ones stay on the shared backdrop, so there's no double-composite on the active pane anymore.

On the Swift Logging check: the `logBackground()` call it points at is pre-existing in `applySurfaceBackground` (this PR only changes the `usesHostLayerFill` condition a few lines above, it doesn't add or touch logging), so I left it alone to keep the diff scoped.

adifsgaid:
@austinywang any chance you could approve CI on this when you get a sec? It's the minimal re-land of #3886 after the #5106 revert, just the opaque-override gate without the Core Image cutout, so #3799 is still open on main. CodeRabbit's approved it and it's only the required workflows that are waiting. Happy to rework it if you'd rather re-land #3886 directly instead.

nevertoday:
Hi! Really looking forward to this fix landing. 🙏

I'm using per-pane OSC 11 background tints to visually distinguish different task contexts in split layouts (e.g., production = red, dev = green, docs = purple). This broke starting from 0.64.8 and the cell-wash workaround doesn't survive the shared window backdrop repaint either.

Currently on 0.64.12 — the tint flashes for a fraction of a second then gets overridden back to the default background on inactive panes, exactly as described in #3799.

Is there anything blocking this PR from being merged? Happy to help test a pre-release build if that would be useful.

nevertoday:
Follow-up with a fresh data point from the latest release.

I retested this on the current latest release:

- `cmux 0.64.14 (94) [1c8f9e261]`
- `background-opacity = 1.0`
- `background-blur = false`
- cmux Ghostty config uses an explicit light theme and can be hot-reloaded with `cmux reload-config`

The issue is still not fixed in the released app because this PR is still not in `v0.64.14`.

What I see now:

1. Native OSC 11 pane background still is not usable for per-pane tints in cmux. The pane may briefly flash the intended color, then the shared/default backdrop wins and the pane no longer keeps its own background.
2. A shell-side fallback using explicit cell backgrounds does partially work, e.g. `ESC[48;2;...m` plus clear/repaint. It colors the terminal cell grid.
3. That fallback cannot color the actual cmux/Ghostty surface host layer: the top/left padding, unpainted surface area, and panel/chrome background remain the default theme color. In my screenshot the terminal cell area is tinted purple, but the surrounding cmux surface/padding is still the default yellow/light background.
4. I also tried writing `background = <pane color>` into the cmux Ghostty theme and calling `cmux reload-config`. That can only change the global terminal background. It cannot restore independent per-pane/split host backgrounds.

So the user-visible failure mode in 0.64.14 is now very clear: workarounds can paint terminal cells, but they cannot paint the whole pane/surface. The missing piece is still exactly what this PR appears to address: an explicit opaque OSC 11 override needs to be owned by the surface host layer instead of being routed through the shared window-root backdrop.

Expected:

- Each split pane that emits an explicit opaque OSC 11 background keeps that color across the whole terminal surface, including padding/unpainted host-layer space.
- Other panes keep their own colors independently.

Actual in the released app:

- OSC 11 flashes then gets overwritten by the shared/default backdrop.
- Cell-background fallback colors only the character grid, not the whole cmux pane/surface.

Could this PR be merged and released, or could you publish a signed build/nightly containing it? I can test immediately on the same setup.

PR #5161, opened 2026-06-02
Render browser focus flash above the portal-hosted webview

lawrencecchen:
## Summary
- The browser focus/attention flash (the blue ring that pulses on navigation and notification arrival) was a SwiftUI overlay on the browser panel. SwiftUI overlays composite **below** the portal-hosted WKWebView, because native web surfaces draw above sibling SwiftUI layers. So once a page was loaded, the flash was hidden behind the web content (the user only saw a partial ring over the SwiftUI address bar, with the sides and bottom swallowed by the webview).
- Move the flash into the AppKit portal layer, the same place the find bar and omnibar suggestions already live so they render above web content. `WindowBrowserSlotView` now owns a passthrough flash overlay view backed by a `CAShapeLayer` ring, animated with the shared `FocusFlashPattern` via `CAKeyframeAnimation`. This mirrors the terminal's flash, which already draws above its Metal surface the same way. `interactionLayerPriority` places the ring above the webview but below the interactive find/omnibar chrome.
- `BrowserPanelView` routes the flash by surface state: panes with a live webview fire `BrowserWindowPortalRegistry.triggerFocusFlash`; the empty new-tab state (no webview mounted) keeps the existing SwiftUI ring, mirroring how the find overlay is already split. The flash is a transient one-shot, so it is an imperative trigger rather than persisted portal `Entry` state.
- Inline DevTools hosting moves the live webview out of the window portal into a local `WindowBrowserSlotView` (the registry entry is discarded), so `triggerFocusFlash(for:)` falls back to walking up from the webview to its hosting slot. This keeps the flash working for window-portal hosting and inline DevTools that is undocked or bottom-docked.

## Behavior change to note
- With a page loaded, the ring now hugs the **web content area**, not the whole panel. The ~28px omnibar/address-bar header is no longer inside the ring, because the portal slot only covers the webview. This is the correct fix for the literal complaint ("doesn't show above the webview") and is consistent with how the find bar / omnibar suggestions are scoped to the slot. Empty new tabs (no webview) keep ringing the whole pane via the SwiftUI path.
- Inline DevTools docked **left/right** (side-dock) still does not flash: the page webview subtree is moved into a sibling `HostedInspectorSideDockContainerView` that has no slot ancestor. This is unchanged from before this PR (the old SwiftUI ring was already hidden behind the webview for every webview-backed pane), so it is a pre-existing non-flash, not a regression. Left out of scope here because covering it means a passthrough overlay inside the fragile DevTools reparenting/z-order path for a niche dev sub-config. Tracked as a follow-up.

## Testing
- Built and ran a tagged Debug build (`reload.sh --tag flash-z`), created a browser pane on `https://example.com`, and triggered the flash via the `surface.trigger_flash` socket command.
- Z-order proof (temporary debug log, since removed): for the webview pane the flash routed `route=portal`, the registry resolved the webView (`HIT`), and in the slot the flash overlay sat at subview index 1 with the hosted webView at index 0 (`flashIdx=1 webIdx=0`) — i.e. the ring is now ordered above the webview.
- Visual proof: window capture of the browser pane shows no ring over the web content at rest, and a full blue rounded ring drawn over the `example.com` content while flashing.
- Empty new-tab path: confirmed the surface reports `in_window=false` (no webview → SwiftUI fallback) and that flashing it does not crash; that fallback code is unchanged apart from the `!shouldRenderWebView` gate.

No automated regression test added: this is a native AppKit z-order fix that only manifests with a live WKWebView composited on screen. A headless test would only re-assert the `interactionLayerPriority` constants and would not exercise the actual native-surface compositing, so per the repo test-quality policy it would be a low-value source-shape test. Verified via the tagged build instead (log + screenshot above).

## Issues
- Task: the blue flash over the browser doesn't show above the webview

---

> [!NOTE]
> **Low Risk**
> UI z-order and animation routing only; no auth, data, or security-sensitive paths. Residual gap for side-docked inline DevTools is pre-existing and documented.
> 
> **Overview**
> The browser **focus/attention flash** no longer relies on a SwiftUI ring when a page is loaded. **WKWebView** draws above sibling SwiftUI, so the pulse was largely hidden; the change adds a **passthrough AppKit overlay** on `WindowBrowserSlotView` (same idea as the terminal flash) with a **`CAShapeLayer`** ring animated via **`FocusFlashPattern`**, stacked **above the web view** but **below** find/omnibar chrome.
> 
> `BrowserPanelView` now triggers **`BrowserWindowPortalRegistry.triggerFocusFlash`** when a web view is mounted, and keeps the existing SwiftUI animation only for the **empty new-tab** state (`!shouldRenderWebView`). The registry walks up to a hosting **`WindowBrowserSlotView`** when portal entries are missing (e.g. inline DevTools).
> 
> With a loaded page, the ring follows the **portal slot (web content area)** rather than the full panel including the address bar.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc2372c400f73d3d44c07dcd17e7a1b36402ddad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added visual focus-flash animation overlay that displays when browser windows receive focus
  * Focus flash animation intelligently adapts to work appropriately with both live web content and placeholder views


lawrencecchen:
Addressed review feedback in bc2372c40:

- **CodeRabbit / cubic #1 (P1, detached-entry no-op):** Valid. `WindowBrowserPortal.triggerFocusFlash(forWebViewId:)` now returns whether it actually fired, and the static `triggerFocusFlash(for:)` only short-circuits on success; otherwise it falls through to the slot-ancestor walk. So a stale `webViewToWindowId` mapping with a detached/pruned entry no longer silently drops the flash.
- **cubic #2 (P2, inline DevTools side-dock):** Declined here, out of scope. When inline DevTools is docked left/right the page WKWebView is moved into a sibling `HostedInspectorSideDockContainerView` with no slot ancestor, so it does not flash. This is not a regression: before this PR the SwiftUI ring composited behind the native webview for every webview-backed pane, so side-docked inline DevTools never showed a visible flash either. Covering it requires a passthrough overlay inside the fragile DevTools reparenting/z-order path for a niche dev sub-config; tracked separately in https://github.com/manaflow-ai/cmux/issues/5170.

austinywang:
Merged the latest `origin/main` (`1f4177ea2b`) and pushed merge commit `ba00c3d5ec` to this PR branch.

Resolved the `BrowserWindowPortal.swift` and `BrowserPanelView.swift` conflicts while retaining current-main browser chrome/attention-ring behavior and the AppKit focus-flash routing. Non-launch validation passed: Swift parser checks, package-resolved policy, and test-wiring lint.

The side-docked inline DevTools gap remains the separately tracked follow-up in #5170.

PR #4786, opened 2026-05-26
Add headless cmux attach foundation

lawrencecchen:
## Summary

Adds the first headless cmux slice:

- `cmuxd-remote serve --unix` runs a named long-lived headless instance with a shared PTY hub.
- `cmuxd-remote headless list` and `headless connect` discover and bridge to named Unix-socket instances.
- `cmux headless attach <destination> --id <instance>` creates a desktop remote workspace attached to an existing headless instance over SSH.
- Remote app configuration now carries `headless_instance_id` and routes daemon RPC through `headless connect --id` instead of a transient `serve --stdio` process.

Prior art checked:

- https://github.com/manaflow-ai/cmux/pull/3004, remote attach foundation.
- https://github.com/manaflow-ai/cmux/pull/3685, Rust cmux multiplexer/server/client foundation.
- https://github.com/manaflow-ai/cmux/pull/4240, tmux session attach surface.
- https://github.com/manaflow-ai/cmux/pull/2620, amux remote daemon/tmux compatibility attempt.
- https://github.com/manaflow-ai/cmux/pull/4748, tmux control panel integration.

## Verification

- `swiftc -parse CLI/cmux.swift Sources/WorkspaceRemoteConfiguration.swift Sources/WorkspaceRemoteSSHBatchCommandBuilder.swift Sources/TerminalController.swift Sources/Workspace.swift`
- `go test -ldflags=-linkmode=external ./cmd/cmuxd-remote`
- `git diff --check`
- Live smoke: built `cmuxd-remote`, ran `serve --unix --id smoke`, verified `headless list --json`, and verified `headless connect --id smoke` returned a valid `hello` RPC response.
- `./scripts/reload.sh --tag headless`

## Dogfood

1. On a remote Mac/Linux host, start a headless instance:
   ```bash
   cmuxd-remote serve --unix --id work
   ```
2. In the tagged desktop app, run:
   ```bash
   cmux headless attach user@host --id work
   ```
3. The new workspace should connect to the existing remote headless instance, open a remote shell through the persistent PTY bridge, and keep the named headless daemon running after local terminal surfaces close.

---
[...]

PR #5162, opened 2026-06-02
headless: place instance socket under per-user runtime dir, 0600

shuhei0866:
Follow-up to the Linux dogfooding on #4786 — turning the socket-path note into a fix.

`runUnixHeadlessServer` placed the instance socket at `/tmp/cmux-headless-<uid>-<id>.sock` unconditionally, with whatever the process umask gave it (I saw `0755`). On a shared host that's a world-traversable directory.

This change:
- prefers a private per-user runtime dir for the socket — `$XDG_RUNTIME_DIR`, falling back to `/run/user/<uid>` when it exists (the systemd-managed `0700` dir) — and keeps `/tmp` as the last resort, so macOS behavior is unchanged.
- `chmod`s the listener to `0600` so the socket is owner-only regardless of umask.

Verified on Ubuntu 24.04 / Go 1.26:
- `go test ./...` green (incl. the existing headless/PTY tests); added a unit test for the `$XDG_RUNTIME_DIR` precedence.
- `serve --unix --id rtcheck` now creates `/run/user/1000/cmux-headless-1000-rtcheck.sock` at `srw-------` (0600), and no longer drops anything in `/tmp`.

Attach still works since the attaching process connects as the same user.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/5162"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782961784&installation_id=133855650&pr_number=5162&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F5162&signature=55c6891ad53b88e5ca7ebc6c56629ec74964efb24d9c775df1f1806debf657bc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Secure the headless Unix socket by placing it in a per-user runtime directory and locking it to `0600`. Falls back to `/tmp` if no private dir is available; macOS behavior is unchanged.

- **Bug Fixes**
  - Prefer `$XDG_RUNTIME_DIR`, then `/run/user/<uid>`, only if the dir is owned by the current user and `0700`; otherwise ignore to avoid attacker-controlled paths.
  - `chmod` the socket to `0600` immediately after bind and remove it on failure; the `0700` dir prevents a meaningful TOCTOU window.
  - Added tests to prefer a private `$XDG_RUNTIME_DIR` and reject group/other-accessible dirs.

- **Migration**
  - On Linux, the default socket path moved from `/tmp/cmux-headless-<uid>-<id>.sock` to `$XDG_RUNTIME_DIR/cmux-headless-<uid>-<id>.sock` (or `/run/user/<uid>/...`). Update any hardcoded paths.

<sup>Written for commit a79a3eb48e812ac9a93497c618e35b142495c92b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/5162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/daemon/remote/cmd/cmuxd-remote/headless.go
+++ b/daemon/remote/cmd/cmuxd-remote/headless.go
@@ -74,6 +74,15 @@ func runUnixHeadlessServer(ctx context.Context, cfg unixHeadlessServerConfig, st
 	if err != nil {
 		return err
 	}
+	// userRuntimeDir keeps the socket in a private 0700 directory, so tightening
+	// it to 0600 here closes the remaining exposure with no meaningful TOCTOU
+	// window (a process-wide umask would not be goroutine-safe). Remove the socket
+	// on failure so a partially-initialized path is not left behind.
+	if err := os.Chmod(absSocketPath, 0o600); err != nil {
+		listener.Close()
+		_ = os.Remove(absSocketPath)
+		return fmt.Errorf("chmod headless socket %s: %w", absSocketPath, err)
+	}
 	defer listener.Close()
 	defer os.Remove(absSocketPath)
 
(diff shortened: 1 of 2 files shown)


shuhei0866:
Pushed a revision addressing the review:

- Dropped the `chmod` in favor of a `0177` umask around `net.Listen`, so the socket is `0600` at bind time — that closes both the leftover-on-`chmod`-error path (there's no `chmod` anymore) and the TOCTOU window where it was briefly group/world-accessible before the `chmod` ran.
- `userRuntimeDir` now uses `isPrivateDir` (owned by the current user, mode `0700`, no group/other access), so an untrusted `$XDG_RUNTIME_DIR` like `/tmp` can't redirect the socket into a shared location; a non-private dir falls through to `/tmp`.
- Added unit tests for the private-dir-is-used and world-accessible-dir-is-rejected cases; re-verified on Ubuntu 24.04 / Go 1.26 (`go test ./...` green, `serve --unix` lands the socket in `/run/user/<uid>` at `0600`).

shuhei0866:
Good catch on the umask — `syscall.Umask` is process-wide, so I reverted to an explicit `chmod` after `Listen`. Because `userRuntimeDir` keeps the socket inside a `0700` directory, the chmod has no exploitable TOCTOU window, and the chmod-error path now removes the socket so nothing is left behind. `go test ./...` and the Linux check (`serve --unix` → `/run/user/<uid>/…sock` at `0600`) are green.

Issue #2824, opened 2026-04-11
Window moves to the right of screen when disconnected from external monitor


aparente:
### cmux version

0.61.0

### macOS version

Sequoia 15.7.4

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Can you reproduce this on cmux NIGHTLY?

I could not test NIGHTLY

### Bug description

When disconnected from an external monitor, cmux goes to the far right of the screen (effectively off screen). The only way to fix it is to either create a new workspace and drag cmux to it or expand the window by dragging to the left and then resize it. Very annoying. 

### Expected behavior

Re-center on screen like other apps do - only cmux is affected by this

### Steps to reproduce

Disconnect monitor

### Shell and environment

Any

### Relevant logs or crash reports

```text

```

ruan11223344:
Confirming this is still happening on **0.64.17** (Apple Silicon, built-in Liquid Retina XDR display).

Just hit it by physically unplugging the cable from an external monitor. The window kept its old coordinates from the external display instead of being repositioned onto the remaining built-in screen, leaving it almost entirely off-screen.

Measured geometry via Accessibility API right after disconnect (built-in display logical bounds are `0,0 → 1512,982`):

```
window: x=956  y=-324   w=1512  h=968
```

So the window was both pushed off the **right** edge (x=956 with width 1512 → extends to 2468, only ~556px visible) and above the **top** edge (y=-324, behind the menu bar). A second cmux window on the same display was unaffected, so it seems tied to which display a given window was last placed on.

Workaround that recovered it without touching the mouse — reposition via AppleScript:

```bash
osascript -e 'tell application "System Events" to tell process "cmux" to set position of window 1 to {0, 25}'
```

**Expected:** on display disconnect, any window whose frame is now wholly/mostly outside the remaining display(s) should be clamped back into visible bounds (re-centered or pinned to `0,25` under the menu bar), the way native AppKit windows are. macOS does fire `NSApplication.didChangeScreenParametersNotification` on display add/remove — handling that to validate/clamp each window's frame against `NSScreen.visibleFrame` would fix this.

SmetDenis:
The same issue. Please help. This is really frustrating... :(

gevky:
Yeah this is really the only bug that made me switch back to ghostty. If i close my macbook or disconnect the widnow goes into weird position that you cant use it anymore.

rifelpet:
[...]

Issue #2666, opened 2026-04-07
Window collapses to narrow vertical sliver on left edge after disconnecting from external displays (clamshell mode)


austinywang:
## Summary

When transitioning from a clamshell-mode multi-monitor setup back to the laptop's built-in display (e.g. unplugging the dock at the end of the day), the cmux window ends up shrunk to a tiny vertical sliver pinned to the left edge of the built-in display. The window is so narrow only the sidebar toolbar buttons are visible — the rest of the UI (workspace list, terminal panes, browser panes) is clipped off-screen to the right.

## Setup

- MacBook with the **lid closed** (clamshell mode)
- **Two external Dell monitors** connected via dock
- cmux running across the external displays (the laptop's internal display is disabled in clamshell mode)

## Steps to reproduce

1. Run cmux in clamshell mode with two external Dell monitors as the active displays
2. Have cmux open with its normal full-sized window on one of the external monitors
3. Disconnect the dock / external displays (or open the laptop lid, removing clamshell mode and forcing the system to fall back to the built-in display)
4. Open the laptop and look at cmux on the built-in display

## Expected

cmux re-fits its window to a sane size on the built-in display — similar to what other macOS apps do when their host display goes away.

## Actual

cmux's window is collapsed to a vertical sliver against the left edge of the built-in display. Only the leftmost few pixels of the sidebar (the toolbar with the sidebar/single-window toggle buttons) are visible. The rest of the window is either clipped off-screen or has had its width forcibly clamped to ~50px. The window is still interactive but unusable until manually resized.

See screenshot.

## Notes / probable cause

- This looks like the saved-geometry restore path computing a window frame whose `width` is near-zero (or whose `origin.x` is far to the left of the new screen and the resulting visible width gets clamped without restoring the original width).
- Possibly related to how cmux persists `lastWindowGeometry.v1` (mentioned in https://github.com/manaflow-ai/cmux/issues/2135). When the external screens go away the saved frame for them is still on disk; on the next display change cmux may be re-applying that frame against a screen that no longer exists, getting clamped to a sliver instead of falling back to a default frame.
- Distinct from:
  - https://github.com/manaflow-ai/cmux/issues/2135 (position not restored on **reconnect**) — this is about size collapsing on **disconnect**.
  - https://github.com/manaflow-ai/cmux/issues/926 (panel layout overlap on resize) — this is about the *window itself* collapsing, not the inner panel layout.
[...]

PR #5193, opened 2026-06-02
Fix main window not refitting after a display reconfiguration (Screen Sharing / resolution change)

domsilvestre:
## Problem

On a Mac accessed via macOS Screen Sharing (for example a headless Mac mini reached from a laptop), connecting or resizing the viewer resizes the host's virtual display. macOS refits ordinary windows to the new geometry, but the cmux window stays where it was and ends up pinned against a screen edge or overflowing the display. The same happens on a plain resolution change or when an external display is connected or disconnected.

## Root cause

cmux main windows set `isMovable = false` for the custom titlebar drag handling (`WindowDragHandleView`, `configureCmuxMainWindowDragBehavior`). A non-movable `NSWindow` is excluded from AppKit's automatic on-screen constraining on display reconfiguration, so nothing repositions or resizes the window when `NSApplication.didChangeScreenParametersNotification` fires. The geometry-resolution helpers already exist (`resolvedWindowFrame`, `clampFrame`, `display(for:in:)`) but only run at launch and session restore, never reactively while the app is open.

## Fix

- Observe `NSApplication.didChangeScreenParametersNotification` in `AppDelegate`.
- For each registered main window, refit it onto the visible frame of the display it now belongs to (matched by display ID, then overlap or proximity, then the fallback display), reusing the existing helpers (`currentDisplayGeometries`, `display(for:in:)`, `clampFrame`).
- The pass is a no-op when a window already fits, so windows on displays the reconfiguration did not affect are left untouched.
- Notifications are coalesced (one reconcile pass per reconfiguration) with a short settle delay so the `NSScreen` geometry is current when read back.

This is intentionally scoped to the missing reactive refit: it clamps the window to fit, it does not restore previously saved per-display geometry (that broader behavior is what #2667 is working toward).

## Tests

Adds a Swift Testing suite for the pure decision (`AppDelegate.reconciledFrameAfterScreenChange`): an oversized window on a shrunk display is refit, an already-fitting window is untouched, an off-screen window is pulled back on-screen, and an empty display list returns nil.

## Verification

- New unit tests pass; the existing window-geometry tests still pass.
- Verified end to end on a Mac mini reached from a MacBook via macOS Screen Sharing: as the viewer (virtual display) resized between 3440x1440, 1920x1080 and 1512x883, the window refit to stay fully on-screen each time instead of staying pinned to an edge.

Related: #2666, #2824, #2667 (same area: window geometry on display configuration change).

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/5193"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782989728&installation_id=133855650&pr_number=5193&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F5193&signature=3663ee7f7286c539160eb0349dbb8a4cf88ffb40bae0e34eb845aaa6f0cd0a39"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fixes main window not refitting after display changes (Screen Sharing, resolution changes, display connect/disconnect). Windows now clamp to the visible screen so they stay fully on-screen.

- **Bug Fixes**
  - Observe `NSApplication.didChangeScreenParametersNotification` and run one coalesced reconcile pass after changes.
  - Refit each main window to its display’s visible frame using `currentDisplayGeometries`, `display(for:in:)`, and `clampFrame` (no-op if it already fits).
  - Handle off-screen and disappeared-display cases; enforce minimum window size. Added `AppDelegate.reconciledFrameAfterScreenChange` with unit tests for oversized, already-fitting, off-screen, and no-display scenarios.
  - Snapshot `mainWindowContexts` before iterating to avoid mutate-during-enumeration crashes, and skip redundant `setFrame` calls when the frame is unchanged.

<sup>Written for commit e7a8920e709a1a74d257eadb3dc43d3a7025c3df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/5193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved window handling when display geometry changes: windows are automatically repositioned and resized into visible areas after monitor connects, disconnects, or resolution changes.

* **Tests**
  * Added tests covering window frame reconciliation across display changes to ensure stable, idempotent behavior.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -1255,6 +1255,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
             name: .feedRequestSendText,
             object: nil
         )
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(handleScreenParametersDidChange(_:)),
+            name: NSApplication.didChangeScreenParametersNotification,
+            object: nil
+        )
 
 #if DEBUG
         // UI tests run on a shared VM user profile, so persisted shortcuts can drift and make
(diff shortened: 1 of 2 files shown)


domsilvestre:
Thanks for the reviews. Addressed in e7a8920:

- Snapshot `Array(mainWindowContexts.values)` before iterating: `resolvedWindow(for:)` can reindex the dictionary, so this avoids the mutate-during-enumeration crash flagged by CodeRabbit and cubic. Matches the snapshot pattern already used elsewhere in the file.
- Documented the 0.1s `asyncAfter` as a debounce, not a correctness dependency: the notification is posted after `NSScreen` updates, the reconcile re-runs on every notification and is idempotent, so the window ends up correct regardless of the exact delay.
- Added docstrings to the reconcile pass and the frame-equality helper.

Verified locally: app builds and the 4 Swift Testing cases for `reconciledFrameAfterScreenChange` pass.

PR #5239, opened 2026-06-02
Inject FocusLogStore instead of a shared singleton

azooz2003-bit:
Follow-up to the CmuxUpdater extraction (https://github.com/manaflow-ai/cmux/pull/5132), which de-singletoned `UpdateLogStore`. `FocusLogStore` (the DEBUG-only focus-event diagnostic logger that lives in the same file) still self-vended a `static let shared`.

This applies the same fix: `AppDelegate` owns the single `FocusLogStore` instance next to `updateLog`, so its lifecycle has one composition root. The in-`AppDelegate` call sites use the instance directly; the `GhosttyTerminalView` and `Workspace` sites reach it via `AppDelegate.shared?.focusLog`, mirroring the existing `AppDelegate.shared?.updateLog` precedent. The type is now a plain injectable class with a documented `@unchecked Sendable` (its mutable state stays confined to the serial queue, unchanged).

The second commit compile-time guards the `FOCUSDBG` `NSLog` in `GhosttyNSView.focusLog` behind `#if DEBUG` (it was runtime-gated only), per the repo debug-log requirement.

No production behavior change. `FocusLogStore.append` is already `#if DEBUG`; this only moves ownership.

## Testing

- `./scripts/reload.sh --tag focuslog-di` builds green.
- Verified no test target references the removed `FocusLogStore.shared` symbol.
- CI `tests-build-and-lag` and the CoreAnimation main-thread startup regression pass on the build.
- No new behavior to exercise at runtime: focus logging (`~/Library/Logs/cmux-focus.log`, the Copy Focus Logs menu action) works exactly as before, now sourced from the AppDelegate-owned instance.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -866,6 +866,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
     private var browserAddressBarBlurObserver: NSObjectProtocol?
     private var browserWebViewFirstResponderObserver: NSObjectProtocol?
     let updateLog = UpdateLogStore()
+    let focusLog = FocusLogStore()
     private lazy var updateController = UpdateController(log: updateLog)
     private lazy var titlebarAccessoryController = UpdateTitlebarAccessoryController(updateLog: updateLog)
     private let windowDecorationsController = WindowDecorationsController()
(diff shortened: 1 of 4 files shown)


azooz2003-bit:
@codex review

PR #5244, opened 2026-06-02
Extract CmuxFileWatch and consolidate file watchers (modular refactor, wave 2)

azooz2003-bit:
Wave-2 infrastructure leaf of the [modular refactor](https://github.com/manaflow-ai/cmux/blob/main/CLAUDE.md#refactor-architecture-layers-coordinatorservicerepository-dependency-inversion). Pulls the FSEvents-backed path watcher out of `TabManager` into a standalone, **git-agnostic** `CmuxFileWatch` package, redesigned onto Swift 6 concurrency. No behavior change intended.

## Scope: generic infra, not a git slice
The extracted code has **no git knowledge** — it just watches a set of paths and yields coalesced events. The git-specific path resolution (which paths matter, dirty-state computation) deliberately stays in `TabManager`, bound for the Wave-3 `CmuxSidebarGit` domain package that will depend on this leaf. This mirrors the `CmuxProcess` infra-leaf precedent (one outside-world capability per package), and `CmuxSettings`' `JSONConfigFileWatcher` is a candidate second consumer to migrate onto it later.

## Redesign at the seam
- `actor RecursivePathWatcher` exposes coalesced changes as `AsyncStream<Void>` instead of the old `init?(descriptor:onChange:)` completion-handler callback (forbidden in new package code). Takes `paths: [String]` directly and exposes `watchedPaths` for caller dedup; the git-flavored `Descriptor` is gone.
- The leading-edge throttle moves off the shared-queue `DispatchWorkItem` onto a stored, cancel-on-stop `Task` driven by an injected `FileWatchClock` (the bounded/cancellable delay carve-out, justified inline). Behavior preserved: first event in a window arms one delay; events during the window coalesce; it reacts during a storm rather than waiting for changes to stop.
- `FileSystemEventStream` keeps the hardened `FSEventStream` internals: one shared serial queue to bound threads, synchronous queue-confined teardown so the stream can't leak before `deinit`. The FSEvents context is now passed **unretained** (lifetime guaranteed by invalidate-before-deinit), which let me delete the callback box and both retain/release C trampolines, leaving only the one irreducible event callback.
- `TabManager` imports the package, resolves watched paths, stores a per-key consumer `Task` that drains `watcher.events` on the MainActor, and tears it down alongside the watcher. Net **-262 lines** from `TabManager`.

## Tests
The old DEBUG semaphore + `Thread.sleep` storm helper (and its `testGitMetadataWatcherRefreshesDuringSustainedEventStorm` XCTest) is replaced by a deterministic Swift Testing test that proves leading-edge coalescing with a gate clock and **no real waiting** (`burstCoalescesAndThrottleRearms`). The submodule-paths test stays (it exercises the git path-resolution builder, which stayed in `TabManager`).

## Gates (all green from real logs)
- `swift build` + 4 `swift test` (package)
- `cmux` scheme: BUILD SUCCEEDED
- `cmux-unit` scheme: BUILD SUCCEEDED (GlobalISel workaround)
- `check-pbxproj.sh` passes; package linked into both `cmux` and `cmux-unit`

Author: @azooz2003-bit (via Aziz Albahar)

## Summary by CodeRabbit

* **Refactor**
  * Switched several subsystems to a path-driven, per-repository filesystem watcher with improved lifecycle/teardown and more deterministic event handling.

* **New Features**
  * Added a new file-watch library with recursive watching, a throttling clock seam, and a concurrency-based file watcher; integrated into builds.
  * Replaced multiple legacy dispatch-source watchers with the new watcher across stores, panels, and explorers.

* **Documentation**
  * Added package README and updated architecture docs for the new watcher.

* **Tests**
  * Added deterministic watcher tests and removed a timing-based flaky test.

## Review feedback addressed (CodeRabbit / Greptile)

- **Synchronous, failable stream start (was deferred).** The earlier revision started the `FSEventStream` in a `Task` inside `init`, which (a) could miss events in the gap before the task ran (`kFSEventStreamEventIdSinceNow`) and (b) swallowed creation failures (TabManager cached a silent watcher). Reworked so the stream is created synchronously in `init?` and returns `nil` on failure, matching the original watcher. The FSEvents sink forwards into a private raw-event `AsyncStream` (capturing a Sendable continuation, not the actor), which is what lets creation happen in-`init`; a single actor-isolated pump drains that raw stream and applies the leading-edge clock throttle.
- **Synchronous teardown (dropped fire-and-forget `Task { await stop() }`).** `TabManager` now drops the watcher reference, which runs the watcher's `deinit` synchronously and invalidates the `FSEventStream` on its shared queue before returning — no window where the old and new streams co-watch. The consumer task captures the events stream (not the watcher), so removal is the last reference.
- **README test pattern.** Added a "Testing" section showing the gate-clock instantiation per the package-docs guideline.
- Re-verified: `swift test` (4), `cmux` + `cmux-unit` BUILD SUCCEEDED; behavior screenshot-verified end-to-end (sidebar git dirty indicator updates on file change) on the tagged build.

## Update: consolidated all file-reload watchers onto CmuxFileWatch (per request)

Beyond the leaf extraction, this PR now also adds a single-file/dir `FileWatcher`
to `CmuxFileWatch` and migrates every hand-rolled file-reload watcher in the app
onto it, deleting ~350 net lines of bespoke `DispatchSource` code:

- `FileExplorerStore` (`FileExplorerDirectoryWatcher`)
- `MarkdownPanel` (file + nearest-ancestor directory watchers)
- `KeyboardShortcutSettingsFileStore` (`ShortcutSettingsFileWatcher` primary + fallbacks)
- `CmuxConfigStore` global-config + per-hook-file watchers (local-config watcher kept — it does search-directory *path re-resolution*, not reload-on-change)
- `CmuxSettings` `JSONConfigStore` (`JSONConfigFileWatcher` replaced + deleted; `CmuxSettings` now depends on `CmuxFileWatch`)

Not migrated, by design (not file-reload watchers): `TerminalController`'s socket-path liveness monitor (auth-path state machine, slated for `CmuxControlSocket`) and process-exit kqueue sources (`FeedCoordinator`, `CMUXWorkstream`).

`FileWatcher` adds nearest-existing-ancestor recovery (subsumes MarkdownPanel's ancestor walk) and an optional leading-edge throttle on the shared `FileWatchClock`.

### Tests + verification
- `CmuxFileWatch`: 11 tests (write→event, create-after-start, ancestor recovery, directory target, atomic inode replace, throttled yield, stop) — all green.
- `cmux` + `cmux-unit` BUILD SUCCEEDED with the cross-package `CmuxSettings → CmuxFileWatch` dependency.
- Live computer-use verification on the tagged build: sidebar git dirty indicator updates on file change (`RecursivePathWatcher`); MarkdownPanel auto-reloads on external edit (`FileWatcher`).


azooz2003-bit:
Thanks — addressed the two Major findings in `adef166`:

- **Deferred FSEvents start → synchronous failable `init?`.** The stream is now created in `init` (no missed-events gap) and returns `nil` on creation failure (no silently-cached watcher). The sink forwards into a private raw `AsyncStream` capturing a Sendable continuation (not the actor), so creation stays in-`init`; an actor pump drains it and applies the throttle.
- **Fire-and-forget `Task { await stop() }` → synchronous deinit teardown.** `TabManager` drops the watcher reference; its `deinit` invalidates the `FSEventStream` synchronously on the shared queue before returning, so there's no stop-then-start window where both streams co-watch. The consumer task captures the events stream, not the watcher, so removal is the last reference.
- Added the README "Testing" section (gate-clock pattern) per the package-docs guideline.

On the **app-level smoke test for the TabManager bridge** (nitpick): respectfully skipping. The consumer bridge lives inside the `@MainActor TabManager` god object, which can't be instantiated in a unit test without booting the app, so a faithful test would be an XCUITest — heavier than the contract warrants while the god object is still un-decomposed. The contract is covered two other ways: the package test proves the `AsyncStream` coalescing, and I verified the full path (file change → sidebar dirty indicator) end-to-end on the tagged build. Per the repo's test-quality policy (skip low-value tests when no practical behavioral test exists), I'd rather add a real seam when `TabManager` is decomposed (the bridge gets an injectable consumer then) than a synthetic unit test now.

PR #5277, opened 2026-06-03
CmuxGit: extract the git/PR subsystem from TabManager (GitMetadataService + PullRequestProbeService)

azooz2003-bit:
Extracts the entire git/PR domain out of `TabManager` into a new Layer-2 package, `CmuxGit` (the blueprint's `CmuxSidebarGit`, generalized for the domain). **TabManager: 11,506 → 9,557 lines (−1,949).** Both services are stateless `Sendable` structs with explicitly `nonisolated async` reads — off the caller's actor on the global executor (SE-0338) and parallel across calls, with no actor serialization (there is no shared state to protect). An executable test pins that contract; if `NonisolatedNonsendingByDefault` is ever adopted, it fails and points at `@concurrent`.

## Stage 1 — `GitMetadataService` (local git, no `git` subprocess)

- `workspaceMetadata(for:)` → branch, dirty state, change-detection signatures (`GitWorkspaceMetadata`)
- `watchedPaths(for:)` → existing paths a watcher should observe, incl. submodule gitlinks
- `repositorySlugs(forDirectory:)` → GitHub `owner/name` remotes, ordered upstream/origin/rest

Behind them: repository resolution (`.git` file + `commondir`), `HEAD`/ref reads, git `config` `include`/`includeIf` + glob matching, binary `index` v2/v3/v4 parse (prefix compression, assume-unchanged/skip-worktree), stat-based dirty detection, slug parsing. TabManager injects the service at its three entry points; the actor-shaped `Task.detached` offloading the old code hand-rolled is replaced by the service's `nonisolated async` execution.

## Stage 2 — `PullRequestProbeService` (GitHub PR badge pipeline)

- `resolveCandidateSeeds(_:gitMetadata:)` → per-panel branch → GitHub slugs
- `fetchRepoResults(...)` → concurrent REST repo fetches + per-branch `head=` lookups, honoring the caller-owned repo cache; auth via `GH_TOKEN`/`gh auth token` (injected `CommandRunning`)
- `resolveRefreshResults(candidates:repoResults:)` + pure selection/policy (preferred-PR ordering, stale-merged filtering, cache/refresh policy)

New `PullRequestStatus` enum bridges the app's `SidebarPullRequestStatus` by identical raw values, so the apply path is unchanged. **TabManager keeps the orchestration** (probe state machines, per-key dictionaries, poll timers, apply path, repo-cache ownership) and calls the services. The `DispatchSourceTimer` poll loops are deliberately not redesigned here — timing changes need their own dogfood and PR.

## Tests

**48 package tests** (`swift test`): repo resolution (normal + `.git`-file worktree), root-escape termination, branch/detached-HEAD, dirty detection, index v2 + v4 prefix-compression (multi-byte strip-length), content-signature stability, watched paths incl. submodule gitlinks, slug parsing + full `include`/`includeIf` rules, preferred-PR selection, stale-merged map filtering, refresh/cache policy, REST decode (`merged_at` → `MERGED` synthesis), branch-endpoint encoding, result resolution, and the off-main execution-contract pin. Pure-static tests migrated out of `cmuxTests` to the package (their proper home); the remaining `WorkspacePullRequestSidebarTests` integration tests exercise the cutover end-to-end through the real `TabManager`.

Two latent issues found and fixed along the way:
- the old `"not a url"` malformed-URL fixture only fails `URL(string:)` under pre-macOS-14-SDK parsing (the lenient parser accepts it) — replaced with an SDK-agnostic fixture;
- **ci.yml's package-test step was never running `CmuxFileWatch`/`CmuxGit`/`CmuxProcess` suites** (compiled only) — all three added to the list, each verified passing standalone.

Also fixes both carryover doc nits from #5244 (stale `CmuxWorkspaceGit` comment; `CmuxSettings` README dependency claim).

## Verification

`swift build` + 48 `swift test` (package), `cmux` (tagged reload, GlobalISel workaround), and `cmux-unit` all green from real logs at every stage. Dogfood: sidebar branch label + dirty dot, and the PR badge (open/merged states + poll refresh) should behave exactly as before.


azooz2003-bit:
CodeRabbit review triage (12 comments, posted against the stage-1 diff before the struct redesign and stage 2):

- **Accepted (2):** strict `nil` assertion in `returnsNilOutsideAnyRepository`; `gitdir:` test condition now uses the git-directory path (`b2c6764c5b`).
- **Outdated (3):** the DocC-coverage comments — `GitMetadataService`'s init + all three reads, `GitWorkspaceMetadata`, and `ResolvedGitRepository` are fully documented in the current diff.
- **Deferred as pre-existing upstream behavior (5):** config case-sensitivity, `includeIf gitdir:` relative rules, index path validation, symbolic-ref escape checks, nested-gitlink recursion. These faithfully preserve TabManager's long-shipped behavior; changing parsing semantics mid-extraction would break the PR's behavior-preservation contract. Tracked in https://github.com/manaflow-ai/cmux/issues/5359.
- **Declined (2):** test-fixture preferences — the `.git`-file form is covered by `dotGitFilePointerResolvesGitDirectory`, and the fixture's normalizing behavior is intentional and documented.

Dismissing the stale CHANGES_REQUESTED per owner instruction; local gate (48 package tests, cmux, cmux-unit) green on the merged tree.

Issue #5309, opened 2026-06-03
Memory leak root cause: CmuxSettings values(for:) parks an uncancellable Task in NotificationCenter.notifications(named:) — leaks a Task + async-sequence per settings observation (regression from #4975)


austinywang:
## Root cause located

The severe memory growth tracked in #5302 / #5305 (live heap: ~1.3M `Task stack`s, **328K `NSNotificationCenter.Notifications`** async-sequences, **207K `AsyncStream<Bool>`**, 53K `AsyncStream<String>`, on `cmux.app` 0.64.12 after ~23h) originates in the new **CmuxSettings** SPM package's `values(for:)` change-observation streams.

The heap classes map 1:1 onto this machinery:
- `values<Value>(for:)` called with `Bool` defaults keys (toggles) → the **207K `AsyncStream<Bool>`**
- called with `String` keys + the secret store → the **53K `AsyncStream<String>`**
- the `NotificationCenter.default.notifications(named:)` inside each → the **328K `NSNotificationCenter.Notifications`** + their `ManagedBuffer<…State, os_unfair_lock>` and `_ContiguousArrayStorage<UnsafeContinuation<…NotificationWrapper>>`
- the `Task { }` wrapping each stream → a large share of the **1.3M `Task stack`s**

No other first-party code produces these Foundation/AsyncStream types (verified by repo-wide grep). This file was **added 2026-05-29 by #4975** ("Reimplement Settings as CmuxSettings/CmuxSettingsUI SPM packages"), so this is a regression from that rewrite — consistent with the symptom reports landing late May / early June (#5209).

## The leaking code (verified)

**`Packages/CmuxSettings/Sources/CmuxSettings/Stores/UserDefaultsSettingsStore.swift:104-128`**

```swift
public nonisolated func values<Value>(for key: DefaultsKey<Value>) -> AsyncStream<Value> {
    AsyncStream<Value>(bufferingPolicy: .bufferingNewest(1)) { continuation in
        let task = Task { [weak self] in
            ...
            let notifications = NotificationCenter.default.notifications(
                named: UserDefaults.didChangeNotification
            )
            for await _ in notifications {        // <-- suspends in the async iterator
                if Task.isCancelled { break }     // <-- only reached AFTER a notification wakes it
                ...
            }
            continuation.finish()
        }
        continuation.onTermination = { _ in task.cancel() }   // <-- cancel() cannot unblock the await
    }
}
```

[...]

Issue #5303, opened 2026-06-03
Browser pane render loop: BrowserPanelView.body re-runs handleBrowserPanelAppear() and re-navigates the WKWebView on every CoreAnimation commit (~39% main-thread CPU, live sample)


austinywang:
## Summary

A 3-second `sample` of a live production `cmux.app` shows the **main thread spending ~39% of its on-CPU time re-evaluating `BrowserPanelView.body` and re-running `handleBrowserPanelAppear()`**, which restores the discarded WKWebView and **re-navigates it** — repeatedly, inside CoreAnimation commit handlers. This is a SwiftUI render/commit feedback loop on the browser pane. It burns CPU continuously (contributes to battery drain) and churns the WebContent helper process (re-navigation), and is a strong candidate engine for the concurrency-object leak reported in #5302.

## Environment

- cmux **0.64.12 (92)**, release build
- macOS 26.4.1, Apple Silicon, 36 GB RAM
- `sample 13331 3` on the live PID; main thread had 937 on-CPU samples in the window

## Hard evidence — `sample` call tree (trimmed)

Under the run-loop's CoreAnimation commit observer:

```
937 Main Thread
  …
  846 CA::Transaction::flush_as_runloop_observer(bool)
    837 CA::Transaction::commit()
      837 CA::Transaction::run_commit_handlers(CATransactionPhase)
        …
        369 closure #4 in BrowserPanelView.body.getter
          336 BrowserPanelView.handleBrowserPanelAppear()
             33 BrowserPanel.noteWebViewVisibility(_:reason:now:recordIfUnchanged:)
               25 closure #1 in BrowserPanel.restoreDiscardedWebViewIfNeeded(reason:)
                 25 BrowserPanel.navigateWithoutInsecureHTTPPrompt(request:…)
                   25 BrowserPanel.performNavigation(request:originalURL:…)
                     25 browserLoadRequest(_:in:)
              8 BrowserPanel.shouldRenderWebView.setter
```

So during CA commits, SwiftUI re-evaluates `BrowserPanelView.body`, which triggers `.onAppear → handleBrowserPanelAppear()`, which calls `noteWebViewVisibility(...)`, which **restores a discarded webview and re-issues navigation** (`restoreDiscardedWebViewIfNeeded → performNavigation → browserLoadRequest`). This repeats every commit pass.

## Relevant code

[...]

PR #5332, opened 2026-06-04
Fix settings-observer memory leak (#5329)

### LINKED ISSUE #5329: cmux.app 0.64.11 reached 53 GB footprint; sample shows UserDefaultsSettingsStore NotificationCenter observer storm
## Summary

A long-running `cmux.app` session on macOS grew to a **53.0 GB physical footprint** with a **61.2 GB peak**. The memory is overwhelmingly in the app process' default malloc zone, not in WebKit or IOSurface/GPU buffers.

This looks directly related to #5309, but the scale on this capture is much larger than #5305 and it occurred on the installed release **0.64.11 (91)**, commit `264455e28`. Filing this as additional live evidence; happy for maintainers to close as a duplicate if #5309 fully covers it.

## Environment

- cmux: **0.64.11 (91)**, commit `264455e28`
- Installed app path: `/Applications/cmux.app`
- macOS: **26.5.1 (25F80)**
- Architecture: `arm64`
- Hardware: Apple Silicon, **Apple M5 Max**, **128 GiB RAM**
- Capture time: 2026-06-03 ~18:55 PDT
- Process launch time: 2026-06-01 15:07 PDT, so uptime was about **2 days 3h 48m**
- Note: latest release visible at capture time was **v0.64.12**; I did not restart/upgrade before capture because I wanted to preserve the leak state.

No serial number, host identifier, user paths, project paths, or command/session identifiers are included here.

## Observed process state

`ps` / `top` showed the main app process as the outlier:

```text
/Applications/cmux.app/Contents/MacOS/cmux
RSS: 46,670,480 KB (~44.5 GiB)
Physical footprint from vmmap: 53.0G
Physical footprint peak: 61.2G
Threads: 59
```

There were several direct shell child processes under cmux, but each was small (roughly 7-10 MB RSS). The app process itself was the memory outlier.

System-wide state at the same time:

```text
Physical memory: 128 GiB
Compressor: ~11G in top output
Swapins/swapouts: 0 / 0 at capture
```

## `vmmap -summary` evidence

The footprint is dominated by the default malloc zone:

```text
Physical footprint:         53.0G
Physical footprint (peak):  61.2G

Writable regions: Total=64.5G written=253.9M resident=44.6G swapped_out=8.6G
TOTAL:            virtual=67.8G resident=46.2G dirty=44.5G swapped=8.6G
```

Largest region / zone entries:

```text
MALLOC_SMALL:        50.0G virtual, 41.1G resident, 41.0G dirty, 8.4G swapped
MALLOC_LARGE:         1.3G virtual,  1.2G resident
IOSurface:          137.6M virtual, 133.4M resident
WebKit Malloc:       32.0M virtual,   1.2M resident
JS VM Gigacage:       4.0G reserved VM, 0K resident
```

Malloc zone table:

```text
DefaultMallocZone:
  size:             52.9G
  resident:         42.6G
  swapped:           8.5G
  allocation count: 216,587,070
  bytes allocated:  50.6G
```

This does not look like the IOSurface/WebKit class of leak for this capture. Recent system logs also showed WebKit memory-pressure messages reporting the WebContent process at about **8 MB** around the capture window.

## `sample 1500 5` evidence

The 5-second sample had repeated cooperative-thread stacks creating `UserDefaultsSettingsStore.values<A>(for:)` streams backed by `NSNotificationCenter.notifications(named:)`:

```text
completeTaskWithClosure
  closure #1 in closure #1 in UserDefaultsSettingsStore.values<A>(for:)
    NSNotificationCenter.notifications(named:object:)
      NSNotificationCenter.Notifications.init(center:name:object:)
        closure #1 in NSNotificationCenter.Notifications.init(center:name:object:)
          -[NSNotificationCenter addObserverForName:object:queue:usingBlock:]
            _CFXNotificationRegisterObserver
              CFXNotificationRegistrarAdd
                _os_unfair_lock_lock_slow
                  __ulock_wait2
```

The same sample also had main-thread work in the title/update path:

```text
GhosttyApp.handleAction(target:action:)
  NSNotificationCenter postNotificationName:object:userInfo:
  TabManager.enqueuePanelTitleUpdate(tabId:panelId:title:)
    NotificationBurstCoalescer.scheduleFlushIfNeeded()
    TabManager.flushPendingPanelTitleUpdates()
      TabManager.updatePanelTitle(tabId:panelId:title:)
```

I am not claiming the title-update path is the allocation source, only that it was active in the same capture. The stronger evidence is the `UserDefaultsSettingsStore.values` / `NSNotificationCenter.notifications(named:)` stack plus the enormous default malloc allocation count.

## Why this seems related to #5309

#5309 describes a leak from `CmuxSettings values(for:)` parking tasks/async sequences around `NotificationCenter.notifications(named:)`. This live sample shows that same code path active in a process whose default malloc zone had grown to **50.6 GB across ~216.6M allocations**.

The scale here suggests either:

1. #5309 also affects 0.64.11 and can grow far beyond the 0.64.12 audit numbers, or
2. there is an additional amplifier around settings observation / notification registration in 0.64.11.

## Ask

Could maintainers confirm whether this is fully covered by #5309, and whether 0.64.12 contains any mitigation for the `UserDefaultsSettingsStore.values` / `NSNotificationCenter.notifications(named:)` path?

If it is the same root cause, it would be useful to have a regression guard that watches one or more of these counts over repeated view remounts / settings observations:

- live `NSNotificationCenter.Notifications`
- live Swift `Task stack` allocations
- `AsyncStream<Bool>` / settings streams
- default malloc allocation count / footprint

I did not attach the raw sample or logs because they can contain local runtime state. I can provide a further sanitized excerpt if helpful.

danielraffel:
## What this fixes

`UserDefaultsSettingsStore.values(for:)` and `SecretFileStore.values(for:)` each started a `Task` that parked inside `NotificationCenter.notifications(named:)` and only checked `Task.isCancelled` *after* a notification woke it. `continuation.onTermination` calls `task.cancel()`, but `cancel()` can't unblock that await — so every time a settings consumer is torn down (a view remount, a settings panel close, a `@LiveSetting` host losing identity) the `Task`, its `AsyncStream`, and the `NSNotificationCenter.Notifications` sequence stay alive until the named notification next fires. For a quiet key — the secret-store change notification, or any `UserDefaults` key on an idle store — that's never, so the trio leaks permanently.

Over a long session these accumulate badly. The report in #5329 captured a **53 GB physical footprint / 61 GB peak** with **~216M live allocations** dominated by `Task stack` / `NSNotificationCenter.Notifications` / `AsyncStream<Bool>`. This is the same root cause already analyzed in #5309.

## The fix

Both stores now run off a single store-owned **block** observer (`addObserver(forName:object:queue:using:)`), bridged into a bounded `AsyncStream<Void>` and fanned out to one bounded signal per live consumer — exactly the subscriber pattern `JSONConfigStore.values(for:)` already uses. Because the internal signal stream is a real, cancellation-aware `AsyncStream`, cancelling a consumer now breaks its loop immediately, deregisters its subscriber, and finishes the stream. No per-consumer notification iterator, nothing parked.

The block registration is wrapped in a small `NotificationObserverHandle` whose own `deinit` removes the observer, so the actor never has to touch the non-`Sendable` observer token from its `nonisolated deinit`.

Net effect: one observer + one fan-out task per store (not per consumer), and teardown reclaims everything synchronously.

## Tests

`SettingsObserverLeakTests` exercises the teardown contract through a new `activeSubscriberCount` seam:
- cancelling a consumer drops the subscriber count back to 0 promptly (UserDefaults + secret store)
- 50 mount/teardown cycles don't accumulate subscribers

`swift test` for the package is green except for one **pre-existing, unrelated** failure — `SettingCatalogTests.userDefaultsStorageKeysAreUnique` (126 keys, 11 duplicates) — which fails on `main` and has nothing to do with these stores.

Closes #5329. Related: #5309, #5303 (remount loop that multiplies the leak).

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/5332"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783131462&installation_id=133855650&pr_number=5332&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F5332&signature=459af449242e533d42023008947ce4c83cdd1179201da1241134c289dc4bca1c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fixes a severe memory leak in settings observers by replacing per-consumer `NotificationCenter.notifications(named:)` loops with a single store-owned block observer and bounded `AsyncStream` fan-out. Streams now subscribe before the first read and cancel cleanly, preventing parked tasks and stale initial values. Closes #5329.

- **Bug Fixes**
  - `UserDefaultsSettingsStore.values(for:)` and `SecretFileStore.values(for:)` use one block-based observer per store with bounded fan-out; subscribers register before the initial read; no per-consumer notification iterators.
  - Added `NotificationObserverHandle` to own and remove the observer in `deinit` safely.
  - Added `SettingsObserverLeakTests` and an `activeSubscriberCount` seam to verify prompt deregistration and no accumulation across 50 remount cycles; strengthened churn assertions.

<sup>Written for commit 3de17f220b0f720dd72e7301dfcbb50491611f46. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/5332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved subscriber/observer cleanup issues so subscriptions are removed promptly and do not accumulate.

* **Refactor**
  * Centralized and optimized change-notification delivery with bounded buffering and deduplication to reduce redundant updates.

* **Tests**
  * Added cancellation and churn tests verifying subscriber counts return to zero after consumer cancellation.

* **Documentation**
  * Updated docs to describe the new single-observer, fan-out, and buffering/deduplication behavior.

### DIFF
--- a/Packages/CmuxSettings/Sources/CmuxSettings/Stores/NotificationObserverHandle.swift
+++ b/Packages/CmuxSettings/Sources/CmuxSettings/Stores/NotificationObserverHandle.swift
@@ -1,0 +1,35 @@
+import Foundation
+
+/// Owns a single block-based `NotificationCenter` observer registration and
+/// removes it on `deinit`.
+///
+/// Storing one of these instead of the raw observer token lets an `actor`
+/// register an observer without ever touching the non-`Sendable` token from its
+/// own `nonisolated deinit`: when the owning actor is deallocated, ARC releases
+/// the handle and the handle's `deinit` removes the observer. The block is the
+/// only `NotificationCenter` surface the store exposes — the leak-prone
+/// `NotificationCenter.notifications(named:)` async iterator (#5329 / #5309) is
+/// never used.
+final class NotificationObserverHandle {
+    private let center: NotificationCenter
+    private let token: any NSObjectProtocol
+
+    /// Registers `block` for `name` on `center` (any object, no specific queue).
+    /// - Parameters:
+    ///   - center: The center to observe. Defaults to `.default`.
+    ///   - name: The notification name to observe.
+    ///   - block: Invoked for each matching notification. Must be `@Sendable`
+    ///     because `NotificationCenter` may deliver on any thread.
+    init(
+        center: NotificationCenter = .default,
+        name: Notification.Name,
+        using block: @escaping @Sendable (Notification) -> Void
+    ) {
+        self.center = center
+        self.token = center.addObserver(forName: name, object: nil, queue: nil, using: block)
+    }
+
+    deinit {
+        center.removeObserver(token)
+    }
+}
--- a/Packages/CmuxSettings/Sources/CmuxSettings/Stores/SecretFileStore.swift
+++ b/Packages/CmuxSettings/Sources/CmuxSettings/Stores/SecretFileStore.swift
@@ -35,11 +35,30 @@ public actor SecretFileStore {
     /// The directory secret files are resolved under.
     public nonisolated let baseDirectory: URL
 
+    /// Per-consumer change signals. One entry per live ``values(for:)`` stream;
+    /// a single ``didChangeNotification`` fans out one `()` to each.
+    private var subscribers: [UUID: AsyncStream<Void>.Continuation] = [:]
+
+    /// The block-observer registration, created lazily on the first subscriber.
+    /// Releasing it (when this actor is deallocated) removes the observer.
+    private var observerHandle: NotificationObserverHandle?
+
+    /// The single long-lived task draining the bridged notification stream and
+    /// fanning out to ``subscribers``. Created lazily with ``observerHandle``.
+    private var observerTask: Task<Void, Never>?
+
     /// Creates a store rooted at `baseDirectory`.
     /// - Parameter baseDirectory: The directory holding the secret files
     ///   (created on first write if absent). The app uses `~/.config/cmux`.
     public init(baseDirectory: URL) {
         self.baseDirectory = baseDirectory
+    }
+
+    deinit {
+        // The observer is removed by `observerHandle`'s own deinit via ARC; we
+        // only stop the long-lived fan-out task here (Task is Sendable, so this
+        // is legal from a nonisolated deinit).
+        observerTask?.cancel()
     }
 
     /// The current secret for `key`, or its ``SecretFileKey/defaultValue`` when absent/empty.
@@ -106,8 +125,12 @@ public actor SecretFileStore {
     /// An `AsyncStream` yielding the current secret and every later change.
     ///
     /// The first element is the current value; subsequent elements arrive when
-    /// ``didChangeNotification`` fires for this key. Buffering is
-    /// `.bufferingNewest(1)`. Cancelling the consuming `Task` ends the stream.
+    /// ``didChangeNotification`` fires. Each signal re-reads the key's value and
+    /// dedups, so a change to an unrelated key never yields here. Buffering is
+    /// `.bufferingNewest(1)`. Cancelling the consuming `Task` synchronously
+    /// deregisters its subscriber and ends the stream — teardown never waits on
+    /// a future notification, so observers do not leak across remounts (the
+    /// #5329 / #5309 leak).
     public nonisolated func values(for key: SecretFileKey) -> AsyncStream<String> {
         AsyncStream<String>(bufferingPolicy: .bufferingNewest(1)) { continuation in
             let task = Task { [weak self] in
@@ -118,21 +141,21 @@ public actor SecretFileStore {
                 var lastYielded = (try? await self.value(for: key)) ?? key.defaultValue
                 continuation.yield(lastYielded)
 
-                let notifications = NotificationCenter.default.notifications(
-                    named: SecretFileStore.didChangeNotification
+                let id = UUID()
+                let (signal, signalContinuation) = AsyncStream<Void>.makeStream(
+                    bufferingPolicy: .bufferingNewest(1)
                 )
-                for await note in notifications {
+                await self.addSubscriber(id: id, continuation: signalContinuation)
+
+                for await _ in signal {
                     if Task.isCancelled { break }
-                    if let changedID = note.userInfo?[SecretFileStore.changedKeyIDKey] as? String,
-                       changedID != key.id {
-                        continue
-                    }
                     let current = (try? await self.value(for: key)) ?? key.defaultValue
                     if current != lastYielded {
                         lastYielded = current
                         continuation.yield(current)
                     }
                 }
+                await self.removeSubscriber(id: id)
                 continuation.finish()
             }
             continuation.onTermination = { _ in task.cancel() }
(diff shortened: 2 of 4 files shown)


danielraffel:
Addressed the review feedback in 3de17f2:

- **Race in `UserDefaultsSettingsStore.values(for:)` and `SecretFileStore.values(for:)`** (coderabbit + cubic P1): the initial value was yielded *before* `addSubscriber`, so a `set`/`reset` landing in that window posted `didChangeNotification` with no subscriber registered and the stream stayed stale. Both streams now create the UUID, make the signal stream, and `addSubscriber` **first**, then read and yield the initial value. A change arriving in the subscribe→read gap buffers a signal that re-reads and dedups, so the first real change is no longer missed.
- **Churn test discarded poll outcomes** (cubic P2): `userDefaultsSubscribersDoNotAccumulateAcrossChurn` now `#expect`s both the register (`>= 1`) and deregister (`== 0`) poll on every cycle, with a per-cycle failure message, so a broken subscription path can't slip through.

On the `SecretFileStore` per-key-filter note from the Greptile summary: dropping the `changedKeyIDKey` filter so all consumers wake + re-read + dedup on any key change is intentional — it mirrors `JSONConfigStore`'s fan-out exactly (the template #5309 points to), and dedup suppresses the false emissions. Secret writes are rare, so the extra reads are negligible.

This keeps the PR aligned with #5309, whose prescribed fix is precisely this `addSubscriber`/`removeSubscriber` fan-out pattern; the secondary items it lists (FeedPanelViewModel re-arm, the #5303 remount loop) are called out there as "not strictly required for the leak" and are out of scope here.

`swift test` passes for the full `CmuxSettings` suite except the pre-existing `SettingCatalogTests.userDefaultsStorageKeysAreUnique` failure (126 vs 115 unique keys), which fails identically with these changes stashed and is unrelated to this PR.

Issue #5617, opened 2026-06-08
Codex: a new session in a workspace silently orphans the prior thread's resume binding (no way back)


lawrencecchen:
A workspace's codex resume binding is overwritten whenever a new codex session starts in that surface, so a later (often throwaway) session permanently displaces the workspace's prior thread. The prior rollout is still on disk but is no longer referenced by any surface, so resuming the workspace restores the wrong session and there is no UI path back to the real one.

Concrete case. A workspace titled "workspace group pin order" hosted a substantive thread (rollout `019e9cbd-1d39-7822-aa1c-58425e4b71ef`: implemented pinned-in-group ordering, ran `$autoreview and merge`, opened https://github.com/manaflow-ai/cmux/pull/5541). After that thread finished cleanly, a fresh `codex` was started in the same surface to ask "what were u working on" (rollout `019ea517-c257-7dc2-b8d0-318b63dc53f8`, a 2-message recap). The workspace's `resumeBinding.checkpointId` flipped to the recap session (`source: agent-hook`), and the pin-order thread became orphaned in every session snapshot. Resuming the workspace now restores the recap, not the real thread.

Root cause:
- The codex/generic `.sessionStart` hook calls `publishAgentSurfaceResumeBinding` unconditionally (`CLI/cmux.swift`), where the claude path gates it behind `shouldPromoteActiveSession` (only `/clear` or replacing a stopped session).
- The app-side write is last-writer-wins with no checkpoint guard (`Workspace.setSurfaceResumeBinding` sets `surfaceResumeBindingsByPanelId[panelId] = binding`).
- There is no per-surface session history, so once the binding is overwritten there is no way back to the prior thread.

"Newest session wins" may be intended, but losing a substantive completed thread to a throwaway, with no way back, is the bug.

The narrow live-clobber subset (a fresh session displacing a still-running different session in the remote/no-TTY slice that https://github.com/manaflow-ai/cmux/issues/5333 punts on) is being fixed separately on branch `feat-codex-resume-displacement`. This issue tracks the broader gap: a completed thread should stay reachable.

Recovery today (manual): `codex resume <prior-session-id>` in the workspace re-publishes the binding to the real thread.

Possible directions:
- Keep a per-surface session history / "resume previous thread" affordance so prior threads stay reachable.
- Refuse to silently overwrite a different session's durable binding without a recoverable path to the displaced thread.

[...]

Issue #5333, opened 2026-06-04
Harden codex hook: PID/TTY truth should override a disagreeing env CMUX_SURFACE_ID


lawrencecchen:
Follow-up to https://github.com/manaflow-ai/cmux/pull/5331 (fixes the spawn-env *source* of codex surface jumbling at `configureTmuxCompatEnvironment`).

The codex/generic hook (`runGenericAgentHook` in `CLI/cmux.swift`) resolves the target surface from `CMUX_SURFACE_ID` in the env (`directSurfaceArg`, ~L27131) and **suppresses the PID/TTY corrector whenever both env ids are present** (`guard directWorkspaceArg == nil || directSurfaceArg == nil else { return nil }`, ~L27163). So the hook trusts the env over process truth.

That's fine now that #5331 fixes the launcher leak and plain `codex`+hook already inherits the matched per-surface env. But it means the codex hook is **not self-correcting**: if any other path ever leaks a wrong `CMUX_SURFACE_ID` into codex's env (a future spawn-env bug, codex spawned as a subprocess inheriting a stale id, etc.), the hook records the wrong surface and PID truth can't override it.

Proposed hardening (from the root-cause workflow): let `resolveAgentProcessTerminalBinding(pid)` / `resolveCallerTerminalBindingByTTY` override a **disagreeing** env surface id when PID resolves to a different *valid* surface, while keeping env as the fallback when PID is unavailable (remote/SSH). Also consider preferring the persisted session mapping over the env on prompt-submit/teardown (mirroring `resolvePreferredSurfaceIdForClaudeHook(preferred: mappedSession?.surfaceId)`).

Riskier change to intricate hook routing — needs focused tests + dogfood. Deferred from #5331 to keep that PR scoped to the source.

lawrencecchen:
Fixed by https://github.com/manaflow-ai/cmux/pull/5351 (codex resume/restore hardening), merged in 5b6325ddfd. The generic hook now overrides a disagreeing/stale ambient surface with the agent's PID/TTY-bound pane, and the omx launcher-env source fix (#5331) was folded in.

PR #5620, opened 2026-06-08
Codex: don't displace a live different session's resume binding (no-TTY slice)

lawrencecchen:
Guards the codex/generic resume-binding publish so a fresh session cannot overwrite a **different, still-live** session's resume binding on the same surface.

### The bug
The codex/generic agent hook published `surface.resume.set` unconditionally on every session-start / prompt-submit / stop (`CLI/cmux.swift`), unlike the claude path which gates it behind `shouldPromoteActiveSession`. Combined with the app-side last-writer-wins write (`Workspace.setSurfaceResumeBinding`), a fresh session routed onto a surface a *different live* session occupies overwrites that session's binding and orphans the running thread across reload.

### Where it bites (and where it doesn't)
Only in the remote/SSH slice that https://github.com/manaflow-ai/cmux/issues/5333's PID/TTY override explicitly punts on: no controlling TTY, so a leaked `CMUX_SURFACE_ID` misroutes the hook onto the live session's pane. In the common single-terminal flow this cannot happen (the prior process is dead before the next starts, and `hasRunningSession(requireLiveProcess:)` self-heals the stale record), so single-terminal behavior is unchanged.

### Fix
Route all four generic-hook publish sites through one local helper that skips the publish when a different session is live on the surface (`requireLiveProcess: true`). A stopped/dead prior session does not count, so normal resume-latest takeover of a finished session is preserved.

### Scope
This does **not** fix the broader incident where a *completed* thread is displaced by a later throwaway session with no way back (the single-terminal case): that is the no-per-surface-history product gap tracked in https://github.com/manaflow-ai/cmux/issues/5617.

### Verification
CI's `tests` job masks the red regression test (separate gap, https://github.com/manaflow-ai/cmux/issues/5641), so red→green is shown via a mock-socket repro driving the real built CLI:

```
unfixed CLI:  LIVE prior -> B clobbers (True, bug)   STOPPED prior -> B publishes (True)   => FAIL
fixed CLI:    LIVE prior -> no displace (False)       STOPPED prior -> B publishes (True)   => PASS
```

Two app-host hook tests added (`CLIGenericHookPersistenceTests`): the live-clobber regression (red without the fix) and a stopped-session-takeover characterization (stays green).

---

> [!NOTE]
> **Medium Risk**
> Changes when resume bindings are written for multi-session / remote routing edge cases; wrong guard logic could block legitimate takeover or still allow clobbering, but scope is hook publish paths with targeted tests.
> 
> **Overview**
> Generic codex/agent hooks no longer call `surface.resume.set` unconditionally. They go through **`publishResumeBindingUnlessLiveSurfaceOwner`**, which skips publishing when another **older** session on the same workspace/surface still has a **live PID** (any `runtimeStatus`, including `needsInput`), so a misrouted fresh session cannot clobber the incumbent’s resume binding in the no-TTY / leaked `CMUX_SURFACE_ID` case.
> 
> **`hasLiveProcessSessionOnSurface`** implements that check: `startedAt` tie-breaking so only the original owner blocks a newcomer, plus **`processStartTime`** (sysctl) vs `updatedAt` so a stale record with a **recycled PID** does not block normal resume-latest takeover when the prior agent is actually gone.
> 
> Four hook paths (session-start, prompt-submit, stop, etc.) are switched from direct `publishAgentSurfaceResumeBinding` to the guarded helper. **`CLIGenericHookPersistenceTests`** adds coverage for stopped-prior takeover, live-prior no-clobber, `needsInput` live prior, and PID reuse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5290f35332a7e88c2abdabf1cc85e9c14873db6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve existing resume bindings when another active agent session owns the same surface, preventing unintended takeover or disruption.
  * Improve robustness against recycled process identifiers so live sessions aren’t mistaken for inactive ones.

* **Tests**
  * Added regression tests for session-start behavior covering live, stopped, and process-reuse scenarios to prevent resume-binding regressions.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -1181,12 +1181,80 @@ private final class ClaudeHookSessionStore {
         }
     }
 
+    /// Whether the given surface is still owned by a *different*, older, live agent session — the one
+    /// whose resume binding the caller must not displace.
+    ///
+    /// "Live" means the recorded PID still has a running process, regardless of `runtimeStatus`: a
+    /// session paused on an approval prompt (`needsInput`), idle at the prompt (`idle`), or in an
+    /// `error` state is still resumable (unlike `hasRunningSession`, which only counts `.running`).
+    /// "Owner" means it started before the current session (`olderThan`): the original occupant wins,
+    /// and a later, misrouted session does not in turn block the owner's own resume updates.
+    func hasLiveProcessSessionOnSurface(
+        workspaceId: String,
+        surfaceId: String,
+        excludingSessionId: String?,
+        olderThan currentSessionStartedAt: TimeInterval?
+    ) throws -> Bool {
+        guard let normalizedWorkspace = normalizeOptional(workspaceId),
+              let normalizedSurface = normalizeOptional(surfaceId) else {
+            return false
+        }
+        let excluded = normalizeOptional(excludingSessionId)
+        return try withLockedState { state in
+            for sessionId in Array(state.sessions.keys) {
+                guard let record = state.sessions[sessionId] else { continue }
+                guard normalizeOptional(record.workspaceId) == normalizedWorkspace,
+                      normalizeOptional(record.surfaceId) == normalizedSurface,
+                      record.sessionId != excluded,
+                      let pid = record.pid,
+                      Self.processExists(pid) else {
+                    continue
+                }
+                // Ownership tiebreaker: only yield to a session that started before the current one.
+                // The original surface owner started first; a misrouted newcomer is younger and must
+                // not, in turn, suppress the owner's own later resume updates (symmetric blocking).
+                if let currentSessionStartedAt, record.startedAt >= currentSessionStartedAt {
+                    continue
+                }
+                // Defend against PID reuse: a crashed agent's record can linger with a PID later
+                // recycled by an unrelated long-lived process, which `kill(pid, 0)` would report as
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Closing without merging. Structured review showed this is not a narrow standalone bug: guarding only `surface.resume.set` leaves the same no-TTY misroute corrupting the live owner's surface via `set_agent_pid` and `setAgentLifecycle`/status, producing an incoherent surface (binding points at the owner, PID/status at the newcomer). That is a facet of the no-TTY routing case https://github.com/manaflow-ai/cmux/issues/5333 deliberately punts on and needs a routing-level decline, not a per-mutation guard.

Refiled precisely as https://github.com/manaflow-ai/cmux/issues/5676. The branch stays as reproduction evidence: a mock-socket repro driving the real CLI hook, plus four app-host regression tests (live running/needsInput preserved; stopped/recycled-PID take over). The resume-binding behavior was verified end-to-end against the built CLI; the multi-mutation suppression was not (the mock-socket harness can't keep a misrouted session's PID live, and the `tests` job masks red app-host tests — https://github.com/manaflow-ai/cmux/issues/5641).

PR #5647, opened 2026-06-08
iOS: fix beta/TestFlight push delivery (production aps-environment + matching gateway)

lawrencecchen:
## Problem

Beta (TestFlight) push notifications never deliver. Two independent breakages, both required to fix:

**A. Wrong entitlement.** The iOS app had a single entitlements file (`ios/Config/cmux.entitlements`) wired for every build config via `CODE_SIGN_ENTITLEMENTS` in `Shared.xcconfig`, hardcoding `aps-environment = development`. A TestFlight/App-Store build with a `development` aps-environment registers a **sandbox** APNs device token, but TestFlight runs against the **production** APNs host (`api.push.apple.com`), which rejects sandbox tokens with `BadDeviceToken`.

**B. The right entitlement gets dropped during export.** Fixing the file is not enough. https://github.com/manaflow-ai/cmux/pull/5496 switched the beta archive to **unsigned** (`CODE_SIGNING_ALLOWED=NO`) to avoid distribution-cert churn. An unsigned archive carries **no** entitlements, so `-exportArchive` re-adds only the provisioning-profile baseline (`application-identifier`, `com.apple.developer.team-identifier`, `get-task-allow`, `beta-reports-active`) and **silently drops every app-capability entitlement**, including `aps-environment`. Verified on the actual signed beta IPA: `aps-environment` was **absent entirely**. So beta/prod push has been dead since #5496 regardless of what the entitlements file says.

This PR fixes both. A config-level entitlements file alone is **behaviorally inert** without the re-sign in part 2.

## Part 1 -- Per-config entitlements file

- Before: one `cmux.entitlements` with `development`, applied to every config via `Shared.xcconfig`. Release/beta got a sandbox token.
- After: split per build config. `CODE_SIGN_ENTITLEMENTS` moved out of `Shared.xcconfig` into the two per-config xcconfigs:
  - `Debug.xcconfig` -> `Config/cmux.entitlements` (`development`, sandbox host) -- local `reload.sh` dev push unchanged.
  - `Release.xcconfig` -> `Config/cmux-release.entitlements` (`production`, `api.push.apple.com`) -- new file, also carries Sign in with Apple and time-sensitive notifications.
- Verified with `xcodebuild -showBuildSettings`: Debug resolves `cmux.entitlements` + `DEBUG`; Release resolves `cmux-release.entitlements` with no `DEBUG`.

This keeps the reported `apnsEnvironment` (`MobileAuthComposition`, `production` for non-DEBUG) in lockstep with the entitlement, and the backend (`web/services/apns`) already routes per stored token environment to the matching host with the production `.p8` provisioned in Vercel. No backend or Swift change.

## Part 2 -- Re-sign the exported IPA with the full entitlements (the part that actually delivers)

Because the archive is unsigned, the only way to guarantee `aps-environment` reaches the signed binary is to re-sign the exported app with it. `ios/scripts/upload-testflight.sh` now, on the **manual** signing path, after `-exportArchive` and before upload:

1. Reads the exported app's current (baseline) entitlements via `codesign -d --entitlements :- --xml`.
2. **Generically** merges in every key from `Config/cmux-release.entitlements` using `PlistBuddy Merge` (copies all keys, skips ones already in the baseline, so `get-task-allow=false` is preserved and future entitlements survive automatically -- nothing hardcodes `aps-environment`).
3. Re-signs the single self-contained `.app` (no `Frameworks/` or `PlugIns/`, GhosttyKit is static, so no nested-code re-sign) with the local `Apple Distribution: Manaflow, Inc. (7WLXT3NR37)` cert + `--timestamp`.
4. **Hard gates:** `codesign -d --entitlements` must show `aps-environment`, and `codesign --verify --strict` must pass. The same two checks re-run on the re-zipped IPA's extracted `.app`, so a bad Payload layout can't slip through.
5. Re-zips `Payload/` into a new IPA and repoints `$IPA_PATH` so the upload ships the re-signed build.

The identity name and entitlements path are overridable via `IOS_DISTRIBUTION_IDENTITY` / `IOS_RELEASE_ENTITLEMENTS`; they default to the distribution cert and `Config/cmux-release.entitlements`.

**Approaches ruled out:** ad-hoc archive (`CODE_SIGN_IDENTITY="-"`) -- the iOS SDK rejects ad-hoc signing for an entitled app (*entitlements require a development certificate*); signing on the fleet -- keeps distribution material off the shared Macs.

**Automatic (CI / cloud-managed) signing path -- fails closed:** there is no named distribution cert in the keychain to re-sign with, so the dropped entitlements cannot be re-added there. `ios-testflight.yml` drives the **primary** beta cut with `--signing automatic`, and the entitlement drop is caused by the *unsigned archive* (which both modes export from), so an automatic export almost certainly drops `aps-environment` too. Rather than warn-and-upload a push-broken IPA (CI warnings are effectively silent), the automatic path now **verifies the exported IPA carries `aps-environment` via `codesign -d` and exits non-zero if it does not** -- so CI fails closed instead of shipping a broken beta. The check is one shared `verify_ipa_has_aps_environment` helper used by both the manual post-re-sign gate and this automatic gate.

## Verification

- `xcodebuild -showBuildSettings` confirms the per-config entitlement + DEBUG/non-DEBUG split stays in lockstep; `plutil -lint` passes on both entitlements.
- Dry-run of the generic merge (fake export baseline + the real `cmux-release.entitlements`): result contains all 4 baseline keys **plus** the 3 release keys (`aps-environment=production`, `com.apple.developer.applesignin`, `com.apple.developer.usernotifications.time-sensitive`), with `get-task-allow` preserved as `false`.
- The shared gate asserts `aps-environment` **== `production`** (not mere presence), so a sandbox/`development` value is rejected on both paths. Unit-tested against real IPAs: passes the production re-signed IPA, fails a baseline export (absent), fails an IPA re-signed with `aps-environment=development`.
- **Verified manually end-to-end:** the re-signed IPA carries `aps-environment=production` (confirmed via `codesign -d --entitlements`), `codesign --verify --strict` is valid, and `altool` **accepted the upload**.
- `shellcheck` clean; autoreview clean.

## User actions required to confirm delivery on the next beta

1. Build/upload a new TestFlight beta from this branch via the manual signing path (the re-sign only runs there).
2. On the Mac: Settings -> Notifications -> enable **Forward notifications to phone** (off by default).
3. On the iPhone beta: grant the iOS notification permission and enable push in the app (off by default).
4. Trigger an agent notification on the Mac and confirm it arrives on the phone.

Do not merge yet -- this needs device dogfood on a real production-signed beta build that actually receives a push.

## Open decision -- how CI cuts a push-working beta (heads-up: this reds CI)

**Merging this as-is turns the `ios-testflight.yml` TestFlight upload job RED.** The automatic export drops `aps-environment`, so the new fail-closed gate refuses the upload. That is intended -- failing closed beats silently shipping a push-broken beta -- but it means **no beta cuts via CI** until you pick one of the options below. (Manual local beta cuts via `--signing manual` already work: the re-sign runs and the gate passes.)

This script change stops CI from shipping a broken beta, but CI alone cannot yet produce a push-*working* one, because `-exportArchive` does not re-add `aps-environment` from the unsigned archive on either signing mode. Pick one:

- **(a)** Import the iOS distribution cert into `ios-testflight.yml` and switch it to `--signing manual`, so the re-sign runs in CI. `nightly.yml` and `release.yml` already import a signing cert on ephemeral runners, so the "no distribution material on CI" line isn't absolute. This is a security-relevant workflow + secrets change, intentionally out of scope for this PR.
- **(b)** Keep cutting betas manually (where the re-sign runs) and let the CI `ios-testflight.yml` automatic path stay fail-closed.

## Residual risk

App Store Connect ingests re-signed binaries (altool already accepted one such upload), so ingestion is not a concern. Two narrower caveats:

- The manual end-to-end verification re-added only `aps-environment`, and that is the only key `altool` has actually validated against the *Beta Distribution* profile. The generic merge re-adds two more keys this profile-stripped export drops (`com.apple.developer.applesignin`, `com.apple.developer.usernotifications.time-sensitive`); they are almost certainly authorized by the App ID (the app ships Sign in with Apple and they live in the committed Release entitlements), but the dogfood upload will confirm the full 7-key set ingests.
- `altool` acceptance proves the entitlements are valid, **not** that a TestFlight install registers and receives a push -- that still requires the device dogfood above.

---

> [!NOTE]
> **High Risk**
> Changes push delivery, code signing/TestFlight upload gates (CI may fail until manual signing), and production auth/API endpoints—high impact on notifications and release pipeline.
> 
> **Overview**
> Fixes **TestFlight/beta push** by aligning production APNs, device-token registration, and signed entitlements end-to-end.
> 
> **Production API URLs** in `AuthConfig` now use `https://cmux.com` instead of `cmux.dev`, avoiding 301 redirects that turn `POST /api/device-tokens` into GET and break silent registration.
> 
> **Per-config push entitlements:** Debug keeps `cmux.entitlements` (`aps-environment: development`); Release/TestFlight uses new `cmux-release.entitlements` (`production`), wired via `Debug.xcconfig` / `Release.xcconfig` instead of a single shared file. Marketing version bumps to **1.0.1**.
> 
> **TestFlight upload script** adds `verify_ipa_aps_environment_production` and, on **manual** signing, re-signs the exported IPA by merging release entitlements into the profile baseline (unsigned archives drop `aps-environment` at export). Upload is blocked unless the signed app has `aps-environment == production`. The **automatic** CI path fails closed the same way if the export lacks production push entitlements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c31ebe1a2ce9a8604d64ad36a072be12db9578b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Packages/CmuxAuthRuntime/Sources/CmuxAuthRuntime/Coordinator/AuthConfig.swift
+++ b/Packages/CmuxAuthRuntime/Sources/CmuxAuthRuntime/Coordinator/AuthConfig.swift
@@ -54,8 +54,13 @@ public struct AuthConfig: Equatable, Sendable {
             callbackURL = "http://localhost:3000/auth/callback"
             defaultAPIBaseURL = "http://localhost:3000"
         case .production:
-            callbackURL = "https://cmux.dev/auth/callback"
-            defaultAPIBaseURL = "https://cmux.dev"
+            // cmux.com is the canonical production origin; cmux.dev is a legacy
+            // alias that 301-redirects everything to cmux.com. A 301 makes
+            // URLSession downgrade POST→GET, so device-token registration
+            // (POST /api/device-tokens) silently failed when pointed at cmux.dev.
+            // Point straight at cmux.com so POSTs are not redirected.
+            callbackURL = "https://cmux.com/auth/callback"
+            defaultAPIBaseURL = "https://cmux.com"
         }
 
         let override = overrides["ApiBaseURL"]
--- a/ios/Config/Debug.xcconfig
+++ b/ios/Config/Debug.xcconfig
@@ -5,3 +5,8 @@
 #include "Shared.xcconfig"
 
 SWIFT_ACTIVE_COMPILATION_CONDITIONS = $(inherited) CMUX_DEV_AUTH
+
+// Local dev / tagged Debug builds register sandbox device tokens, so use the
+// "development" aps-environment (sandbox APNs host). Mirrors the "sandbox"
+// apnsEnvironment reported by MobileAuthComposition under #if DEBUG.
+CODE_SIGN_ENTITLEMENTS = Config/cmux.entitlements
--- a/ios/Config/Release.xcconfig
+++ b/ios/Config/Release.xcconfig
@@ -10,3 +10,10 @@ DEVELOPMENT_TEAM = 7WLXT3NR37
 // Distinguish the TestFlight beta on-device from a future App Store "cmux"
 // (overrides PRODUCT_DISPLAY_NAME = cmux from Shared.xcconfig). Debug stays "cmux".
 PRODUCT_DISPLAY_NAME = cmux BETA
+
+// TestFlight/App Store builds run against the production APNs host
+// (api.push.apple.com), so use the "production" aps-environment. A "development"
+// entitlement here yields a sandbox token the production host rejects, which is
+// why beta push silently failed to deliver. Mirrors the "production"
+// apnsEnvironment reported by MobileAuthComposition for non-DEBUG builds.
+CODE_SIGN_ENTITLEMENTS = Config/cmux-release.entitlements
--- a/ios/Config/Shared.xcconfig
+++ b/ios/Config/Shared.xcconfig
@@ -44,5 +44,13 @@ INFOPLIST_KEY_NSCameraUsageDescription = Scan cmux pairing QR codes from your Ma
 // ==========================================
 // Entitlements
 // ==========================================
-// AI agents can modify Config/cmux.entitlements to add capabilities
-CODE_SIGN_ENTITLEMENTS = Config/cmux.entitlements
+// CODE_SIGN_ENTITLEMENTS is set per-config, not here, so the aps-environment
+// can differ between local dev and distribution builds:
+//   - Debug.xcconfig   -> Config/cmux.entitlements         (aps-environment: development / sandbox APNs)
+//   - Release.xcconfig -> Config/cmux-release.entitlements (aps-environment: production / api.push.apple.com)
+// TestFlight/App Store run against production APNs, so a "development"
+// aps-environment there yields a sandbox token the production host rejects
+// (silent push failure). Keep each file's aps-environment in lockstep with the
+// reported apnsEnvironment in MobileAuthComposition.
+// AI agents can modify Config/cmux.entitlements (and cmux-release.entitlements)
+// to add capabilities; mirror capability changes across both files.
--- a/ios/Config/Shared.xcconfig
+++ b/ios/Config/Shared.xcconfig
@@ -13,7 +13,7 @@ PRODUCT_BUNDLE_IDENTIFIER = dev.cmux.ios
 // release with ios/scripts/bump-ios-version.sh. Keep this human-meaningful and
 // tellable; CFBundleVersion (CURRENT_PROJECT_VERSION) is the monotonic build id
 // (a UTC timestamp stamped by ios/scripts/upload-testflight.sh on TestFlight).
-MARKETING_VERSION = 1.0.0
+MARKETING_VERSION = 1.0.1
 CURRENT_PROJECT_VERSION = 1
 
 // Dev-build identity, surfaced in-app (Settings > About) so a DEBUG dogfood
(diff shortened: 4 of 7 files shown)


lawrencecchen:
Follow-up: verifying the cut beta surfaced a **deeper** bug than the entitlement value. Inspecting the *signed* app showed `aps-environment` was **absent entirely** (only application-identifier / team-identifier / get-task-allow / beta-reports-active). Root cause: `upload-testflight.sh` archived UNSIGNED (`CODE_SIGNING_ALLOWED=NO`), which skips Xcode's entitlements processing, so the archive carried no entitlements and `-exportArchive` re-added only the profile baseline, silently dropping `aps-environment`. That, not just dev-vs-prod, is why beta/prod push never delivered.

Pushed the load-bearing fix to this branch: archive with **ad-hoc** signing (`CODE_SIGN_IDENTITY="-"`) instead of unsigned. Ad-hoc mints no certificate (keeping the cert-churn protection) but runs the entitlements step, so the archive captures `aps-environment` and export carries it into the IPA. Verified on the signed binary (config-level `-showBuildSettings` does NOT prove this). A re-cut beta is uploading with a hard `codesign -d --entitlements` gate before upload.

lawrencecchen:
Superseded by https://github.com/manaflow-ai/cmux/pull/5729, which carries this PR's post-export re-sign mechanism as a strict superset (plus the fleet archive lane and main's --external support, which this branch's stale base reverted). The open CI signing decision (cert-in-CI vs manual cuts) is unchanged and now tracked against the merged lane: until decided, ios-testflight.yml's automatic path stays fail-closed.

PR #5660, opened 2026-06-08
Fix status pills not rendering on workspaces created after launch

### LINKED ISSUE #5659: Status pills don't render on workspaces created after app launch
## Repro

1. Launch cmux (status pills work fine on all pre-existing workspaces)
2. Create a new workspace (e.g. workspace:11 "Benchmarks")
3. Set a status pill on that workspace: `cmux set-status test_probe "VISIBLE?" --icon=star.fill --color=#FF0000 --priority=200 --workspace workspace:11`
4. `cmux list-status --workspace workspace:11` confirms the entry exists in the model
5. The pill does **not** render in the sidebar

Control test: the same `set-status` command targeting a pre-existing workspace (e.g. workspace:3 or workspace:6) renders immediately.

## Evidence

- `list-status` shows the correct entries on workspace:11 (data is in the `Tab.statusEntries` model)
- Switching away from workspace:11 and back doesn't trigger rendering
- Setting a completely different value (to force a Combine publisher emission) doesn't help
- The issue affects ALL status keys on that workspace — including cmux's own native `copilot=Running` pill set by the copilot agent hooks
- Other workspaces in the same window render pills fine

## Likely cause

`TabItemView.body` renders metadata entries from a cached `workspaceSnapshotStorage` (`@State`) that's rebuilt by `refreshWorkspaceSnapshot()` when `tab.sidebarObservationPublisher` fires. For a newly-created workspace, the `TabItemView`'s `.onReceive(tab.sidebarObservationPublisher)` subscription may not be delivering emissions to the snapshot refresh path. The data mutation lands on the `Tab` model (confirmed by `list-status`), but the SwiftUI view never rebuilds its snapshot.

Possibilities:
- The `lazy var sidebarObservationPublisher` materializes after the first status mutation, and `.dropFirst()` on `$statusEntries` drops the actual update instead of the initial empty value
- The `ForEach` identity (`\.id`) for the new workspace row doesn't change after the first render, so `.equatable()` prevents body re-evaluation
- The snapshot cache key (`PresentationKey`) doesn't include status-content-sensitive fields, so the cache hit at line 14068-14070 returns the stale empty snapshot

## Environment

- macOS 26, cmux release build
- Single window, 10+ workspaces
- workspace:11 created via sidebar "+" button during this session

obvioussean:
Status pills set on a workspace created after app launch are stored in the `Tab` model but never render in the sidebar — while pre-existing workspaces render immediately. `list-status` confirms the data; the UI shows nothing.

### Why

`.onAppear` fires synchronously during view insertion, before `.onReceive` subscriptions are live. Sidebar data set by socket commands (e.g. agent hook `set_status`) between `.onAppear` and subscription goes unobserved: the observation publishers use `.dropFirst()` to skip the current-value replay, making them change-only — so a mutation that lands before the subscriber connects isn't replayed, and the `.onAppear` snapshot captured the pre-mutation (empty) state. Pre-existing workspaces work because their `.onReceive` subscriptions were established before any status was set.

### What didn't work: removing `.dropFirst()` globally

The initial approach removed `.dropFirst()` from `sidebarObservationSignal` and the `$panels` chain, turning the publishers from change-only into replay-on-subscribe. This fixed the immediate bug but introduced three problems caught by adversarial review (GPT-5.5 + GPT-5.4, independently):

1. **Extension sidebar refresh loop** — `extensionSidebarDebouncedObservationPublisher` is a computed function that creates a new `MergeMany` per body evaluation. Subscription-time replay → handler increments `@State` token → body re-eval → new subscription → replay → self-sustaining churn.
2. **Immediate publisher fires ~7 un-debounced replays per row** on the typing-latency-sensitive path, each calling `makeWorkspaceSnapshot()`.
3. **Test breakage** — existing tests assume change-only semantics (`publishCount == 0` after subscribe + heartbeat mutation).

### What works: targeted deferred refresh

Add a one-shot `.task` to `TabItemView` that calls `refreshWorkspaceSnapshot()` — running asynchronously on the next main-actor tick after the view's modifier chain (including `.onReceive` subscriptions) is fully established. It re-reads tab state at that point, catching any mutations that landed in the `.onAppear`-to-subscription gap. `SidebarWorkspaceSnapshotRefreshPolicy.decision` no-ops when nothing has changed, so the cost for pre-existing workspaces is one cheap snapshot comparison per row appearance.

The observation publishers keep their `.dropFirst()` and change-only contract. No other consumers are affected.

### Validation

Couldn't build locally (SPM resolution failures in the worktree) — CI will validate. The change is +11 lines in one file (a `.task` modifier on `TabItemView`).

Closes #5659

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/Sources/ContentView.swift
+++ b/Sources/ContentView.swift
@@ -15829,6 +15829,17 @@ struct TabItemView: View, Equatable {
         .onAppear {
             refreshWorkspaceSnapshot(force: true)
         }
+        // Re-read current state after the modifier chain is fully established.
+        // .onAppear fires synchronously during view insertion, before .onReceive
+        // subscriptions are live. Sidebar data (status pills, metadata, etc.) set
+        // by socket commands between .onAppear and subscription can be missed by
+        // the change-only observation publishers. This one-shot async refresh runs
+        // on the next main-actor tick — after .onReceive is subscribed — so it
+        // picks up any state that .onAppear saw as empty but that was mutated in
+        // the gap. The snapshot comparison no-ops when nothing changed.
+        .task {
+            refreshWorkspaceSnapshot()
+        }
         .task(id: finderDirectoryCacheKey) {
             let cache = await WorkspaceFinderDirectoryResolver.cache(for: finderDirectoryCacheKey)
             guard !Task.isCancelled else { return }


obvioussean:
Closing in favor of #5662 which fixes this at the publisher level with `.prepend(())` — cleaner than my consumer-side `.task` approach, benefits all subscribers, and includes a regression test.

PR #5690, opened 2026-06-09
Add cmux tmux attach command

lawrencecchen:
## Summary
- Add a `cmux tmux` CLI command that opens a cmux terminal backed by a tmux control-mode IO bridge.
- Route tmux pane output through normal terminal stdout so cmux/Ghostty own native scrollback and Cmd-F search.
- Support attach-or-create, strict attach, strict new, workspace/pane/surface/split/cwd/focus routing, and localized help.
- Add CLI socket regression coverage for surface creation and split creation.

## Implementation
- `cmux tmux ...` creates a terminal whose startup command is `cmux __tmux-io-backend ...`.
- The hidden backend puts stdin in raw mode, starts `tmux -C`, decodes `%output` and `%extended-output` into terminal bytes on stdout, forwards input bytes back with `send-keys -H`, and polls terminal size into `refresh-client -C`.

## Testing
- `git diff --check`
- `python3 -m json.tool Resources/Localizable.xcstrings >/tmp/localizable-check.json`
- `xcodebuild test -project cmux.xcodeproj -scheme cmux-unit -configuration Debug -destination 'platform=macOS' -derivedDataPath /tmp/cmux-tmuxmanual -only-testing:cmuxTests/CMUXCLIErrorOutputRegressionTests/testTmuxOpenCreatesSurfaceWithAttachOrCreateCommand -only-testing:cmuxTests/CMUXCLIErrorOutputRegressionTests/testTmuxAttachSplitCreatesPaneWithStrictAttachCommand`
- Manual backend smoke: piped `echo hi` through `cmux __tmux-io-backend` and verified decoded tmux pane output.
- `./scripts/reload-cloud.sh --tag tmuxio`

## Dogfood
- Tagged build: http://127.0.0.1:17320/tmuxio
- Check `cmux tmux work` opens a new terminal backed by the tmux control-mode bridge and attaches or creates session `work`.
- Check Cmd-F finds text printed inside that tmux session.
- Check native scrollback contains output from commands run inside tmux.
- Check `cmux tmux attach -s work --split right` opens a right split attached to the same session.

## Task
Build tmux attach mode where we can attach a terminal to a tmux session; first interface should just be a cmux tmux command that lets me create new or attach to existing tmux session, implemented with a custom manual IO backend so native find and scrollback work.

## Summary by CodeRabbit

* **New Features**
  * Added a `cmux tmux` subcommand with `open`, `attach`, and `new` actions for working with tmux sessions from the CLI.
  * Added options for session selection, working directory, workspace, pane or surface targeting, split direction, window selection, and focus behavior.

* **Documentation**
  * Added localized CLI help text for the new tmux commands in English and Japanese.

* **Tests**
  * Added regression coverage for tmux session creation and attachment flows.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -2773,7 +2773,7 @@ struct CMUXCLI {
         "--order", "--out", "--pane", "--panel", "--path", "--profile", "--property",
         "--provider", "--relay-port", "--script", "--selector", "--session",
         "--shell", "--source", "--subtitle", "--surface", "--tab", "--target-pane",
-        "--text", "--timeout", "--timeout-ms", "--title", "--transcript",
+        "--text", "--timeout", "--timeout-ms", "--title", "--tmux", "--transcript",
         "--turn", "--type", "--url", "--url-contains", "--value", "--window",
         "--workspace", "--checkpoint", "--checkpoint-id",
     ]
@@ -2996,6 +2996,7 @@ struct CMUXCLI {
         if command == "docs" { try runDocsCommand(commandArgs: commandArgs, jsonOutput: jsonOutput); return }
         if command == "welcome" { printWelcome(); return }
         if command == "diff-viewer-server" { try runDiffViewerServerCommand(commandArgs: commandArgs); return }
+        if command == "__tmux-io-backend" { try runTmuxIOBackendCommand(commandArgs: commandArgs); return }
 
         if command == "settings",
            settingsCommandDoesNotNeedSocket(commandArgs) {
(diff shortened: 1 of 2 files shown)


robertnisipeanu:
Hi @lawrencecchen
I'm also working on a similar feature, but for remote hosts (over SSH): https://github.com/manaflow-ai/cmux/pull/5553
However, in there I replicate the entire tmux server running, mapping tmux sessions to local workspaces, windows to tabs, panes to new type of panes that can render per tab (since originally cmux uses tabs in panes; while tmux uses panes in tabs and there was a discrepancy).

I feel like it's worth mentioning it since they're similar commands, but with a different target host (local vs remote). From UX perspective it would be nice if "ssh-tmux" and "tmux" feel similar (and has similar capabilities) to the user.

Issue #5757, opened 2026-06-10
Beachball under heavy agent load: one-NSThread-per-socket-connection + synchronous v2MainSync serializes all clients on the main thread (1030 threads, 446 client handlers observed)


lawrencecchen:
## Summary

Under a heavy local agent fleet, the control socket's **one-NSThread-per-connection** model plus **synchronous main-actor command execution** (`v2MainSync` → `DispatchQueue.main.sync`) serializes all clients' main-actor commands on the main thread and explodes the thread count. Live `sample` of a beachballing stable build (v0.64.14) shows **1030 live threads, 446 of them client-handler NSThreads**.

## Profile evidence

`sample` of stable pid 670 (this machine: 487 `codex` + 19 `claude` agent processes all talking to one cmux socket):

- **446** client-handler NSThreads from `TerminalController.spawnClientHandler(socket:peerPid:)` (one `Thread.detachNewThread` per accepted connection, no pool). Total process threads: **1030**.
- Main thread, when not idle in `mach_msg`, is ~100% in `v2MainSync` → `MainActor.assumeIsolated` → `processV2Command`. Largest single main-actor command: `v2SurfaceReadText` (`read-text`) formatting full terminal scrollback (~135/289 of the main thread's busy samples).
- Two client threads parked in `FeedCoordinator.ingestBlocking` on a `DispatchSemaphore` (blocking a worker thread per in-flight `feed.push` with a wait timeout).

## Root cause

`Sources/TerminalController.swift`:

- `spawnClientHandler` detaches a fresh `NSThread` per socket connection; `handleClient` blocks that thread in a read loop for the connection lifetime. No thread pool / bound. Each thread also carries a stack, contributing to scheduler pressure and memory.
- `v2MainSync` runs every non-`socketWorker` command via `DispatchQueue.main.sync` + `MainActor.assumeIsolated`. With N agents each issuing `read-text` / workspace / surface commands, all of it serializes on the main thread. A single expensive command (e.g. `read-text` over large scrollback via Ghostty FFI) stalls the run loop → beachball.
- `system.top` polling (see https://github.com/manaflow-ai/cmux/issues/5756) compounds the main-actor hop rate.

This is the cmuxterm-hq control-center workload (hundreds of agents on one host), but any user running many agent tabs hits the same serialization.

## Directions (not yet decided)

1. Bound client handling with a worker pool / async accept loop instead of one detached thread per connection.
[...]

PR #5759, opened 2026-06-10
Cache negative cmux-scope results so system.top stops re-probing every process per poll

### LINKED ISSUE #5756: system.top re-probes every non-cmux process with 3 sysctls per poll (negative scope results never cached) — top CPU consumer under heavy agent load
## Summary

`system.top` re-probes every non-cmux process with 3 `sysctl` syscalls on **every** poll, because the scope-attribution cache only stores positive results. Under a heavy local agent fleet (here: ~500 processes, 487 of them `codex`), this is the single largest on-CPU consumer in a live `sample` of a beachballing stable build (v0.64.14).

## Profile evidence

`sample` of the running stable app (pid 670, ~170% CPU) shows a socket-worker thread spending ~770 samples in `v2SystemTop`, of which ~400 are in `CmuxTopProcessSnapshot.allProcesses` → `cachedCMUXScope` → `cmuxScope(for:expectedCacheKey:)` → `kinfoProc` → `sysctl`:

```
465 TerminalController.v2SystemTop          (on-CPU)
400 CmuxTopProcessSnapshot.allProcesses
392 CmuxTopProcessSnapshot.cmuxScope
353 CmuxTopProcessSnapshot.kinfoProc        (sysctl KERN_PROC_PID)
```

`v2SystemTop` / `v2SystemMemory` are polled repeatedly, so this scan runs on a loop.

## Root cause

`Sources/CmuxTopSnapshotScopeCache.swift` `cachedCMUXScope(for:cacheKey:)`:

```swift
guard let scope = cmuxScope(for: pid, expectedCacheKey: cacheKey) else {
    return nil          // <-- nil result is NOT cached
}
cmuxTopScopeCache.withLock { cache in
    cache[cacheKey] = CmuxTopProcessScopeCacheValue(scope: scope)
}
```

`cmuxScope(for:)` does two `sysctl(KERN_PROC_PID)` calls + one `sysctl(KERN_PROCARGS2)`. For any process **without** a cmux scope (the vast majority: system daemons, helpers, and agent processes that don't export `CMUX_WORKSPACE_ID`/`CMUX_SURFACE_ID` or run `cmux hooks ... monitor`), it returns nil and is never cached, so every subsequent poll re-runs all 3 syscalls for that process. The cache is keyed by `(pid, startSeconds, startMicroseconds)` and pruned to active PIDs each capture, so positive entries persist — only negatives leak.

A process's cmux scope is derived from its environment and argv, both fixed at `exec`, so it never changes during the process lifetime. Caching the negative ("this PID has no cmux scope") is correct.

## Fix

Cache negative results too: distinguish "key present, scope == nil" from "key absent (needs compute)". Store `CmuxTopProcessScope?` in the cache value and only call `cmuxScope(for:)` when the key is absent. Reduces the per-poll cost from O(non-cmux processes × 3 syscalls) to O(new processes since last poll).

## Impact

Eliminates the dominant `system.top` CPU cost under many-agent load (the cmuxterm-hq control-center workload), reducing both worker-thread CPU and the cadence of main-actor hops from `v2SystemTop`'s base-payload `v2MainSync`.

lawrencecchen:
## Summary
`system.top` re-probed **every** non-cmux process with up to 3 `sysctl` calls on every poll, because `CmuxTopProcessSnapshot.cachedCMUXScope` only cached positive scope lookups. Under a heavy local agent fleet this was the single largest on-CPU consumer in a live `sample` of a beachballing stable build (v0.64.14) — hundreds of processes × 3 syscalls × poll rate.

This caches negative results too, with the nuance needed to stay correct:

- **Negative TTL (15s), not lifetime.** The cache key is `(pid, process start time)`, which an `exec` does **not** change, while scope is derived from argv/environment, which an exec *can* change. A lifetime negative cache would permanently mis-attribute a process first sampled in its fork-before-exec window, or one that execs into a `cmux hooks … monitor` later (e.g. a launchd-parented monitor). The TTL bounds attribution latency to 15s while still collapsing the steady-state per-poll storm. Positive scopes stay cached indefinitely (inherited-env / stable-argv scope doesn't disappear).
- **Permanent procargs denial is cached.** `KERN_PROCARGS2` fails permanently for other-user / protected processes (not just on exit races). The probe now distinguishes a permanent denial (process still alive → cache nil) from a transient exit/pid-reuse race (stays uncached). Without this, every root/other-user daemon would remain a permanent cache miss and keep the sysctl fan-out unbounded.
- **Concurrency-safe write.** `capture()` runs concurrently (async task-manager sampling and sync `system.top` socket handling) and the probe runs outside the cache lock, so the post-probe write refuses to overwrite a positive scope a concurrent capture already discovered, and returns that positive.

Drops steady-state per-poll cost from O(non-cmux processes × 3 syscalls) to roughly O(non-cmux processes × 3 syscalls ÷ (TTL ÷ poll interval)).

## Testing
New wired Swift Testing suite `cmuxTests/CmuxTopSnapshotScopeCacheTests` (`.serialized`, probe-injection + monotonic-clock seam): negatives cached within the TTL (probed once per window, not per poll), positives cached indefinitely, transient failures retried, an expired negative re-probed after exec, and a stale negative never clobbering a concurrently discovered positive. Two-commit red/green: commit 1 adds the seam + tests with the buggy behavior; commit 2 enables the TTL negative caching.

Three Codex autoreview rounds drove the design: a lifetime negative cache (exec problem) → 15s TTL; an unconditional write (concurrency clobber) → conditional write; transient-only failure classification (procargs denial leak) → permanent-denial caching.

## Notes
The only `workflow-guard-tests` failure is the pre-existing `AgentLaunchSanitizer.swift` file-length-budget overflow on `main` (untouched here; `main`'s own CI already fails this step). This PR's additions stay within budget in a separate wired test file.

## Issues
- Closes https://github.com/manaflow-ai/cmux/issues/5756
- Related (architectural umbrella): https://github.com/manaflow-ai/cmux/issues/5757

### DIFF
--- a/Sources/CmuxTopProcessEnumeration.swift
+++ b/Sources/CmuxTopProcessEnumeration.swift
@@ -89,7 +89,7 @@ nonisolated extension CmuxTopProcessSnapshot {
         let path = includeProcessDetails ? processPath(pid: pid) : nil
         let rawTTY = Int64(bsdInfo.e_tdev)
         let ttyDevice = rawTTY > 0 ? rawTTY : nil
-        let cmuxScope = cachedCMUXScope(for: pid, cacheKey: cacheKey)
+        let cmuxScope = cachedCMUXScope(for: pid, cacheKey: cacheKey, nowNanoseconds: sampledAtNanoseconds)
         let rawProcessGroupID = Int(bsdInfo.pbi_pgid)
         let processGroupID = rawProcessGroupID > 0 ? rawProcessGroupID : nil
         let rawTerminalProcessGroupID = Int(bsdInfo.e_tpgid)
--- a/Sources/CmuxTopSnapshot.swift
+++ b/Sources/CmuxTopSnapshot.swift
@@ -117,7 +117,7 @@ nonisolated struct CmuxTopProcessInfo: Sendable {
     }
 }
 
-nonisolated struct CmuxTopProcessScope: Sendable {
+nonisolated struct CmuxTopProcessScope: Sendable, Equatable {
     let workspaceID: UUID?
     let surfaceID: UUID?
     let attributionReason: String
--- a/Sources/CmuxTopSnapshotScopeCache.swift
+++ b/Sources/CmuxTopSnapshotScopeCache.swift
@@ -84,19 +84,27 @@ nonisolated extension CmuxTopProcessSnapshot {
 
         switch probe(pid, cacheKey) {
         case .resolved(let scope):
-            guard let scope else {
-                // BUG (issue 5756): a resolved "no cmux scope" is not cached, so
-                // every poll re-probes every non-cmux process. Fixed in the next
-                // commit by caching the negative result under a TTL.
-                return nil
-            }
-            cmuxTopScopeCache.withLock { cache in
+            // Cache the resolved result. Positive scopes are kept indefinitely;
+            // negative scopes are kept only until the TTL elapses so a later exec
+            // is eventually attributed. The key is pruned to live pids each
+            // capture, so a recycled pid gets a fresh key.
+            //
+            // capture() runs concurrently (async task-manager sampling and sync
+            // system.top socket handling), and the probe happened outside the
+            // lock. Never let a stale negative from an older capture clobber a
+            // positive scope a newer capture already discovered for the same
+            // process: if we probed nil but a positive entry now exists, keep and
+            // return it.
+            return cmuxTopScopeCache.withLock { cache -> CmuxTopProcessScope? in
+                if scope == nil, let existing = cache[cacheKey], let existingScope = existing.scope {
+                    return existingScope
+                }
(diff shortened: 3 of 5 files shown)


lawrencecchen:
Note on `workflow-guard-tests`: the Swift file-length-budget step fails on this PR, but the only offender is `Packages/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentLaunchSanitizer.swift` (566 lines, untracked, ≥500 threshold), which is **pre-existing on `main`** and not touched by this change — recent `main` CI runs already fail this same step. This PR's own additions stay within budget: the new regression tests live in a separate, wired `cmuxTests/CmuxTopSnapshotScopeCacheTests.swift` (96 lines) instead of growing the already-capped `CmuxTopSnapshotScopeTests.swift`. Compile + lag (`tests-build-and-lag`) and the unit `tests` job exercise the new red/green coverage.

Issue #5796, opened 2026-06-10
fish: no shell integration ships, so claude bypasses the wrapper — session restore degrades to `bash --resume <id>` (invalid option) and loses cwd


STRML:
### Bug description

On a system where the user's shell is **fish**, quitting and reopening cmux with "Resume Agent Sessions on Reopen" enabled relaunches Claude panes as:

```
bash: --resume: invalid option
Usage:  bash [GNU long option] [option] ...
```

…followed by a fresh fish login shell in the **wrong working directory**. The agent session is not resumed.

### Environment

- cmux 0.64.14 (latest release at time of filing), macOS (Darwin 25.4.0)
- User login shell: `/opt/homebrew/bin/fish`
- Real claude binary: `~/.local/bin/claude` (native install); `claudeCodeCustomClaudePath` set correctly in Settings → Automation

### Root cause

`Resources/shell-integration/` ships **bash and zsh integration only** — there is no fish integration:

```
$ ls /Applications/cmux.app/Contents/Resources/shell-integration/
.zlogin .zprofile .zshenv .zshrc
cmux-bash-bootstrap.bash  cmux-bash-integration.bash
cmux-zsh-integration.zsh  ghostty-integration.zsh
```

So under fish, the `claude` wrapper shim (`Resources/bin/claude`) never gets onto PATH and is never invoked. Verified on a live cmux-launched Claude process: it has the full pane environment (`CMUX_SURFACE_ID`, `CMUX_SOCKET_PATH`, `CMUX_SHELL_INTEGRATION_DIR`, …) but **none** of the wrapper's exports (`CMUX_AGENT_LAUNCH_KIND/EXECUTABLE/ARGV_B64`, `CMUX_CLAUDE_PID`), and `Resources/bin` is absent from its PATH. Hooks are therefore never injected either (same family as #5427 / #5639, but at initial launch rather than on resume).

Without an `AgentLaunchCommandSnapshot` from the wrapper, the resume binding is built from the pane's bootstrap process instead. cmux spawns panes via `login … /bin/bash --noprofile --norc -c 'exec -l <user-shell>'` (see the ancestry dump in #4858), so `commandParts()` in `Sources/RestorableAgentSession.swift` resolves the executable to `bash`, `AgentLaunchSanitizer.preservedArguments(kind: "claude")` keeps the `-c` (a valid claude flag) and drops the rest, and `resumeWithOption` produces:

```
'bash' '--resume' '<session-id>' '-c'
```

[...]

PR #5848, opened 2026-06-10
fix: never use a shell bootstrap executable in a resume binding

STRML:
**Status:** updated onto `main` (`78f4c82504`) as merge commit `e8098b4d7c`. No conflicts.

**Reviewer's guide**
- `Packages/macOS/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentResumeArgv.swift`: the shared seam plus the codex branch.
- Tests: `AgentResumeArgvTests.swift`.
- Worth knowing for review: the app-target duplicate of this builder, `Sources/RestorableAgentSession.swift`, still has the unguarded shape. That path is not in this diff, and it is the one that most plausibly produced the literal `'bash' '--resume'` string in #5796.

---

## Summary

Hardens resume-binding construction so a shell can never be replayed as the agent executable. When a captured launch snapshot resolves to a shell or login bootstrap (`sh`, `bash`, `zsh`, `fish`, `csh`, `tcsh`, `ksh`, `dash`, `login`, including dash-prefixed login `argv[0]` forms like `-fish`), `AgentResumeArgv.commandParts()` now falls back to the kind's known executable and drops the captured argv — those arguments belong to the shell, not the agent.

Fixes the defensive half of #5796: on fish (no shell integration before #5678), the snapshot degraded to the pane's bootstrap (`/bin/bash --noprofile --norc -c 'exec -l fish'`) and every Claude pane persisted a `'bash' '--resume' '<id>' '-c'` binding — a guaranteed `bash: --resume: invalid option` on reopen, which also dropped the binding's cd prefix and left the fallback shell in the wrong directory. #5678 closed the fish integration gap; this PR makes the failure mode impossible for any future snapshot path that captures a shell (agents launched via wrapper scripts, future integration gaps).

## Why now

Hit in the wild on 0.64.14 + fish (full forensics in #5796). The guard sits in the shared `commandParts` seam, so it covers `builtInKind` (all built-in kinds) and `launcherResolution` (`claude-teams`/`codex-teams`/`omo`) in one place.

## Commits (two-commit regression policy)

1. `test:` failing regression test only — 4 parametrized cases (claude with the real-world bash bootstrap argv, codex/zsh, grok/`-fish` login argv[0], opencode/`login`). Red without the fix.
2. `fix:` the guard. Full `CMUXAgentLaunch` suite green: 80 tests, 11 suites.

## Not touched

- The app-target duplicate `commandParts` in `Sources/RestorableAgentSession.swift` (fork paths + antigravity resume). Same theoretical flaw, but the claude fork path already emits a literal `claude`, and I kept this PR to the pure package layer where the live bug path runs and tests execute in isolation. Happy to follow up there if you want the same guard applied.
- The stale-binding shadowing described in #5796 (a poisoned `resumeBinding` surviving relaunch-then-quit and shadowing a fresh `terminal.agent` snapshot) — that's a lifecycle question, flagged separately in the issue.

## Test plan

- [x] Commit 1 alone: new test fails (4/4 cases red, verified locally)
- [x] Commit 2: `swift test --package-path Packages/CMUXAgentLaunch` — 80 tests, 11 suites, all green
- [x] No behavior change for legitimate snapshots: existing `executablePathOverridesFallback` and all other suites untouched and passing

### DIFF
--- a/Packages/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentResumeArgv.swift
+++ b/Packages/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentResumeArgv.swift
@@ -283,12 +283,37 @@ public struct AgentResumeArgv: Sendable, Equatable {
         return [parts.executable, option, sessionId] + preserved
     }
 
+    /// Executable names that are shells or login bootstraps, never an agent binary.
+    ///
+    /// A launch snapshot can degrade to the pane's bootstrap process when no agent
+    /// launch metadata was captured (e.g. the agent was started without the shell
+    /// integration's wrapper). cmux spawns panes via
+    /// `login … /bin/bash --noprofile --norc -c 'exec -l <shell>'`, so such a snapshot
+    /// replays as `bash --resume <id>` — a guaranteed parse error that also drops the
+    /// binding's cd prefix when the child shell exits.
+    /// https://github.com/manaflow-ai/cmux/issues/5796
+    private static let shellBootstrapExecutableNames: Set<String> = [
+        "sh", "bash", "zsh", "fish", "csh", "tcsh", "ksh", "dash", "login",
+    ]
+
+    private func isShellBootstrapExecutable(_ executable: String) -> Bool {
+        var name = URL(fileURLWithPath: executable).lastPathComponent
+        // Login shells rewrite argv[0] with a leading dash ("-bash", "-fish").
+        if name.hasPrefix("-") { name.removeFirst() }
+        return Self.shellBootstrapExecutableNames.contains(name)
+    }
+
     private func commandParts(
         executablePath: String?,
         arguments: [String],
         fallbackExecutable: String
     ) -> (executable: String, tail: [String]) {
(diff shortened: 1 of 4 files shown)


STRML:
Rebased onto `main` (78f4c82504), no conflicts. Reviewer's guide at the top of the description. One note for review: the app-target duplicate of this builder (`Sources/RestorableAgentSession.swift`) still has the unguarded shape; it is not in this diff but is the path most likely behind the literal `'bash' '--resume'` in #5796. Only red check is CLA. @lawrencecchen

STRML:
recheck

STRML:
Rebased onto `main` (`0be8ab34ad`, 2 commits, linear). Force-pushed `fix-shell-executable-resume-binding` @ `ce2264638eb`. Siblings: #12009 #11483 #10856.

STRML:
Rebased onto `main` (623e5bcbb80), 2 commits, linear. Force-pushed fix-shell-executable-resume-binding @ 00f1a8bc473 (supersedes the 0be8ab34ad stack).

STRML:

STRML:

PR #5869, opened 2026-06-11
iOS render-grid liveness watchdog: probe before teardown, consume during subscribe ack (fixes the 10.5s false-fire replay loop)

lawrencecchen:
Fixes the render-grid liveness watchdog false-fire found during the 2026-06-10 Release-sim bisect: the phone logged `render-grid stream silent for 10499ms, re-subscribing` every ~10.5s forever, plus `subscribe failed reason=start: requestTimedOut`, while the Mac demonstrably kept delivering (`mobile.emit -> connection delivered=true`). Every false fire made the Mac re-replay the full grid, a constant repaint every 10.5s (plausibly the reported "weird cursor paints") plus wasted battery and bandwidth.

## Root cause

Two decouplings between the liveness recorder and reality:

1. **Silence is the normal state of a healthy idle terminal.** The Mac dedupes render-grid emits by row signature and stateSeq (cursor blink changes neither), so an idle shell pushes zero events for arbitrarily long. The watchdog treated wall-clock silence alone as death, so it tore down and full-grid-replayed every healthy idle subscription once per threshold window, forever. The original watchdog commit (`e3f418560`) assumed an actively-streaming connection and never accounted for idle.
2. **The liveness clock was stamped only inside the listener's `for await` consumer loop**, which did not start until `resolveTerminalOutputTransport` and the `mobile.events.subscribe` ack both completed. Events delivered during that establishment window piled up unconsumed in the subscription stream's buffer, invisible to the clock. Each watchdog fire then cancelled its own in-flight start subscribe; `MobileCoreRPCSession.cancelPendingRequest` surfaces that cancellation as `requestTimedOut` (the real wire timeout is 30s, so the 10.5s cadence of that log was the watchdog cancelling itself, not the network), and the dying generation called `markMacConnectionUnavailable()` underneath the fresh one.

## Fix

- **One liveness ownership point.** `recordTerminalEventStreamLiveness()` is stamped by every consumed envelope, by a successful probe, and (as the generation reset) when a watchdog generation is armed. The watchdog reads the same record; `listenerID` guards every transition; `stopRenderGridLivenessWatchdog` resets cleanly.
- **Silence is a suspicion, not a verdict.** A threshold crossing now runs a bounded idempotent `mobile.events.subscribe` probe (same `stream_id`, current topics) instead of tearing down. A completed round-trip proves the transport the events ride on is alive AND that the registration is installed. Only a failed probe runs the existing teardown + re-subscribe + replay recovery, so the original ~85s silent-death case still heals (probe deadline default 3s, runtime-injectable, bounds the whole attempt including pre-wire token work).
- **Repaired registrations catch up.** The Mac's subscribe ack now reports `already_subscribed`. When the probe reinstalls a registration the host had lost (`already_subscribed: false`), delta continuity is broken, so the phone replays every mounted surface and re-fetches the workspace list, without restarting the intact phone-side listener. Older Macs omit the field; their registrations cannot outlive the connection, so `nil` is treated as already-active.
- **The start subscribe ack runs concurrently with consumption** (`beginTerminalEventSubscriptionStart`). The ack is a server-side enable handshake, not a delivery precondition, so the consumer loop starts immediately and the clock stays coupled to actual event arrival. Ack success/failure is acted on only while the generation is current, so a superseded ack can no longer mark the connection unavailable, and a cancelled ack logs as cancelled instead of a fake wire timeout.
- **Probe slot ownership token** so a superseded probe completing late cannot clear or clobber a newer generation's single-flight slot.

## Red/green

Commit 1 (https://github.com/manaflow-ai/cmux/commit/4c57e66ba) adds the failing tests plus a DEBUG-only seam to run one watchdog evaluation deterministically against the injected clock; the scripted host drives the real `MobileCoreRPCSession` transport and the real consumer path. `renderGridEventsArrivingDuringStartSubscribeAreConsumed` and the healthy-idle watchdog test fail against pre-fix code (the false fire reproduces 1:1: `sync.liveness re-subscribe silentMs=10000` plus the replay storm); `watchdogStillResubscribesGenuinelyDeadStream` pins the original recovery so the fix cannot regress it. The probe redesign retargeted the idle test's teardown signals to listener restarts and replay traffic (the probe itself legitimately re-sends `mobile.events.subscribe`); both liveness tests were re-verified red against the commit-1 sources after the amendment.

`swift test --package-path Packages/CmuxMobileShell`: 59 tests green. `swift test --package-path Packages/CmuxMobileRPC`: 23 green. Build-verified Debug and Release for arm64 simulator (`/tmp/cmux-watchdog`) and the macOS app (`/tmp/cmux-watchdog-mac`, host-side ack change).

## Residual risk

A Mac-side producer wedge with the registration intact (demand tracker desync deeper than a re-subscribe replace can re-arm) is not phone-detectable and is not recovered by this watchdog; the input-driven seq self-heal still covers it during typing. Dead-stream detection now takes threshold + probe deadline (~12s) instead of ~9s.

## Review fixes (autoreview round 1)

- A successful probe now also calls `markMacConnectionHealthy()`: the round-trip is positive proof of the client/host connection, so a transient foreground RPC failure can no longer leave the status UI stuck unavailable on an idle terminal.
- `mobile.events.subscribe`/`unsubscribe` are now non-interactive in `MobileHostRequestActivity` accounting (joining `mobile.host.status` and the replays), so the idle probe cannot keep the Mac permanently "recently active" and starve background work gated on mobile quiet (TabManager git/PR metadata refresh).

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/5869"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783745716&installation_id=133855650&pr_number=5869&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F5869&signature=794449c8729a1e38a6645b04b691b12ac64e67ace93057c3c34de0a2db4b5460"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches mobile sync connection recovery and Mac RPC subscription semantics; behavior is heavily regression-tested but wrong probe/repair logic could still cause stale UI or excess replay traffic.
> 
> **Overview**
> Fixes the iOS render-grid **liveness watchdog** falsely tearing down healthy idle Mac connections (~10.5s full-grid replay loop).
> 
> **Watchdog behavior:** Prolonged silence no longer immediately triggers re-subscribe + replay. It first runs a bounded idempotent `mobile.events.subscribe` probe (configurable via `livenessProbeTimeoutNanoseconds`, default 3s). Only a failed probe runs the existing resync path. A successful probe refreshes the liveness clock and marks the connection healthy; if the host reports **`already_subscribed: false`**, mounted surfaces are replayed and the workspace list is refreshed without restarting the listener.
> 
> **Subscribe handshake:** The start `mobile.events.subscribe` ack runs **concurrently** with the event consumer so buffered events during establishment update liveness. Stream end before that ack completes now marks the connection unavailable instead of reconnect looping. Cancelled subscribe requests log as cancelled, not wire timeouts.
> 
> **Mac host:** Subscribe RPC responses include `already_subscribed`. `mobile.events.subscribe` / `unsubscribe` are treated as non-interactive so idle probes do not block background Mac work.
> 
> **Tests:** New `MobileShellRenderGridLivenessTests` plus scripted host fixtures cover establish-window consumption, healthy idle, lost-registration repair, dead-stream recovery, and pre-ack stream close.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab8f139565c4c055406f209b2c9f25a374f47777. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes the iOS render‑grid liveness watchdog false‑fires by probing before teardown and consuming events while the start‑subscribe ack is in flight. Also handles a stream that ends before the start ack by marking the Mac unavailable, preventing a reconnect loop.

- **Bug Fixes**
  - Silence triggers a bounded, idempotent `mobile.events.subscribe` probe (same `stream_id`/topics) before recovery; only a failed probe re‑subscribes and replays. Deadline via `MobileSyncRuntime.livenessProbeTimeoutNanoseconds` (default 3s).
  - Start‑subscribe ack now runs in parallel with stream consumption; liveness stamps on each consumed envelope and on probe success. Cancelled acks log as cancelled, not `requestTimedOut`.
  - If the event stream ends before the start‑subscribe ack completes, treat it as a failed start: stop and mark the connection unavailable (prevents a reconnecting livelock).
  - Subscribe ack includes `already_subscribed` (decoded as `alreadySubscribed`); when `false`, replay mounted surfaces and refresh the workspace list without restarting the listener. `nil` is treated as already active for older hosts.
  - Single‑flight probe with an ownership token; successful probes mark the Mac healthy. `mobile.events.subscribe`/`unsubscribe` are counted as non‑interactive host activity. Probe deadline uses a one‑shot `DispatchSourceTimer`.

- **Tests**
  - Added regression coverage for establish‑window consumption, healthy idle (no teardown), lost‑registration repair (replay + workspace refresh), dead‑stream recovery, and the “stream ends before start ack” case (converges to unavailable).

<sup>Written for commit ab8f139565c4c055406f209b2c9f25a374f47777. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/5869?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Subscription responses now indicate whether a subscription was already present.
  * Runtime exposes a configurable liveness probe timeout with a sensible default.

* **Bug Fixes**
  * Reduced false liveness timeouts by decoupling subscription acknowledgments from event consumption.
  * Host-side subscription loss now reinstates subscriptions, replays mounted terminal surfaces, and refreshes workspace lists without restarting listeners.
  * Watchdog uses bounded, cancellable probes to verify stream health and avoid late-state mutation.

* **Tests**
  * Added comprehensive tests and test-support utilities for render-grid liveness and recovery scenarios.


lawrencecchen:
Addressed the Greptile P2: the probe deadline now uses a one-shot DispatchSourceTimer (the same sanctioned primitive as the watchdog tick) instead of Task.sleep, cancellation wired both directions. 59 package tests green.

lawrencecchen:
Merge-train adjudication of the two open review findings, deferred as follow-ups rather than blockers:

1. CodeRabbit: start-ack collapses the subscription ack to isSubscribed and skips the alreadySubscribed == false replay/repair. For reason=start a fresh install is the expected outcome of a brand-new listener generation, and the resync path replays the requested surfaces unconditionally right after refreshTerminalEventSubscription; the input_seq_wait refresh escalates to a full restart+replay via the pending-seq tracker when a surface stays behind, so missed deltas self-heal. Converging the start path onto the probe path is the right refactor and belongs to the liveness-subsystem extraction follow-up.

2. Codex: a stale start-task timeout can stop a listener whose events already flow. Real but narrow edge (start ack lost while events ride the same healthy connection). A correct guard needs a saw-event-this-generation flag because arming the watchdog stamps the liveness clock, which is new state and new test surface; bolting it onto this stabilization PR risks more than the edge costs. Same follow-up as (1).

lawrencecchen:
Update to the adjudication above: the start-path concern family was partially RIGHT, with a different mechanism than either bot described. The ipad run failed macConnectionStatusMarksUnavailableWhenEventStreamCloses deterministically: when the transport drops before the start ack lands, the stream-end restart supersedes the listener generation, the parked ack's failure verdict gets dropped by its generation guard, and recovery livelocks in reconnecting (iphone passed only because the ack failure sometimes wins the race). Fixed in two commits, red test (98dbf2c53) then fix (ab8f13956): a stream that ends before its handshake completes is treated as a failed start (stop + unavailable) instead of restarting. Deferral of the two follow-ups stands; the convergence refactor now has a pinned regression test.

Issue #5641, opened 2026-06-08
CI: `tests` job masks unexpected unit-test failures (only checks the last suite's "(0 unexpected)" summary)


lawrencecchen:
The `tests` job can report green even when a unit test fails unexpectedly, which silently defeats the repo's two-commit red→green regression policy.

`.github/workflows/ci.yml` (the "Run unit tests" step) decides pass/fail like this when xcodebuild exits non-zero:

```sh
SUMMARY=$(echo "$OUTPUT" | grep "Executed.*tests.*with.*failures" | tail -1)
if echo "$SUMMARY" | grep -q "(0 unexpected)"; then
  echo "All failures are expected, treating as pass"
else
  echo "Unexpected test failures detected"; exit 1
fi
```

It only inspects the **last** per-suite summary line. xcodebuild prints one "Executed N tests, with M failures (K unexpected)" line per suite, in suite order. If an unexpected failure happens in any suite that is not the last one printed, and the last suite's line reads "(0 unexpected)", the whole job is treated as pass.

Observed live: run https://github.com/manaflow-ai/cmux/actions/runs/27116321138 (job `tests`) logged `All failures are expected, treating as pass` at 04:55:52 even though xcodebuild exited non-zero, so a real unexpected failure was masked. (Surfaced while adding a deliberately-red regression test for the codex resume-binding work, https://github.com/manaflow-ai/cmux/issues/5617 — the red test could not be observed failing via this job.)

Impact: a "red test first" commit can show green, so the red→green proof the policy depends on is unreliable for any suite that is not alphabetically/order last.

Possible fix: aggregate across all summary lines (fail if any line shows `([1-9]… unexpected)`), or parse xcodebuild's final overall result / the `.xcresult` instead of `tail -1` of per-suite summaries.

austinywang:
Hit this again on https://github.com/manaflow-ai/cmux/pull/5923 (red commit for https://github.com/manaflow-ai/cmux/issues/5913), with two additional findings from the job log of https://github.com/manaflow-ai/cmux/actions/runs/27384993177 (job `tests`):

[...]

PR #5912, opened 2026-06-11
Forward notifications to iPhone only when away from the Mac

lawrencecchen:
Forward terminal notifications to the iPhone only when the user is actually away from the Mac. Lawrence: "If you are already on the MacBook you probably don't need a notification on iOS. We should make this an option. The default should be to NOT send iOS notifications if the MacBook is active."

## Behavior

Phone forwarding stays opt-in: the existing "Forward notifications to my iPhone" master toggle is unchanged. This PR adds a forwarding mode under it that refines WHEN forwards happen:

- **Only when away from this Mac** (default): a notification is pushed to the phone only if the Mac is not actively in use at delivery time.
- **Always**: the previous behavior, forward regardless of presence.

The default applies to everyone, including users who already had forwarding enabled before this PR. That is intentional: the new default matches what forwarding should have been, and anyone who wants every push can pick "Always" in the Notifications page.

## Heuristic

The Mac counts as ACTIVE (suppress the phone forward) iff ALL of:

1. The console session belongs to the current user and is UNLOCKED (no login window, no fast-user-switch away, no lock screen).
2. Displays are awake and the screensaver is not running.
3. Last HARDWARE user input was within the last 120 seconds, via `CGEventSource.secondsSinceLastEventType(.hidSystemState, eventType:)`, taking the min across keyboard, mouse-moved, mouse-down (left/right/other), and scroll-wheel events.

Why `.hidSystemState` and not `.combinedSessionState`: the combined session state is refreshed by synthetic posted events, so cmux agents driving the debug socket, accessibility automation, or any event-posting tool would keep the Mac "present" forever. The HID system state tracks hardware input. cmux's own synthetic input (debug-socket `simulate_shortcut`, mobile RPC key events from the phone) is dispatched in-process and never reaches CGEvent counters at all, so driving the Mac from the phone correctly counts as away.

Why lock flips to away instantly: a locked screen, sleeping display, or running screensaver is definitive evidence the user is not watching the Mac, regardless of how recent the last keystroke was. Only the input-recency rule has the 120 s window; the threshold lives in one place, `MacPresenceMonitor.recentHardwareInputThreshold`, and the Settings copy derives from it.

Lock state has no public synchronous macOS API, so the live provider consults two de-facto sources and ORs them: the `CGSSessionScreenIsLocked` key in `CGSessionCopyCurrentDictionary()` (nil dictionary = no WindowServer session = away; `kCGSessionOnConsoleKey` covers fast-user-switch and login window) and a `ScreenLockObserver` tracking the `com.apple.screenIsLocked`/`com.apple.screenIsUnlocked` distributed notifications. If both ever miss a lock, the 120 s idle rule still bounds the suppression window.

Suppressed notifications are NOT retroactively forwarded when the Mac later goes away; the push mirrors "what would buzz the phone right now". Mac-side behavior (notification list, unread accounting, sounds, desktop notification) is completely unchanged; only the phone push is gated.

## Mechanism

- `Sources/MacPresenceMonitor.swift`: one small type with an injected clock and injected signal provider; the heuristic is a pure function over a signals snapshot (`MacPresenceMonitor.verdict(for:)`) returning a reasoned verdict (active / locked / displays asleep / screensaver / no recent input).
- The gate sits inside the existing forward gate, `PhonePushClient.forward(_:)`, decided per notification at delivery time, so the phone never receives a suppressed push. `TerminalNotificationStore.deliverNotificationSideEffects` is untouched.
- Hot-path bounding, settled across review rounds: the read-only send-throttle check runs before presence sampling (storms in away mode probe at most once per workspace+surface per second); the throttle slot is consumed only after the gate passes (a suppressed notification never blocks a forwardable one); suppressed bursts are bounded by `MacPresenceDecisionCache`, which reuses ACTIVE decisions for up to 1 s but NEVER reuses AWAY decisions. Any bounded design is up to 1 s stale on one side of an active/away transition; the staleness is deliberately on the suppression side (documented invariant in `MacPresenceDecisionCache`): a stale active answer can only suppress a push within 1 s of the user leaving a Mac they were provably using, while a stale away answer would push terminal content to the phone with the user sitting at the Mac. `.always` mode skips sampling entirely.
- Mode is stored like its sibling settings (`forwardNotificationsToPhone`, `forwardNotificationsHideContent`): a UserDefaults key (`forwardNotificationsToPhoneMode`) surfaced via `@AppStorage` on the Notifications page, in `Sources/Cloud/PhoneForwardingMode.swift`. The sibling forwarding settings are deliberately not exposed in `~/.config/cmux/settings.json` (keyboard shortcuts only) or `cmux.json` (notification hooks only) and are not in the configuration docs, so the mode follows the same pattern; there is no docs page listing them to update.

## Tests

Two-commit red/green for the gate (commit 1 adds the behavior spec with the gate permissive, commit 2 implements it), plus review-round hardening commits. The CI `tests` job currently stops partway through the alphabetical suite list and judges only the last summary line (https://github.com/manaflow-ai/cmux/issues/5641), so red/green was additionally proven by running `-only-testing:cmuxTests/PhonePushPresenceGateTests` on the AWS M4 Pro builder at the red commit (suppression tests fail) and at head (all pass); evidence in the PR comments.

Covered through the injected seams, deterministic, no sleeps: active Mac suppresses; idle 121 s forwards; locked/display-sleep/screensaver forward instantly despite 1 s-old input; synthetic-input-only forwards; "Always" forwards regardless; idle exactly 120 s counts as active; unknown idle counts as away; lock-source matrix (nil session dictionary, dictionary lock key, observed-notification lock with the dictionary key absent, off-console session); cache reuses active decisions and never reuses away decisions; mode parsing defaults and garbage fall back to only-when-away. Tests use Swift Testing per repo policy.

## Review

`autoreview` (Codex engine) ran to a clean result: "patch is correct, no actionable defects". Findings fixed along the way: throttle-before-sampling ordering, never-stale-away cache, MainActor isolation for `ScreenLockObserver`, dual lock-state sources, Swift Testing conversion, one-type-per-file splits. One Aziz policy-scanner finding is consciously rejected: it flags the nested `Signals`/`Verdict`/`Decision` types inside `MacPresenceMonitor`, but `skills/cmux-architecture/SKILL.md` explicitly allows closely-bound nested types to stay with the parent, and these are the monitor's own vocabulary types.

## Localization audit

New user-facing strings (mode picker label, two options, away explanation with the threshold-derived minutes argument) use `String(localized:defaultValue:)` with explicit `en` and `ja` entries in `Resources/Localizable.xcstrings` (verified by parsing the file: every `notifications.forwardToPhone.*` key has both locales). No bare string literals in the changed views.


lawrencecchen:
Red/green evidence for the gate, executed on the AWS M4 Pro builder with `xcodebuild -scheme cmux-unit -only-testing:cmuxTests/PhonePushPresenceGateTests test` (the hosted CI `tests` job currently stops partway through the alphabetical suite list and judges only the last summary line, see https://github.com/manaflow-ai/cmux/issues/5641, so this class does not execute there).

RED at https://github.com/manaflow-ai/cmux/commit/96db2115e30c34ef9737d77da42a7d949c4abe0a (behavior spec with the gate permissive):

```
Test Case '-[cmuxTests.PhonePushPresenceGateTests testActiveMacSuppressesForwardInOnlyWhenAwayMode]' failed
Test Case '-[cmuxTests.PhonePushPresenceGateTests testIdleExactlyAtThresholdCountsAsActive]' failed
Executed 14 tests, with 2 failures (0 unexpected)
```

Exactly the two suppression assertions fail; the 12 spec tests that match pre-gate behavior pass.

GREEN at https://github.com/manaflow-ai/cmux/commit/a043549ac (branch head, Swift Testing):

```
◇ Suite PhonePushPresenceGateTests started.
✔ Suite PhonePushPresenceGateTests passed after 0.028 seconds.
✔ Test run with 22 tests in 1 suite passed after 0.028 seconds.
** TEST SUCCEEDED **
```

Issue #5189, opened 2026-06-02
Select Workspace 1…9 shortcut can't be rebound — remapping in Settings has no effect (⌘1–9 stays hardcoded)


austinywang:
## Bug

The **Select Workspace 1…9** keyboard shortcut (default `⌘1`–`⌘9`) cannot be rebound. Changing it in **Settings → Keyboard Shortcuts** has no effect: the default `⌘1–9` keeps selecting workspaces regardless of what you set, so the binding is effectively hardcoded. The sibling **Select Surface/Tab 1…9** shortcut (default `⌃1`–`⌃9`) appears to have the same problem.

These numbered shortcuts are special — a single setting stands in for the whole `1…9` range (the stored key is normalized to a `"1"` digit placeholder, `usesNumberedDigitMatching`). The remap path needs explicit exception handling for that range, and something along that path isn't honoring a rebind.

## Steps to reproduce

1. Open **Settings → Keyboard Shortcuts**.
2. Find **Select Workspace 1…9** (defaults to `⌘1–9`).
3. Rebind it to a different modifier+digit (e.g. `⌃1–9`, or `⌥1–9`).
4. Press the newly assigned shortcut, and separately press the old `⌘1–9`.

## Expected

Only the newly bound shortcut selects workspaces by number; the old `⌘1–9` no longer does (unless that's what you bound).

## Actual

The rebind doesn't take effect — `⌘1–9` still selects workspaces (and/or the new binding does nothing). The shortcut behaves as if hardcoded.

## Starting points (verify, don't trust blindly)

- `Sources/KeyboardShortcutSettings.swift` — action `.selectWorkspaceByNumber` ("Select Workspace 1…9", default `StoredShortcut(key: "1", command: true)`) and `.selectSurfaceByNumber`. Both have `usesNumberedDigitMatching == true`. The recorder normalizes a recorded shortcut via `resolvedNumberedDigitShortcut(_:)` (requires a `1…9` digit key, normalizes the key to `"1"`, should preserve modifiers) → `normalizedRecordedShortcutResult` → `normalizedRecordedShortcut`. Check whether a modifier change is actually accepted and persisted here, and whether `conflicts(...)` / `conflictingAction(...)` rejects the new binding.
[...]

PR #5939, opened 2026-06-12
Browser back/forward get cmd+opt+[ / cmd+opt+]; cmd+[ / cmd+] stay focus history in every pane

lawrencecchen:
## Problem

Focus history (https://github.com/manaflow-ai/cmux/commit/2aa7e7428) bound cmd+[ / cmd+] app-wide, and the dispatch chain checks it before browserBack/browserForward, so keyboard page back/forward in browser panes died: both actions defaulted to the same chord and focus history always won.

## Fix

One chord, one meaning, in every pane (dogfood explicitly rejected an interim context-routed design where cmd+[ changed meaning between terminal and browser panes):

- **cmd+[ / cmd+] = focus history everywhere**, including focused browser panes. Unchanged from current behavior outside browsers; now intentional rather than accidental inside them.
- **cmd+opt+[ / cmd+opt+] = browser page back/forward**, new defaults for `browserBack`/`browserForward` (browser panes only). They consume with a beep at the page-history boundary instead of a silent no-op. cmd+opt matches cmux's focus-modifier family and is unclaimed by macOS, Ghostty, and cmux.
- **Menu truthfulness:** `browserBack`/`browserForward` menu equivalents stay suppressed (replacing the old equality-based hack with an unconditional rule plus rationale): the View menu items no-op without a focused browser, and a static equivalent would bypass the focus gate (same hazard class as https://github.com/manaflow-ai/cmux/issues/5189). The History menu keeps truthful cmd+[ / cmd+] badges.
- **Migration safeguard:** a factory-default binding now yields to any user-configured binding on the same keystroke, so the new cmd+opt+[ / cmd+opt+] defaults can never shadow an existing user binding; dispatch source order no longer decides collisions with newly shipped defaults. Read-side only; user config files are never written.

Safari-style recipe (documented in the shortcut docs) for users who want cmd+[ as page-back inside browser panes: `{ "shortcuts": { "bindings": { "browserBack": "cmd+[", "browserForward": "cmd+]" }, "when": { "focusHistoryBack": "!browserFocus", "focusHistoryForward": "!browserFocus" } } }`.

## Commits

Commit 1 added failing regression tests for an earlier context-routed design; the final design supersedes them with `testFocusHistoryOwnsCommandBracketsEverywhereAndBrowserHistoryHasItsOwnChords`. The history retains the iteration.

## Tests

- `KeyboardShortcutContextTests`: chord ownership and non-conflict, menu suppression, user-customization detection.
- `AppDelegateShortcutRoutingTests.testFactoryDefaultBindingYieldsToUserConfiguredBindingOnSameChord`: runtime proof that a user binding beats a factory default earlier in the dispatch chain.

## Preflight (tagged build cmdbr, driven via debug socket)

In a focused browser pane: cmd+opt+[ went back a page, cmd+opt+] forward, cmd+[ / cmd+] moved focus history browser↔terminal exactly as in terminal panes, and page navigation never hijacked pane focus.

## Summary by CodeRabbit

* **New Features**
  * User-configured keyboard shortcuts now take precedence over factory defaults when using the same keystroke.
  * Browser navigation shortcuts beep when back/forward is unavailable.

* **Changes**
  * Updated default browser back/forward shortcuts from Command+[ and Command+] to Command+Option+[ and Command+Option+].


lawrencecchen:
@codex review

PR #5942, opened 2026-06-12
[codex] Make dev auth and QR setup automatic

azooz2003-bit:
## Summary

Makes tagged dev builds ready for auth and mobile pairing by default:

- resolves macOS dev auth origin automatically, using local web only when reachable and otherwise production auth/API origins
- enables the tagged iOS pairing host before launch so QR generation has routes without a separate defaults write
- preserves host-supplied compact attach URLs when QR helper filtering keeps the original route set
- aligns iOS DEBUG/dev builds with the same production Stack auth default, while preserving `CMUX_DEV_AUTH_ORIGIN=local` / `CMUXAuthEnvironment=development` for intentional local auth testing

## Why

Dev builds could route sign-in to localhost or the development Stack project while the paired Mac was using production Stack auth. That let QR transport succeed, but the Mac then rejected the iPhone token with an account verification failure. The default dogfood path now uses one auth realm on both Mac and iOS.

## Validation

- `bash -n scripts/reload.sh scripts/dev-setup.sh scripts/mobile-attach-qr.sh scripts/mobile-dev-launch.sh`
- `bash -n ios/scripts/reload.sh scripts/reload.sh scripts/dev-setup.sh scripts/mobile-attach-qr.sh scripts/mobile-dev-launch.sh`
- `node --test scripts/lib/attach-url.test.mjs`
- `./scripts/reload.sh --tag ios-close-workspace --launch`
- `ios/scripts/reload.sh --tag ios-close-workspace --device-only`
- verified tagged Mac auth status is signed in
- verified mobile host status is running with routes
- verified rebuilt iOS app Info.plist has `CMUXAuthEnvironment=production`
- generated a fresh mobile QR artifact

Note: direct relaunch after reinstall could not complete because CoreDevice reported the iPhone unavailable/locked; the device build and install succeeded.

### DIFF
--- a/Packages/CMUXAuthCore/Sources/CMUXAuthCore/Models/CMUXAuthEnvironment.swift
+++ b/Packages/CMUXAuthCore/Sources/CMUXAuthCore/Models/CMUXAuthEnvironment.swift
@@ -4,7 +4,7 @@ import Foundation
 ///
 /// Decided by the composition root (development for DEBUG builds, production
 /// otherwise) and injected; this package never reads build flags itself.
-public enum CMUXAuthEnvironment: Sendable {
+public enum CMUXAuthEnvironment: Equatable, Sendable {
     /// The development Stack project (DEBUG builds, local web stack).
     case development
     /// The production Stack project (Release/nightly builds).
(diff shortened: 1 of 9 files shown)


azooz2003-bit:
Superseded by focused account-verification fix in #5944. This PR mixed QR/dev setup changes with the iOS auth-realm fix, which made the review target unclear.

PR #5972, opened 2026-06-12
Fix 100% CPU render loop when an extension sidebar is selected (#5970)

### LINKED ISSUE #5970: Selecting a bundled extension-sidebar preset pins cmux at ~100% CPU (sustained SwiftUI re-render loop in VerticalTabsSidebar)
## cmux version
`0.64.14-nightly.2740219108001` (commit `e2d6c44`). Also reproduces on the prior nightly `0.64.14-nightly.2739628906001` (commit `cb631e0`).

## macOS version
macOS 26.0.1 (25A362)

## Mac chip
Apple Silicon (M-series)

## Installation method
Direct download (DMG), NIGHTLY. Also reproduced in a from-source build.

## Can you reproduce on NIGHTLY?
Yes, it still reproduces on NIGHTLY.

## Bug description
Selecting any bundled extension-sidebar preset (e.g. "Project Worktrees", provider id `com.example.cmux.sidebar.project-worktrees`) puts the app into a sustained ~100% CPU SwiftUI re-render loop. The window becomes unresponsive (spinning beachball).

Because the selection is persisted to the `cmuxExtensionSidebar.providerId` default, the loop also fires immediately on the **next launch**, so the app is effectively bricked on startup until that pref is cleared, with no obvious way out from the UI.

## Expected behavior
Selecting a bundled extension-sidebar provider should render it once and then idle, exactly like the default workspaces sidebar. No render loop, no pegged CPU.

## Steps to reproduce
Scriptable (most reliable):
1. Quit cmux NIGHTLY.
2. `defaults write com.cmuxterm.app.nightly cmuxExtensionSidebar.providerId -string "com.example.cmux.sidebar.project-worktrees"`
3. Launch cmux NIGHTLY. The main thread pegs at ~100% CPU and stays there indefinitely.

Via UI:
1. Open the sidebar provider switcher and choose a bundled preset (e.g. "Project Worktrees").
2. CPU immediately pegs at ~100% and stays there.

(For the stable build, substitute bundle id `com.cmuxterm.app`.)

## Relevant logs
`sample` of the pegged process. The hot path is a continuous SwiftUI transaction flush where every body pass rebuilds the full settings catalog:

```
GraphHost.flushTransactions -> runTransaction -> AG::Subgraph::update   (repeating)
  ViewBodyAccessor.updateBody
    VerticalTabsSidebar.body.getter
      closure #11 in VerticalTabsSidebar.body.getter
        CmuxExtensionSidebarSelection.descriptor(for:)
          CmuxExtensionSidebarSelection.descriptors.getter
            CmuxExtensionSidebarSelection.customSidebarDescriptors.getter
            CmuxExtensionSidebarSelection.isEnabled.getter
              SettingCatalog.init() -> AutomationCatalogSection.init() -> SecretFileKey.init(...)
```

## Additional context (source analysis)
This looks like two compounding issues in `Sources/ContentView.swift`:

1. **The body re-evaluates continuously** when the selected provider is anything other than the default workspaces sidebar. With the default provider the sidebar idles; selecting a bundled preset flips `VerticalTabsSidebar.body` from `workspaceScrollArea` to `extensionSidebarScrollArea` (`ContentView.swift:11103-11106`), and that subtree keeps invalidating the body. A from-source build from ~Jun 10 with the *same* pref does **not** loop, so this appears to be a regression introduced in the sidebar reworks landed since.

2. **Each body pass does very expensive work.** The branch condition calls `CmuxExtensionSidebarSelection.descriptor(for:)` -> `descriptors`, and both `isEnabled` (`:10164`) and `customSidebarsEnabled` (`:10181`) construct a full `SettingCatalog()` on every access. The doc comment at `:10153-10162` explicitly notes these synchronous reads are meant for paths that "run outside the SwiftUI update cycle", but here they execute on every body evaluation, turning the invalidation into a CPU catastrophe.

I have a fix in progress and will open a PR referencing this issue.

## Workaround
```
defaults delete com.cmuxterm.app.nightly cmuxExtensionSidebar.providerId
```
(or `com.cmuxterm.app` for the stable build), then relaunch.

kevinsslin:
## Summary

Fixes #5970: selecting a bundled extension-sidebar preset (e.g. "Project Worktrees") pinned the main thread at ~100% CPU in an unbounded SwiftUI re-render loop, and because the selection persists in `cmuxExtensionSidebar.providerId`, the app re-entered the loop on every launch.

**Root cause.** `VerticalTabsSidebar` observed workspace state with `.onReceive` on a publisher built inline on every `body` pass:

```swift
.onReceive(extensionSidebarImmediateObservationPublisher(renderContext: renderContext).receive(on: RunLoop.main)) { _ in
    refreshExtensionSidebarSnapshot()
}
```

That factory returns a fresh `AnyPublisher` each render, so SwiftUI re-subscribed every render. Each publisher is a `CombineLatest` of `@Published` workspace fields, which replays its current value on every subscription. The replay called `refreshExtensionSidebarSnapshot()` -> bumped the `@State` `extensionSidebarUpdateToken` -> re-invalidated `body` -> re-subscribed -> replayed again. (The default workspaces sidebar doesn't wire a recreated-per-render publisher into a body-level `@State` bump, so it was unaffected.)

**Fix.** Observe via `.task(id: renderContext.tabIds)`, bridging the existing observation publishers through `.values`. SwiftUI owns the subscription lifecycle: the task restarts only when the observed workspace set changes and is torn down on disappear, so a subscription is established once per genuine change instead of every frame. No new Combine view state, no shadow tab-id tracker. The publishers themselves are unchanged, so their on-subscribe replay (relied on by a late-mounting sidebar row, covered by `WorkspaceSidebarObservationTests`) still fires when a subscription is genuinely established.

## Testing

- **Repro (before):** built a Debug app at the affected commit, set `cmuxExtensionSidebar.providerId = com.example.cmux.sidebar.project-worktrees` with a workspace open -> sustained ~100% CPU. `sample` shows a continuous `GraphHost.flushTransactions` -> `VerticalTabsSidebar.body` -> `CmuxExtensionSidebarSelection.descriptor(for:)`.
- **After:** rebuilt with this change, same pref + workspace -> CPU settles to idle (~0.5%); the extension sidebar renders and provider switching works.
- `WorkspaceSidebarObservationTests` (the publishers' emit-on-subscribe contract) is unaffected since the publishers are untouched.

## Notes

- Reworked from an earlier `@State`-publisher-caching approach to `.task(id:)` per reviewer feedback (`cmux-swift-concurrency-modernization` / `cmux-swift-architectural-rethink`): no new Combine `@State`, and the runtime owns the subscription lifecycle.

## Checklist

- [x] Existing publisher tests (`WorkspaceSidebarObservationTests`) remain valid; the change is in view-subscription lifecycle, not the publishers
- [ ] Demo video (can add if useful)

## Summary by CodeRabbit

* **Bug Fixes**
  * Optimized extension sidebar observation to avoid repeated re-subscriptions, reducing unnecessary refreshes. This improves stability and responsiveness of the sidebar, resulting in smoother interactions when switching views or tabs. The update harmonizes observation behavior across related areas to prevent redundant updates and lessen UI jitter.

### DIFF
--- a/Sources/ContentView.swift
+++ b/Sources/ContentView.swift
@@ -11414,18 +11414,26 @@ struct VerticalTabsSidebar: View {
                     CmuxExtensionSidebarSelection.setProviderId(CmuxSidebarProviderDescriptor.defaultWorkspacesID)
                 }
             )
-            .onReceive(
-                extensionSidebarImmediateObservationPublisher(renderContext: renderContext)
-                    .receive(on: RunLoop.main)
-            ) { _ in
-                refreshExtensionSidebarSnapshot()
+            // Observe workspace state via `.task(id:)` so SwiftUI owns the
+            // subscription lifecycle: restarted only when the observed workspace
+            // set changes, torn down on disappear. Bridging the Combine
+            // publishers through `.values` (instead of `.onReceive` on a publisher
+            // rebuilt every body pass) avoids re-subscribing on every render,
+            // which replayed each `@Published`-backed publisher's current value,
+            // bumped `extensionSidebarUpdateToken`, re-invalidated the body, and
+            // span the main thread at 100% CPU once an extension sidebar was
+            // selected (#5970).
+            .task(id: renderContext.tabIds) {
+                for await _ in extensionSidebarImmediateObservationPublisher(renderContext: renderContext).values {
+                    refreshExtensionSidebarSnapshot()
+                }
             }
-            .onReceive(
-                extensionSidebarDebouncedObservationPublisher(renderContext: renderContext)
-                    .receive(on: RunLoop.main)
+            .task(id: renderContext.tabIds) {
+                let debounced = extensionSidebarDebouncedObservationPublisher(renderContext: renderContext)
                     .debounce(for: Self.extensionSidebarObservationCoalesceInterval, scheduler: RunLoop.main)
-            ) { _ in
-                refreshExtensionSidebarSnapshot()
+                for await _ in debounced.values {
+                    refreshExtensionSidebarSnapshot()
+                }
             }
             // Fade the extension's content out at the bottom so it dissolves behind the
             // sidebar footer instead of overlapping it sharply, matching the default
@@ -11586,18 +11594,22 @@ struct VerticalTabsSidebar: View {
             }
             .background(Color.clear)
             .modifier(ClearScrollBackground())
-            .onReceive(
-                extensionSidebarImmediateObservationPublisher(renderContext: renderContext)
-                    .receive(on: RunLoop.main)
-            ) { _ in
-                refreshExtensionSidebarSnapshot()
+            // Stable observation via `.task(id:)` (see the matching block above):
+            // SwiftUI restarts it only when the observed workspace set changes, so
+            // the publishers are not rebuilt and re-subscribed every body pass,
+            // which replayed their current value and span the main thread at 100%
+            // CPU once an extension sidebar was selected (#5970).
+            .task(id: renderContext.tabIds) {
+                for await _ in extensionSidebarImmediateObservationPublisher(renderContext: renderContext).values {
+                    refreshExtensionSidebarSnapshot()
+                }
             }
-            .onReceive(
-                    extensionSidebarDebouncedObservationPublisher(renderContext: renderContext)
-                        .receive(on: RunLoop.main)
-                        .debounce(for: Self.extensionSidebarObservationCoalesceInterval, scheduler: RunLoop.main)
-                ) { _ in
-                refreshExtensionSidebarSnapshot()
+            .task(id: renderContext.tabIds) {
+                let debounced = extensionSidebarDebouncedObservationPublisher(renderContext: renderContext)
+                    .debounce(for: Self.extensionSidebarObservationCoalesceInterval, scheduler: RunLoop.main)
+                for await _ in debounced.values {
+                    refreshExtensionSidebarSnapshot()
+                }
             }
             .onReceive(
                 NotificationCenter.default.publisher(for: BrowserStackSidebar.stateDidLoadNotification)


kevinsslin:
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review

kevinsslin:
Thanks for the review. Agreed on the `cmux-swift-concurrency-modernization` / `cmux-swift-architectural-rethink` point: storing `AnyPublisher` in `@State` with a shadow `tabIds` guard was a symptom patch.

Reworked to the suggested Swift Concurrency shape: two `.task(id: renderContext.tabIds)` modifiers that bridge the existing observation publishers via `.values` and call `refreshExtensionSidebarSnapshot()` on each element. SwiftUI now owns the subscription lifecycle (restarted only when the observed workspace set changes, torn down on disappear), so this removes:

- the three new `@State` fields (`extensionSidebarImmediateObservation`, `extensionSidebarDebouncedObservation`, `extensionSidebarObservedTabIds`),
- the `updateExtensionSidebarObservation` guard helper, and
- the `.onChange(of:initial:)` lifecycle hook.

This also addresses the `renderContext`-staleness note: the task is keyed on `renderContext.tabIds`, so the captured `renderContext` always matches the id that (re)started the task. No new Combine view state and no shadow source of truth.

Verified locally on a from-source build at the affected commit: with `cmuxExtensionSidebar.providerId = com.example.cmux.sidebar.project-worktrees` and a workspace open, the main thread now settles to idle (was a sustained ~100% before), and the extension sidebar still renders and refreshes.

(The two failing `Vercel` checks are `Authorization required to deploy` on fork PRs, not related to this change.)

kevinsslin:
@coderabbitai review
@greptile-apps review

azooz2003-bit:
Reviewed this and reproduced the CPU issue before/after.

The CPU fix is real: with `cmuxExtensionSidebar.providerId = com.example.cmux.sidebar.project-worktrees`, the merge-base build `r5972b` sustained high CPU, `ps` samples were mostly ~130-300%, `top` showed 173.9% after startup, and `sample` showed `GraphHost.flushTransactions -> VerticalTabsSidebar.body -> CmuxExtensionSidebarSelection.descriptor`. The PR build `r5972p` with the same provider settled to 0-4% in `top`, and its sample did not show that SwiftUI body loop.

I think there is still one blocker before merge: the new `.task(id: renderContext.tabIds)` observation fixes the resubscription loop, but it also removes the old incidental refresh path for hosted extension snapshots when only `SidebarUnreadModel` changes. `extensionWorkspaceSnapshot` includes `sidebarUnread.unreadCount(...)` and `latestNotificationText`, while `CMUXInstalledExtensionSidebarHostView` only sends `sidebarSnapshotDidChange()` when `snapshotUpdateToken` or snapshot `sequence` changes. Unread-only changes can therefore leave hosted extension sidebars stale until another workspace publisher emits.

Suggested fix: add an explicit bounded refresh trigger for the unread projection consumed by hosted extension snapshots, without reintroducing body-level Combine resubscription.

azooz2003-bit:
CI follow-up: `workflow-guard-tests` is failing on the Swift file length budget, not on the CPU behavior.

Failure:

```text
Swift file length budget exceeded.
+12 Sources/ContentView.swift
   actual=19277 budget=19265
Split the file, reduce the new growth, or refresh the budget only when accepting known debt.
```

Since this PR only changes `ContentView.swift`, the lowest-risk path is probably to reduce the added comment/code lines while adding the hosted-extension unread refresh fix, rather than updating the budget for this bug fix.

PR #5984, opened 2026-06-12
iOS: Telegram-style chat surface for agent sessions

azooz2003-bit:
A native chat UI for AI agent sessions on iOS, rendered from the agent's transcript JSONL instead of the terminal grid. Workspaces with a chat-capable session get an Agent Chat toolbar button that opens the conversation as Telegram-style bubbles: user prompts outgoing, agent prose incoming with markdown, terminal/diff cards, collapsed thought and tool rows, actionable permission and question cards with frozen receipts, a typing indicator driven by hook state, and a composer with photo attachments, Esc/Ctrl-C chips, and a stop button.

Architecture, mac side: AgentChatTranscriptService (new, Sources/Mobile/AgentChat/) tracks sessions from agent hook events plus the ~/.cmuxterm hook session stores, resolves and tails each session's transcript with CmuxFileWatch (bounded backfill, incremental parse, truncation recovery, pid liveness sweeps), serves mobile.chat.sessions/history RPCs, and pushes chat.message frames to subscribed phones. Send, interrupt, and answer reuse the existing mobile terminal injection paths (bracketed paste plus submit key), so chat input behaves exactly like composer input.

Architecture, iOS side: two new packages. CmuxAgentChat is platform-free domain: wire DTOs, claude and codex JSONL parsers (fail-open: unknown line types become an unsupported row, never dropped), ChatConversationStore (seq-windowed transcript, optimistic pending sends reconciled against transcript echoes, subscribe-before-fetch so no event falls in the handshake gap), and a scripted FixtureChatEventSource. CmuxAgentChatUI is the SwiftUI surface, dark-first with adaptive light mode (terminal cards stay dark in both schemes). 99 package tests cover parsers, projector, store, ANSI sanitizing, and wire coding.

Verified end to end on simulator against a tagged mac build with a live claude session: pairing, session listing, history, live push (transcript append renders on the phone in about 200 ms), and phone-to-terminal send. A DEBUG-only demo (Settings > Developer > Agent Chat Demo) renders the full taxonomy from fixtures.

Known gaps, deliberate for v1: history is bounded to the newest 2000 transcript lines (the UI shows "Earlier history is on your Mac" past that); codex sessions need hook-launched terminals for the surface binding; answer() injects the option's number key, which matches claude's TUI prompts; permission cards are fixture-only until the hook wire carries permission requests. Follow-ups filed from review that are not in this PR: wire enum forward-compat (unknown raw values currently drop a frame), Dynamic Type and VoiceOver sweeps, scroll-anchor preservation on history prepend, per-event hook ordering via AsyncStream, retry-with-attachments.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/5984"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783878175&installation_id=133855650&pr_number=5984&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F5984&signature=e45659fd94f8554910ff747f91e890ea5f1531e65663cae700eed1094e46cb14"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Adds a Telegram‑style chat on iOS for coding‑agent sessions with push‑driven transcripts, block‑level markdown, and a photo‑aware composer. Scope is agent‑only; terminal/diff output renders as cards inside agent transcripts, and chat toggles in place on the current workspace tab (session list updates via push, no polling).

- **New Features**
  - Platform packages: `CmuxAgentChat` (typed message model with fail‑open wire coding, Claude/Codex JSONL parsers, OSC‑133 terminal blocks, ISO‑8601 via `ChatWireCoding`, `ChatConversationStore` with optimistic/queued/cancelable sends that flush one per idle turn, `ChatSessionListReducer`) and `CmuxAgentChatUI` (bubble UI with block‑level markdown and code‑language labels, terminal/diff cards, actions, typing indicator, composer with photo attachments and real pending thumbnails, bounded markdown/content/diff caches, en+ja localization, DEBUG demo and scroll‑geometry overlay).
  - Push‑based session list via `MobileChatEventSource` (no polling). The opened session is pinned while chat is on; the in‑place toggle binds to the visible workspace tab. Ended sessions are read‑only. Header shows workspace/tab with a leading state symbol.

- **Bug Fixes**
  - Stability/perf: fixed a 100%‑CPU scroll freeze; at‑bottom derives from scroll geometry; keyboard pin keeps the tail visible; tap‑to‑dismiss is limited to the transcript; header title width is bounded; error banners auto‑dismiss.
  - Detection/adoption/resolution: live agent adoption on title changes and workspace list refreshes (throttled); robust `/tmp`↔`/private/tmp` matching; same‑directory agents disambiguated (skip transcripts already claimed); `$HOME` skipped; inode‑based rotation detection with explicit reset frames; strict pin prevents wrong‑session switches; spinner‑forever on initial load replaced by a retry path.
  - Composer/send reliability: attachments removable by picked‑item id and persist on retry; pending thumbnails show real previews (wrap to two per row); sending is blocked while photos stage or disconnected; queued sends flush one per idle turn and can be canceled; retry while working re‑queues; stale delivered text‑only pendings are swept; subscribe before fetch to avoid gaps; reconnect with bounded backoff; localized chat RPC errors (en + ja).
  - Accessibility: VoiceOver labels/values/hints across bubbles, tool/terminal outcomes, and diff lines; larger touch targets; Reduce Motion stops infinite pulses.

<sup>Written for commit 75274f5dd78266f51960c8671aeb1995aea81003. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/5984?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Full Agent Chat integrated across Mac/iOS: session list, history paging, live events, interrupt/answer controls, demo screen, and terminal-backed sends.

* **UI**
  * Composer with multiline input, image attachments, optimistic send/retry/discard, stop/hard-stop; rich transcript UI with theming, cached markdown, scroll-to-bottom, typing indicator, localized strings.

* **Parsing & Reliability**
  * Robust transcript parsing, ANSI-sanitized terminal output, truncation/budgeting, stable IDs, and resilient wire decoding.

* **Tests**
  * Extensive unit/integration tests covering store, parsers, UI sanitizer, and projector.

* **Chores**
  * New Swift package manifests and package wiring for agent chat modules.


azooz2003-bit:
@codex review

azooz2003-bit:
Iteration log since the PR opened (multi-agent review loop, each round = specialist reviews → fixes → adversarial re-verification → live E2E re-check):

Round 1 (5a39f3e66e): conventions, correctness, and UX reviews. Fixed the silent event-loss gap between history fetch and subscribe, eternal typing indicator from dead agent processes, session-ended composer state, double-tap hazards on permission cards, tailer leaks, sidechain transcript pollution, and attachment-only sends never submitting.

Round 2 (39ec981b2b): an adversarial pass proved two round-1 fixes broken and both were refixed with regression tests (window dedup spanned only a 64-message suffix and double-inserted under tool storms; ended-session tailer eviction was instantly resurrected by the same hook event). Also: wire forward-compat now fails open end to end with golden-fixture tests, reconnects back off exponentially, hook-store file reads throttled off the per-event path, descriptor pushes coalesced, drafts survive cover dismissal, the terminal escape hatch lands on the session's actual surface.

Round 3 (728cc61c20): a second adversarial pass found the attachment reconcile matched only fixture-shaped echoes (real echoes are pasted image paths in prose) and that claude --resume's reused session id kept a stale pid that flapped live sessions back to ended. Both fixed, plus seq-regression re-anchoring after transcript rewrites, gap-safe resync after long disconnects, and failed sends' retry rows are never consumed by other echoes.

Functionality was independently re-verified each round by a codex agent driving the tagged mac build over its debug socket: package suites green (now 70+14), live append-to-push latency ~250ms, subscription handshake present, no emit storms.

azooz2003-bit:
Rounds 6-9 (continued multi-agent loop + live simulator dogfood):

**Round 6** — fixed a user-reported stuck spinner ("Unknown chat session or transcript" with no escape) when opening chat on a session whose transcript file doesn't exist yet: the Mac history handler now refreshes bindings and retries, and the client shows an actionable Retry instead of spinning. At-bottom detection switched to real scroll visibility (the pill could hide and steal scroll up to a viewport early). Five shipping strings that fell back to English got Japanese.

**Round 7** — full accessibility pass: VoiceOver labels/values/hints on delivery state, terminal/tool outcomes, and expansion state; bubbles read as one element; agent replies and errors announced; Dynamic Type on reading content; 44pt touch targets; Reduce Motion; 15 new localized keys. Plus adversarial hardening (dead-Retry edge, pid-adoption only when absent, skip pointless second resolver scan).

**Round 8** — a user-reported hard freeze (app fully unresponsive after scrolling) traced via process sample to a container-level SwiftUI animation animating the whole lazy transcript on every at-bottom flip; scoped to the pill transition. Error banner auto-dismiss, photo-staging send guard, and a `chat.sessions.dump` debug verb added to diagnose state inconsistencies.

**Round 9** — fixed the "Session ended while the agent is running" inconsistency: the chat button picked sessions by recency, so a dead session could shadow a live one. Selection now prioritizes needs-input > working > idle and never opens an ended session when a live one exists (`ChatSessionDescriptor.openable`, unit-tested). Diagnosed codex's intermittent send failure: sends fired while the agent is mid-response paste into Claude Code's input box but the running task swallows the submit; the normal idle turn-taking flow submits reliably (verified with a live, responding agent). Auto-submit-on-idle for busy sends is a tracked follow-up.

86 package tests, CI green. Still gated on dogfood approval before merge.

azooz2003-bit:
Dogfood finding (chat toggle missing for Cloud sessions) — root-caused.

Reported: opening a Cloud session from the phone lands in a terminal with no toggle to switch to the chat GUI.

Root cause: the agent-chat registry is populated **only from local hook events** — the `cmux hooks` CLI writes `~/.cmuxterm/<agent>-hook-sessions.json` and posts `feed.push` to the local socket, which `AgentChatTranscriptService` ingests and whose transcript it tails from the local `~/.claude` (or `~/.codex`) projects dir. Two session kinds therefore never register, so `mobile.chat.sessions` returns nothing for their workspace and the toggle (correctly, given no transcript) stays hidden:

1. **Cloud VM sessions** — the agent runs on the remote VM; its hooks/transcript live on the VM, not the Mac. The Mac proxies only the PTY for the terminal view.
2. **Native `agentSession` surfaces** (`Claude Code · React`) — cmux's in-process agent panels don't run the hooks CLI, so they don't register either (these already have their own GUI, so less of a gap).

This is an architectural limitation, not a one-line visibility bug: making chat work for Cloud requires the remote daemon to surface the VM's agent session + stream its transcript/hook events back to the Mac (or phone), then inject them into the registry tagged with the proxying local workspace id. That spans cmuxd (VM) + the Mac remote client + the registry and needs a live Cloud VM to build and verify, so it's scoped as the next major piece rather than an autonomous quick fix.

Current chat support is accurate for **local, terminal-hosted, hook-wired agent sessions** (the common `cmux` + `claude`/`codex` case), where the toggle appears once the agent fires its first hook. Tracking Cloud/agentSession chat as a follow-up.

azooz2003-bit:
Spec-fidelity review (codex, read-only) against docs/cmux-ios-telegram-chat/PLAN.md. The shipping chat features match the spec except three known deltas, surfaced here for a product decision rather than implemented unprompted (flagging since recent curation reverted a speculative status feature):

1. Two-tick delivery (PLAN.md:48) — spec wants two ticks (Mac ack / agent processing); the pending row shows one tick (delivered = Mac accepted) and is then reconciled away by the transcript echo. A true second tick needs an "agent processing" signal plus persisting send-status onto the real message (a status-model change). Deferred pending your call — it's adjacent to the unread divider you reverted, so I didn't want to add another status indicator unasked.

2. Attach flow (PLAN.md:18) — spec describes an attach sheet with PhotosPicker + recent screenshots; current composer uses a direct PhotosPicker button only (no recent-screenshots source). Functional, just missing that tier.

3. Off-main markdown/ANSI (PLAN.md:48) — attribution is cached by message id (markdown, ANSI, prose segments, blocks, diff lines), but the caches are @MainActor so a cache *miss* parses on the main actor. With content-hash caching a miss is one-time per message (not per scroll), and the round-23 perf pass found the scroll path otherwise clean, so this is bounded; moving misses off-main is a larger render-pipeline change (async placeholder-then-update).

No correctness deviations found in the shipping UI. Happy to implement any of these if you want them in this PR — otherwise they stand as documented deltas.

azooz2003-bit:
Live verification (post 8pm PT limit reset)

Ran the first true live agent round-trip today now that the Claude weekly limit has reset. Drove a fresh hooked-terminal Claude session (Opus 4.8) with a deterministic prompt; it replied correctly, and the reply was written to the session JSONL as an `assistant` record (`message.content[].text`), exactly the shape `ClaudeTranscriptParser` consumes. So the source side of the chat pipeline (prompt to live reply to transcript to parse) is confirmed against real live agent output, not just fixtures.

Verified green this round: CmuxAgentChat 82 tests, CmuxAgentChatUI 30 tests. Full lens rotation (adversarial, performance, architecture, product-vs-spec, accessibility, localization, memory-safety) came back clean. The memory-safety pass found no retain cycles or leaks across the ChatScreen/store/event-source lifecycle; the one flagged item (the fire-and-forget unsubscribe in `MobileChatEventSource.onTermination`) is the correct idiom for a best-effort cleanup RPC issued after the pump is already cancelled.

Still gated on environment, not on PR code:
- Fresh-session registration into the chat list is driven by the `cmux hooks` hook writing `claude-hook-sessions.json`. The dogfood machine's global `~/.claude/settings.json` currently has no hooks block, so a bare-launched session writes a transcript without registering. Sessions launched through the normal cmux-managed path still register (the registry holds 11 live sessions with on-disk transcripts).
- A full phone round-trip (sim paired, tap a session, watch a streaming reply) needs sim-to-Mac pairing; no attach URL is configured on this machine.

No code changed this round.

azooz2003-bit:
Overnight progress (sim-only; iPhone unplugged until morning)

Dogfood fixes from the device session:
- Tap anywhere dismisses the keyboard (rewritten via didMoveToWindow; the first version never attached the recognizer).
- Composer material now bleeds through the bottom safe area, so no dark gap below the bar when the keyboard is down.
- Header state is now a leading symbol (working = pulsing green, idle = gray dot, needs-input = orange question mark, ended = hollow ring) instead of a word, so line two is the full tab name. Sim-verified.

#5 (chat for plain terminals) started, building in slices:
- Slice A: OSC 133 command-block model + a pure incremental parser (segments the PTY stream into command/output/exit blocks, strips ANSI, folds CR progress redraws, flags alt-screen programs). 10 unit tests, 92 total green.
- Slice B: the single-column monospace LOG UI (the differentiator vs agent bubbles): a command row with the typed command over its output, long-output collapse, exit-status footer (green check / red exit code + a red fail rail), running spinner, and an interactive-program card (Open Terminal) for full-screen TUIs. Verified on the simulator via a new DEBUG Settings > Developer > Terminal Log Demo.

Next: terminal session identity (composer/header/toggle) and the Mac-side OSC 133 shell-integration + PTY parsing to serve real terminal sessions (needs a live Mac; likely finished after the iPhone is back).

azooz2003-bit:
Overnight progress: #5 iOS data-plane complete (terminal sessions through the store)

Following an architecture review, the iOS side now carries terminal command-blocks through the existing chat store/event-source seam, additive and gated on session kind so the agent path is untouched:

- TerminalCommandBlock is Codable; a new ChatSessionEvent.terminalBlocks case and an optional ChatHistoryPage.terminalBlocks field carry blocks over the wire (whole-block values, upserted by id, idempotent on reconnect). Agent payloads still decode unchanged.
- ChatConversationStore gained a terminal mode: a flat command-log projection (bypassing the bubble-grouping projector), an upsert-by-id apply arm, and history seeding. The agent reproject/apply paths are verbatim.
- The Terminal Log Demo is now a real ChatScreen over a terminal fixture store, so the full path (history -> rows -> monospace log + terminal composer + header) is verified end-to-end on the simulator.

106 unit tests pass (6 new: wire round-trips incl agent backward-compat; store history-seeds-rows and upsert-by-id).

The only remaining piece for #5 is the Mac source: OSC 133 shell-integration injection, host-side PTY parsing into blocks (reusing the parser), and serving terminal sessions via the existing mobile.chat.* RPC. That needs a live Mac (and an empirical check of whether the embedded Ghostty already emits OSC 133 marks), so it's the next step once the device is back.

Still sim-only tonight (iPhone unplugged). Phone-only blockers for the morning: on-device confirmation of the earlier dogfood fixes, and the Mac-side terminal end-to-end.

azooz2003-bit:
Overnight autonomous run — consolidated handoff (sim-only; iPhone was unplugged)

Branch at 6422a79c7d, clean, never merged. All work below is pushed.

Dogfood fixes (from the device session before the iPhone was unplugged):
- Scroll-up freeze fixed (the 100% CPU hang): the scroll-to-bottom pill's at-bottom state was flipped inside withAnimation, which re-laid-out the LazyVStack every animation frame while the anchor oscillated across the visibility threshold. Decoupled it (set state without animating the list).
- Tap anywhere dismisses the keyboard (rewritten via didMoveToWindow; the first attempt never attached the recognizer). Also fixed a HIGH bug found in review: the recognizer leaked on the shared window and dismissed the keyboard app-wide on other screens — now removed on detach + deinit.
- Composer material bleeds through the bottom safe area (no dark gap when the keyboard is down) and behind the keyboard's rounded corners.
- Header shows the workspace name + tab as a subtitle, with state as a leading symbol (working = pulsing green, idle = gray dot, needs-input = orange question mark, ended = hollow ring) instead of a word, so the tab name isn't truncated. Sim-verified.
- Chip-row soft edge fade; failure-rail alignment.
- Toggle binds to the visible tab (matches the session's surface id), and the toggle shows only when the visible tab has a chat session (the heuristic you asked about).

#5 (chat for plain terminals) — iOS side complete, sim-verified end-to-end, 111 unit tests:
- Slice A: TerminalCommandBlock model + a pure incremental OSC 133 parser (command/output/exit segmentation; strips ANSI; folds CR progress redraws; flags alt-screen; handles split escapes and split CRLF across chunks; O(n) over a command, not O(n^2)).
- Slice B: single-column monospace LOG UI (the differentiator vs agent bubbles): command rows, output blocks with collapse, exit-status footer (green check / red exit + fail rail), running spinner, and an interactive-program card (Open Terminal) for full-screen TUIs.
- Slice C: terminal blocks are first-class transcript rows, rendering through the real ChatTranscriptListView.
- Slice C2: ChatSessionKind on the descriptor + a terminal composer (monospace "> command").
- Data-plane: TerminalCommandBlock/ChatSessionEvent/ChatHistoryPage carry blocks over the existing ChatEventSource seam (additive, agent payloads still decode); ChatConversationStore has a terminal mode (flat log projection, upsert-by-id, history seeding, reset-clear, send-pending reconcile). The agent path is untouched.
- Hardened across adversarial / performance / accessibility / localization review (fixed: reset stale-render, invisible/leaking terminal sends, the O(n^2) fold, batched alt-screen modes, unbounded escape buffer, a keyboard-leak, header pulse, Equatable field order, outputLines re-split).
- Verified via a real ChatScreen over a terminal fixture store (Settings > Developer > Terminal Log Demo on the sim).

Remaining for #5 — the Mac source only (needs a live Mac; deferred deliberately overnight to avoid a heavy app build that would disrupt the pairing host):
- TerminalChatTranscriptService: per-surface OSC133CommandParser fed by the EXISTING MobileTerminalByteTee (256KB ring), diff blocks -> emit .terminalBlocks on the chat.message topic; serve block history.
- TerminalChatSessionRegistry: live terminal surfaces -> descriptors (kind: .terminal); merge into v2MobileChatSessions/History.
- OSC 133 injection is essentially a no-op: cmux already emits OSC 133 marks via its own shell integration (cmux-zsh-integration.zsh etc.), so the marks are already in the PTY stream.
- Capability/default: a tab is chat-capable iff OSC 133 marks were observed; chat is the default only when marks are present, else raw terminal.
- Two perf follow-ups belong here: cap retained per-block output (infinite producers + per-tick wire), and single-row reproject on upsert.

Phone-only blockers for the morning: on-device confirmation of the tap-dismiss + keyboard-gap fixes, and the Slice D Mac end-to-end (needs the Mac + iPhone).

PR #6017, opened 2026-06-12
iOS Devices sheet: always-available Add device entry that re-enters the pairing flow

lawrencecchen:
Lawrence: "in the devices menu we should also be able to add a new device and go through the flow again."

## Which side

The "devices menu" is the iOS Devices sheet (`DeviceTreeView`, titled "Devices", opened from the workspace list toolbar). Once a phone was paired, the add-device pairing flow became unreachable from there: the root `PairingView` sheet only mounts on the disconnected branch, and the only connected-state path was Settings -> Switch Mac, which offers a bare QR scanner, not the full flow.

The Mac side already has two re-entry points into its pairing window (Settings -> Mobile -> "Pair…" and the command palette "Connect iPhone/iPad") and has no Devices menu to extend, so no Mac changes here. A dedicated Mac menu item is a possible follow-up; left out to avoid colliding with the in-flight pairing-window changes.

## Shared action and entry points

One shared `beginAddDevice()` action in `DeviceTreeView`, invoked from two entry points per the shared-behavior policy:

- an always-visible "Add device" list row (also present in the empty state), `MobileDeviceTreeAddDevice`
- a toolbar plus button, `MobileDeviceTreeAddDeviceToolbar`

Both present the existing first-run `PairingView` (QR scan or manual host) as a sheet over the tree, driven by the same store mutations as the root pairing path (`connectPairingInput` / `connectManualHost`). No keyboard shortcut added. Pairing window internals (`MobilePairingView`/`Model`/`WindowController`, iOS URL scheme) are untouched.

Re-entering the flow over a live connection needed explicit policy, extracted into `DeviceTreeAddDevicePolicy` (a documented public value type in `CmuxMobileShellModel`, next to `MobileConnectionState`):

- Cancelling a freshly opened sheet must not call `cancelPairing()` while still connected; that call sets `connectionState = .disconnected` and would tear down the live connection.
- The sheet dismisses only when the attempt left the shell connected (success), then refreshes `loadPairedMacs` + `loadRegistryDevices` so the new device appears in the tree.
- A cancelled attempt (disconnected, no `connectionError` — the store's cancellation path sets none) restores the previously live Mac via the existing `switchToMac(macDeviceID:)`, so backing out of the sheet never strands the user.
- A failed attempt (error set) deliberately does not auto-restore: the reconnect path begins a fresh pairing attempt, which clears `connectionError` and would erase the failure reason while the user reads it on the auto-presented pairing sheet.

## Review notes (autoreview, 3 rounds)

Fixed from review: the policy is a value type with an initializer (was a namespace enum), all public API has DocC, and the cancel-restore behavior above came out of round 1. `cmux-policy-check` is clean.

Consciously deferred (one root cause, store-owned): `MobileShellComposite`'s pairing attempt is destructive — it can replace the live client before success, any new attempt clears the store-owned `connectionError`, and there is no way to invalidate an in-flight attempt without clearing the live context. Consequences the reviewer flagged: a genuinely failed add attempt still drops the live session (error stays readable, recovery is Settings -> Switch Mac), and a cancel during a mid-flight attempt cannot guarantee the attempt stops. Both are pre-existing in the shipped Settings -> Switch Mac -> "Pair Another Mac" path, which calls the same destructive `connectPairingURL` while connected with no restore and no cancellation at all; this PR's entry point is strictly safer than that path. The real fix is an attempt-scoped pairing API in the store (do not tear down the live client until the new connection succeeds; non-destructive cancel; failure reporting that survives a reconnect) — follow-up, and it intersects the concurrent pairing-flow work. Malformed input never reaches the destructive path: `PairingView` validates manual hosts locally and the QR scanner accepts only `cmux-ios://` pairing links.

## Tests

`DeviceTreeAddDevicePolicyTests` (8 tests) in `CmuxMobileShellModelTests`; `swift test --package-path Packages/CmuxMobileShellModel` green (38 tests, 7 suites). The policy lives in the model package because `CmuxMobileShellUI` declares iOS-only platforms and cannot host-run `swift test`.

Builds verified: iOS simulator arm64 (`ios/cmux.xcworkspace`, scheme `cmux-ios`) and macOS (`cmux.xcodeproj`, scheme `cmux`, build-only; the Mac app is behaviorally unchanged).

## Localization audit

No new user-facing strings. The row and the toolbar button's accessibility label reuse `mobile.addDevice.title` ("Add device"); `mobile.addDevice.title`, `mobile.common.done`, `mobile.deviceTree.title` verified present with en + ja in `ios/cmux/Resources/Localizable.xcstrings`. Accessibility identifiers are not user-facing.

## Dogfood

1. Pair the iPhone/simulator (`ios/scripts/reload.sh --tag adddev`), open the workspace list, tap the Devices toolbar icon.
2. Previously bad: the Devices sheet had no way to add a device. Now: an "Add device" row sits below the tree and a plus button sits in the toolbar; either opens the full Add device flow (QR scan or manual host).
3. Cancel the sheet without starting an attempt: the live connection must survive.
4. Pair a second Mac through it: on success the sheet closes and the new Mac appears in the tree.

---

> [!NOTE]
> **Medium Risk**
> Changes mobile connection lifecycle around destructive pairing while connected; mitigated by explicit policy, restore via existing `switchToMac`, and unit tests, with known pre-existing limits on mid-flight cancel documented in the PR.
> 
> **Overview**
> The iOS **Devices** sheet (`DeviceTreeView`) now exposes **Add device** at all times—a list row (including empty state) and a toolbar **+**—both driving one `beginAddDevice()` path that presents the existing **`PairingView`** sheet over the tree.
> 
> A new **`DeviceTreeAddDevicePolicy`** in the model package encodes behavior when pairing re-enters over a live Mac session: **Cancel** must not call `cancelPairing()` while still connected; the sheet **dismisses and refreshes** the device list only on success; a **cancelled** attempt (disconnected, no `connectionError`) **restores** the prior Mac via `switchToMac`; **failed** attempts keep the sheet and error visible and do **not** auto-restore (reconnect would clear the error).
> 
> **`DeviceTreeAddDevicePolicyTests`** (8 cases) lock in those rules for `swift test` on the model package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48ec65652f29c9b8751c0444830cefa2c93e4058. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Add-device entry in the device tree (toolbar button and list row) opens a pairing sheet.
  * Pairing flow: safer cancel behavior that preserves live connections, restores previous Mac when appropriate, and refreshes paired/registry devices after successful pairing.
  * Accessibility identifier added to the device tree view.

* **Tests**
  * Tests covering cancel/dismiss/restore behavior across connection states.

### DIFF
--- a/Packages/CmuxMobileShellModel/Sources/CmuxMobileShellModel/DeviceTreeAddDevicePolicy.swift
+++ b/Packages/CmuxMobileShellModel/Sources/CmuxMobileShellModel/DeviceTreeAddDevicePolicy.swift
@@ -35,18 +35,28 @@ public struct DeviceTreeAddDevicePolicy: Sendable {
     /// Whether a completed add-device attempt should reconnect the Mac that
     /// was live when the attempt started.
     ///
-    /// The underlying pairing path is destructive: a failed (or cancelled)
-    /// attempt leaves the shell `.disconnected` even when a healthy connection
-    /// existed before it began. When the attempt started over a live
-    /// connection (`previousMacDeviceID` non-nil) and ended disconnected,
-    /// restore that Mac via `switchToMac` instead of stranding the user —
-    /// the same "never strand the user" contract `switchToMac` documents for
-    /// its own failure path. A successful attempt is connected (to the new
-    /// device), so no restore happens.
+    /// The underlying pairing path is destructive: a cancelled attempt leaves
+    /// the shell `.disconnected` even when a healthy connection existed before
+    /// it began. When the attempt started over a live connection
+    /// (`previousMacDeviceID` non-nil), ended disconnected, and reported no
+    /// connection error (the store's cancellation path sets none, while every
+    /// real failure does), restore that Mac via `switchToMac` instead of
+    /// stranding the user — the same "never strand the user" contract
+    /// `switchToMac` documents for its own failure path.
+    ///
+    /// A failed attempt (`hasConnectionError`) deliberately does NOT restore:
+    /// the reconnect path begins a fresh pairing attempt, which clears
+    /// `connectionError` and would erase the actionable failure reason while
+    /// the user is reading it. The disconnected shell auto-presents the
+    /// pairing sheet with that error for an in-place retry; reconnecting the
+    /// previous Mac without losing the error needs a store-owned error
+    /// channel that survives reconnects (follow-up). A successful attempt is
+    /// connected (to the new device), so no restore happens either.
     public func restoresPreviousConnection(
         connectionState: MobileConnectionState,
-        previousMacDeviceID: String?
+        previousMacDeviceID: String?,
+        hasConnectionError: Bool
     ) -> Bool {
-        connectionState != .connected && previousMacDeviceID != nil
+        connectionState != .connected && previousMacDeviceID != nil && !hasConnectionError
     }
 }
--- a/Packages/CmuxMobileShellModel/Sources/CmuxMobileShellModel/DeviceTreeAddDevicePolicy.swift
+++ b/Packages/CmuxMobileShellModel/Sources/CmuxMobileShellModel/DeviceTreeAddDevicePolicy.swift
@@ -6,8 +6,11 @@ import Foundation
 ///
 /// Lives in the model package (with ``MobileConnectionState``) so it stays
 /// platform-neutral and `swift test` exercises it on any host, following the
-/// `MobileShellRouteAuthPolicy` pattern.
-public enum DeviceTreeAddDevicePolicy {
+/// `MobileShellRouteAuthPolicy` pattern. A value type (not a static-helper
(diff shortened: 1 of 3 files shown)


lawrencecchen:
The workflow-guard-tests failure (Validate Swift file length budget) is inherited from main, not this PR: https://github.com/manaflow-ai/cmux/pull/5651 grew Sources/TerminalNotificationStore.swift and Sources/Feed/FeedCoordinator.swift past the checked-in ledger. Fix: https://github.com/manaflow-ai/cmux/pull/6018. Once that lands, merging main here turns the guard green; this PR's own files are all under the 500-line threshold.

PR #6102, opened 2026-06-14
iOS composer: image attachments, attach button, autocorrect, padding

lawrencecchen:
## What changed

Adds iMessage-style image attachments to the iOS mobile composer and fixes a few composer ergonomics issues. Reuses the existing terminal.paste_image transport (no new wire protocol).

### Store (CmuxMobileShell)
- New `MobilePendingAttachment` value type (image data + lowercase format hint + stable id), host-testable (no UIKit).
- Per-terminal pending-attachment state keyed by terminal id (like the text draft), with add/remove/clear/read methods.
- `composerCanSend(forTerminalID:)`: true when text is non-empty OR at least one attachment is staged (images-only send allowed).
- `submitComposer()`: sends staged images in pick order (awaited, so the injected file paths land first), then submits the text, then clears the staged set for the submitted terminal.

### Composer UI (TerminalComposerView)
- Replaced the chevron.down (hide-composer) button left of the field with a paperclip attach button that opens a PhotosUI photo picker (multi-select). Composer dismissal still lives on the accessory toolbar's compose toggle (`ComposerDockIntent.closeComposer` -> `dismissComposer()`), so no dismiss path is lost.
- Picked images are encoded the same way the clipboard paste path encodes them (PNG, JPEG fallback over ~8MB) and staged.
- Staged attachments render as a horizontal row of removable thumbnail chips above the text field.
- Send enabled on text OR attachments; routes through `submitComposer()` and re-measures the band height.
- Composer now uses normal iOS text assistance (autocorrect on, sentence-case) since it is natural language to an agent. The raw terminal input field (TerminalInputTextView) is unchanged and keeps autocorrect off.
- Reduced top padding above the field (was 8pt vertical, now 2pt top / 8pt bottom).

### Entitlements / localization
- Added `NSPhotoLibraryUsageDescription` to `ios/Config/Info.plist`.
- Added en/ja strings for the attach and remove-attachment labels.

## How to test
- Host unit tests: `cd Packages/CmuxMobileShell && swift test --filter ComposerPendingAttachmentTests` (10 tests, all green).
- On device: open a terminal, tap the paperclip, pick one or more photos, confirm chips appear above the field and each has a working x. Send with text and with empty text (images only); confirm images arrive before the message and chips clear after send.

## Caveats
- The SwiftUI view + PhotosPicker cannot be host-compiled (CmuxMobileShellUI depends on the GhosttyKit xcframework, iOS-only); the final Xcode build validates it. API usage was parse-checked against the iOS SDK.
- `mobile.composer.close` localization key is now unused (the chevron it labeled was removed) but left in place to avoid extraction churn.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/6102"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784012687&installation_id=133855650&pr_number=6102&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F6102&signature=c373141fe31c960c1ca947573c8c52e8f75a80c67c8d3748c4a77d6c83abb884"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Large in-memory attachment state and async multi-RPC submit interact with account switch, reconnect, and terminal topology; mitigated by caps, pruning, identity guards, and broad unit/routing tests, but still user-data and sync-sensitive.
> 
> **Overview**
> Adds **iMessage-style photo attachments** to the iOS terminal composer: pick from the library, stage per-terminal chips, then send **images first** (existing `terminal.paste_image`) followed by the composed text via new **`submitComposer()`**.
> 
> **`MobileShellComposite`** gains per-terminal `MobilePendingAttachment` staging with atomic per-terminal and **global** count/byte caps, topology validation/pruning on workspace sync, sign-out wipe, and session-generation guards for in-flight picks. Submit paths capture workspace/terminal/text up front, block double-send, abort on sign-out or connection swap mid-flight, and keep staged content on RPC failure (bool results from paste-image).
> 
> **`TerminalComposerView`** swaps the band chevron for a **paperclip** + `PhotosPicker`, chip row, Send gated on text or attachments, file-backed import with ImageIO downsampling and cancellable staging tasks, plus composer autocorrect/sentence case and tighter top padding. Adds photo-library usage strings and updates UI tests to dismiss via the toolbar compose toggle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02e276a1440b188bb2e5b55912418bd4c0ed37a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Adds iMessage-style photo attachments to the iOS composer with a paperclip picker, per-terminal staging chips, and images-first sending via `submitComposer()`. Reliability is tightened across routing, memory, cancellation, and now aborts mid-send if the session or connection changes.

- **New Features**
  - Pick photos (multi-select) via `PhotosUI`; stage per terminal as removable chips; Send works with text or attachments (images first via `submitComposer()`).
  - Store: `MobilePendingAttachment`, `composerCanSend`, whole-submit re-entrancy guard, and targeted RPCs that capture workspace/terminal/text at start for correct routing.
  - UI: replace chevron with attach; enable autocorrect and sentence case; tighter top padding; add `NSPhotoLibraryUsageDescription` and localized attach/remove labels.

- **Bug Fixes**
  - Memory/perf: file-backed import with on-disk size gate; bounded ImageIO downsample/re-encode; cached thumbnails; structured cancellation with propagation; cancel on re-pick, terminal switch, and view disappear; caps enforced atomically (per terminal: 10 items, 8 MB per image, 32 MB total; global: 20 items, 64 MB).
  - Topology/privacy: validate terminal before staging; prune attachments when terminals disappear; clear staged attachments on sign-out; drop picks for stale sessions or missing terminals.
  - Reliability: snapshot target terminal and text at submit start; abort the submit if sign-in or connection changes mid-send (keep attachments and text for retry); skip mid-send deletions; remove after ack; on image failure, stop and keep remaining attachments and text.

<sup>Written for commit 02e276a1440b188bb2e5b55912418bd4c0ed37a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **New Features**
  * Added photo attachment support to the terminal composer with a photo picker, localized attach/remove strings, and iOS photo library permission.
  * Enable images-only sending with removable thumbnail chips, attachment count/size limits, and send button gating.

* **Bug Fixes**
  * Improved reliability with per-terminal staged attachments, send-time terminal targeting, images-then-text submission, and ack-based failure handling that preserves unacknowledged attachments.
  * Prevented cross-account/session carryover by clearing staged attachments on sign-out and guarding in-flight operations.

* **Tests**
  * Added unit and end-to-end coverage for per-terminal staging, session behavior, re-entrancy, and routing snapshots.


lawrencecchen:
Addressed the three P1 autoreview findings (commit 4ba9f3f9e):

**1. Wrong-terminal routing after a mid-send switch.** `submitComposer()` now captures the target workspace + terminal id once and threads them through both the image sends and the text send, so a terminal switch while an awaited image send is in flight can no longer reroute later images or the text. New internal targeted variants `submitTerminalPasteImage(_:format:workspaceID:terminalID:)` and `submitComposerInput(workspaceID:terminalID:)`; the public selection-based overloads delegate to them, preserving the clipboard-paste path.

**2. Failed sends no longer lose photos.** The image send path returns `Bool` (threaded through `sendRemoteTerminalPasteImage`). `submitComposer()` removes each attachment only after its send is acknowledged; on a failure it stops, keeps the remaining and failed attachments staged, and does not submit the text, matching the existing text-keep-on-failure semantics.

**3. Bounded staging, no per-render full-image decode.** Picker capped at `maxSelectionCount: 10` plus a 32 MB total byte budget enforced at stage time. Each attachment gets a small downsampled thumbnail built once off the main thread (ImageIO `CGImageSourceCreateThumbnailAtIndex`) and cached by id in a side cache; the chip renders the cached thumbnail instead of decoding the full `Data` in the view body on every keystroke.

**Final API shape (image/text send):**
- `submitTerminalPasteImage(_:format:) async -> Bool` (selection) -> `submitTerminalPasteImage(_:format:workspaceID:terminalID:) async -> Bool` (captured) -> `sendRemoteTerminalPasteImage(...) async -> Bool`
- `submitComposerInput() async` (selection) -> `submitComposerInput(workspaceID:terminalID:) async -> Bool` (captured) -> `sendRemoteTerminalPaste(...) async -> Bool`
- `addPendingAttachment(...)` now returns the new attachment's id so the UI can key the thumbnail cache.

**Tests:** `ComposerPendingAttachmentTests` updated (failed send keeps attachments). New `ComposerSubmitRoutingTests` drives `submitComposer()` end to end over the real `terminal.paste` / `terminal.paste_image` RPC frames: captured-terminal routing, mid-send switch does not reroute, full image-send rejection keeps everything, and a partial failure clears only the acknowledged attachment. All composer store tests pass on host. Swift file-length budget guard is green (TSV refreshed for the two grown files; the iOS-only view cannot be host-compiled, so CI ios-simulator / Xcode is the final compile gate).

lawrencecchen:
Round-2 autoreview fixes (96dcff1f6):

1. P1 Swift 6 Sendable: `PreparedAttachment` no longer stores a `UIImage` across the detached thumbnail-prep task. The off-main path keeps the ImageIO downsampling but hands back the thumbnail as PNG `Data` (Sendable); the `UIImage` is built on the main actor when populating the chip cache. No UIKit reference crosses the concurrency boundary.
2. P1 Privacy: `pendingAttachmentsByTerminalID` is now cleared in `signOut()`, right beside the text-draft wipe, so a previous account's staged photo bytes cannot resurface on a reused terminal id. (Drafts have no per-terminal-disappear prune hook, so sign-out is the matching account-bound reset.)
3. P2 Double-submit: added an `isSubmittingComposer` re-entrancy guard around the whole `submitComposer()` (images + text), set at entry and cleared in a defer. A double tap on Send no longer re-uploads the still-staged attachments while the first RPC awaits; failure still keeps attachments for retry.
4. P2 UI test: `testComposerSurvivesRepeatedOpenCloseCycles` now closes the composer via the accessory compose toggle instead of the removed `MobileComposerClose` chevron; dropped the unused `bandClose` constant.
5. P3 Localization: added `NSPhotoLibraryUsageDescription` (en + ja) to `InfoPlist.xcstrings`.

Tests: added host tests for sign-out clearing and the double-submit guard; `swift test --filter "ComposerPendingAttachmentTests|ComposerSubmitRoutingTests"` is 16/16 green. File-length budget guard is green (refreshed the two touched files).

lawrencecchen:
Round-3 autoreview fixes (7853a83f3):

**Finding 1 (text not captured):** `submitComposer()` now snapshots `terminalInputText` before any await and threads it through `submitComposerInput(...capturedText:)`. The text send and the post-send draft reconcile both key on that snapshot, so a terminal switch or field edit during the image `paste_image` awaits can no longer skip the composed text or paste a different terminal's draft. A mid-send edit is preserved (the field clears only when it still equals the sent snapshot). Text-only paths pass nil (no prior await, no drift); images-only sends snapshot empty text and no-op the text submit.

**Finding 2 (re-stage after sign-out):** Added a `signInGeneration` token bumped in `signOut()`. The picker captures it before its off-main load+encode and the store re-checks it in a new `addPendingAttachment(...ifSessionGeneration:)` path that drops the result if the token moved (sign-out) or the target terminal no longer exists. The recheck lives in the store mutation path so it holds even if the picker view is gone.

Tests (`ComposerSubmitRoutingTests | ComposerPendingAttachmentTests`): 21 pass, including 5 new ones covering snapshot-survives-edit, snapshot-survives-switch, guarded-add dropped-after-signout / dropped-when-terminal-gone / succeeds-when-unchanged. File-length budget green (actuals refreshed for both touched files).

lawrencecchen:
Round-3 hardening (autoreview findings):

1. **Full-size PNG OOM** — `TerminalComposerView.prepare` no longer calls `pngData()` on a full-res `UIImage`. It now downsamples the picked item via `CGImageSourceCreateThumbnailAtIndex` (send payload capped at 2048 px longest edge, thumbnail at 168 px) and re-encodes the bounded `CGImage` through `CGImageDestination`. The send payload tries PNG at the send size, then JPEG at decreasing quality, then smaller dimensions, all under the 8 MB per-image cap. A large HEIC/panorama is never materialized as a full raster. Returns `Data` only across the task boundary (Sendable-safe).

2. **Concurrent picker batches bypass caps** — the count (10), total-byte (32 MB), and per-image (8 MB) caps are now enforced atomically inside `MobileShellComposite.addPendingAttachment`, computed against the current staged set at mutation time on the `@MainActor`. Two racing batches can no longer both append past the cap. The view tracks the staging `Task` and cancels the prior one on a new batch; it treats the store as the authoritative cap (its own checks are a pre-filter only). Sign-in generation guard kept.

3. **Removed attachment still uploaded mid-send** — `submitComposer` re-checks each attachment is still present in `pendingAttachmentsByTerminalID` for the captured terminal before uploading, so a chip deleted mid-send is skipped. Failure-keeps-attachments and re-entrancy behavior unchanged.

Tests: added store-enforced count/byte cap coverage (including racing adds against the same starting budget) in `ComposerPendingAttachmentTests` and a mid-send-removal-skips-upload test in `ComposerSubmitRoutingTests`. `swift test --filter "ComposerPendingAttachmentTests|ComposerSubmitRoutingTests"`: 28/28 pass. File-length budget green.

lawrencecchen:
Addressed the two remaining memory-safety findings (commit 42847cf99).

**Finding 1 (orphaned bytes for stale terminal ids):** The base `addPendingAttachment` now validates the target id is in the current topology via a shared `terminalExistsInTopology` check, so both add paths (base and session-guarded) funnel through one existence gate and there is no unchecked entry point. Added `prunePendingAttachmentsForMissingTerminals()`, called from the `workspaces` `didSet` (the single topology funnel: every sync/merge/replace goes through it), which drops dictionary entries whose terminal id is no longer present, releasing multi-MB staged bytes on a topology change instead of holding them until sign-out. Sign-out clear unchanged. New host tests: base-path add for a missing id is rejected; a topology update that drops a terminal prunes its attachments while a surviving terminal (and one moved to another workspace) keeps its.

**Finding 2 (picker slurped full original asset):** Replaced `loadTransferable(Data.self)` with a file-backed `ImportedImageFile` Transferable (`FileRepresentation(contentType: .image)` copying the import to a temp URL), so a huge ProRAW/DNG/panorama never enters memory as Data. The picker size-gates the file on disk (60 MB raw bound, above the 8 MB per-image cap since HEIC decodes larger) before reading any bytes, downsamples directly from the URL via `CGImageSourceCreateWithURL` (full-res raster never materialized), and deletes the temp file after encoding. Per-image/total caps, atomic store enforcement, the session-generation guard, and staging Task cancellation are all preserved; only Data/String/URL cross concurrency boundaries.

Tests: `ComposerPendingAttachmentTests|ComposerSubmitRoutingTests` -> 31/31 pass. File-length budget respected (refreshed actuals for the two touched files).

lawrencecchen:
Addressed the two remaining autoreview findings (commit a8461d04a).

Finding 1 (global byte cap): added `maxPendingAttachmentTotalBytesAllTerminals` (64 MB) and `maxPendingAttachmentCountAllTerminals` (20), enforced atomically in `addPendingAttachment` after the per-terminal checks as a hard reject. The sum across all terminal keys is consistent because the add path runs on @MainActor. Per-terminal caps remain. New host tests cover the multi-terminal case: global byte and count budgets bind while each terminal stays under its own cap, and the per-terminal cap still binds under global headroom. `swift test --filter "ComposerPendingAttachmentTests|ComposerSubmitRoutingTests"` is green (34 tests).

Finding 2 (cancelled picker batches): replaced the fire-and-forget `Task.detached` in `prepare` with a structured background-priority child task group so cancellation propagates from the staging task into the ImageIO decode. Added `Task.isCancelled` checks before launching, before the heavy `CGImageSourceCreateWithURL`, and before the thumbnail encode. The staging task is now cancelled on composer `.onDisappear` and on a `terminalID` change, in addition to the existing cancel-on-new-batch. Temp files are still cleaned on the cancellation path via the existing per-iteration `defer`. Only Sendable `Data`/`URL` cross concurrency boundaries; UIImage is still built on the main actor.

File-length budget guard green (bumped the two touched files to their new actuals).

lawrencecchen:
Fixed the P1 stale-session continuation finding in `submitComposer()`.

The submit loop awaited a `paste_image` RPC per image, and `sendRemoteTerminalPasteImage` returns true even when a superseded connection answered, so a sign-out / account switch / Mac switch / reconnect that landed mid-flight let the loop send the next staged image and then the captured text through whatever `remoteClient` is now current, leaking the previous user's or previous Mac's unsent content into a different session.

Fix: capture a submit-time identity once at the start of the run and re-check it before every image send and before the text send via `isComposerSubmitIdentityCurrent(signIn:connection:)`. Identity is `signInGeneration` (sign-out / account switch, bumped in `signOut()`) plus `connectionGeneration` (Mac switch / reconnect / disconnect, bumped wherever the remote client/transport is replaced). Both already existed. On mismatch the whole submit aborts (loop stops, text is not sent) and attachments and text stay staged for a retry; the abort itself clears nothing.

Tests (ComposerSubmitRoutingTests, over the real RPC wire): a sign-out and a connection swap each between the first and second image send abort the submit so no further image and no text reach the new session, and the connection-swap case asserts the unsent attachment and text stay staged. Verified red without the guard (text reaches the new session, staged content lost), green with it. 36 tests pass; file-length budget respected.

PR #6305, opened 2026-06-17
Stop main window drifting down on sleep/wake

sergej-koscejev:
> _Note: this PR was prepared by an AI agent (Claude) on the author's behalf._

## Problem

After the Mac sleeps and wakes (e.g. you lock with `Ctrl+Cmd+Q` and come back later, or the display sleeps), the cmux main window shifts downward, and it accumulates across sleep/wake cycles. The lock keystroke itself is not the trigger — the sleep→wake is.

## Cause

This is AppKit re-constraining the window, not cmux repositioning it. None of cmux's own observers move the window here (`sessionDidResignActive` only saves a snapshot; `didWake` only restarts the socket listener; `applicationDidBecomeActive` only re-places windows when none is visible).

On a display/system sleep→wake, AppKit re-runs its constrain pass (`constrainFrameRect(_:to:)`) over every window. The default implementation does **not only** clamp off-screen windows back into view — it also repositions windows that are *already fully on-screen*, which is what shows up as the per-cycle creep. The exact reposition is AppKit-internal and depends on the display arrangement and each screen's menu-bar / safe-area insets, so:

- it is **not** a fixed titlebar-height nudge (the move is larger), and
- it is **not** limited to a window whose titlebar sits under the menu bar — it also hits e.g. a window occupying the bottom half of an external display, and likely other arrangements.

The main window is `.fullSizeContentView`, and cmux disables OS window restoration (`isRestorable = false`), re-applying its saved frame only at startup. Because nothing re-asserts the saved frame after wake, whatever AppKit's re-constrain produced sticks and accumulates.

## Fix

Override `CmuxMainWindow.constrainFrameRect(_:to:)` to leave an already-reachable frame untouched, only falling back to AppKit's default constraining when the proposed frame would otherwise be stranded off-screen (e.g. a display was disconnected), so a lost window can still be pulled back into view. cmux already owns and clamps its own window placement, so deferring to AppKit's re-constrain here only produced the drift. The reachability decision is factored into a pure, testable helper.

## Tests (two-commit red/green)

1. **Failing test** — `testConstrainPreservesOnScreenFrameOverlappingMenuBar` asserts that constraining an on-screen frame leaves it untouched (a titlebar-under-menu-bar frame is used as one easy, deterministic on-screen case). Verified red-by-assertion without the fix (inherited `NSWindow.constrainFrameRect` moves it).
2. **Fix** — adds the override plus deterministic, screen-agnostic coverage of the `shouldPreserveFrameDuringConstrain` helper (inside / menu-bar-overlap / stranded / barely-peeking / no-screens).

A third commit corrects code/test comments that had mis-stated the cause. All 6 tests pass green locally on macOS 26.5.1 (M2 Max). Test file is wired into `project.pbxproj` and passes `lint-pbxproj-test-wiring.sh`.

## Verification

Reproduced and confirmed fixed on the reporter's macOS (26.5.1 Tahoe) via a tagged Debug build and a real sleep/wake cycle — the window now stays put with no cumulative creep.

See also the note below on overlap with #2667.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window positioning during sleep/wake to prevent the app from subtly shifting the window when the proposed location is already reachable on-screen, especially when the title bar overlaps the system menu bar area.

* **Tests**
  * Added debug-only regression and unit tests to verify frame-preservation behavior for on-screen, off-screen, edge-overlap, and empty visible-region scenarios.

### DIFF
--- a/Sources/App/CmuxMainWindow.swift
+++ b/Sources/App/CmuxMainWindow.swift
@@ -97,6 +97,58 @@ final class CmuxMainWindow: NSWindow {
         guard !isSoftHiddenForVisibilityController else { return }
         super.flagsChanged(with: event)
     }
+
+    /// cmux owns main-window placement: it persists and restores window frames
+    /// itself and disables AppKit window restoration (`isRestorable = false`),
+    /// re-applying the saved frame only at startup.
+    ///
+    /// On a display/system sleep→wake (the kind a locked Mac eventually goes
+    /// through — the lock keystroke itself is not the trigger) AppKit re-runs
+    /// its constrain pass over every window. The default implementation does not
+    /// only clamp off-screen windows back into view; it also repositions windows
+    /// that are *already fully on-screen*, which is what we observe as the
+    /// window creeping each sleep cycle. The exact reposition is AppKit-internal
+    /// and depends on the display arrangement and each screen's menu-bar /
+    /// safe-area insets, so it is neither a fixed titlebar-height nudge nor
+    /// limited to a window whose titlebar sits under the menu bar — it also hits
+    /// e.g. a window in the bottom half of an external display, and likely other
+    /// arrangements. Because cmux never re-asserts the saved frame after wake,
+    /// whatever the re-constrain produced sticks and accumulates.
+    ///
+    /// Fix: refuse the re-constrain for any frame that is already reachable on
+    /// some screen, and defer to AppKit's default only when the frame would
+    /// otherwise be stranded off-screen (e.g. a display was disconnected), so a
+    /// genuinely lost window can still be pulled back into view.
+    override func constrainFrameRect(_ frameRect: NSRect, to screen: NSScreen?) -> NSRect {
+        if Self.shouldPreserveFrameDuringConstrain(
+            frameRect,
+            visibleFrames: NSScreen.screens.map(\.visibleFrame)
+        ) {
+            return frameRect
+        }
+        return super.constrainFrameRect(frameRect, to: screen)
+    }
+
+    /// Whether `proposedFrame` is reachable enough across `visibleFrames` that
+    /// AppKit's constraining pass should be skipped. The frame qualifies when it
+    /// overlaps some screen's visible area by at least `minimumVisibleExtent`
+    /// points in both dimensions (or its full extent, when smaller) — i.e. a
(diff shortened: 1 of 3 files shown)


sergej-koscejev:
The PR is 100 % vibe-coded by Claude Code (Opus 4.8), let me know if that's a problem. I have no experience with Swift whatsoever, I just was annoyed by the bug. Thanks!

sergej-koscejev:
> Commit adds multiple artifact files violating source-control-artifacts.md: .vercel/ (explicitly prohibited), .cursor/, .greptile/, .coderabbit.yaml (tool configs), .agents/skills, and .claude/ 

Huh? It does no such thing?

sergej-koscejev:
> _Note: this comment was written by an AI agent (Claude) on the author's behalf._
>
> _Edit: corrected the cause framing — the trigger is sleep→wake, not the lock keystroke; see the updated PR description._

### Overlap with #2667

Heads-up for reviewers: #2667 ("Fix window collapsing to sliver after disconnecting external displays") works in the same window-frame-on-display-reconfiguration machinery, so the two should be reconciled even though they don't conflict textually.

- **No merge conflict.** #2667 changes `AppDelegate.swift` / `Workspace.swift` / `SessionPersistenceTests.swift`; this PR changes `CmuxMainWindow.swift` + a new test file.
- **Behavioral interaction to be aware of.** #2667 adds a `NSApplication.didChangeScreenParametersNotification` handler that reconciles main-window frames and calls `window.constrainFrameRect(targetFrame, to: screen)`, relying on AppKit's clamping. This PR overrides `CmuxMainWindow.constrainFrameRect` so it returns an already-reachable frame **unchanged**, only falling back to `super` (normal clamping) when the frame would be stranded off-screen. So #2667's sliver/off-screen recovery cases still clamp as before, but its on-screen targets pass through untouched (which is what it already intends).
- **Possible solution overlap.** A display/system sleep→wake can itself post `didChangeScreenParameters`. If so, #2667's handler would independently re-assert the persisted frame on wake — addressing the same drift via a more active mechanism than this passive constrain override. If #2667 is revived, it may be cleaner to fold the sleep/wake case into its broader screen-parameter reconciliation rather than keep a separate override.

Flagging so whoever lands either PR knows `CmuxMainWindow.constrainFrameRect` no longer always clamps.

PR #6392, opened 2026-06-18
Eliminate flaky tests: deterministic rewrites across Swift, Python, shell

azooz2003-bit:
One PR that fixes **every flaky test discovered in this run**: 95 findings across 60 files, found and fixed via Claude Code dynamic workflows (parallel discovery → parallel fix → adversarial review → parallel rewrite) over ~950 test files.

Flaky = nondeterministic pass/fail on identical correct code. Deterministic test sleeps (explicitly allowed by the repo review rules) are left untouched.

## Fix categories
- **sleep-as-sync / sleep-then-assert** → wait on the real completion signal (await it, inject a virtual clock, or a deadline-bounded poll of a real predicate).
- **tight-timeout** → widened (failure path only; passing runs return immediately).
- **wall-clock-assert / assert-on-duration** → best-of-N minima + deterministic work-count, or a causality assertion on the logical outcome instead of latency.
- **order-dependence** → `.serialized` suites / per-test state reset.
- **network / port races** → local/ephemeral servers, retries + soft-skip.
- **activation perf gate** (`scripts/perf-activation-session.py`) → best-of-N timing; wall-clock budgets made advisory, structural budgets stay blocking.

## Verification
- Determinism checker (from the gate PR #6399): **0 active findings**.
- Python/shell: all changed files pass `py_compile` / `bash -n`.
- Swift: all 8 changed SPM package test targets compile via `swift build --build-tests`.
- `cmuxTests/`/`cmuxUITests/` (Xcode project): CI.
- Adversarial review of all uncompiled Swift diffs: 0 blockers.

Test/CI-infra only; no app/runtime/UI behavior change (no dogfood required).

Related: #6394 (runner stale-lock self-heal), #6399 (CI gate that bans new flaky primitives).

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability across the codebase by replacing fixed timing delays with polling-based synchronization and deadline-bounded waits, reducing flakiness in asynchronous operations.
  * Tightened performance benchmarks and adjusted timing expectations to better account for CI load variations.
  * Enhanced test determinism through explicit event-driven coordination and sampling strategies.


azooz2003-bit:
Addressed review feedback in 5d4c282 (test-only):

- **test_browser_goto_split.py** — `_wait_url_loaded` now raises `AssertionError` at the deadline instead of silently returning (no more false pass when the page never leaves `about:blank`), and both tests manage the local HTTP server via `with _local_test_server() as page_url:` so the server thread/socket cannot leak on setup exceptions. (CodeRabbit + Greptile)
- **test_surface_move_reorder_api.py** — the cleanup re-select of `ws0` now asserts `_wait(...)` converged and raises `cmuxError` otherwise, preventing state leak into later runs. The two earlier `_wait` calls are already followed by explicit re-read assertions, so they stay loud.
- **test_homebrew_sha.sh** — soft-skip is now scoped to transient transport failures (`000|408|429|5??`); deterministic client errors like 404 (missing/renamed release asset) fail hard.
- **test_visual_typing_char_by_char.py** — deviated from the literal `endswith` suggestion on purpose: this test types into the user's real interactive shell, where zsh-autosuggestions render trailing glyphs after the cursor, so `endswith` would itself flake. Instead I capture the static prompt line once and validate the typed prefix against the post-prompt remainder, which excludes a `cmux` already in the prompt/path (the real concern) while staying robust to trailing autosuggestion glyphs.

All four files pass `py_compile` / `bash -n`.

azooz2003-bit:
Round 2 (7e75ad4) — addressed the remaining 5 comments:

- **test_pane_break_swap_preserve_focus.py (🔴 Critical)** — real `NameError`: the lambda read `p` in the ternary condition before the walrus assigned it. Fixed by binding the walrus inside the condition (`p if len(p := _panes(...)) == 2 else None`), keeping a single socket read and returning the panes. Verified at runtime.
- **perf-activation-session.py:646** — `actual is None` now appends to `failures` unconditionally; a missing measurement is a benchmark-contract failure, not a load-sensitive timing value, so it stays blocking even in advisory timing mode.
- **perf-activation-session.py:698** — store `dict(real_scrollback)` so `best_of_snapshot_timing`’s in-place mutation no longer clobbers the raw `snapshot_with_real_scrollback` capture (same dict object).
- **NotificationAndMenuBarTests.swift:285** — did not take the literal `XCTWaiter().wait` suggestion: a synchronous wait risks starving the evaluation task. Instead, the result is stored in a lock-guarded box; on timeout the test fails fast (`XCTFail` + cancel) without awaiting `evaluationTask.value`, which could otherwise hang until the suite timeout if the stalled hook ignores cancellation.

All Python/shell touched files pass `py_compile`/`bash -n`; the Swift test compiles on CI.

PR #6452, opened 2026-06-19
Flaky-test follow-up: review fixes + second-pass team sweep

azooz2003-bit:
Follow-up to #6392 (the first flaky-test sweep), which merged with only the original 44-file content. These 5 commits were authored after that merge and never landed — this PR brings them onto current `main`.

### What's here

1. **Review-comment fixes for #6392** (`9b39678`, `178db58`) — the CodeRabbit/Greptile feedback on the original sweep, including a 🔴 critical walrus `NameError` the sweep itself introduced (`(p := …) if len(p)==2` read `p` before assignment). Also: `_wait_url_loaded` raises on timeout instead of returning silently; `with`-managed local HTTP server; `homebrew_sha` soft-skip scoped to transient codes (404 fails hard); `perf-activation-session.py` missing-measurement is now always blocking + a `dict()` copy so best-of mutation can't clobber the raw capture; the notification stall test fails fast via a lock-guarded result box.

2. **`reports.md`** (`de48ffb`) — consolidated output of an in-session discovery workflow (20 parallel agents over the 419 signal-bearing test files of 2080). 101 candidate findings, concentrated in `Packages/` unit tests the first sweep never touched.

3. **37 verified second-pass fixes** (`5c7c76c`) — applied by a coordinated area-team workflow (per-area lead verifies + plans shared-helper edits; single owner applies them; parallel members own one file each). 9 candidates were adversarially refuted and left (deterministic-allowed sleeps, factually-wrong premises) — documented in `reports.md`. Highlights: injected `ManualTestClock` for real-clock timeout tests; `_ = await poll` → `#expect(await poll)`; per-PID/UUID socket+port paths; order-dependent session-restore now scans all workspaces; latent `NameError`s fixed; surface-targeted sends to avoid focus races.

4. **MobileCoreRPC cancelled-while-queued race** (`7d257a5`) — fixed deterministically. The writer-gate state is private to the production `MobileCoreRPCSession` actor, so there is no test-only signal; this adds a `#if DEBUG`, read-only `debugQueuedRequestCount()` accessor (in the file's existing `#if DEBUG` test-support extension, next to `debugWithRequestTimeout`). It is compiled out of Release and changes no shipping behavior.

### Scope / validation

All changes are test-only except the one `#if DEBUG` read-only accessor (zero shipping-build behavior). Locally: all changed Python pass `py_compile`, shell passes `bash -n`, and all 5 touched packages pass `swift build --build-tests`. No test was executed against a live socket.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/6452"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784486276&installation_id=133855650&pr_number=6452&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F6452&signature=76cc143c6750a5401703e83ef24665034ffcdde87e96e59fea1f75b3e15f889a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Follow-up to the first flaky-test sweep: restores missed commits, fixes 37 more flaky tests, exposes the `CmuxMobileRPC` writer-queue state to tests via `@testable` (widened to internal; no debug hooks, no shipping change), refreshes the Swift file-length budget, and applies review cleanups that remove wall-clock asserts and fixed sleeps. Outcome: more reliable CI with no production behavior changes.

- **Bug Fixes**
  - Fail-loud guards and correctness: fixed walrus `NameError`; `_wait_url_loaded` now raises on timeout; `homebrew_sha` soft-skips only transient HTTP codes (404 fails hard); local HTTP server managed with `with`.
  - Deterministic timing: injected `ManualTestClock` into `CmuxAuthRuntime` sign-out and hosted sign-in slow-hint; sleeps replaced with pollers/RunLoop waits; UITest socket waits use shared helpers.
  - Reduced flakiness: `CmuxCommandPalette` benchmarks are advisory (BENCH logs only); redundant wall-clock asserts and fixed sleeps removed (ShellStartupMatrix, MultiWindowNotifications, tab dragging); asserts moved to causal points.
  - Isolation and cleanup: per-PID/UUID tmp files and ports; `mkstemp` focus files; safer temp/FD handling; `RemoteCLIRelayServerTests` capture errno before close; robust server/subprocess teardown.
  - Focus/input races and order assumptions: surface-targeted sends; restore checks scan all workspaces; waits validate real selection/focus signals; typed-text checks ignore static prompt content.
  - Performance tooling: missing perf measurements are now blocking; snapshot metrics copied to avoid clobbering.
  - `CmuxMobileRPC` cancel-while-queued race: tests observe `session.queuedRequestIDs` via `@testable import` after widening from `private` to internal (no `#if DEBUG` accessors in production sources).
  - Added `reports.md` with second-pass discovery results; updated Swift file-length budgets for touched tests.

<sup>Written for commit b64a0c14f5487ddf0483943cddaab39ef3446d7d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6452?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **Tests**
  * Improved reliability by replacing real-time sleeps with deterministic manual-clock timing and explicit state polling across multiple scenarios.
  * Strengthened failure signaling by converting “silently waited” checks into assertions and tightening timeouts/diagnostics when expectations aren’t met.
  * Reduced benchmark CI noise by downgrading brittle performance comparisons to advisory BENCH-only output.
  * Enhanced isolation and cleanup with safer resource handling (sockets/process FDs), more resilient teardown, and per-run temporary files/paths.


azooz2003-bit:
Addressed review in 5eb118b (test-only):

- **ShellStartupMatrixTests** — dropped the redundant `duration < 5.0` ceiling; the bootstrap already runs under a 5s `runProcess` timeout, so `status==0` + `!timedOut` is the causal completion signal. (CodeRabbit + Greptile)
- **MultiWindowNotificationsUITests** — assert no-foreground at the causal point (`waitForCommandCompletionWhileBackgrounded`) instead of polling a fixed 2s window. (CodeRabbit)
- **test_tab_dragging.py** — removed the two `time.sleep(1.5)` waits; the file-content poll loop below is the readiness signal. (CodeRabbit)
- **test_terminal_focus_routing.py** — `tempfile.mktemp` → `mkstemp` (atomic, clears Ruff S306). (CodeRabbit)
- **test_session_restore_stress_kill_cycles.py** — fixed 0.1s settle → deadline-bounded poll on the is-selected signal. (CodeRabbit)
- **RemoteCLIRelayServerTests** — capture `errno` before `close()`. (CodeRabbit)

Two I am intentionally NOT taking:

- **CommandPaletteSearchEngineTests ratio bound (Greptile, → `* 1.10`)** — these are relative optimized-vs-reference comparisons (both measured back-to-back in the same run), already loosened here to `1.25`/`2.0` precisely because the tighter band flaked under CI scheduling noise. Tightening to `1.10` reverses the de-flake and would reintroduce the exact flakiness this PR removes, so I am keeping the noise-tolerant bound.
- **`reports.md` is generated output (Greptile)** — fair point in general, but this file is the explicit consolidated-findings deliverable this audit task asked for, so it is intentional rather than incidental tool output. Happy to drop it or move it under `docs/`/`scripts/audit/` if maintainers prefer.

azooz2003-bit:
Follow-up on the CommandPalette benchmark point (5e4d2fc): you were right that the loosened ratio bands (esp. `0.80` → `1.25`) stopped validating anything. Rather than pick a magic multiplier, I made all four search benchmarks **advisory**: keep the `BENCH …` diagnostic prints for trend tracking, drop the flaky wall-clock `#expect` ratio/dropped-frame asserts. Rationale: the engine exposes no preparation/work counter, so there is no causal signal to assert on here, and a same-run wall-clock ratio is load-sensitive on shared CI (tight → flakes, wide → meaningless). This matches the repo test-time policy and the activation-session perf gate (which is where real latency regressions are actually gated). Each test still exercises both code paths end to end.

Issue #6381, opened 2026-06-18
Main-thread deadlock in session autosave tick via ghostty_surface_needs_confirm_quit (0.64.16)


liugangsheng:
## Bug description

cmux 0.64.16 fully hangs (spinning beachball, never recovers) when the session autosave tick fires while a terminal surface is wedged on its internal `os_unfair_lock`. The main thread parks in `__ulock_wait2` and never returns.

A `sample` of the wedged process shows the main thread permanently blocked here:

```
2253 Thread_108112   DispatchQueue_1: com.apple.main-thread  (serial)
+ 2253 specialized AppDelegate.finishSessionAutosaveTick(source:generation:) + 568
+   2253 AppDelegate.saveSessionSnapshot(...) + 544
+     2253 AppDelegate.buildSessionSnapshot(...) + 1028
+       2253 TabManager.sessionSnapshot(...) + 1340
+         2253 Workspace.sessionSnapshot(...) + 4948
+           2253 Workspace.sessionPanelSnapshot(panelId:includeScrollback:restorableAgent:resumeBinding:) + 8592
+             2253 ghostty_surface_needs_confirm_quit + 92
+               2253 _os_unfair_lock_lock_slow + 172
+                 2253 __ulock_wait2 + 8   <-- main thread parked forever
```

So the autosave tick walks every panel, calls `terminalPanel.needsConfirmClose()`, which calls `ghostty_surface_needs_confirm_quit(surface)` synchronously on the main thread. That ghostty C API takes an internal `os_unfair_lock` that is also contended by the surface's renderer/io threads. When those threads are themselves blocked on the same lock (e.g. a wedged surface mid-teardown), the main thread parks in `_os_unfair_lock_lock_slow` -> `__ulock_wait2` and never returns, beach-balling the whole app.

This is a hard deadlock, not a transient hitch. `kill -9` is the only recovery.

## Expected behavior

The autosave tick should never block the main thread on a per-surface lock. A wedged surface should be skipped (or time out), and the app must keep responding.

## Steps to reproduce

Hard to reproduce deterministically since it requires the surface's renderer/io threads to be stuck on the lock at the moment the autosave tick fires, but the trigger is:

1. Open several workspaces/tabs with shells running (I had a mix of long-running shells and agent sessions).
2. Leave cmux running for a few hours.
3. Eventually the session autosave tick fires while a surface is wedged -> full hang.

[...]

Issue #2387, opened 2026-03-31
Regression: session persistence no longer restores workspaces/terminals on relaunch


austinywang:
## Bug

Session persistence has regressed — relaunching cmux no longer restores the previous workspace layout, terminal sessions, or scrollback. This previously worked.

## Expected behavior

On quit + relaunch, cmux should restore:
- All workspaces and their names
- Pane/split layout within each workspace
- Terminal scrollback content
- Working directories
- Browser pane URLs

## Actual behavior

Relaunching cmux starts with a blank state — no workspaces, terminals, or layouts are restored.

## Notes

- This is a regression from previously working behavior
- Related but distinct from https://github.com/manaflow-ai/cmux/issues/480 (original feature), https://github.com/manaflow-ai/cmux/issues/1984 (Sparkle-specific), https://github.com/manaflow-ai/cmux/issues/2072 (macOS update)
- Need to identify what change broke the save/restore path

## Priority

High — session persistence is critical for daily use.

walid-mos:
Hello ! Is someone working on this ? Got the same problem here ? 

Could see with AI what is broking down but swift is not a language I am really familiar with sadly

fpigeonjr:
Confirming this regression on macOS 26.4.1. After logging out and back in, cmux launches with a blank state — all workspaces are gone. Previously they were restored on relaunch.

imjanehsieh-maker:
Still reproducing on cmux `0.64.3 (83) [aea6cfcde]`, macOS Darwin 24.6.0, with new evidence that this is more than the workingDirectory loss described in #2125: **an entire panel entry is dropped from the session JSON** during the restore→re-save cycle.

### Concrete diff

Same workspace ("主編"), before vs. after Cmd+Q + relaunch:

**Before quit (`session-com.cmuxterm.app-previous.json`)** — 6 panels:
- 文章規劃 — `69379D9C-741D-402A-92B0-D01E6EB25CE9`
- 內容優化 — `DC55D44F-0374-4C9E-BD87-C55D27B9FEA2`
- **Threads — `5A445B4A-1453-4500-8EFF-ADFCA0430C00`**
- 內容磁鐵 — `D6589686-2CD1-4673-9D26-8D050E18B41F`
- EDM — `E0F8F8D0-03C3-472E-9668-E0BDD191BEC6`
- 優惠推廣 — `E8DE6517-7570-44B6-945E-63E30AD8F155`

[...]

Issue #1340, opened 2026-03-13
Save scrollback on autosave timer (not just clean quit)


vitto32:
Currently scrollback is only persisted on clean termination (`applicationShouldTerminate`, `willPowerOff`). The autosave timer (every 8s) passes `includeScrollback: false`, so a crash or force-kill loses all scroll history.

tmux-resurrect saves scrollback every 15 minutes by default — a good middle ground between performance and durability.

**Proposal:** add a separate, less-frequent timer (e.g. every 5–15 min) that saves with `includeScrollback: true`. The existing autosave (8s, layout-only) stays unchanged.

This way a crash loses at most 15 min of scrollback instead of everything.

PR #6615, opened 2026-06-22
Fix terminal scrollback not restoring on quit/relaunch

### LINKED ISSUE #2016: Restore terminal scrollback content across app restarts (like iTerm2 / macOS Terminal.app)
## Feature Request

When cmux is quit and restarted, all terminal panes come back blank — the layout and tabs restore, but the scrollback content is gone.

**iTerm2** and **macOS Terminal.app** both persist and restore the scrollback buffer text across app restarts, so you can scroll up and see your previous output. This is extremely useful for:

- Reviewing previous command output after an app update or crash
- Continuing where you left off without re-running commands
- Keeping context when working across multiple agent sessions

## Current Behavior
- Tabs, layout, and working directories restore ✅
- Terminal scrollback content is blank ❌

## Expected Behavior
- Scrollback buffer text should be persisted to disk and restored on relaunch

## Upstream

This is a limitation inherited from Ghostty's `libghostty`, which also doesn't persist scrollback. There's an open issue upstream: [ghostty-org/ghostty#1847](https://github.com/ghostty-org/ghostty/issues/1847).

Ideally this lands in libghostty so cmux gets it for free — filing here mainly as a signal that this is a top-requested feature for agent-heavy workflows.

### LINKED ISSUE #2823: Sessions are lost after app restart
### Problem or motivation

User sessions are not being persisted correctly. When the application is closed and reopened, the existing session is destroyed, forcing users to log in again.

This behavior negatively impacts user experience, especially for apps expected to maintain persistent login state.

### Proposed solution

This seems to be very impacting to productivity to loose the sessions created so far.

### Alternatives considered

_No response_

### Additional context

_No response_

grantland:
## Summary

Fixes #2823. Likely also resolves #2016 (restore terminal scrollback content across app restarts) — confirm scope and change to `Fixes #2016` if it fits.

(The earlier `150dc9036` addressed the *glued prompt* replay sub-symptom of #2823; this fixes the remaining half — scrollback was never *captured/persisted*, so terminals restored empty. Capture + the existing replay fix together make scrollback restore.)

Related / adjacent (not closed by this PR):
- #2016 — restore terminal scrollback across app restarts (likely the canonical ask; see above)
- #1340 — save scrollback on the autosave timer, not just clean quit (this PR is quit-time only)
- #2387 — regression: session persistence no longer restores workspaces/terminals on relaunch
- #6381 — main-thread deadlock in session autosave via `ghostty_surface_needs_confirm_quit` (same close-confirm machinery this PR decouples scrollback from)
- Separate internal defect found during this work — prod delivers `report_shell_state` but `shellState` stays `.unknown` at snapshot time — tracked as its own follow-up.

**What changed?**
Terminal scrollback is now persisted at session save unless the shell is *positively* running a command. Previously persistence was gated on `!closeConfirmationRequired`, which coupled it to idle/close-confirmation detection: when a panel's shell-activity state was unavailable (`.unknown`/`nil`), the conservative `needsConfirmClose` fallback skipped persistence and the terminal restored **empty**.

- `WorkspaceSessionRestorePolicyService.shouldPersistSessionScrollback` now takes `shellActivityState` and returns `(state ?? .unknown) != .commandRunning` (persist for `.promptIdle` **and** `.unknown`/`nil`; skip only `.commandRunning`).
- The session-snapshot gate in `Workspace.swift` passes the panel's shell state directly (the now-unused `closeConfirmationRequired` local is removed).
- The `Workspace.shouldPersistSessionScrollback` static wrapper keeps its signature; `fallbackNeedsConfirmClose` is explicitly retired (documented).
- `shouldReplaySessionScrollback` still excludes agent/tmux/resume terminals; the close-confirmation **warning** path (`resolveCloseConfirmation`) is unchanged.

**Why?**
Regression from `5f074f810` (2026-03-13, *"Fix session restore replay for transient terminal states"*). Before it, snapshots always captured scrollback when `includeScrollback` was set. That commit added a gate to avoid persisting transient/mid-command screens, but it over-applied: it treated `.unknown` (no shell-integration report yet) the same as transient, so any terminal without a fresh prompt-idle report stopped persisting scrollback. This fix narrows the gate to its original intent — skip only states we *positively* know are mid-command — so scrollback is no longer silently dropped when the shell state is merely unknown.

Note: this is the resilient behavior fix. There is a separate, deeper record-side issue (see Testing) where prod *delivers* prompt-idle reports but `shellState` still ends up `.unknown` at snapshot time; that is tracked as a follow-up and is not addressed here.

## Testing

**How did you test this change?**
- Added a red/green pair: a failing app-level regression test (`SessionPersistenceTests.testSessionScrollbackPersistsUnlessCommandRunning`) that asserts the corrected contract, then the fix. Added a package-level unit test (`WorkspaceSessionRestorePolicyServiceTests.scrollbackPersistencePolicy`) covering `.promptIdle`/`.unknown`/`nil`/`.commandRunning`.
- Empirically verified the behavior: with the gate change, a build whose `shellState` is always `.unknown` (the exact failing condition) **now captures** scrollback at quit (session snapshot gains a `scrollback` key with content) where it previously produced no `scrollback` key at all.

**What did you verify manually?**
- Reproduced the bug: prod session snapshots have no `scrollback` key for plain idle terminals (live prod: 7 plain terminals, 0 with scrollback).
- Instrumented the snapshot gate and confirmed capture was skipped because `shellState == nil → closeConfirmationRequired == true → shouldPersistScrollback == false`; the read path (VT export / live-surface) was never reached.
- Confirmed prod's shell integration delivers reports (socket live, `CMUX_TAB_ID`/`CMUX_PANEL_ID` set, `_CMUX_SEND_TOOL=nc`, socket replies `OK`) — so the bug is record-side, not send-side.

Manually tested locally: built the fix into a local Release-config app and confirmed at quit that a plain terminal with `shellState=nil` now writes a `scrollback` key with content to the session snapshot (previously the key was absent). Caveat: the automated red/green suites (`swift test`/`xcodebuild test`) run on CI — please confirm green there. Harness note: builds via `scripts/reload.sh` have working shell-integration reporting (`shellState` becomes `.promptIdle`), while plain `xcodebuild` builds do not — Debug-vs-Release was a red herring. Verify persistence with `shellState` forced (the unit tests), not by eyeballing a local build.

## Demo Video

For UI or behavior changes, include a short demo video (GitHub upload, Loom, or other direct link).

- Video URL or attachment: _TODO_ — record: open a plain terminal, run a few commands, Cmd+Q, relaunch → scrollback contents restored (before this change they were blank).

```text
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review
```

## Checklist

- [x] I updated docs/changelog if needed (not needed — changelog is generated from commits at release; no documented behavior changes)
- [x] I requested bot reviews after my latest commit (trigger block posted as a PR comment)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced terminal session restoration to intelligently preserve scrollback based on shell activity. Session history is now retained when the shell is idle or in unknown states, and only excluded when commands are actively running.

### DIFF
--- a/Packages/macOS/CmuxWorkspaces/Sources/CmuxWorkspaces/Session/WorkspaceSessionRestorePolicyService.swift
+++ b/Packages/macOS/CmuxWorkspaces/Sources/CmuxWorkspaces/Session/WorkspaceSessionRestorePolicyService.swift
@@ -184,8 +184,15 @@ public struct WorkspaceSessionRestorePolicyService<Binding: WorkspaceSurfaceResu
         ).restorableTmuxStartCommand(rawCommand)
     }
 
-    /// Returns whether terminal scrollback should be persisted when closing/restoring.
-    public func shouldPersistSessionScrollback(closeConfirmationRequired: Bool) -> Bool {
-        !closeConfirmationRequired
+    /// Returns whether terminal scrollback should be persisted at session save.
+    ///
+    /// Persistence depends only on whether a command is *positively* running:
+    /// persist for `.promptIdle` and for `.unknown`/`nil` (no shell-integration
+    /// report), and skip only for `.commandRunning`. It must NOT be coupled to
+    /// close-confirmation — an unknown shell state would otherwise route through the
+    /// conservative `needsConfirmClose` fallback and drop scrollback entirely, so
+    /// terminals restored with no contents.
+    public func shouldPersistSessionScrollback(shellActivityState: PanelShellActivityState?) -> Bool {
+        (shellActivityState ?? .unknown) != .commandRunning
     }
 }
--- a/Packages/macOS/CmuxWorkspaces/Tests/CmuxWorkspacesTests/Session/WorkspaceSessionRestorePolicyServiceTests.swift
+++ b/Packages/macOS/CmuxWorkspaces/Tests/CmuxWorkspacesTests/Session/WorkspaceSessionRestorePolicyServiceTests.swift
@@ -290,4 +290,18 @@ struct WorkspaceSessionRestorePolicyServiceTests {
         #expect(service.restorableTmuxStartCommand("hudson omx") == nil)
         #expect(service.restorableTmuxStartCommand("omx hud") == "omx hud")
     }
+
+    @Test("scrollback persists unless a command is positively running")
+    func scrollbackPersistencePolicy() {
+        let service = makeService()
+
+        // Persist for idle shells.
+        #expect(service.shouldPersistSessionScrollback(shellActivityState: .promptIdle))
+        // Regression: persist when shell-integration state is unavailable, so
+        // terminals don't restore empty. (.unknown and never-reported nil.)
+        #expect(service.shouldPersistSessionScrollback(shellActivityState: .unknown))
+        #expect(service.shouldPersistSessionScrollback(shellActivityState: nil))
+        // Skip only when a command is positively running.
+        #expect(!service.shouldPersistSessionScrollback(shellActivityState: .commandRunning))
+    }
 }
--- a/Sources/Workspace.swift
+++ b/Sources/Workspace.swift
@@ -481,12 +481,8 @@ extension Workspace {
                 promptForApproval: false,
                 approvalStoreURL: SurfaceResumeApprovalStore.defaultURL()
             )
-            let closeConfirmationRequired = Self.resolveCloseConfirmation(
-                shellActivityState: panelShellActivityStates[panelId],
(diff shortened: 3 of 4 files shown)


grantland:
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review

grantland:
Addressed in cc515c74a — removed the dead `fallbackNeedsConfirmClose` parameter from the static wrapper and updated the test call sites to the single-argument form.

grantland:
@coderabbitai review
@greptile-apps review

PR #6686, opened 2026-06-23
Fix idle terminals stuck at shellState .unknown (record side of #6618)

### LINKED ISSUE #6618: Shell-activity reports are delivered over the control socket but `shellState` stays `.unknown` (idle terminals treated as unknown/busy)
### cmux version

0.64.16

### macOS version

26.5.1 (25F80)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Built from source

I could not test NIGHTLY

### Bug description

A terminal's shell-activity state (`PanelShellActivityState`) stays `.unknown` even though the
shell **does** report `prompt`/`running` over the control socket. The report is delivered and
accepted (the socket replies `OK`), but it never becomes `shellState = .promptIdle` by the time
it's read (e.g. at session-snapshot time).

This is the record-side half of the scrollback-restore work in #2823. The persistence change in
that PR makes scrollback robust to `.unknown` (so the user-visible "scrollback doesn't restore"
symptom is fixed), but this underlying reporting→record gap is a separate defect that also degrades
anything else keyed on `shellState`:
- close-confirmation warnings (`resolveCloseConfirmation` falls back to the conservative                                `needsConfirmClose` when state is `.unknown` → can prompt to confirm closing an idle terminal)
- PR probing (`scheduleWorkspacePullRequestRefresh` fires on `.promptIdle`)
Send-side is confirmed working; the defect is on the record side (received but not stored as
`.promptIdle`). The handler returns `OK` regardless of recording — recording is gated on                              `socketFastPathState.shouldPublishShellActivity` (dedup) → `DispatchQueue.main.async` → tab/panel                     lookup in `controlSurfaceReportShellState`.

Suspected causes (unconfirmed):
1. First-report / registration race: a fresh terminal's *first* `prompt` report arrives before the                       panel is registered → dropped. The shell's own dedup (`_CMUX_SHELL_ACTIVITY_LAST=prompt`,                             cmux-bash-integration.bash:616) then never re-sends → `shellState` stuck `.unknown`.
2. `shouldPublishShellActivity` dedup desync (marks published while the main-actor write is skipped).                 3. Surface-id keying mismatch between record and read.

### Expected behavior

A reported `prompt` (idle) should result in `shellState = .promptIdle` for that panel, so
idle terminals are recognized as idle (no spurious close-confirmation, PR probing fires,
scrollback policy sees an idle shell).

### Steps to reproduce

1. Open cmux, open a plain terminal, run a couple of commands, leave it idle at the prompt.
2. In that terminal, confirm the shell is reporting and delivery works:
   ```bash
   echo "tool=[$_CMUX_SEND_TOOL] sock=$CMUX_SOCKET_PATH live=$([ -S "$CMUX_SOCKET_PATH" ] && echo y || echo n) tab=$CMUX_TAB_ID panel=$CMUX_PANEL_ID last=$_CMUX_SHELL_ACTIVITY_LAST"
   printf '%s\n' "report_shell_state prompt --tab=$CMUX_TAB_ID --panel=$CMUX_PANEL_ID" | nc -w 1 -U "$CMUX_SOCKET_PATH"                                                                                                                        ```
   Observed: `tool=[nc]`, `live=y`, `tab`/`panel` set, `last=prompt`, socket replies `OK`.                            3. Quit cmux (Cmd+Q) and inspect the session snapshot — plain idle terminals have no `scrollback`
   key, which (pre-fix) means `shellState` was `.unknown` at snapshot:                                                   ```bash
   jq '[.windows[].tabManager.workspaces[].panels[]|select(.terminal!=null and .terminal.agent==null)]
       | {plain: length, withScrollback: (map(select(.terminal.scrollback!=null))|length)}' \
     "$HOME/Library/Application Support/cmux/session-com.cmuxterm.app.json"
   ```
   Observed: `{ "plain": 7, "withScrollback": 0 }` despite the shell reporting `prompt`.

### Shell and environment

bash login shell (`/bin/bash -lic`) with cmux shell integration sourced from                                          `$CMUX_SHELL_INTEGRATION_DIR`. `_CMUX_SEND_TOOL=nc` (`/usr/bin/nc`). No tmux.

                                         ## Relevant logs or crash reports
No crash. Diagnosed by instrumenting the snapshot gate and `controlSurfaceReportShellState`
(Release-capable logging, since `cmuxDebugLog` is `#if DEBUG`): the gate saw `shellState == nil`;
in a delivery-capable build `report_shell_state` reached the handler, but in prod the reported
state never lands as `.promptIdle` by snapshot time.

```text

```

_No response_

### Additional context

- Cannot be confirmed by instrumenting prod (official build) directly; reproducing the *drop*
  needs a build that both delivers (so not a bare `xcodebuild` build, which never delivers) and
  reproduces the miss (so not a `reload.sh` build, which records fine). That needs a proper                             build/instrumentation path (solve the `reloads.sh` Release signing wall, or use the AWS macOS                         builders per the `regression-hunt` playbook).
- Related: #2823 (scrollback restore — user-visible symptom fixed by the persistence PR), #2016,
  #1340, #2387, #3712, #6381 (same close-confirm autosave machinery).- Regression that surfaced the gate dependency: `5f074f810`
  ("Fix session restore replay for transient terminal states", 2026-03-13).

lawrencecchen:
## Summary

Fixes https://github.com/manaflow-ai/cmux/issues/6618.

A terminal's shell-activity state (`PanelShellActivityState`) stays `.unknown` even though the shell reports `prompt`/`running` over the control socket and the socket replies `OK`. The report is accepted but never lands as `.promptIdle`, which degrades everything keyed on `shellState`: close-confirmation warnings (`resolveCloseConfirmation` falls back to the conservative `needsConfirmClose`), PR probing (`scheduleWorkspacePullRequestRefresh` fires on `.promptIdle`), and scrollback policy.

This is the record-side half of the scrollback work in https://github.com/manaflow-ai/cmux/issues/2823. https://github.com/manaflow-ai/cmux/pull/6615 only made scrollback persistence *tolerant* of `.unknown`; it explicitly left this record-side defect as a follow-up.

## Root cause

The shell reports each transition exactly once and dedupes locally (`_CMUX_SHELL_ACTIVITY_LAST` is set *before* the send in `cmux-bash-integration.bash`, so it never re-sends an identical state). On the app side, `controlSurfaceReportShellState` dispatches to the main actor and resolves the workspace via `AppDelegate.tabManagerFor(tabId:)`. During session restore the `Workspace` and its panels already exist, but the workspace is not yet inserted into any window's `TabManager`, so the lookup returns `nil` and the report is silently dropped. The shell never re-sends, so `shellState` is stranded at `.unknown` until the next state transition (never, for an idle terminal). This is why it is build/timing dependent and why it hit all of the reporter's restored terminals.

Ruled out the surface-id keying suspicion: `CMUX_PANEL_ID == surface.id == TerminalPanel.id`, so record/read keys match.

## What changed

- Reports that cannot be applied yet are buffered per `(workspaceId, surfaceId)` on `AppDelegate` (bounded, like the socket fast-path dedupe) via a single shared `recordReportedShellActivity` entrypoint that both racy socket reporters now use.
- `TabManager.workspaceTabsWillChange` (the single chokepoint where a workspace joins a manager) replays the buffer against the now-reachable workspace. Only landed reports are cleared, so a surface still missing its panel retries on the next registration.
- `updatePanelShellActivityState` / `updateSurfaceShellActivity` now return whether the report landed, so the buffer knows when to retry vs clear.

No change to `SocketFastPathState`: nothing is lost anymore, so its optimistic dedupe stays correct.

The fix is principled: it makes the model the single source of truth and guarantees a reported state is never lost to a registration race, rather than papering over timing with a delay or retry.

## Testing

Two-commit red/green (`TabManagerShellActivityReplayTests.testBufferedShellStateReplaysWhenWorkspaceRegisters`):
- Commit 1 buffers the report but does not replay it → the buffered state never lands → red.
- Commit 2 wires the replay → green.

The test reports shell state for a workspace that is not reachable through `AppDelegate` (the exact unreachable condition that occurs mid session-restore), asserts it is buffered rather than dropped, then triggers a tab-set change and asserts the state replays onto the live panel.

## Review Trigger

```text
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review
```

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/6686"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784803797&installation_id=133855650&pr_number=6686&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F6686&signature=8f20e44842a07f3a510e335477b7e3ff25efb4cc057053bf4fe4f9a04a35d811"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches session-restore timing and terminal shell-state plumbing used by close confirmation and PR probing, but behavior is narrowly scoped to the pre-registration race with bounded buffers and new regression tests.
> 
> **Overview**
> Fixes **#6618**: control-socket `report_shell_state` updates were dropped when the workspace existed but was not yet reachable via `AppDelegate.tabManagerFor`, leaving panels stuck at `.unknown` after session restore (shell does not resend the same state).
> 
> **`AppDelegate.recordReportedShellActivity`** is now the single entrypoint for socket reporters: apply immediately when a `TabManager` owns the workspace, otherwise buffer the latest state per workspace/surface with bounded eviction. **`TabManager`** replays and clears each workspace bucket after `tabs` updates (`workspaceTabsDidChange` on `WorkspacesModel` / `WorkspacesHosting`) and again when a main window registers via **`flushPendingShellActivityForRegisteredWorkspaces`**.
> 
> **`updatePanelShellActivityState`** / **`updateSurfaceShellActivity`** return whether the report landed (including a workspace-based overload for replay). Stale surfaces are discarded when the bucket drains. **`TabManagerShellActivityReplayTests`** covers buffer→replay and stale discard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74f517a707727c120a18a130e03bf18e9fd23f11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes #6618 by buffering control-socket shell-state reports that arrive before a workspace registers and replaying them after tabs update or when the window registers. This lets idle terminals reach `.promptIdle`/`.running`, restoring close-confirmation prompts, PR probing, and scrollback behavior.

- **Bug Fixes**
  - Added per-workspace/surface buffer in `AppDelegate.recordReportedShellActivity`; buffers only when `tabManagerFor` is nil, applies immediately when reachable. Bounded at 1024 workspaces and 256 surfaces per workspace; exposes `hasPendingShellActivityReports`.
  - Replays run after tabs update via `WorkspacesHosting.workspaceTabsDidChange` and on window registration via `TabManager.flushPendingShellActivityForRegisteredWorkspaces`; `TabManager` drains the workspace bucket and drops unapplied entries as stale. Ensures `.promptIdle` PR probes fire.
  - `Workspace.updatePanelShellActivityState` and `TabManager.updateSurfaceShellActivity` now return `Bool`; added a workspace-based overload; `.promptIdle` still schedules PR refresh.
  - Both control-socket reporters now use the shared entrypoint; regression tests in `TabManagerShellActivityReplayTests` cover buffer→replay, registration-triggered replay, and stale discard.

<sup>Written for commit 74f517a707727c120a18a130e03bf18e9fd23f11. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6686?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Shell activity reports that arrive before workspaces are reachable are now properly buffered and replayed once the workspace becomes accessible, ensuring accurate and consistent state updates.

* **Tests**
  * Added comprehensive test coverage for shell activity buffering and replay behavior under timing edge cases.


lawrencecchen:
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review

lawrencecchen:
Addressed review feedback in `7b8b07fb1`:

- **Codex (P2, PR-probe pruned):** moved the replay from the `tabs` `willChange` hook to a new `workspaceTabsDidChange` hook, so it runs after `TabManager.tabs` holds the new value. `scheduleWorkspacePullRequestRefresh` now sees the restored workspace instead of pruning the just-scheduled prompt-idle key. The shell state already landed before (it applies against the explicit `Workspace` object); this restores the PR-probe side effect for restored idle terminals too.
- **Greptile (P2, O(N×buffer) drain + wholesale eviction):** regrouped the pending buffer as `[UUID: [UUID: PanelShellActivityState]]`. A replay now drains only that workspace's surfaces (O(surfaces)), and the overflow cap is on distinct workspaces so any eviction is confined to a single bucket rather than clearing the whole buffer.
- **Greptile (P2, dirty global test state):** the regression test now clears the shared `AppDelegate` buffer via `defer` so a mid-test failure can't leak into later tests.

Re-verified red→green on the AWS M4 test host (`cmux-unit`): the test passes on HEAD and still fails on the test-only commit.

PR #6734, opened 2026-06-24
Remote tmux: gate shared ControlMaster ready before attach burst (fixes #6732)

### LINKED ISSUE #6732: Remote tmux: only ~2 of N sessions mirror on first attach when host has many sessions (race in bulk attach)
### Summary
With the **Remote tmux** beta (PR #5553), `cmux ssh-tmux <host>` only mirrors a couple of sessions when the remote host has many tmux sessions. Discovery finds all of them, but the per-session control-stream attach drops most of them — leaving (in my case) only the 2 alphabetically-first sessions as workspaces.

### Environment
- cmux: Remote tmux beta (Settings → Beta Features → Remote tmux enabled)
- Remote host: macOS (Apple Silicon, Darwin arm64), tmux 3.6b, login shell zsh
- Remote had **22 tmux sessions** at attach time

### Repro
1. On a remote host, create ~22 tmux sessions (several with spaces in their names, e.g. `Beyond Carat Hub`, `Meta Google`).
2. `cmux ssh-tmux user@host`
3. Only 2 sessions appear as workspaces (the two alphabetically-first: `22`, `Ads`). The other 20 are missing.

### Expected
All 22 sessions mirror as workspaces (sessions → workspaces, windows → tabs).

### Actual
2 of 22 mirror. The two that survive are the only ones that end up `(attached)` in `tmux list-sessions`; the rest are silently absent from the sidebar.

### It's intermittent (race signal)
Closing the dedicated cmux window and re-running `cmux ssh-tmux user@host` mirrored the rest on the **second** attempt. So this looks like a race/timeout in the bulk attach, not a hard cap — re-running with a fresh burst eventually wins more sessions. (Re-running *without* closing the window just reuses the existing dedicated window and adds nothing.)

### Server side is healthy — ruled out as the cause
I verified the remote independently; none of these is the bottleneck:
- **tmux reachable for every session**, including names with spaces — `tmux has-session -t '<name>'` returns rc=0 for all 22 over a shared SSH ControlMaster.
- **Discovery returns all 22** — the exact command cmux runs, `tmux list-sessions -F '#{session_id}:#{session_windows}:#{session_attached}:#{session_created}:#{session_name}'`, lists all 22 and the parser handles the spaces/colons fine.
- **No SSH channel cap** — opened **22 simultaneous** long-lived channels over a single ControlMaster with zero `open failed` / `administratively prohibited` / MaxSessions rejections (default sshd `MaxSessions 10` was not hit for multiplexed channels on this macOS host).

So discovery is complete and the transport is fine; the loss is in the per-session `tmux -CC attach` control-stream bring-up (`RemoteTmuxController.mirrorHostInNewWindow` → `mirrorSession` → `attach`). Connections that don't reach `%enter` get torn down (so they don't leave empty workspaces), which is what removes the other 20.

### Suggestions
- Bound/stagger the concurrent control-stream attaches, or retry failed `%enter` handshakes, instead of firing all N at once.
- After the initial mirror, reconcile against `%sessions-changed` so any sessions that lost the initial race still get picked up without a manual close + re-attach.
- Surface attach failures (currently only logged under `#if DEBUG`) so users can tell a partial mirror from a complete one.

Related: #2673, #560.

harshkukadia:
Fixes #6732.

## Problem
On the first `cmux ssh-tmux <host>` against a host with many tmux sessions, only **one** session mirrors; closing the window and re-attaching mirrors the rest. (Reported in #6732.)

## Root cause — cold-ControlMaster creation race
`mirrorHostInNewWindow` fires all N `tmux -CC attach` control connections in a tight concurrent burst, each spawned with `ControlMaster=auto` (`RemoteTmuxHost.controlModeArguments`), with **no guarantee the shared master is up yet**. On a cold first attach the master socket isn't ready, so the N connections all race to *create* it at the same `ControlPath`. One wins; the rest fail with:

```
ControlSocket /Users/<you>/.cmux/ssh/tmux-…sock already exists, disabling multiplexing
```

…and only one session ends up mirrored. The **second** attach worked only because the master was still warm from `ControlPersist=180`.

### Reproduced against a real host (20 sessions), using cmux's exact ssh flags
| Scenario | Result |
|---|---|
| Warm master + `ControlMaster=auto` attaches | **20/20** ✅ |
| **Cold** master + concurrent `auto` attaches (= cmux's first attach) | **0–1/20**, 9× "ControlSocket already exists, disabling multiplexing" ❌ |
| Gated/ready master, pure-slave `ControlMaster=no` | 10/20 (this is the separate `MaxSessions=10` ceiling; under `auto` the overflow self-heals via direct-connection fallback, hence 20/20 above) |
| ControlMaster **pool** (multiple masters) | worse — rapid master creation trips server `MaxStartups` ("Connection closed by … port 22") ❌ |

So a master pool is *not* the answer; gating one shared master is.

## Fix
Add `RemoteTmuxSSHTransport.ensureMasterReady()`:
- opens the shared master (a trivial no-op command under `ControlMaster=auto` + `ControlPersist`), then
- polls `ssh -O check` (local control socket, no network round-trip) until the master accepts sessions (bounded: ≤ ~1 s, then proceeds best-effort).

It's awaited before **both** bulk attach loops (`mirrorHostInNewWindow` and `mirrorHost`). The subsequent `ControlMaster=auto` attaches then connect to a *ready* master instead of racing to create it. It's idempotent (returns immediately if a master is already live) and never throws, so discovery (`listSessions`) remains the auth/reachability gate and the interactive-auth path is unchanged.

No `sshd` changes required on remote hosts; no connection sprawl.

## Notes / scope
- A secondary ceiling exists for very large session counts: past `MaxSessions` (default 10), the extra `auto` attaches fall back to direct connections (works, validated 20/20, but more connections). Bounding the attach-burst concurrency could further harden huge-N cases; left out here to keep the fix focused on the race that actually breaks the common case.

## Testing
- New code is parse-checked; the full app build (Xcode) was not run in my environment (Command Line Tools only) — relying on CI / maintainer build for the full typecheck.
- Behavior validated out-of-band by reproducing the exact ssh invocation against a 20-session host (table above): the gated-master recipe is the one that yields a clean 20/20 from cold.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/6734"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784870451&installation_id=133855650&pr_number=6734&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F6734&signature=829518f8dcb81ed3e016be1ddaa21c67f8cfe384a97ac0e2bc5fa553f9d450d7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Prevents the cold-start SSH ControlMaster race that caused only one tmux session to mirror on the first attach. Fixes #6732 and makes first-time mirrors reliably attach all sessions.

- **Bug Fixes**
  - Added `RemoteTmuxSSHTransport.ensureMasterReady()` to open the shared master once and confirm with `ssh -O check` (local socket). Idempotent; both it and its helper now rethrow `CancellationError`.
  - Both bulk attach flows await readiness and fail closed on `false`: `mirrorHostInNewWindow` cleans up and discards the window; `mirrorHost` throws.
  - No change to auth/discovery flow; no server-side changes; avoids multiple masters or extra connections.

<sup>Written for commit 3895250cfe66720c8d57fac4437116d9a468f4e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6734?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when mirroring remote sessions by ensuring the SSH ControlMaster is ready before starting attaches.
  * Reduced race conditions and startup bursts that could cause intermittent failures on cold first connections.
  * Added a readiness check to confirm multiplexing is available before beginning mirroring, aborting cleanly when it isn’t.

### DIFF
--- a/Sources/RemoteTmuxController.swift
+++ b/Sources/RemoteTmuxController.swift
@@ -354,12 +354,19 @@ final class RemoteTmuxController {
         windowRegistry.bind(host: host, windowId: windowId)
 
         let bootstrapWorkspaceId = manager.tabs.first?.id
-        // Establish & confirm the shared ControlMaster before the attach burst, so
-        // the per-session `tmux -CC attach` connections (each ControlMaster=auto)
-        // attach to a ready master instead of all racing to create it — which on a
-        // cold first attach made all but one session fail to mirror. See
-        // RemoteTmuxSSHTransport.ensureMasterReady.
-        await transport(for: host).ensureMasterReady()
+        // Confirm the shared ControlMaster is ready before the attach burst, so the
+        // per-session `tmux -CC attach` connections (each ControlMaster=auto) attach
+        // to a ready master instead of all racing to create it — which on a cold
+        // first attach made all but one session fail to mirror. Fail closed: if
+        // readiness can't be confirmed, tear down the window rather than firing the
+        // burst against a cold master. See RemoteTmuxSSHTransport.ensureMasterReady.
+        guard try await transport(for: host).ensureMasterReady() else {
+            windowRegistry.unbind(hostHash: host.connectionHash)
+            transportRegistry.remove(connectionHash: host.connectionHash)
+            RemoteTmuxSSHTransport.spawnControlMasterExit(host: host)
+            appDelegate.discardMainWindowWithoutClosedHistory(windowId: windowId)
+            throw RemoteTmuxError.unreachable("SSH ControlMaster not ready for \(host.destination)")
+        }
         for session in sessions {
             do {
                 try mirrorSession(host: host, sessionName: session.name, into: manager)
@@ -403,10 +410,12 @@ final class RemoteTmuxController {
             throw RemoteTmuxError.unreachable("app not ready")
         }
         let sessions = try await listSessions(host: host)
-        // Warm the shared ControlMaster before the burst (see ensureMasterReady):
+        // Confirm the shared ControlMaster before the burst (see ensureMasterReady):
         // without it, concurrent ControlMaster=auto attaches race to create the
-        // master and all but one fail on a cold first mirror.
-        await transport(for: host).ensureMasterReady()
+        // master and all but one fail on a cold first mirror. Fail closed.
+        guard try await transport(for: host).ensureMasterReady() else {
+            throw RemoteTmuxError.unreachable("SSH ControlMaster not ready for \(host.destination)")
+        }
         for session in sessions {
             // One session failing to attach must not abort mirroring the rest.
             do {
--- a/Sources/RemoteTmuxController.swift
+++ b/Sources/RemoteTmuxController.swift
@@ -354,6 +354,12 @@ final class RemoteTmuxController {
         windowRegistry.bind(host: host, windowId: windowId)
 
         let bootstrapWorkspaceId = manager.tabs.first?.id
+        // Establish & confirm the shared ControlMaster before the attach burst, so
+        // the per-session `tmux -CC attach` connections (each ControlMaster=auto)
+        // attach to a ready master instead of all racing to create it — which on a
+        // cold first attach made all but one session fail to mirror. See
+        // RemoteTmuxSSHTransport.ensureMasterReady.
+        await transport(for: host).ensureMasterReady()
         for session in sessions {
             do {
                 try mirrorSession(host: host, sessionName: session.name, into: manager)
@@ -397,6 +403,10 @@ final class RemoteTmuxController {
             throw RemoteTmuxError.unreachable("app not ready")
         }
         let sessions = try await listSessions(host: host)
+        // Warm the shared ControlMaster before the burst (see ensureMasterReady):
+        // without it, concurrent ControlMaster=auto attaches race to create the
+        // master and all but one fail on a cold first mirror.
+        await transport(for: host).ensureMasterReady()
         for session in sessions {
             // One session failing to attach must not abort mirroring the rest.
             do {
--- a/Sources/RemoteTmuxSSHTransport.swift
+++ b/Sources/RemoteTmuxSSHTransport.swift
@@ -109,7 +109,7 @@ actor RemoteTmuxSSHTransport {
     @discardableResult
     func ensureMasterReady() async throws -> Bool {
         try? host.ensureControlSocketDirectory()
-        if await masterIsRunning() { return true }
+        if try await masterIsRunning() { return true }
         // Open the master exactly once (single creator → no creation race), then
         // confirm. Cancellation propagates; any other launch failure means the
         // master could not be brought up, i.e. not ready.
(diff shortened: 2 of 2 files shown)


harshkukadia:
Thanks for the review — addressed both in 5fcecf3.

**Greptile (P1, blocking-runtime / cancellation):** Removed the `Task.sleep` + `ssh -O check` polling loop entirely, so the `cmux-swift-blocking-runtime` pattern no longer applies. The master is created synchronously by a single `run(["true"])` (which awaits the ssh process exit, by which point `ControlMaster=auto`+`ControlPersist` has backgrounded it), so a poll loop was unnecessary — it now opens the master once and does one authoritative `ssh -O check`. Cancellation propagates: `run()`/`runProcess` are cancellation-aware and `ensureMasterReady` rethrows `CancellationError` (no more `try?`-swallowed sleep rushing through iterations). As you noted, the warm path (after `listSessions`) returns `true` on the first check and never opens a second connection.

**CodeRabbit (Major, fail closed):** `ensureMasterReady()` now returns `Bool` (and rethrows `CancellationError`); non-cancellation launch failures map to `false`. `ssh -O check` is the single authoritative readiness signal with no best-effort fallthrough. Both bulk-attach paths fail closed — `mirrorHostInNewWindow` tears the just-created window down via the existing cleanup path and throws, `mirrorHost` throws — so the `tmux -CC attach` burst never runs against an unconfirmed master.

On tests: there's no existing fake-`ssh` harness for the process-spawning paths in `RemoteTmuxSSHTransport` (current RemoteTmux tests cover pure logic — argv/stderr classification, parsers), and the readiness path is entirely process-driven. Happy to add a fixture-based test (inject `sshExecutablePath` pointing at a stub) if maintainers want that harness introduced.

harshkukadia:
Addressed in 3895250.

**Greptile P1 / CodeRabbit (cancellation in `masterIsRunning()`):** Good catch — the `try?` collapsed `CancellationError` to `false`, so a task cancelled during the check (or the post-create recheck) would take the fail-closed teardown path and throw `RemoteTmuxError.unreachable` instead of surfacing cancellation. `masterIsRunning()` is now `async throws -> Bool`: it rethrows `CancellationError` and collapses only non-cancellation launch/socket failures to `false`. Both call sites in `ensureMasterReady` now use `try await`, so a cancelled task can no longer fall through into `run(["true"])` / master creation.

**CodeRabbit (localize readiness error strings):** Skipping intentionally for consistency. Every existing `RemoteTmuxError.unreachable(...)` call site in this file uses a raw, non-localized string (`"no tmux sessions on …"`, `"could not create window"`, `"could not mirror any tmux session on …"`, `"already attaching …"`, etc.) — these are diagnostic `Error` messages, not localized UI surfaces. Localizing only the two new ones would be inconsistent and out of scope for this fix; happy to do a separate pass that localizes all of them if maintainers want `RemoteTmuxError` messages internationalized.

PR #6741, opened 2026-06-24
Pin Claude auto-resume binding to the launch cwd (#4256)

### LINKED ISSUE #4256: Resume launch command uses panel's current cwd, but session may have been created in a different cwd → session-not-found after restart

endmoseung:
## Summary

Fixes the Claude **auto-resume binding** path (`source: agent-hook`) so a restored session `cd`s into the directory Claude filed the transcript under, not a runtime cwd the agent drifted into mid-session. Resolves #4256.

Claude namespaces each transcript under `<config>/projects/<encoded launch cwd>/<id>.jsonl`, and `claude --resume <id>` only locates it from that same cwd. When a session starts in one dir and the agent later `cd`s elsewhere (repo root → worktree, or `$HOME` → a subdir), the persisted binding pinned the **drifted** cwd, so auto-resume ran `cd '<drift>' && claude --resume <id>` and failed with `No conversation found`.

This is the follow-up that #5154 explicitly deferred:

> The same root-cause drift also reaches the auto-resume binding path (`source: agent-hook`, `--resume`), whose cwd originates in the CLI hook (`CLI/cmux.swift`) … intentionally left as a focused follow-up.

#5154 fixed the snapshot fork/restore path (`Sources/RestorableAgentSession.swift`); the binding published by `publishAgentSurfaceResumeBinding` in `CLI/cmux.swift` was untouched.

## Why a candidate match isn't enough

In the real failure (captured from a live `0.64.9` session, see #4256) **both** inputs to `AgentResumeWorkingDirectory().resolve(...)` are the drift by the time the binding is published: the runtime `cwd` is the drift, and the captured `launchCommand.workingDirectory` *also* collapsed to the drift (SessionStart fired after the agent moved, and the hook process had no trusted `CMUX_AGENT_LAUNCH_CWD`, so `agentLaunchCommandFromEnvironment` fell back to `parsedInput.cwd`). The true launch cwd is therefore not among the candidates.

What is reliable: every Claude transcript record carries the launch cwd in a top-level `"cwd"`, and `transcript_path` is in the hook payload. So the launch cwd is recovered from the transcript content and accepted only when it re-encodes to the transcript's project dir (no lossy decode).

## Changes

- **New shared type** `ClaudeResumeWorkingDirectory` + `ClaudeProjectDirEncoding` in `CMUXAgentLaunch` (single source of truth for both targets). `verifiedWorkingDirectory(...)`:
  1. matches a candidate dir's encoding to the transcript path's project-dir segment;
  2. else recovers the launch cwd from the transcript file's top-level `"cwd"`, accepted only if it re-encodes to that same project dir;
  3. else probes the Claude config roots on disk.
- **CLI binding builder** `publishAgentSurfaceResumeBinding` resolves the Claude resume cwd through it before the existing namespacing fallback, gated to `kind == "claude"` (other agents unchanged). `transcriptPath` is threaded through all 8 call sites.
- **App snapshot resolver** delegates to the same shared type, removing the duplicated transcript-verification logic (no behavior change; the existing #5154 tests cover it).

## Testing

- `CMUXAgentLaunch` package: `swift test` green (added `ClaudeResumeWorkingDirectoryTests` — candidate match, transcript-content recovery, round-trip guard rejection, config scan, empty inputs; full suite 161 tests pass).
- Added a CLI hook integration test (`cmuxTests/CLIGenericHookPersistenceTests.swift`) that reproduces the drift (both candidates = drift, transcript under the launch dir) and asserts the published `surface.resume.set` cwd/command pin the launch dir.

> ⚠️ **Disclosure:** I could not run the full Xcode app/CLI build or the `cmuxTests` target locally (this machine is missing `CoreSimulator.framework`, so `xcodebuild` fails to load `IDESimulatorFoundation` before compiling). The shared-package logic and its unit tests are verified via `swift test`; the CLI/app integration compiles against the same shared API but its full build + the new integration test should be confirmed by CI.

## Residual notes for reviewers

- If Claude ever changes its project-dir encoding, `ClaudeProjectDirEncoding` is now the one place to update (used by both app and CLI).
- The round-trip guard is intentionally strict: it only accepts a recovered cwd that re-encodes to the transcript's project dir, so a foreign/spoofed `cwd` in transcript content can't be honored.
- The transcript read is a bounded 64 KB head read with per-line JSON parsing; a mid-codepoint UTF-8 cut yields `nil` and falls through safely.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/6741"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784887225&installation_id=133855650&pr_number=6741&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F6741&signature=ed26aaa0dd3fcaebb0e6c6da467f62633ed5eafe6fda5d61577e9fc0389ffcb5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Pins Claude auto-resume to the original launch directory and verifies it against the transcript to prevent “No conversation found” after mid-session `cd`. Resolves #4256.

- **Bug Fixes**
  - Verifies the resume cwd with `ClaudeResumeWorkingDirectory`: match the transcript’s project dir, else recover from transcript content with a round-trip check, else scan config roots; supports nested `<id>/messages/<id>.jsonl`.
  - Updates CLI `publishAgentSurfaceResumeBinding` (Claude only) to resolve via transcript before namespacing and threads `transcriptPath`; fixes non-optional `sessionId` handling.

- **Refactors**
  - Extracts `ClaudeProjectDirEncoding` and centralizes logic in `@macOS/CMUXAgentLaunch`.
  - App snapshot resolver now delegates to one shared resolver per index-load pass with memoized config-root/transcript lookups, removing duplicate verification code.

<sup>Written for commit daa09dd5aedf89aa20f147d97e0b8da5ec7754c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6741?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved Claude resume handling by verifying and resolving the correct working directory from the session transcript, even when the current CWD has drifted.
  * Resume/session flows now better preserve and propagate transcript path context for more consistent continuation behavior.

* **Bug Fixes**
  * Fixed resume targeting the wrong folder by validating the expected project-dir encoding and, when needed, recovering the launch CWD from transcript JSONL records.
  * Added more reliable fallback behavior (including honoring `CLAUDE_CONFIG_DIR`) when transcript lookup isn’t provided.

* **Tests**
  * Added coverage for transcript-derived CWD recovery and Claude project directory encoding rules.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -23086,6 +23086,7 @@ struct CMUXCLI {
                         displayName: String(localized: "cli.claude-hook.notification.title", defaultValue: "Claude Code"),
                         sessionId: sessionId,
                         cwd: parsedInput.cwd,
+                        transcriptPath: parsedInput.transcriptPath,
                         launchCommand: launchCommand
                     )
                 }
@@ -23211,6 +23212,7 @@ struct CMUXCLI {
                         displayName: String(localized: "cli.claude-hook.notification.title", defaultValue: "Claude Code"),
                         sessionId: sessionId,
                         cwd: parsedInput.cwd ?? mappedSession?.cwd,
+                        transcriptPath: parsedInput.transcriptPath ?? mappedSession?.transcriptPath,
                         launchCommand: mappedSession?.launchCommand
                     )
                 }
@@ -23334,6 +23336,7 @@ struct CMUXCLI {
                     displayName: String(localized: "cli.claude-hook.notification.title", defaultValue: "Claude Code"),
                     sessionId: sessionId,
                     cwd: parsedInput.cwd ?? mappedSession?.cwd,
+                    transcriptPath: parsedInput.transcriptPath ?? mappedSession?.transcriptPath,
                     launchCommand: mappedSession?.launchCommand ?? firstSightingLaunchCommand
                 )
             }
@@ -26721,6 +26724,7 @@ struct CMUXCLI {
         displayName: String,
         sessionId: String,
         cwd: String?,
+        transcriptPath: String?,
         launchCommand: AgentHookLaunchCommandRecord?
     ) {
         let resumeEnvironment = agentSurfaceResumeEnvironment(kind: kind, environment: launchCommand?.environment)
@@ -26728,11 +26732,26 @@ struct CMUXCLI {
         // runtime cwd: cwd-namespaced agents (Claude, Grok, Gemini, …) file their session under the
         // launch dir, so resuming from a worktree the agent later `cd`'d into fails with "No
         // conversation found".
-        let resumeWorkingDirectory = AgentResumeWorkingDirectory().resolve(
-            kind: kind,
-            runtimeCwd: cwd,
-            launchWorkingDirectory: launchCommand?.workingDirectory
-        )
+        //
+        // For Claude, the captured launch cwd can itself be the drifted runtime cwd when this hook
+        // process lacks a trusted `CMUX_AGENT_LAUNCH_CWD` (`agentLaunchCommandFromEnvironment` then
+        // falls back to the hook-reported cwd). So before the namespacing rule, verify against the
+        // transcript on disk which candidate directory actually holds the session — the same policy
+        // PR #5154 applied to the snapshot path, now shared via `ClaudeResumeWorkingDirectory`.
+        let verifiedClaudeWorkingDirectory: String? = kind == "claude"
+            ? ClaudeResumeWorkingDirectory().verifiedWorkingDirectory(
+                sessionId: sessionId,
+                transcriptPath: transcriptPath,
+                claudeConfigDir: launchCommand?.environment?["CLAUDE_CONFIG_DIR"],
+                candidateWorkingDirectories: [launchCommand?.workingDirectory, cwd].compactMap { $0 }
+            )
+            : nil
+        let resumeWorkingDirectory = verifiedClaudeWorkingDirectory
+            ?? AgentResumeWorkingDirectory().resolve(
+                kind: kind,
+                runtimeCwd: cwd,
+                launchWorkingDirectory: launchCommand?.workingDirectory
+            )
         guard let command = agentSurfaceResumeCommand(
             kind: kind,
             sessionId: sessionId,
@@ -29913,6 +29932,7 @@ export default function cmuxPiSessionExtension(pi: ExtensionAPI) {
                         displayName: def.displayName,
                         sessionId: sessionId,
                         cwd: hookCwd ?? mapped?.cwd,
+                        transcriptPath: input.transcriptPath ?? mapped?.transcriptPath,
                         launchCommand: launchCommand
                     )
                 }
@@ -29980,6 +30000,7 @@ export default function cmuxPiSessionExtension(pi: ExtensionAPI) {
                     displayName: def.displayName,
                     sessionId: sessionId,
                     cwd: latest.cwd,
+                    transcriptPath: latest.transcriptPath,
                     launchCommand: latest.launchCommand
                 )
                 if let lifecycle = latest.agentLifecycle {
@@ -30180,6 +30201,7 @@ export default function cmuxPiSessionExtension(pi: ExtensionAPI) {
                     displayName: def.displayName,
                     sessionId: sessionId,
                     cwd: hookCwd ?? mapped?.cwd,
+                    transcriptPath: input.transcriptPath ?? mapped?.transcriptPath,
                     launchCommand: launchCommand ?? mapped?.launchCommand
                 )
                 if codexPromptTurnWentTerminal() {
@@ -30454,6 +30476,7 @@ export default function cmuxPiSessionExtension(pi: ExtensionAPI) {
                     displayName: def.displayName,
                     sessionId: sessionId,
                     cwd: cwd,
+                    transcriptPath: input.transcriptPath ?? mapped?.transcriptPath,
                     launchCommand: launchCommand ?? mapped?.launchCommand
                 )
             }
@@ -30635,6 +30658,7 @@ export default function cmuxPiSessionExtension(pi: ExtensionAPI) {
                     displayName: def.displayName,
                     sessionId: sessionId,
                     cwd: hookCwd ?? mapped?.cwd,
+                    transcriptPath: input.transcriptPath ?? mapped?.transcriptPath,
                     launchCommand: launchCommand ?? mapped?.launchCommand
                 )
             }
--- a/Packages/macOS/CMUXAgentLaunch/Sources/CMUXAgentLaunch/ClaudeTranscriptResume.swift
+++ b/Packages/macOS/CMUXAgentLaunch/Sources/CMUXAgentLaunch/ClaudeTranscriptResume.swift
@@ -30,10 +30,14 @@ public enum ClaudeProjectDirEncoding {
 public struct ClaudeResumeWorkingDirectory {
     private let fileManager: FileManager
     private let homeDirectory: String
+    // A reference-type cache so that reusing one instance across a session-index load loop shares the
+    // (expensive) config-root scan and per-project transcript probes instead of redoing them per call.
+    private let cache: TranscriptLookupCache
 
     public init(fileManager: FileManager = .default, homeDirectory: String = NSHomeDirectory()) {
         self.fileManager = fileManager
         self.homeDirectory = homeDirectory
+        self.cache = TranscriptLookupCache(homeDirectory: homeDirectory, fileManager: fileManager)
     }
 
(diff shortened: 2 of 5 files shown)


endmoseung:
Thanks for the reviews — pushed fixes for the actionable ones (sessionId `?? ""` that wouldn't compile, nested `<id>/messages/<id>.jsonl` project-dir inference, shared per-load resolver/cache so account roots are scanned once, and a session-start success assertion + nested-layout test).

A few I deliberately skipped, with reasoning — happy to change if you disagree:

- **Fail closed (skip/clear the binding) when Claude verification returns nil** — I kept the fallback to `AgentResumeWorkingDirectory().resolve(...)`. For the common, non-drifted session both candidates already equal the launch cwd, so `resolve` (which prefers the launch dir for `.byDirectory`) returns the correct dir; failing closed there would regress the happy path by dropping a valid binding. `verifiedWorkingDirectory` only overrides when it can *prove* a better dir against the transcript on disk, so the fallback is strictly the prior behavior, not a new guess.

- **Move `ClaudeProjectDirEncoding` off the caseless-enum namespace** — kept it as is to match the existing `ClaudeConfigDirectoryPath` / `AgentLaunchEnvironmentPolicy` enums already in this package. Glad to fold it into `ClaudeResumeWorkingDirectory` if you'd prefer the static-namespace rule enforced here.

- **Lossy `/a.b` vs `/a/b` ambiguity in recovery** — the recovery path never decodes; it reads the launch cwd from the transcript's own `cwd` field and accepts it only when `projectDirName(forPath:)` re-encodes to the transcript's project dir, so an ambiguous/foreign path fails closed rather than being chosen.

Note: I couldn't run the full Xcode build locally (this machine lacks `CoreSimulator.framework`, so `xcodebuild` fails to load before compiling). The `CMUXAgentLaunch` package + its tests pass via `swift test`; the app/CLI compile and the integration test should be confirmed by CI.

PR #6787, opened 2026-06-26
Fix sleep/wake window geometry restore

xiahuaaaa:
## Summary

This is a smaller, sleep/wake-focused version of the approach in #4825.

- Capture main-window geometry before screen/session sleep transitions.
- Restore that intended geometry after `screensDidWake`, `didWake`, or session reactivation if AppKit reports a shrunken or displaced frame.
- Use the cached pre-sleep geometry for session snapshots and autosave fingerprints while the wake restore window is active, so transient bad frames are not persisted.
- Include display metadata in the autosave fingerprint when geometry is hashed.

## Why

On macOS, display sleep/wake can cause AppKit to report a temporary constrained window frame. A window that previously filled the visible display can come back much smaller, and the autosave path may persist that transient frame before the user notices.

This patch keeps the scope narrower than #4825: it does not implement full display disconnect/reconnect home tracking. It only protects sleep/wake transitions and keeps the existing session geometry resolver in charge of clamping/remapping.

## Validation

- `git diff --check`
- Added geometry decision tests for ratio-based shrink, stable-size system move, already-restored frames, invalid tiny frames, and larger current frames.

Could not run `xcodebuild test` locally because this machine only has Command Line Tools selected, not a full Xcode installation.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/6787"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785030710&installation_id=133855650&pr_number=6787&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F6787&signature=95c0d353a1d78798d3172ce959640aeb27e3a47c80a44945641075db1cd2abc2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fixes macOS sleep/wake window geometry by capturing main-window frames before sleep and restoring them after wake to avoid shrunken or shifted windows. Prevents transient frames from being persisted or used in autosave fingerprints.

- **Bug Fixes**
  - Capture per-window geometry on `screensDidSleep`; schedule restore after `screensDidWake`, `didWake`, and session activation with short delayed retries; clear cached data after a few seconds or on window close.
  - Restore only when `shouldRestoreWindowFrameAfterWake` flags shrink or system move; ignore invalid frames and larger current frames.
  - Use cached pre-sleep geometry for session snapshots and autosave fingerprinting during the wake restore window; include display metadata in the fingerprint.
  - Add tests for shrink-by-ratio, already-restored frames, invalid tiny frames, stable-size moved frames, and larger current frames.

<sup>Written for commit b3be6b7b67d4f2ecc86157683bdacce9afd2da30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6787?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -660,6 +660,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         let display: SessionDisplaySnapshot?
     }
 
+    private struct PreSleepWindowGeometry {
+        let frame: SessionRectSnapshot
+        let display: SessionDisplaySnapshot?
+    }
+
     nonisolated static let persistedWindowGeometrySchemaVersion = 2
     private nonisolated static let persistedWindowGeometryDefaultsKey = "cmux.session.lastWindowGeometry.v2"
 #if DEBUG
@@ -805,6 +810,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
     var shortcutLayoutCharacterProvider: (UInt16, NSEvent.ModifierFlags) -> String? = KeyboardLayout.character(forKeyCode:modifierFlags:)
     private var workspaceObserver: NSObjectProtocol?
     private var lifecycleSnapshotObservers: [NSObjectProtocol] = []
+    private var preSleepWindowGeometries: [UUID: PreSleepWindowGeometry] = [:]
+    private var wakeWindowGeometryRestoreGeneration = 0
     private var windowKeyObservers: [NSObjectProtocol] = []
     private var shortcutMonitor: Any?
     private var shortcutDefaultsObserver: NSObjectProtocol?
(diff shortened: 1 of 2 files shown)


xiahuaaaa:
Hi maintainers, a gentle follow-up on this PR.

This fixes the macOS display sleep/wake case where a cmux window that was filling the display comes back at a reduced size. The branch has been rebased onto the latest `main` and updated at commit `b3be6b7`; the restore logic uses display-relative ratios and tolerances rather than 4K-specific constants, with focused regression tests for shrink, movement, no-op, invalid, and larger-frame cases.

The available CI checks are passing. I could not run the full Xcode test suite locally because this machine only has Command Line Tools. Please review and merge if the approach is acceptable; I am happy to address any requested changes promptly.

PR #6815, opened 2026-06-26
Fix zsh shell integration printing `file exists` under noclobber (#6714)

### LINKED ISSUE #6714: zsh shell integration prints `file exists` when noclobber is enabled
## Summary

After upgrading to cmux 0.64.17, new zsh tabs can print this on startup when `noclobber` is enabled:

```text
_cmux_install_cli_command_shim:13: file exists: /var/folders/.../T//cmux-cli-shims/<surface-id>/claude
```

## Cause

`Resources/shell-integration/cmux-zsh-integration.zsh` generates a per-surface CLI shim at:

```text
${TMPDIR:-/tmp}/cmux-cli-shims/${CMUX_SURFACE_ID:-$$}/claude
```

The shim writer uses normal zsh redirection:

```zsh
} >"$shim_path" 2>/dev/null || return 0
```

When the target shim already exists and the user's interactive zsh has `noclobber` enabled, zsh refuses to overwrite the generated file and emits `file exists` during shell startup.

## Reproduction

```zsh
setopt noclobber
TMPDIR=/tmp CMUX_SURFACE_ID=test-cmux-shim-repro source Resources/shell-integration/cmux-zsh-integration.zsh
TMPDIR=/tmp CMUX_SURFACE_ID=test-cmux-shim-repro source Resources/shell-integration/cmux-zsh-integration.zsh
```

The second source attempts to rewrite the existing shim and prints the error.

## Expected behavior

cmux should silently refresh its own generated temporary shim even when the user has `noclobber` enabled.

## Suggested fix

Use zsh's explicit clobber redirection for this generated file:

```zsh
} >|"$shim_path" 2>/dev/null || return 0
```

This preserves the user's global `noclobber` setting while allowing cmux to overwrite the shim file it owns.

austinywang:
Fixes #6714

## Problem

With `setopt noclobber` enabled in their interactive zsh, users see this error printed on every new zsh tab's startup:

```text
_cmux_install_cli_command_shim:13: file exists: /var/folders/.../T//cmux-cli-shims/<surface-id>/claude
```

## Root cause

`Resources/shell-integration/cmux-zsh-integration.zsh` writes a per-surface CLI shim with a plain `>` redirection:

```zsh
} >"$shim_path" 2>/dev/null || return 0
```

The shim writer (`_cmux_install_cli_command_shim`) runs more than once per shell:

1. At source time, via the top-level `_cmux_install_cli_wrapper claude` call.
2. On the first prompt, via the `_cmux_fix_path` precmd hook (which re-installs the cmux-owned wrappers after user startup files run).

The second write targets a shim that already exists. Under the user's global `noclobber`, zsh refuses to overwrite it and emits `file exists` (the `2>/dev/null` does not suppress it — the failure is reported by the shell's redirection machinery on the compound-command redirect itself). The `|| return 0` then *skips* the write, so the shim is also left stale.

## Fix

Use zsh's explicit clobber redirection (`>|`) for this cmux-owned generated file:

```zsh
} >|"$shim_path" 2>/dev/null || return 0
```

This refreshes the shim regardless of the user's global `noclobber` setting, and is the **same operator this integration already uses everywhere else** for its own generated marker/cache files (e.g. the git active-pwd marker, PR-status cache, branch-watch signal files). The user's `noclobber` preference for their own files is untouched.

## Testing

Two-commit red/green structure:

1. **Commit 1** adds `tests/test_issue_6714_zsh_shim_noclobber.py` (failing test only) and wires it into the app-host focused-regression shard next to the other zsh-sourcing regression tests (`test_issue_2448`). The test drives the real integration file through zsh with `noclobber` enabled, writes the `claude` shim twice with distinct wrapper paths, and asserts the second write is **silent** *and* **actually refreshes** the shim contents. It is deterministic (no PTY/sleep/network) and locale-pinned.
2. **Commit 2** applies the one-line redirection fix, turning the test green.

Verified locally: test is red without the fix (reproducing the exact `file exists` symptom), green with it; the issue's double-source repro no longer prints anything. `zsh -n` parse check passes; `py_compile` and `check-test-determinism.py --strict` pass.

## Localization

No user-facing strings added or changed — this removes a spurious zsh-emitted error message and changes a shell redirection operator only. No `Localizable.xcstrings` / web message catalog changes required.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/6815"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785052500&installation_id=133855650&pr_number=6815&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F6815&signature=7f9dda72a3fae793dbd1351c7537fc919d4b370e263d3711882a78d4333cbf54"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fixes the zsh “file exists” startup error under `setopt noclobber` by writing the cmux-owned CLI shim with `>|` so it always refreshes and stays silent. Adds a regression test and wires it into CI. Fixes #6714.

- **Bug Fixes**
  - Use `>|` when writing the per-surface CLI shim in `cmux-zsh-integration.zsh` to bypass user `noclobber` and refresh on reinstall.
  - Add `tests/test_issue_6714_zsh_shim_noclobber.py` and run it in CI to verify the second write is silent and updates the shim.

<sup>Written for commit d3621ef50e9bc44e095739bedcbbab84e703848c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6815?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI shim generation so it now overwrites existing files reliably, even when shell overwrite protection is enabled.
  * Fixed a shell integration issue where refreshed command shims could fail to update correctly.
  * Added a new regression test to cover this scenario on macOS and prevent future breakage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -644,6 +644,7 @@ jobs:
           CMUX_CLI_BIN="$CLI_BIN" python3 tests/test_cli_omx_fallback_path.py
           CMUX_CLI_BIN="$CLI_BIN" python3 tests/test_cli_omc_fallback_path.py
           python3 tests/test_issue_2448_shell_claude_wrapper_dispatch.py
+          python3 tests/test_issue_6714_zsh_shim_noclobber.py
           python3 tests/test_shell_git_branch_stale_cwd.py
           python3 tests/test_shell_git_config_remote_url_parsing.py
           CMUX_CLI_BIN="$CLI_BIN" python3 tests/test_claude_hook_stop_last_assistant.py
--- a/Resources/shell-integration/cmux-zsh-integration.zsh
+++ b/Resources/shell-integration/cmux-zsh-integration.zsh
@@ -292,7 +292,12 @@ _cmux_install_cli_command_shim() {
         else
             printf 'exec "%s" "$@"\n' "$escaped_wrapper"
         fi
-    } >"$shim_path" 2>/dev/null || return 0
+    # Use zsh's explicit clobber redirection (>|) so cmux always refreshes its
+    # own generated shim, even when the user's interactive zsh has `noclobber`
+    # set. A plain `>` is refused under noclobber and prints `file exists` on
+    # startup (the writer runs again from the _cmux_fix_path precmd hook after
+    # the shim already exists). See issue #6714.
+    } >|"$shim_path" 2>/dev/null || return 0
     /bin/chmod 0700 "$shim_path" >/dev/null 2>&1 || return 0
 
     if [[ "$command_name" == "claude" ]]; then
--- a/tests/test_issue_6714_zsh_shim_noclobber.py
+++ b/tests/test_issue_6714_zsh_shim_noclobber.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python3
+"""
+Regression coverage for https://github.com/manaflow-ai/cmux/issues/6714.
+
+When a user enables ``setopt noclobber`` in their interactive zsh, cmux's shell
+integration prints a spurious error on startup:
+
+    _cmux_install_cli_command_shim:13: file exists: \
+        /var/folders/.../T//cmux-cli-shims/<surface-id>/claude
+
+Root cause: ``Resources/shell-integration/cmux-zsh-integration.zsh`` writes a
+per-surface CLI shim with a plain ``>`` redirection::
+
+    } >"$shim_path" 2>/dev/null || return 0
+
+``_cmux_install_cli_command_shim`` runs more than once per shell (once at source
+time via the top-level ``_cmux_install_cli_wrapper claude`` call, and again on
+the first prompt via the ``_cmux_fix_path`` precmd hook). The second write
+targets a shim that already exists, so under ``noclobber`` zsh refuses to
+overwrite the file and emits ``file exists``. The ``2>/dev/null`` does not
+suppress the message because the no-clobber failure is reported by the shell's
+redirection machinery on the compound-command redirect itself, and the
+``|| return 0`` then *skips* the write entirely -- so the shim is also left
+stale (not refreshed to the latest wrapper path).
+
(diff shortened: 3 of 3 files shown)


austinywang:
Follow-up audit for #7474: I re-ran the regression at the test-only commit, fix commit, and current `main`, and rechecked this PR's review record.

The two CodeRabbit warning-only pre-merge notes require no follow-up change: the standalone behavioral regression already has module-level documentation and does not introduce an undocumented production API, while the description-template warning is PR metadata rather than a correctness finding. There are no inline CodeRabbit or human review findings, no unresolved review threads, and all recorded checks are green.

PR #6982, opened 2026-06-26
Add anonymous rate limit to iOS analytics proxy

### LINKED ISSUE #5569: Add anonymous rate limit to iOS analytics proxy (/api/analytics/events)
Follow-up from the iOS analytics Phase A PR (https://github.com/manaflow-ai/cmux/pull/5567).

`web/app/api/analytics/events/route.ts` accepts anonymous requests by design (pre-auth install/sign-in/pairing events must flow before the user is authenticated) and forwards allowlisted batches to PostHog with the server-side project key. It currently has only per-request shape bounds (event-name allowlist, 64 KB body cap, ≤100 events/batch, ≤64 properties/event). There is no per-IP / per-client / repeat-request rate limit, so a client could repeatedly POST valid batches and burn cmux Vercel compute and PostHog ingestion quota.

Why it was deferred in Phase A: the only reusable limiter in the codebase (`web/services/apns/rateLimit.ts`) is a per-`userId` Postgres-transaction gate, which does not fit an anonymous-first, high-volume telemetry path (no user id pre-auth; a DB write + advisory lock per analytics batch would defeat a lightweight proxy). In-memory limiting does not work on Vercel's per-instance stateless functions. The PostHog-pollution angle is already neutralized because the PostHog key is public (the web client posts to r.cmux.com directly); the residual is compute/egress DoS, which every public Next route shares and none of the sibling anonymous routes currently gate.

Proposed fix: an edge/WAF anonymous rate limit (Vercel edge config or equivalent) keyed on IP, or a shared distributed limiter, applied before `forwardToPostHog`. Optionally a lightweight signed app token to distinguish real app traffic from arbitrary clients.

austinywang:
Fixes #5569

## Summary
- gate `/api/analytics/events` with a Vercel Firewall rate limit before auth, body parsing, or PostHog forwarding
- add `CMUX_ANALYTICS_RATE_LIMIT_ID` to the web env contract and local/test defaults
- add route coverage for the anonymous rate-limit block path

## Validation
- `bun run typecheck`
- `bun test`
- `python3 scripts/check-test-determinism.py --strict`

---
## Summary by cubic
Adds an anonymous, pre-auth rate limit to the iOS analytics proxy using `@vercel/firewall` to block repeated requests before any work. Implements #5569, tightens env validation for the limiter ID, and allows previews to omit it.

- **New Features**
  - Gate `/api/analytics/events` with `@vercel/firewall` at the start; on Vercel return 429 when limited/blocked and 503 on not-found, unknown, or thrown errors; skip the check when not on Vercel or when the limiter ID is unset in preview.
  - Add `CMUX_ANALYTICS_RATE_LIMIT_ID` to the env contract with local/test defaults; required and validated in Vercel production; optional in previews and locally; if missing in production, fail closed at runtime with 503 and logging.
  - Add tests to ensure the limiter runs before auth/body parsing/PostHog, that allowed batches forward, that error states return 503, and that previews can omit the limiter ID.

- **Migration**
  - Create a Vercel Firewall rate limit and set `CMUX_ANALYTICS_RATE_LIMIT_ID` in Vercel production. Previews may omit this.

<sup>Written for commit 3cc769c0a97fc3b94fb539c45a188612219abf18. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/6982?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


austinywang:
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review

PR #7021, opened 2026-06-28
remote-tmux: linked-view mirror mode (one SSH session for MaxSessions=1 hosts)

ejc3:
## Summary

An optional **linked-view** beta mode for remote-tmux, for SSH hosts whose `sshd` caps each connection to a single concurrent session (`MaxSessions 1`, e.g. hosts with per-connection 2FA). The normal mirror opens one `tmux -CC` control client per session, which needs one SSH session each and so fails past the first. Linked-view instead creates ONE hidden, cmux-owned aggregate "view" session, `link-window`s every real session's windows into it, and drives a SINGLE `-CC` client attached to the view — so all workspaces stream live over the one allowed connection (one auth), while the real tmux sessions stay intact for `tmux ls` / `attach` interop.

Default off. Adds the `remoteTmux.linkedView` beta flag.

## Highlights

- **Decision core (pure, tested):** view-session identity/ownership, window→workspace regrouping, the link/unlink reconciler (never unlinks the placeholder or a non-owned window; never empties the view), and the composed planner.
- **Live coordinator:** drives discovery/link/reconcile over the single stream, plus `new-workspace` / new-tab / split / rename / close / paste all routed through that shared connection (commands target tmux's global `@window` / `%pane` ids). New tabs anchor on the home session by name and trigger a reconcile so they surface.
- **Per-window sizing:** each linked-view window is sized to its cmux pane via `window-size manual` + `resize-window`, debounced and re-applied on reconnect so every tab fits.
- **Browser behavior in mirror windows:** interactive browser requests (the globe button / palette) open a browser in a new **local** workspace; the Browser New-Workspace/button path stays local in the mirror window (restoration/automation behavior unchanged).
- **Settings toggle:** a "Remote tmux linked view" beta toggle under Settings → Beta Features, with en/ja localization (`Localizable.xcstrings`).
- **Docs:** the remote-tmux docs page now covers linked-view setup and how to mirror multiple servers into one window.
- **Robust parsing** for non-UTF-8 control clients (printable delimiters) and CRLF.

## Tests

Unit suites for the reconciler, view-session identity, workspace model, and planner; plus live-verified end to end against a `MaxSessions 1` test server.

---

**Stacked on #7020.** This PR's actual change is the commits *after* the proxy-retry fix; until #7020 merges, the compare view here also includes its 4 commits. Review the `remote-tmux: …linked-view…` commits.

---
[...]

PR #7026, opened 2026-06-28
remote-tmux: upload pasted/dropped files in band over the control connection

ejc3:
## Problem

Pasting a screenshot (or dropping a file) onto a remote-tmux **mirror** terminal uploaded it with `scp`, which opens a second multiplexed SSH channel. Hosts that cap each connection to one session (`MaxSessions 1`, e.g. per-connection 2FA) refuse it — the `-CC` control client already holds the one allowed channel — so the upload fails and the user only gets an unusable macOS-local path.

## Fix

Upload **in band** over the existing `tmux -CC` control connection: no second SSH channel, no remote helper binary (works on `MaxSessions 1` hosts). The file is base64-encoded locally and streamed to a remote temp file as a series of `run-shell "printf %s <chunk> >> <file>"` commands, then decoded and verified remotely.

`run-shell` stdout is **not** readable over control mode, so the result (size + POSIX `cksum`) is reported back through a tmux user option that cmux reads with `show-options` and checks against the local file.

- **`RemoteTmuxInBandUpload`** — pure, unit-tested protocol helpers: extension sanitizer, base64 chunking (32 KiB), command builders that contain no `"`/`\`/`#` (so they survive tmux double-quoting), ack parser, and a POSIX cksum that matches `cksum(1)`.
- **`RemoteTmuxControlConnection.uploadFileInBand`** — setup → chunked stream → decode/verify/atomic-move → ack, using a `queryWithTimeout` that reconnects a wedged control stream on timeout so it can't hang. Encode/checksum run off the main actor.
- **`remoteUploadTarget`** prefers in-band for a connected mirror; scp remains the fallback when the stream isn't live.

## Frozen-pane fix (follow-up commit)

The upload originally issued `run-shell -t %<pane>`. Targeting a pane makes tmux display the command's (empty) output in that pane's copy/view-mode, which swallows keystrokes — the pane looked frozen until you exited the viewer. `run-shell` runs on the server and needs no pane target, so all upload commands now go through `RemoteTmuxInBandUpload.runShellCommand` (no `-t`). Regression test asserts the wrapper never emits `-t`/`%pane`.

## Verification

- Unit tests for the protocol helpers (POSIX cksum vs `cksum(1)`, chunking, sanitizer, the no-`-t` invariant).
- Empirically validated against tmux 3.7: run-shell stdout is not queryable; the option→show-options ack channel works; the decode pipeline round-trips.

## Known v1 limitations

- In-band uploads are capped at 25 MiB (streaming a larger file over the control stream is slow); the upload streams to completion even if cancelled mid-flight; a timeout reconnects (a remote `/tmp/cmux-ul-*` temp dir may then linger); a multi-file batch isn't transactional.

Stacks on the linked-view core (#7021). ⚠️ Draft: needs live verification against a real `MaxSessions 1` host before un-drafting.


ejc3:

PR #7049, opened 2026-06-29
Eagerly publish this cmux instance's tmux identity before exec so ssh-tmux reuses the current window

ejc3:
## Problem

When a cmux surface attaches a remote `tmux` session whose **global environment was seeded by a different cmux instance** (an older launch, or a separate app), `cmux ssh-tmux` opens a brand-new window instead of reusing the current one.

## Root cause

The shell integration syncs each instance's `CMUX_*` identity through the tmux session's **global** environment:

- An app-spawned shell *publishes* its workspace-scoped identity (`CMUX_SOCKET_PATH`, `CMUX_WORKSPACE_ID`, …) into the tmux global env (`set-environment -g`), and clears surface-scoped keys (`set-environment -gu`).
- A shell that starts **inside** tmux *pulls* that global env back (`_cmux_tmux_refresh_cmux_environment`).

If instance B attaches a session that instance A seeded, B's in-tmux pull reads **A's** stale identity — wrong `CMUX_SOCKET_PATH` / `CMUX_WORKSPACE_ID` — so `ssh-tmux` resolves the wrong instance and can't reuse the window.

## Fix

Eagerly, authoritatively publish **this** instance's identity into the tmux global env **before** the login shell is `exec`'d. Three parts:

1. **Eager entrypoint.** New `_cmux_tmux_publish_cmux_environment_eager()` (zsh + bash), a thin wrapper over the existing `_cmux_tmux_publish_cmux_environment` so the published key set stays single-sourced (no-op when `$TMUX` is already set).

2. **Authoritative.** The publish now **clears** (`set-environment -gu`) a managed workspace-scoped key this instance does *not* have set, instead of skipping it. Previously a value seeded by another instance (e.g. a stale `CMUX_TAG` from a tagged debug build) survived in the session and was fed back via the in-tmux pull. The publish is gated by a set-key signature, so the clear runs at most once per process.

3. **Hermetic & ordered.** The app-generated remote bootstrap invokes the eager publish from a one-shot subshell after exporting the instance's `CMUX_*` identity but before `exec`. A plain `"$CMUX_LOGIN_SHELL" -c` would still source the user's startup files first (zsh `$ZDOTDIR/.zshenv` chaining to `~/.zshenv`; bash `$BASH_ENV`), which could auto-attach tmux or mutate `CMUX_*` before the publish — so the subshell runs hermetically: **zsh `-f`** (NO_RCS) and **bash `BASH_ENV= … --noprofile --norc`**, sourcing the integration explicitly and inheriting the already-exported identity. The zsh/bash one-shot lines are single-sourced through `eagerIdentityPublishLine`.

This generalizes the existing publish path rather than special-casing window reuse; the shared helper and its `$TMUX` guard remain the single authority.

## Tests

Two-commit red/green (per the repo's regression-test policy). Verified locally with `xcodebuild test` (scheme `cmux-unit`): all 5 tests **fail** at the test-only commit and **pass** at the fix commit.

- `testBootstrapEagerPublishWritesFullWorkspaceIdentityAndClearsSurfaceScope` — eager publish writes the full workspace-scoped key set (`-g`) and clears surface-scoped keys (`-gu`).
- `testBootstrapEagerPublishIsNoOpInsideTmux` — no-op when `$TMUX` is set.
- `testRemoteBootstrapPublishesTmuxIdentityBeforeExecingLoginShell` — the bootstrap runs the eager publish before `exec` in both branches, and does so **hermetically** (`zsh -f`; `bash BASH_ENV= --noprofile --norc`).
- `testEagerPublishClearsManagedWorkspaceKeyUnsetByThisInstance` — a managed workspace key this instance lacks (e.g. `CMUX_TAG`) is **cleared** (`-gu`), not skipped.
- `testEagerPublishBashPublishesIdentityAndClearsUnsetKey` — bash executable twin: exercises the bash integration's eager publish + authoritative clear at runtime.

### DIFF
--- a/Resources/shell-integration/cmux-bash-integration.bash
+++ b/Resources/shell-integration/cmux-bash-integration.bash
@@ -433,8 +433,15 @@ _cmux_tmux_publish_cmux_environment() {
     local key value
     for key in "${_CMUX_TMUX_SYNC_KEYS[@]}"; do
         value="${!key}"
-        [[ -n "$value" ]] || continue
-        tmux set-environment -g "$key" "$value" >/dev/null 2>&1 || return 0
+        if [[ -n "$value" ]]; then
+            tmux set-environment -g "$key" "$value" >/dev/null 2>&1 || return 0
+        else
+            # Authoritative: clear a managed key THIS instance does not have, so a
+            # value seeded by a DIFFERENT instance (e.g. a stale CMUX_TAG from a
+            # tagged debug build) can't survive in the session and feed back via
+            # the in-tmux pull. Skipping it (the old behavior) left it stale.
+            tmux set-environment -gu "$key" >/dev/null 2>&1 || return 0
+        fi
     done
 
     for key in "${_CMUX_TMUX_SURFACE_SCOPED_KEYS[@]}"; do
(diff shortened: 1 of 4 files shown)


ejc3:

PR #7107, opened 2026-06-30
remote-tmux linked-view: follow the remote's active window, and fix window-teardown data loss

ejc3:
Two problems I hit dogfooding linked-view mirrors.

**A mirrored tab could take your work with it when it closed.** Mirror a host, then let its remote tmux session end (or aggregate a second host into the same window). The now-dead mirror tab gets converted into a plain local shell so you can keep using it — you `cd` somewhere and start typing. Then the host's teardown runs, still treats that tab as a mirror, and closes it: your local shell and everything in it is gone.

The teardown now checks whether the tab is still a live mirror before closing it (converted-to-local tabs are left alone). Aggregating a host into a window clears the "disposable" mark on that window's local shells so a later teardown can't discard them, and if an empty host's freshly-created session vanishes out from under us the tab is re-bootstrapped instead of left dead.

**The mirror now tracks the remote's active window.** Switch windows on the remote and cmux's active tab follows. The catch: a mirror rebuild fires on every command and every resize, so re-selecting on each rebuild would snap you back to the remote's active window and clobber a tab you clicked by hand. So re-selection only happens on an actual active-window change or when a new mirror tab appears — the linked-view `list-windows` format carries `#{window_active}` and the model tracks `activeWindowId` to tell the difference.

Also: programmatic `close-window` now takes the same close path as every other close, and `close-window` / `focus-window` accept an index, ref, or UUID for the target window.

The linked-view plan/model unit tests and the mirror render-fidelity harness pass.


ejc3:

PR #7209, opened 2026-07-02
New Workspace on a mirrored tmux session creates a local workspace instead of a session on its host

ejc3:
Press ⌘N while sitting on a mirrored tmux session and cmux creates a plain local workspace. It used to create a new tmux session on that workspace's host and mirror it back as a new workspace.

Repro: `cmux ssh-tmux <host>`, select one of the mirrored session workspaces, press ⌘N (or double-click the sidebar's empty area, whose code still documents "a new workspace becomes a new tmux session"). A local workspace appears; nothing happens on the host.

The remote routing hook disappeared in #7406: when mirror placement moved from dedicated windows into the current window, `performNewWorkspaceCreationAction` lost its remote-tmux branch, so every entry point that funnels through it (⌘N, the titlebar plus button, the sidebar double-click) falls through to local creation.

The old hook can't return as it was, because it looked up the window's one registered host and that concept is gone — mirrors from several hosts share one window by default now. Instead the host comes from the workspace the user is actually sitting on:

- `newSessionHost(activeTabId:entries:)` is the pure core: the host of the active workspace's live mirror, or nil to create a local workspace. A local workspace next to mirrors, a mirror mid-teardown (weak workspace gone), or no selection at all each fall through to local creation.
- `handleNewWorkspaceRequested(in:)` claims the action synchronously, then runs `new-session -d -P -F '#{session_name}'` on the host and mirrors the session back into the same window. After the ssh round trip it revalidates: a closed window skips the mirror (the detached session is picked up on the next attach), and if the user switched tabs meanwhile the new workspace is added without stealing selection.
- `mirrorSession` gains a `select:` parameter so a user-initiated create selects its workspace through the normal path, while bulk attach keeps `select: false` and never steals focus.
- The hook sits before the configured new-workspace action and only for terminal workspaces — a new browser workspace never spawns a tmux session.

Out of scope, deliberately: workspace-group placement and cwd inheritance don't apply to the remote path (a new remote session starts in tmux's default cwd).

The first commit adds only the regression test so CI shows it failing without the fix: New Workspace on a selected mirror must not add a local workspace, and must still add one when a local workspace is selected. The second commit adds the fix plus unit tests for the host-derivation truth table (including two hosts' mirrors sharing a window) and the controller seam.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -7338,6 +7338,16 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
         let context = livePreferredContext
             ?? preferredMainWindowContextForWorkspaceCreation(event: event, debugSource: debugSource)
 
+        // On a remote-tmux mirror workspace, a new terminal workspace means
+        // "create a new tmux session on that workspace's host" — route it to the
+        // remote and mirror it back instead of creating a local workspace. The
+        // browser variant always stays local.
+        if initialSurface == .terminal,
+           let context,
+           remoteTmuxController.handleNewWorkspaceRequested(in: context.tabManager) {
+            return true
+        }
+
         let workspaceGroupTarget = context.flatMap { workspaceGroupNewWorkspaceTarget(in: $0) }
         // The configured new-workspace action is the user's override for the
         // plain New Workspace behavior; the browser variant keeps its own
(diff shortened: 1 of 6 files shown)


ejc3:
Closing in favour of #8428, which carries this same routing (`handleNewWorkspaceRequested` / `newSessionHost`) built on the shared view connection, plus the two pieces this PR had: the `RemoteTmuxNewWorkspaceHostRoutingTests` suite and the "Couldn't Create a tmux Session" alert, now wired on both the multiplexed and the one-shot create paths.

Keeping both open only splits review of the same fix.

PR #7210, opened 2026-07-02
Design: New Local Workspace in the command palette

ejc3:
Design placeholder for a follow-up that adds a **New Local Workspace** entry to the command palette — the palette companion to the New Local Workspace File-menu item.

In a remote-tmux mirror window, plain New Workspace (⌘N) spawns a tmux session on the window's host instead of a local workspace. The menu-item change adds a "New Local Workspace" action that forces local creation, shown only when New Workspace would otherwise go remote. This specs the same thing for the command palette, gated the same way and running the same action.

**No code — this PR is just the design doc**, opened as a draft so the implementation can be reviewed against a written plan (and referenced from the menu-item PR). The doc was adversarially reviewed; it calls out:

- the testability seam — expose the contribution via a `static` factory (like the auth commands) so the `when:` gate is unit-testable;
- the menu/palette reactivity gap and the invalidation hook that keeps the two surfaces in lockstep while the palette stays open across a tab/window switch;
- the fail-safe when there is no host / the beta is off (gate reads false → hidden);
- an auth-command-style gate test in the app-target test suite (not the search-engine suites, which don't evaluate `when:`).

Depends on the New Local Workspace menu-item PR #7214, which adds `performNewLocalWorkspaceAction` and `wouldNewWorkspaceSpawnRemote` (the shared action and the gate this design reuses).

### DIFF
--- a/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/ShortcutAction+Defaults.swift
+++ b/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/ShortcutAction+Defaults.swift
@@ -44,6 +44,7 @@ extension ShortcutAction {
         case .quit: return ShortcutStroke(key: "q", command: true)
         case .toggleSidebar: return ShortcutStroke(key: "b", command: true)
         case .newTab: return ShortcutStroke(key: "n", command: true)
+        case .newLocalWorkspace: return ShortcutStroke(key: "n", command: true, control: true)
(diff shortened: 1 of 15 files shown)


ejc3:
Closing this one. It's a design doc for a command-palette entry whose code doesn't exist yet, so there's nothing here to review or land on its own. When the palette entry gets built, the reasoning can come with the code.

The File-menu version of New Local Workspace is #7214.

Issue #2673, opened 2026-04-07
Remote workspace mode: SSH ControlMaster + first-class remote tmux/shell support


benpayne:
## Problem

cmux currently treats remote SSH connections as plain terminal panes. When a user runs `cmux ssh user@host`, they get a single SSH session in a pane — and any subsequent operation against that host (opening another shell, listing tmux sessions, querying file tree, etc.) requires a fresh SSH connection with fresh auth.

This becomes especially limiting once #560 (tmux control panel) is in place: the new tmux sidebar shows local sessions only, even when the user is actively working on a remote host. The natural ask is "show me tmux sessions on the remote I'm SSH'd to, and let me click to attach" — same as local sessions, just on the remote.

More broadly, cmux needs a first-class concept of a **remote workspace**: an authenticated connection to a host that the rest of cmux can route operations through, without re-prompting for credentials and without opening a new TCP/TLS handshake per command.

## Why not a VSCode-style server?

VSCode Remote works by deploying `vscode-server` (a Node.js daemon, ~80–200MB) to the remote host on first connection. All subsequent operations talk to that daemon over a single SSH tunnel. It enables file watching, language servers, port-forwarding daemons, extensions running remote code — none of which cmux needs.

A cmux server would be:
- A maintenance burden (signed binaries, multiple architectures, version sync with the local client)
- A new trust surface (a daemon running as the user's account on every server)
- Months of work for capabilities cmux doesn't actually need

cmux's remote feature surface is much smaller than VSCode's:
- Open a new shell on the remote in directory X
- List/create/attach/kill tmux sessions on the remote
- Maybe: detect listening ports for forwarding
- Maybe: surface git/cwd/file metadata for the sidebar

**Every one of those is a single shell command** that can run as `ssh -S <sock> host <command>`.

## Proposed approach: SSH ControlMaster + remote tmux

OpenSSH's `ControlMaster`/`ControlPath` provides connection multiplexing out of the box. cmux already uses it in `Sources/TerminalSSHSessionDetector.swift` for file uploads.

The pattern:

```sh
# Open a master once, with auth (interactive prompt happens here)
ssh -M -S ~/.cmux/ssh/<alias>.sock -fnNT user@host

# Subsequent commands reuse the master — no auth, no handshake, subsecond
ssh -S ~/.cmux/ssh/<alias>.sock host tmux list-sessions -F '...'
ssh -S ~/.cmux/ssh/<alias>.sock -t host tmux attach -t mywork
[...]

PR #7264, opened 2026-07-03
ssh-tmux: mirror into current window by default (--new-window keeps dedicated window)

0xJord4n:
## Summary

`cmux ssh-tmux <host>` now mirrors a remote host's tmux sessions as plain workspaces in the **current window** by default — no group, no dedicated new window. A `--new-window` flag preserves the old dedicated-window behavior.

Previously the command always opened a dedicated new window that quarantined the remote sessions away from your local workspaces. Now remote sessions sit alongside local ones in the same sidebar.

## What changed

- **`RemoteTmuxController.mirrorHostInCurrentWindow`** — hardened from the old `mirrorHost`: auth-required handling, no-window fallback (creates a plain window), idempotent reuse, `beginAttach` guard, cancellation, window-raise on activate, and a fail-loud guard that throws (without discarding the window) when zero sessions mirror.
- **New socket command `remote.tmux.attach_here`** (registered at dispatch + worker allowlist + execution policy); retired `remote.tmux.mirror`.
- **CLI**: `cmux ssh-tmux` routes to the current-window mirror by default; `--new-window` routes to the old `remote.tmux.window` path verbatim.
- **Docs + localization**: remote-tmux docs page updated; `cli.help.ssh-tmux` and docs strings localized in English + Japanese.

## Behavior decisions

- **Session end (network loss)** → reconnect, same as SSH workspaces (existing path, unchanged).
- **Session end (tmux session killed)** → the workspace closes (nothing to reconnect to); the window and other workspaces survive.
- **New tmux session creation from cmux** → not offered in v1; new sessions appear via the live feed.

## Testing

- **Unit**: 61 remote-tmux tests pass (`cmux-unit` scheme). Includes a new guard asserting `remote.tmux.attach_here` stays dispatched and validates params before touching the network, and an updated capabilities assertion for the method rename.
- **Live E2E** against a real tmux host: default mirrors sessions into the current window (no new window, local + remote coexist); `--new-window` opens a dedicated window; killing a remote session closes only its workspace while the window survives.

## Notes

Behind the existing `remoteTmux` beta flag. No new keyboard shortcut. No schema/config changes.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/7264"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785680166&installation_id=133855650&pr_number=7264&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F7264&signature=61c4eb11ed35b0e683826079a9cb651ca13425b578d9d50e525fa2d5176934fe"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
`cmux ssh-tmux <host>` now mirrors remote tmux sessions into your current window by default; use `--new-window` for a dedicated window. Added `remote.tmux.window`; socket calls stay focus‑neutral unless `activate: true`, and closing a mirrored workspace detaches (never kills).

- **New Features**
  - CLI/docs: default to current-window; `--new-window` opens a dedicated window. Help updated and localized.
  - Socket/controller: added `remote.tmux.window` (dedicated window) alongside `remote.tmux.mirror` (current window). Focus is opt‑in; creates the window after SSH preflight, consolidates any existing mirrors, removes the bootstrap tab, and throws if no sessions mirror.

- **Bug Fixes**
  - Closing a mirrored tab/window or via v2 socket now detaches without killing the remote session; killing only happens via an explicit disconnect.
  - Failure handling: clean up partial mirrors, discard a just-created dedicated window, never fabricate a `window_id`, and return a localized “could not create a new window” error; both mirror entry points validate params and fail early with structured errors.
  - Tests/CI: added regressions for detach‑on‑close, dedicated‑window consolidation, focus neutrality, and param validation; focused gate runs before flaky suites.

<sup>Written for commit f6b650e43d8787168443d3587094e34025355336. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/7264?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * `cmux ssh-tmux` now mirrors remote tmux sessions into the **current window** by default.
  * Added `--new-window` to mirror into a **dedicated new window** (with updated focus/activation behavior).
  * Added socket command `remote.tmux.attach_here` for current-window mirroring (and updated docs to reflect the default entry point).

* **Bug Fixes**
  * Improved attach preflight with structured auth-required results and safer, idempotent mirroring.
  * Mirroring now isolates per-session failures and validates that workspaces were actually created.
  * Closing a mirrored remote tmux workspace now **detaches** instead of killing the remote session.

* **Documentation / Tests**
  * Updated help text, Remote tmux beta docs, localized strings, and socket-command examples.
  * Added/updated capabilities tests and new regression coverage for attach/close semantics.


0xJord4n:
Addressed the bot review findings in `df562509be`:

- **`?? UUID()` fabricated window_id** (CodeRabbit) — fixed. Both the reuse and success paths now resolve `windowId(for:)` once and **throw** `RemoteTmuxError.unreachable(...)` when it's nil, instead of returning a fabricated UUID the caller would treat as a live window. This matches the method's existing fail-loud posture and also removed a redundant double lookup.
- **`attachCli` omits `--no-focus`** (CodeRabbit, en + ja) — fixed. Both message catalogs now list `--no-focus` alongside `--port`, `--identity`, and `--new-window`. (The `cli.help.ssh-tmux` help text already had `--no-focus` in its `Flags:` list, so no change needed there.)
- **Spec drift: `attach-here` hyphen + `Status: Draft`** (Greptile) — fixed. The design spec now uses `remote.tmux.attach_here` (underscore, matching production) and `Status: Implemented (PR #7264)`.

Build green, 61 unit tests pass (incl. the `attach_here` socket guard), and the live E2E behavior described in the PR body is unchanged.

robertnisipeanu:
When I introduced the ssh-tmux feature, the idea to keep them in the same window also sparkled in my head.

In this case, you would loose a very important feature of working remotely: being able to create new remote workspaces (sessions). It's also not clear for me if you can cleanly disconnect from the remote session (without closing the workspaces, since that would trigger the remote tmux sessions to also be killed)?
Yes, I know the old default would still be usable via a flag, but I think the solution should not degrade the "default" experience.

I actually liked the idea from this issue, under `Sidebar UX example`, but I did not want to bring an UX change together with the new ssh-tmux as it would have decreased the likelihood of it making it into master (so I kept it out of MVP1): https://github.com/manaflow-ai/cmux/issues/2673
With the example from the issue, there could be even dedicated + button for creating a new workspace and X button for disconnecting from that remote session.

Ofcourse, this is not my call, by the maintainers call; it's just my opinion as someone that does uses this feature daily.

But just to be clear: I love the idea, and I definitely would find it useful to have both local and remote sessions in the same window; but it would need to retain the 2 basic features: being able to detach from the remote; and being able to create more workspaces in the remote easily.

0xJord4n:
@robertnisipeanu thanks for this — you're right on both counts, and the detach point is the one that actually mattered.

**Detach, not kill (fixed in `54417d76`):** you nailed a real data-loss trap. With mirrors living as plain workspaces in the current window, closing a workspace ran `kill-session` on the remote — so the natural "get this off my screen" gesture silently killed your live tmux session. That's now fixed: **closing a mirror workspace (or its window, or quitting) detaches and leaves the session alive on the server for resume.** Killing is no longer ever a close side-effect.

- `closeWorkspace` routes mirrors to `detachMirrorWorkspaceKeptOpenLocally` instead of `handleWorkspaceClosed`.
- `markRemoteTmuxKillOnWindowCloseIfNeeded` is now a no-op, so the last-tab / window-close / app-quit paths all fall through to the existing detach handlers.
- The kill machinery (`handleWorkspaceClosed` + the window kill-marker) is retained but dormant — it's exactly what a future **explicit "disconnect / kill session" action** should call.

Verified live against a real tmux host: mirrored 9 sessions into the current window, closed every mirror workspace, and all 9 sessions survived (host session set unchanged). Added a RED/GREEN regression test (`RemoteTmuxMirrorCloseDetachTests`) so the kill-on-close behavior can't come back.

**Creating new remote sessions:** agreed this is a genuine gap, not just a nice-to-have — and the [#2673 "Sidebar UX example"](https://github.com/manaflow-ai/cmux/issues/2673) direction (grouped `REMOTE: host` section with a dedicated `+` for new-session and `X` for clean disconnect) is the right home for both "create session" and an explicit "kill this session" affordance. I kept it out of this PR deliberately to land the detach fix cleanly, same reasoning you gave for keeping it out of MVP1 — but the dormant kill path above is left in place precisely so that `X` has something to call. Happy to follow up with that as a separate PR.

So this PR now retains detach; new-session creation is the tracked follow-up. Ultimately the maintainers' call on sequencing, but I think this keeps the default experience whole in the way you're describing.

robertnisipeanu:
@0xJord4n not sure this is a step in the right-direction either, because now detach is a possibility, but a synced close is not possible anymore. So while now detaching is possible again, closing & cleaning up sessions/tabs on the remote is degraded (not possible anymore).

0xJord4n:
@robertnisipeanu agreed — swinging from "always kill" to "always detach" trades one gap for the other. The close gesture is simply ambiguous: it can't express both "get this off my screen" and "I'm done with this session."

I think the two verbs need to be separate, like tmux itself (`detach` vs `kill-session`):

- **Default close = detach.** The asymmetry is risk: an accidental detach costs a re-attach; an accidental kill is unrecoverable work loss. The destructive verb shouldn't be reachable by the same gesture as the safe one.
- **Explicit kill = its own affordance.** The kill path (`handleWorkspaceClosed` + the window kill-marker) was deliberately kept dormant for exactly this. Concretely: a context-menu item on mirror workspaces — **"Kill Remote Session and Close"** — next to the existing "Reconnect" item, so a synced close is one right-click away. Host-level cleanup (per-host `X`, batch kill) belongs with the #2673 sidebar design as the follow-up.

That would restore synced close without making it the default blast radius — and it's a small diff since the machinery is already in place. Does that shape work for you? If so I'll add it to this PR.

PR #7544, opened 2026-07-07
fix(settings): merge shortcut edits into cmux.json instead of clobbering other bindings

JiseokJang:
## Summary

Editing **any** keyboard shortcut in Settings (**⌘,** → Keyboard Shortcuts) rewrote the **entire** `shortcuts.bindings` object in `~/.config/cmux/cmux.json`, silently dropping any binding the Settings UI did not load into its typed model — including bindings authored directly in the file in a schema-valid form the UI never emits (e.g. a plain string stroke). Editing action A destroyed unrelated action B.

### Repro (before)

```jsonc
// ~/.config/cmux/cmux.json
{ "shortcuts": { "bindings": { "focusDown": "cmd+shift+down" } } }
```

1. `cmux reload-config` → `focusDown` (⌘⇧↓) works.
2. Settings → change a *different* action (e.g. Global Search). Save.
3. `focusDown` is **gone** from `cmux.json` and falls back to default.

## Root cause

- **Full-block replace on save:** `ShortcutListModel.write` / `GlobalHotkeySection.write` handed the whole in-memory model dict to `JSONConfigStore.set`, which `JSONPath.assign`-replaced the entire `shortcuts.bindings` leaf.
- **Lossy typed model:** the package `StoredShortcut` decoded only the object form, and `Dictionary` decode is all-or-nothing — so a single string/array/sentinel entry dropped the **whole** map from the UI model (the UI showed no shortcuts), which was then written back without it.

## Fix

Both halves, so bindings survive **and** display correctly:

**1. Merge, don't replace (stops the data loss).** Save now merges only the changed entries into the existing on-disk bindings object, preserving every other entry verbatim (any schema-valid form, exact authored shape):
- **`JSONConfigStore.merge(_:for:)`** — per-key merge into the raw on-disk dictionary read by `mutateRoot`, overwriting only the passed keys and leaving untouched siblings byte-for-form intact.
- **`ShortcutListModel.write`** — per-action edits merge only the keys that actually changed; **Reset Defaults** still intentionally replaces the whole object.
- **`GlobalHotkeySection.write`** — merges only the global-hotkey key.

**2. Lossless load (fixes the blank UI).** `StoredShortcut.decodeFromJSON` now parses every schema-valid form — string stroke, chord array, and unbind sentinels (`""`/`none`/`clear`/`unbound`/`disabled`/`null`) — in addition to the object form, so one hand-authored entry no longer blanks the whole map and the binding shows in the UI. The grammar mirrors the app-target runtime reader, so a UI-loaded binding and a runtime-resolved one agree on the canonical key form (arrows as glyphs, Return as `"\r"`, etc.). It's a lenient *decode* (modifier-less strokes decode for display); the recorder still enforces modifier requirements at record time.

## Testing

- `ShortcutListModelTests` (18, incl. new `editingOneActionPreservesHandAuthoredStringBinding` and `handAuthoredStringBindingLoadsIntoModel`; existing `resetAllWritesEmpty`, `failedWriteRollsBackOptimisticBinding`, `consecutiveAssignmentsMergeBeforeObserverEcho` still pass).
- `SettingCodableTests` (12, incl. new StoredShortcut decode of object/string/chord/sentinel/null forms, invalid-string rejection, and mixed-form dictionary survival).
- Full app build green.

---

> [!NOTE]
> **Medium Risk**
> Changes how user keyboard bindings are read and written to `cmux.json`; incorrect merge or decode logic could still corrupt or mis-display shortcuts, though behavior is covered by new regression tests.
> 
> **Overview**
> Fixes a regression where saving **any** keyboard shortcut in Settings replaced the entire `shortcuts.bindings` object in `cmux.json`, silently removing bindings the UI never loaded (e.g. hand-edited string strokes like `"cmd+shift+down"`).
> 
> **Merge-on-save:** Adds `JSONConfigStore.merge(_:for:)` to patch only changed keys into the on-disk map. `ShortcutListModel` and `GlobalHotkeySection` now merge changed entries on normal edits; **Reset Defaults** still replaces the whole bindings object via `replacingAll: true`.
> 
> **Lossless load:** `StoredShortcut.decodeFromJSON` now accepts schema-valid string strokes, chord arrays, unbind sentinels, and `null`, with parsing aligned to the app runtime so mixed-format maps decode instead of failing all-or-nothing and blanking the shortcuts UI.
> 
> **UI ingest:** `ShortcutListModel.ingestBindings` filters out bare first strokes on modifier-required actions (matching runtime) while leaving those keys on disk untouched.
> 
> Regression tests cover clobber preservation, mixed-form decode, and bare-stroke presentation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d298946ecd9b0bac956bfce15b2d9fce89514f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Improved shortcut handling to support more on-disk formats, including hand-edited strings, chord shortcuts, and unbound values.
  * Updated shortcut saving to preserve existing entries when only part of the set changes.

* **Bug Fixes**
  * Editing one shortcut no longer overwrites other manually configured shortcuts.
  * Resetting shortcuts now fully clears the saved bindings.

* **Tests**
  * Added regression coverage for mixed shortcut formats and lossless round-tripping.

### DIFF
--- a/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Stores/JSONConfigStore.swift
+++ b/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Stores/JSONConfigStore.swift
@@ -120,6 +120,31 @@ public actor JSONConfigStore {
         }
     }
 
+    /// Merges `entries` into the dictionary already stored at `key`'s path,
+    /// overwriting only the provided keys and leaving every other on-disk entry
+    /// byte-for-form untouched.
+    ///
+    /// Unlike ``set(_:for:)`` — which replaces the whole object at the leaf —
+    /// this preserves sibling entries the caller did not pass, including entries
+    /// authored in a schema-valid form this process would never emit itself
+    /// (e.g. a hand-written string stroke `"cmd+shift+up"` rather than the
+    /// object form). Use it for user-editable maps like `shortcuts.bindings`,
+    /// passing only the entries that actually changed, so editing one entry can
+    /// never rewrite or silently drop the others.
+    public func merge<Element: SettingCodable>(
+        _ entries: [String: Element],
+        for key: JSONKey<[String: Element]>
+    ) throws {
+        guard !entries.isEmpty else { return }
+        try mutateRoot { root in
+            var existing = (key.path.lookup(in: root) as? [String: Any]) ?? [:]
+            for (entryKey, element) in entries {
+                existing[entryKey] = element.encodeForJSON()
+            }
+            key.path.assign(existing, in: &root)
+        }
+    }
+
     /// Removes the key's entry from the file. Parent objects that become
     /// empty are pruned. The file itself is not deleted even when no entries
     /// remain.
--- a/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/StoredShortcut.swift
+++ b/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/StoredShortcut.swift
@@ -42,11 +42,30 @@ public struct StoredShortcut: Sendable, Equatable, Hashable, Codable, SettingCod
     }
 
     public static func decodeFromJSON(_ raw: Any?) -> StoredShortcut? {
-        guard let raw, !(raw is NSNull) else { return nil }
-        guard let data = try? JSONSerialization.data(withJSONObject: raw, options: .fragmentsAllowed) else {
-            return nil
+        guard let raw else { return nil }
+        // Accept every schema-valid binding form (`$defs.shortcutBinding`) so a
+        // hand-authored string stroke, chord array, or unbind sentinel loads
+        // into the UI model instead of being dropped. Dropping any one entry
+        // makes the all-or-nothing dictionary decode blank the entire bindings
+        // map, which is what left the Settings UI showing no shortcuts. The
+        // grammar mirrors the app target's runtime reader so a UI-loaded binding
+        // and a runtime-resolved one agree on the canonical key form.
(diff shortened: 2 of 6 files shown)


JiseokJang:
Thanks for the reviews — both addressed in the latest commits.

- **Observer echo wipes saved shortcut UI** (cursor, Medium): resolved by the lossless-load change. `StoredShortcut.decodeFromJSON` now parses the string / chord-array / unbind-sentinel forms too, so `jsonStore.values(for:)` no longer all-or-nothing decodes the map to `[:]`. A shortcut saved via Settings stays in the UI after the post-write observer echo, and the whole list no longer blanks when the file contains a hand-authored non-object entry.

- **Match runtime parsing** (greptile, P1): the value decoder is intentionally lenient (it has no action context), but `ShortcutListModel.ingestBindings` now applies the runtime's action-aware first-stroke rule — a modifier-less ("bare") first stroke is only surfaced for actions that allow it (or for Space). So a hand-authored `"focusDown": "down"` is no longer presented as the active binding while the app ignores it; the UI shows the same effective shortcut the runtime resolves. The raw entry still stays in `cmux.json` (merge-on-save never drops untouched keys) — it's just not shown as loaded. Test: `bareFirstStrokeBindingIsNotPresentedForModifierRequiredAction`.

Verification: `ShortcutListModelTests` (19) and `SettingCodableTests` (12) pass; full app build green.

PR #8012, opened 2026-07-13
Fix iOS recovery from stalled mobile RPC writes

### LINKED ISSUE #8005: iOS: constant "Connection lost - Retry" / all workspaces Disconnected while the Tailscale path is healthy (typing a few chars kills the session)
### Problem or motivation

The iOS app constantly drops its live connection to the paired Mac even when the underlying Tailscale link is healthy. This is now hitting external beta users, not just us.

Beta user report (email, 2026-07-12):

> I'm having lots of issues maintaining connection via tailscale between my mac and phone when using the iOS app. I get a "Connect lost - retry" popup consistently after entering just a couple characters into the terminal line in any of my Cmux workstations (see screenshot). Is this an issue that is common in the current beta?

Our own dogfooding shows the same symptoms:

- "Connection lost / Retry" pill appears over the terminal right after typing a few characters (reporter was on LTE; also reproduces on Wi-Fi).
- The Computers list shows "Disconnected. The live connection dropped. The selected cmux build may still be online. Tap Reconnect." with every workspace row marked Disconnected.
- A "Reconnecting..." overlay sits on top of live terminal content for long stretches; input gets queued ("Queued follow-up inputs") instead of delivered.
- At the same moment, the Tailscale app pings the same Mac with a direct connection at ~63 ms, so the tailnet path itself is fine; it is the cmux session layer that gives up.

Two screen recordings of the behavior exist (2026-07-10 12:29 PT, 2026-07-11 02:47 PT) and can be provided on request.

### Repro steps

1. Pair an iPhone with a Mac (standard Tailscale pairing).
2. Open any workspace with a terminal in the iOS app.
3. Type a couple of characters into the terminal.
4. Observe the "Connection lost - Retry" popup; the Computers list then shows everything Disconnected until a manual Reconnect.

### Expected

The mobile session layer should ride out transient stalls, roaming (Wi-Fi to cellular), and brief backgrounding without declaring the connection dead, should auto-reconnect without a manual tap, and should never let a healthy direct tailnet path surface as a constant "Connection lost". Durable, minimally-throwing connectivity is the top mobile priority right now.

### Likely area

- `Packages/iOS/CmuxMobileTransport` (`CmxNetworkByteTransport`, NWConnection viability/timeout handling)
- `Packages/iOS/CmuxMobileRPC` (`MobileCoreRPCSession` keepalive, write acks, reconnect policy)
- `Sources/Mobile/MobileHostService.swift` (host-side idle/ticket expiry, per-connection teardown)
- Reconnect UX: banner state machine and auto-retry backoff

### Related

- #7230 (trusted manual-host pairing for subnet-router setups)
- #6700 (pairing to a cloud VM, removing the Mac + Tailscale requirement)
- #8000 / #8001 (cmux hive umbrella; multi-device); the planned iroh migration does not replace fixing session durability on the current Tailscale TCP path.

austinywang:
## Summary

- track the active serialized RPC write and recycle the persistent transport only when that write itself exceeds the request deadline
- distinguish a blocked transport send from an ordinary response-only timeout, leaving healthy multiplexed sessions intact
- route write stalls, failed liveness probes, network transitions, and manual Retry through one transport-rotation path that preserves the paired-Mac ticket, route, workspace state, terminal surfaces, and stable event stream ID
- keep the iOS UI in automatic `Reconnecting…` until the replacement subscription is acknowledged, falling back to the existing Retry state only when the fresh subscription fails
- bind read/write loop teardown to its connection generation so a late failure from an old socket cannot close its replacement

## Root cause

`MobileCoreRPCSession` serializes frames through one writer task. If `CmxByteTransport.send` never received its `contentProcessed` callback, the request timeout removed the response continuation but left the writer parked forever. Later keystrokes, keepalives, and re-subscriptions queued behind that write. Existing recovery paths then re-subscribed on the same wedged session, which explains why Tailscale remained healthy while every cmux workspace became Disconnected.

The transport already lets Network.framework own transient `.waiting` retry/backoff. This change does not add sleeps or inflate a timeout; it replaces only a session proven stale by a blocked send or failed positive-liveness probe.

## Regression proof

Commit `eef873e72f` adds the failing test without the fix. Before commit `86ea04298b`, the first send timed out, the second request also timed out, only one transport was created, and the stalled transport never closed.

The fixed test additionally releases the old stalled writer after transport #2 is live and proves that late failure cannot tear down the replacement connection.

## Verification

- `/usr/bin/arch -arm64 /usr/bin/xcrun swift test --package-path Packages/iOS/CmuxMobileRPC` — 79 tests passed
- `/usr/bin/arch -arm64 /usr/bin/xcrun swift test --package-path Packages/iOS/CmuxMobileShell` — 482 tests passed
- `python3 scripts/swift_file_length_budget.py --base-ref origin/main` — passed
- `python3 scripts/check-package-resolved-policy.py` — passed
- `./scripts/lint-ios-package-conventions.sh` — passed with baseline warnings only
- `git diff --check` — passed

I chose deterministic injected-transport coverage over cloud/device video: the reported failure is a physical tailnet timing issue, while the regression directly stalls `transport.send`, exercises the real RPC writer/session state machine, and verifies automatic shell recovery without relying on radio conditions. No XCUITest or local app build was run.

## Localization audit

No user-facing copy was added or changed. The existing localized Reconnecting, Connection lost, and Retry surfaces are reused, so `Resources/Localizable.xcstrings` and the English/Japanese catalogs require no changes.

## Related work

This is complementary to #7876: that PR centralizes broader connectivity lifecycle ownership, while this change fixes the lower-level persistent-writer wedge and ensures recovery rotates a stale transport rather than repairing streams on the same session.

Closes #8005

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8012"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786578194&installation_id=133855650&pr_number=8012&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8012&signature=f0e5a8aaf1e386988e9fc06e4e016f2b57942fff3e249cfa820fb4034601de02"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fixes iOS drops when an RPC write stalls by detecting blocked `transport.send`, rotating only the stale transport, and auto‑resubscribing without losing session state. If the replacement subscription fails, we now refresh the latest stored route and reconnect before falling back to Retry. Addresses #8005.

- **Bug Fixes**
  - `CmuxMobileRPC`: Track the active write; if a request times out while still in `send`, cancel it, surface `transportWriteTimedOut`, and recycle only the transport (healthy multiplexed sessions stay intact).
  - `CmuxMobileRPC`: Split read/write loops and bind teardown to a connection generation; serialize detached transport cleanup (`enqueueTransportClose`) so one close is in‑flight and one queued, backpressuring new connections to avoid leaks or livelock; add `MobileCoreRPCClient.resetConnectionForRecovery()` to drop just the transport.
  - `CmuxMobileShell`: Route write stalls, failed liveness probes, event‑stream end, network changes, and manual Retry through one single‑flight recovery that rotates the transport and preserves workspace/terminals; keep automatic Reconnecting until the new subscription ACK.
  - `CmuxMobileShell`: If the replacement subscription fails, escalate to reconnect using the latest stored route; Retry from Unavailable also refreshes the stored route.
  - Error handling/UI: Add `MobileShellConnectionError.transportWriteTimedOut`; classify it as a Mac‑availability failure. Plain `requestTimedOut` is response‑scoped and does not trigger availability recovery.
  - Tests: Cover stalled‑write timeout/cancel and late‑failure isolation, serialized/queued transport close with connection backpressure, manual Retry rotating the transport, automatic recovery on availability failures, single‑flight across concurrent triggers, failed replacement subscription escalating to the stored route, and generation‑scoped workspace failures.

<sup>Written for commit 96bb9d3e5b4fb1423ef881b639bf9898fd1352cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8012?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCClient.swift
+++ b/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCClient.swift
@@ -65,6 +65,16 @@ public final class MobileCoreRPCClient: MobileSyncing, Sendable {
         await session.tearDown(error: .connectionClosed)
     }
 
+    /// Drops the current transport while preserving this client's route and ticket.
+    ///
+    /// The next request lazily establishes a fresh transport. Call this only
+    /// after a liveness probe or availability failure proves the
+    /// persistent session is stale; ordinary response timeouts remain isolated
+    /// to their request.
+    public func resetConnectionForRecovery() async {
+        await session.tearDown(error: .connectionClosed)
+    }
+
     /// Subscribe to server-pushed events. Returns a stream of envelopes
     /// matching any of the requested topics. Cancel by terminating iteration.
     public func subscribe(to topics: Set<String>) async -> AsyncStream<MobileEventEnvelope> {
--- a/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCClient.swift
+++ b/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCClient.swift
@@ -59,15 +59,17 @@ public final class MobileCoreRPCClient: MobileSyncing, Sendable {
         )
     }
 
-    /// Tear down the persistent transport (called when the client is
-    /// replaced or the user signs out).
+    /// Atomically detaches the persistent transport (called when the client is
+    /// replaced or the user signs out). Physical close cleanup is serialized
+    /// separately so a transport callback cannot block logical disconnect.
     public func disconnect() async {
         await session.tearDown(error: .connectionClosed)
     }
 
     /// Drops the current transport while preserving this client's route and ticket.
     ///
-    /// The next request lazily establishes a fresh transport. Call this only
+    /// The next request lazily establishes a fresh transport while physical
+    /// cleanup of the detached transport proceeds separately. Call this only
     /// after a liveness probe or availability failure proves the
     /// persistent session is stale; ordinary response timeouts remain isolated
     /// to their request.
--- a/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCSession+ReadLoop.swift
+++ b/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCSession+ReadLoop.swift
@@ -1,0 +1,47 @@
+internal import CMUXMobileCore
+import Foundation
+
+extension MobileCoreRPCSession {
+    func readLoop(
+        transport: any CmxByteTransport,
+        connectionID: UUID
+    ) async {
+        var buffer = Data()
+        while !Task.isCancelled {
+            let chunk: Data?
+            do {
+                chunk = try await transport.receive()
+            } catch {
+                await tearDownIfInstalled(
+                    connectionID: connectionID,
+                    error: .connectionClosed
+                )
+                return
+            }
+            guard let chunk, !chunk.isEmpty else {
+                if chunk == nil {
+                    await tearDownIfInstalled(
+                        connectionID: connectionID,
+                        error: .connectionClosed
+                    )
+                    return
+                }
+                continue
+            }
+            buffer.append(chunk)
+            let frames: [Data]
+            do {
+                frames = try MobileSyncFrameCodec.decodeFrames(from: &buffer)
+            } catch {
+                await tearDownIfInstalled(
+                    connectionID: connectionID,
+                    error: .invalidResponse
+                )
+                return
+            }
+            for frame in frames {
+                dispatch(frame: frame)
+            }
+        }
+    }
+}
--- a/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCSession+ReadLoop.swift
+++ b/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCSession+ReadLoop.swift
@@ -28,6 +28,10 @@ extension MobileCoreRPCSession {
                 }
                 continue
             }
+            guard !Task.isCancelled,
+                  installedConnectionID == connectionID else {
+                return
+            }
             buffer.append(chunk)
             let frames: [Data]
             do {
--- a/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCSession+TransportClose.swift
+++ b/Packages/iOS/CmuxMobileRPC/Sources/CmuxMobileRPC/MobileCoreRPCSession+TransportClose.swift
@@ -1,0 +1,30 @@
+internal import CMUXMobileCore
+import Foundation
+
+extension MobileCoreRPCSession {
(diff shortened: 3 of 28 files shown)


azooz2003-bit:
Ran down the `mobile-core-package` / `ios-tests` failure (https://github.com/manaflow-ai/cmux/actions/runs/29301396629). Verdict: suite-parallelism flake, not a bug in this PR's recovery logic. `ios-tests` is only the aggregator re-reporting it.

`manualRetryRotatesPersistentTransportAndResubscribes` passes 110/110 in isolation at 96bb9d3e5b on an M-series Mac: 40 plain runs, 40 with breadcrumb instrumentation, 30 under 10x CPU-hog load. The whole-suite run is what flakes, with a rotating victim: three local full-suite runs produced three different single-test wall-clock timeouts, two of them in files this PR does not touch and one pair on current `main` (details and data in https://github.com/manaflow-ai/cmux/issues/8143). In the CI failure the entire 485-test run took 10.8s on a 6-vCPU runner, so the test's 10s `waitForCount` deadlines expired while its recovery task's main-actor continuations sat behind the rest of the parallel suite.

Rerun is currently impossible (`test-ios` is one of the workflows disabled by the CI experiment until 2026-07-21) and the check is not merge-required this week, so the red X is frozen but not blocking. If you want the new tests immune before the suite-wide fix, giving their waits more headroom than the suite's worst-case wall-clock, or marking the recovery suite `.serialized`, would do it.

One spot worth an eyeball while you are in there (I could not make it fire in 110 runs, so no repro claim): `beginTerminalEventSubscriptionStart`'s ack path calls `markMacConnectionHealthy(completingRecovery: true)` whenever the listener generation is still current. If that ack lands while a newer recovery sits in `.resettingSession`, before its resync supersedes the listener, `completeMobileConnectionRecovery()` nils `recoveryID` and the in-flight recovery task then fails its `recoveryID` guards and abandons without rotating the transport. Completing recovery only from the awaiting-subscription states would close that window.

PR #8064, opened 2026-07-14
Fix three terminal clipboard copy failures (Cmd+C swallowed on agent TUI panes, empty writes clearing the clipboard, VT-export capture race)

a05031113:
## Summary

Copying terminal text failed intermittently — and on agent TUI panes (Claude Code) almost always. Dogfooding with DEBUG probes on a live Claude Code pane isolated three independent bugs that stack:

### 1. Cmd+C swallowed by stale selection validation (main user-facing failure)

On a pane whose TUI redraws continuously, `ghostty_surface_has_selection` reports `false` while the runtime still holds a live selection — the highlight stays visible and copy-on-select still produces the full text. `validateUserInterfaceItem` gated Copy on that flag, so the Edit ▸ Copy item goes disabled, and **a disabled menu item swallows ⌘C at key-equivalent dispatch** — the event never reaches the surface, so the runtime's own `copy_to_clipboard` binding can't run either.

**Repro fingerprint:** select text in a pane running Claude Code → Copy is greyed in the Edit menu and ⌘C is a silent no-op — but *holding the Edit menu open* and pressing ⌘C copies fine (menu tracking lets the key fall through to the surface).

Fix: enable Copy whenever a surface exists. Copying with no selection remains a harmless no-op.

### 2. Empty clipboard writes destroy the clipboard

`TerminalPasteboardService.writeString` ran `clearContents()` before writing even when the payload was empty (e.g. copy-on-select firing after the selection was already invalidated by a redraw). The user's previous clipboard content is silently destroyed — paste turns up empty and clipboard managers record nothing. Empty standard writes are now ignored (and logged).

### 3. VT-export clipboard capture swallows concurrent user copies

The one-shot `standardClipboardWriteCapture` armed by `readTerminalTextFromVTExportForSnapshot` (session snapshots, mobile replay, agent naming — all frequent on agent panes) diverted the *next* standard-clipboard write from **any** source. A user copy racing an in-flight VT export was captured instead of reaching `NSPasteboard`. The capture now carries a predicate and only consumes writes that match the export payload shape (a single absolute path to an existing file); non-matching writes pass through with the capture left armed. Also checks the previously ignored `setString` result (retry once + log), and injects the standard pasteboard so tests cover the write path without touching the real general pasteboard.

## Tests

- `CmuxTerminalTests`: non-matching writes pass through an armed capture and leave it armed; empty writes don't clear existing content; standard writes land on the injected pasteboard after the capture disarms.
- `cmuxTests/SessionPersistenceTests`: `isPlausibleExportedScreenPath` accepts an existing file path / file URL and rejects user text, missing paths, and directories.

Note on the two-commit red/green policy: the regression tests exercise seams introduced by the fixes themselves (`matching:` predicate, injected standard pasteboard), so a tests-only first commit would not compile. Happy to restructure where feasible if preferred.

## Localization audit

No user-facing strings added or changed (code comments and DEBUG-only `cmuxDebugLog` probes only), so no `Localizable.xcstrings` / web message catalog changes are needed.

## Verification

Built with `./scripts/reload.sh --tag fix-clipboard` and dogfooded on a live Claude Code pane on macOS 26.5: select → ⌘C → paste now works with the Edit menu closed; copy-on-select content survives subsequent selection-clear events; package syntax verified and behavior confirmed by the reporting user across multiple rounds.

https://claude.ai/code/session_01SwGyYRt9QjDUhimcp2xB7E

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8064"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786640566&installation_id=133855650&pr_number=8064&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8064&signature=290a3e6d8f3e7a054fdf2b24588aa03a92dee8a5b2100b3b5db78f07fa400b7a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fixes three clipboard copy failures in terminal panes, especially fast‑redrawing agent TUIs. Cmd+C now works reliably, VT‑export no longer swallows user copies, empty writes don’t wipe the clipboard, and overlapping captures are rejected; fixes now live in `writeRepresentations` and cover all text formats.

- **Bug Fixes**
  - Always enable Copy when a terminal surface exists; copying with no selection is a no‑op.
  - Ignore empty standard-clipboard writes; on write failure after clear, retry once if changeCount is unchanged and log.
  - VT‑export capture only diverts writes that match a single absolute path or file URL under the process temp directory (accepts both /var and /private/var); non‑matching writes pass through and leave the capture armed; capture claim is atomic.
  - Reject overlapping clipboard captures; arming while one is in flight returns nil.
  - Apply capture and empty‑write guards in `writeRepresentations` (match on preferred `text/plain`); inject the standard pasteboard in tests.

<sup>Written for commit 3b2442629aec70a52370674eb0b94337eb9ed163. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8064?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Clipboard export capture is now more conservative and won’t intercept unrelated standard clipboard writes.
  - Empty clipboard writes no longer clear existing clipboard contents.
  - Empty/invalid exported-screen payloads are rejected unless they look plausible and are under the app’s temp directory.
  - Copy menu/shortcut is enabled whenever a terminal surface is available.
- **Enhancements**
  - Clipboard capture supports predicate-based matching, only diverting the first write that satisfies the predicate.
  - Improved clipboard write diagnostics for troubleshooting.
- **Tests**
  - Added coverage for exported-screen path plausibility rules.


a05031113:
Addressed the automated review findings in 8eb7394:

- **One-shot capture consumed twice / TOCTOU** (Greptile P1, CodeRabbit critical): the capture is now claimed atomically — only the write that clears the slot captures and returns; a concurrent matching write (or one racing the disarm in `captureNextStandardClipboardWrite`'s defer) falls through to the real pasteboard instead of being swallowed.
- **Retry clobbering a newer concurrent write** (Greptile P1): the retry only runs while `changeCount` still equals the value returned by our `clearContents`, so it can never destroy a value another writer landed in between.
- **Any existing absolute path matching the export predicate** (Greptile P1) and **filesystem I/O in the runtime callback** (Greptile P2): `isPlausibleExportedScreenPath` is now a pure string prefix check against the process temporary directory (accepting both the `/var/...` and `/private/var/...` spellings), with interior newlines rejected. No `fileExists` on the hot path, and copied paths like `/etc/hosts` no longer match. Tests updated to the new contract.

a05031113:
Addressed the follow-up finding in 8f81791:

- **Overlapping captures replace ownership** (Greptile P1): `captureNextStandardClipboardWrite` now rejects a new capture (returns `nil` without arming) while another is in flight, instead of replacing the slot — so concurrent VT exports can no longer cross-deliver one pane's export to another pane's snapshot. Every caller already treats `nil` as "fall back to a non-capture read" (`readTerminalTextForSnapshot` falls back to the plain base64 read; mobile replay falls back to byte-tee state), so overlap degrades gracefully. Regression test added.

The earlier inline comments on `TerminalController.swift:702` (existing-path match, `fileExists` in the runtime callback) were already addressed in 8eb7394 — the predicate is now a pure string prefix check against the process temporary directory.

bogdanovich:
Bumped into the same issue. Can this get merged?

a05031113:
@bogdanovich Thanks for confirming — sorry you hit it too. Quick workaround until this lands: after selecting, hold the Edit menu open and press ⌘C (a disabled Copy menu item otherwise swallows the key equivalent before the terminal's own copy binding can run), or enable copy-on-select in settings.

Branch is refreshed against latest main (f440d86, no conflicts; the pasteboard files had no upstream changes) and re-verified locally: GhosttyKit + app build green, clipboard capture tests pass. Ready for maintainer review / CI approval.

PR #8132, opened 2026-07-15
remote-tmux: give a freshly split mirror pane key focus so you can type without clicking

ejc3:
## What you see

Split a pane in a mirrored tmux session and the new pane lights up with the selection highlight but takes no keyboard input — you have to click it before you can type. Both panes render a hollow cursor, so neither one actually holds key focus.

Repro:
1. Open a mirrored session.
2. Split a window.
3. Try to type. Nothing lands until you click the new pane.

## Root cause

A window mirror renders each tmux pane as its own `TerminalPanel` inside the mirror's own split tree. Those panels are not workspace split-tabs, so the app's normal focus path — `focusPanel` and the surface's active- and visibility-transition first-responder edges — can't resolve them. A freshly split pane is born already active and already visible inside an already-visible tab, so none of those edges fire and key focus is never established. Only a click (`mouseDown` -> `becomeFirstResponder`) sets it.

## The fix

Track the panes the mirror just created. The first time such a pane becomes the active pane, drive the same first-responder establishment a click does (`moveFocus`), on the creation event edge. Every attempt re-checks that the mirror is still on screen, that this pane is still the active one, and that its window is key — so switching panes cancels a pending focus instead of stealing it, and a background or headless mirror never moves the responder. It's consumed once per created pane, so a later active-pane echo won't re-drive it.

## Tests

- A mirror-layer regression test (red before the fix) that asserts the mirror drives key focus onto the newly split pane's own panel — that panel was empty before.
- A control test that a routine active-pane switch does not re-drive focus.
- Pane input-mapping guards.

The tests assert that the mirror drove focus rather than leaning on AppKit responder timing, so they're timing-independent.

## Known follow-ups

- Establishment currently uses a short bounded retry (~6 x 16ms) to wait for the new pane's surface to mount into the key window. That's a soft timer. The cleaner design drives it off the surface's mount/visibility event edge, or better, makes key focus reconcile to tmux's active pane as a first-class invariant (tmux is the authority) so split, close, zoom, and click are all handled the same way.
- Runtime GUI verification of the split-then-type gesture is still worth doing before this comes out of draft.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8132"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786682358&installation_id=133855650&pr_number=8132&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8132&signature=8a47c798dbed8f7af2f89fb7fc8f9e81529ed914585d93082ff797c7b94c6400"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Newly split panes in a tmux mirror now get keyboard focus on mount, so you can type right away without clicking. Focus is driven from the pane view’s attach edge, with guards to prevent steals in background tabs, across pane switches (A→B→A), or when a non-terminal control is focused.

- **Bug Fixes**
  - Add a one-shot `onDidAttachToWindow` hook on the pane’s `GhosttySurfaceScrollView`; it’s consumed before the callback runs, re-arms only while unattached or in the parking window, and then drives `moveFocus` — no timers or polling.
  - Extract pure policy `paneKeyFocusDecision` (`focusNow`/`waitForMount`/`skip`): wait only when unattached or parked; skip for hidden/non-key windows or when the first responder isn’t in the terminal layer.
  - Track `paneActivationGeneration`; bump on every active-pane transition (including N→O→N and teardown) and validate pending intents against the armed generation to block stale steals.
  - Tests: new `RemoteTmuxMirrorPaneKeyFocusDecisionTests` (truth table) and expanded `RemoteTmuxMirrorNewPaneKeyFocusTests` (attach-hook one-shot and generation bump).

<sup>Written for commit bbd1999e4772d42475a7d6c546b5b5c2342af107. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8132?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Sources/GhosttyTerminalView.swift
+++ b/Sources/GhosttyTerminalView.swift
@@ -8183,6 +8183,13 @@ final class GhosttySurfaceScrollView: NSView {
     private var searchOverlayMutationGeneration: UInt64 = 0
     private var observers: [NSObjectProtocol] = []
     private var windowObservers: [NSObjectProtocol] = []
+    /// One-shot hook consumed the next time this view attaches to a window
+    /// (`viewDidMoveToWindow` with `window != nil`). A single owner arms it to
+    /// act on the mount edge — the remote-tmux mirror uses it to establish key
+    /// focus on a freshly created pane whose view SwiftUI has not yet inserted.
+    /// Consumed before it runs, so reparent churn after the first attach never
+    /// re-fires it.
+    var onDidAttachToWindow: (() -> Void)?
     private var scrollbarTrackingArea: NSTrackingArea?
     private var isLiveScrolling = false
     private var lastSentRow: Int?
@@ -9163,6 +9170,10 @@ final class GhosttySurfaceScrollView: NSView {
         windowObservers.forEach { NotificationCenter.default.removeObserver($0) }
         windowObservers.removeAll()
         guard let window else { return }
+        if let attachHook = onDidAttachToWindow {
+            onDidAttachToWindow = nil
+            attachHook()
+        }
         windowObservers.append(NotificationCenter.default.addObserver(
             forName: NSWindow.didBecomeKeyNotification,
             object: window,
--- a/Sources/RemoteTmuxWindowMirror.swift
+++ b/Sources/RemoteTmuxWindowMirror.swift
@@ -128,6 +128,12 @@ final class RemoteTmuxWindowMirror: RemoteTmuxControlPaneMutationOwner {
     /// active pane, so exactly one creation drives key focus — never a later
     /// active-pane echo or a co-attached client's pane switch.
     @ObservationIgnored private var panesAwaitingCreationFocus: Set<Int> = []
+    /// Bumped on every active-pane transition, including to `nil`. A pending
+    /// creation-focus intent captures this when it arms; any intervening
+    /// transition — even one that returns to the same pane (N→O→N) —
+    /// invalidates the intent, so an attach hook that fires later can never
+    /// replay a stale activation as a focus steal.
+    @ObservationIgnored private(set) var paneActivationGeneration: UInt64 = 0
     @ObservationIgnored var isApplyingRemoteLayout = false
     @ObservationIgnored var isApplyingTmuxFocus = false
     @ObservationIgnored var lastDividerPositions: [UUID: CGFloat] = [:]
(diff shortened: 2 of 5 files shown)


ejc3:
Closing this, with the measurements that make it unnecessary.

**Its target suite passes without it.** `RemoteTmuxMirrorNewPaneKeyFocusTests` on a tree that does *not* contain this PR:

```
Test run with 2 tests in 1 suite passed
** TEST SUCCEEDED **   restarts=0
```

The job asserted the tree lacks this PR's marker (`cmuxOwningGhosttyView` in `RemoteTmuxWindowMirror.swift`) before running, so the green cannot be credited to the patch being present. `main` solved the focus problem another way.

**And it no longer applies to `main`.** This PR contains both halves of one change: it widens `cmuxOwningGhosttyView` in `Sources/App/ShortcutRoutingSupport.swift` and adds a cross-file call at `Sources/RemoteTmuxWindowMirror.swift:772`. On current `main` that helper is still `private func` and its signature has become `cmuxOwningGhosttyView(for:includingHostedSurfaceDescendants:)`. Applying this branch onto `main` takes the call site and silently drops the widening, with no conflict reported, and the build then fails with:

```
Sources/RemoteTmuxWindowMirror.swift:772:16: error: cannot find 'cmuxOwningGhosttyView' in scope
```

That was the only error in a combined build of 24 branches, which is how I found it.

If the mount-edge focus approach is still wanted for its own sake rather than for that suite, it needs rewriting against the current signature rather than a rebase, since the shape it patched no longer exists.

PR #8188, opened 2026-07-15
iOS: add pane-grouped deck and Safari-style pane map

azooz2003-bit:
## Summary

- Publish pane topology through the existing shared `workspace.layout.v1` contract and add the authoritative `workspace.pane_reorder.v1` mutation path.
- Route iPhone drag-to-relayout to the paired Mac, move complete pane contents on macOS, publish one reconciled layout, and roll back optimistic iOS order with a localized failure state when the mutation fails.
- Replace the separate full-screen pane presentation with one workspace-local native navigation stack. It starts with the prior focused terminal already installed, so entering a workspace never flashes the pane map first.
- Keep one shared toolbar across terminal and pane-map endpoints. Back, title, menu, and disappearing items participate in the same native zoom transition.
- Use native collection-view drag/drop plus explicit gesture arbitration so scroll, swipe, drag, and absorbed movement cannot resolve as pane selection; only an intentional resolved tap focuses a pane.
- Preserve Safari-style split ratios, card sizing, tab grouping, live previews, and older-Mac fallback behavior.
- Localize the reorder failure in English and Japanese.

## Mechanism

The iOS pane map keeps a single optimistic `PaneMapReorderState` snapshot per request. `MobileShellComposite` sends the requested pane order through the shared workspace mutation client, then refreshes from the authoritative Mac layout. A failed or mismatched response restores the prior snapshot and exposes the localized error.

The Mac owner-routed handler validates the requested pane IDs against the workspace, reorders full pane contents through the shared terminal-controller move path, installs placeholders during the transaction, rolls back on failure, and publishes exactly one final mobile layout without stealing focus.

`PaneZoomPresentationState` owns a workspace-local native navigation path initialized to `.terminal`. Pane-map and terminal endpoints therefore share one stack and toolbar, use the same matched zoom source, remain interruptible in both directions, and restore the prior focused surface on the first presented frame. Compact workspace navigation owns tab-bar visibility while a workspace is pushed.

`PaneMapSelectionArbitration` absorbs movement before selection resolution. UIKit collection drag/drop owns relayout; a tap is accepted only after the gesture resolves without scroll or drag movement.

## Verification

- Final focused iOS simulator result: 15/15 passed (`PaneMapCollectionLayoutEngineTests`, `PaneMapReorderStateTests`, `PaneZoomPresentationStateTests`, and both pane-map XCUITest flows).
- Mobile terminal preview regression suite: 9/9 passed.
- Final focused macOS `cmux-unit` result: 25/25 passed, including authoritative publication, full-pane moves, stale-topology rejection, and failure reconciliation.
- Workspace grouping, `Package.resolved` policy, pbxproj test wiring, localization JSON parsing, and `git diff --check` passed.
- The broad iOS convention lint still reports its existing repository-wide baseline; no new finding originates in the final routing follow-up.
- Both supplied recordings were extracted and inspected frame by frame. A new real paired-simulator recording confirms workspace entry goes directly to the restored `alpha` terminal with no pane-map flash or primary tab bar.
- Physical Aziz dogfood recording was inspected across all 508 frames: map/deck correlation, swipe absorption, drag reorder, intentional tap focus, partial-edge interruption/cancel, map return, and terminal return completed without the prior exit ghost.
- Physical XCTest before the assertion-marker stabilization: 1 passed, 1 failed only because the marker center was covered after the behavior completed. The assertion was stabilized afterward and passes in the final simulator run.
- Final tagged macOS build `cmux DEV panes` succeeded from pushed HEAD and is running on LG HDR 4K with the persisted three-pane dogfood workspace.

## Device status / residual risk

Aziz became unavailable to CoreDevice before the final two physical tests and optimized reinstall could be repeated. The final signed build and physical rerun remain pending only on the phone reconnecting/unlocking; simulator and prior physical recording evidence are complete.

## Dictionary

- Authoritative reconciliation means iOS replaces its optimistic order with the Mac-confirmed topology after each mutation.
- Gesture absorption means movement claims the interaction so it cannot later become a tap.

---

> [!NOTE]
> **Medium Risk**
> Touches workspace sync, mutation reconciliation, and terminal replay side channels; race handling is explicit but layout + reorder paths are user-visible and cross Mac/iOS contract boundaries.
> 
> **Overview**
> Adds **end-to-end pane topology** on mobile: workspaces can carry a synced `layout` (splits, panes, surface tabs) through state sync, RPC list decoding, and `MobileWorkspacePreview`, with topology hashing that ignores divider ratios and titles so drags don’t spam invalidation.
> 
> **Pane map visuals** build on shared terminal render-grid code: `paneMapPreview` resolves styled cells and wide glyphs; VT replay reuses `resolvedCharacterCellWidths`. The shell fetches read-only `mobile.terminal.replay` with `grid_only` (bounded concurrency) for map thumbnails without resizing the Mac grid.
> 
> **Mutations and races**: `workspace.pane.reorder` (capability `workspace.pane_reorder.v1`) applies the RPC’s authoritative workspace list immediately; a **foreground list authority generation** drops stale in-flight `mobile.workspace.list` snapshots so optimistic UI can’t roll back a successful reorder. Opening a workspace picks the **focused pane’s** ready terminal (with sensible fallbacks).
> 
> **UI layer** in this diff adds pane-map collection layout, `PaneMapOverlay`, surface deck / toolbar reserves for pane-map controls, DEBUG panes-tabs preview + autoplay, and connection-recovery overlay gating so it doesn’t double-draw under terminal covers.
> 
> CI runs **CmuxMobileRPC** package tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51bf1abd3876601a55abbbfc98255c98a6bebc85. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


azooz2003-bit:
V4 pane UX verification update — `3217729c70366f25e568078812b02ce0947f6ff4`

The final fix mounts the Surface Deck inset on the terminal destination inside `PaneZoomNavigationStack`, so the UIKit-hosted navigation destination no longer covers the deck.

Evidence:
- Hosted exact selector: [deck/map correlation + toolbar](https://github.com/manaflow-ai/cmux/actions/runs/30692223583/job/91348952786) — 1 test, 0 failures, 38.089s.
- Hosted exact selector: [drag reorder + terminal zoom both directions](https://github.com/manaflow-ai/cmux/actions/runs/30692224894/job/91348937066) — 1 test, 0 failures, 39.121s.
- Those workflow runs are red only from unrelated package-convention/mobile-core jobs; both iPhone selector jobs above are green.
- Recorded simulator transition contacts cover enter, exit, second exit, and direct restore. The settled app frame was pixel-identical after cropping the changing status bar from 25.515s through 55.145s (FFmpeg PSNR: infinite), ruling out delayed exit ghost/reflow.
- Final tagged simulator build succeeded with embedded `CMUXGitSHA=3217729c70`, launched on isolated `cmux-dev-panes` (`E58149E1-DE6E-4EC7-997B-3665476601A0`), auto-signed-in/paired, and reached a live Mac subscription after reconnect settling.
- Tagged Mac cloud build [30693084277](https://github.com/manaflow-ai/cmux/actions/runs/30693084277) succeeded and launched; its socket lists `iOS panes dogfood`, and its window is on LG HDR 4K. Embedded `CMUXCommit=7515e623d` is the Blacksmith ephemeral one-parent child of this PR head; its only tree delta is the cloud credential denylist removing two tracked `.env.example` files.
- Signed arm64 device app succeeded with bundle `dev.cmux.ios.panes`, embedded `CMUXGitSHA=3217729c70`, Apple Development signature/team `7WLXT3NR37`. Aziz is currently unavailable, so `panes` is queued for device `4A52829D-6427-599F-A166-4058881D2DF4` at `2026-08-01T03:18:28-0700`; the install LaunchAgent is running with last exit code 0 and will auto-install + launch on reconnect.

Guards:
- `git diff --check`, Swift parse, PBX test wiring (644 files), and workspace package grouping passed.
- Final fix changes no user-facing strings; localization catalog audit found no required delta.
- Worktree is clean except preserved `azizalbahar-blog/`.
- PR is OPEN, MERGEABLE, CLEAN. It has not been merged.

azooz2003-bit:
Exact-HEAD follow-up verification for 77e66b5c7b:

- Added main-actor isolation to PaneZoomPresentationState.BackgroundSnapshot. This is behavior-neutral and removes the Swift 6 isolation warning on the pane zoom background restoration path.
- Full isolated simulator rebuild and launch passed with tag panes on cmux-dev-panes. The app connected to the paired panes Mac build and subscribed successfully.
- Focused Swift Testing coverage passed: PaneZoomNavigationBackgroundBridgeTests/restoresOwnedAncestorBackgroundsWithoutClobberingNewOwners, 1 test, 0 failures.
- Physical-device archive run panes-7047542b6b84 succeeded for generic/platform=iOS with CMUX_GIT_SHA=77e66b5c7b, tag panes, bundle dev.cmux.ios.panes, and staging API plus Iroh origins. The local debugging export succeeded.
- Aziz was unreachable, so the signed build replaced the stale panes queue entry at 2026-08-01 06:22 PDT. Independent verification shows embedded SHA 77e66b5c7b, valid designated requirement, Apple Development signature for team 7WLXT3NR37, and the queue LaunchAgent running with last exit code 0. It is configured to auto-install, auto-sign-in, pair, and launch when the phone reconnects.
- Localization audit: no user-facing string delta.

Manual retry if needed: ios/scripts/reload-cloud.sh --tag panes --device-id 4A52829D-6427-599F-A166-4058881D2DF4

azooz2003-bit:
V6 takeover round — merged current main (resolving 5 conflicts incl. restoring main's ghostty/bonsplit submodule bumps), triaged every Cubic finding against runtime, and landed fixes as 7cbecf3b45.

**Fixed (P1/P2):** stale pane-map drags now fail closed — the phone sends the layout revision it observed (`base_layout_version`) and the Mac rejects mismatches as `conflict` (regression test verified red without the fix); mobile reorder/move route to the workspace's owning window before honoring a stale `window_id`; reorder placeholders force-close so a Mac close-confirmation sheet can't leave phantom tabs; in-flight `workspace.list` fetches can no longer overwrite a mutation's authoritative layout (foreground: authority generation dropped on arrival + trailing refresh re-armed; secondary Macs: refresh pending re-armed); `PaneMapReorderState` ignores lower-revision (stale) lists; workspace entry prefers a ready terminal over an unready focused-pane selection; a pane-map refresh cancelled before start no longer wedges the toolbar spinner; browser streams keep the surface deck as their exit path; the pane zoom stack no longer overrides the shell tab-bar policy (fixes iPad split-view hiding); duplicate pane IDs no longer trap the coordinator; VoiceOver ordinals follow optimistic reorders; DEC reverse-video previews resolve the final mode setting; sign-out retires in-flight dialing transports; `grid_only` replay skips unbounded VT fallbacks for hibernated pane-map previews; dead legacy geometry-reorder API removed; fixed sleeps/unbounded waits removed from pane tests.

**Already fixed before this round (documented, no change needed):** focus-follows-content after reorder (6545e29c7e); `pendingConnectClient` teardown (superseded by main's `connectionAttemptClient` in merge e844e989e9); all flagged localization keys have en+ja entries in `ios/cmux/Resources/Localizable.xcstrings` (`L10n.string` resolves via `Bundle.main`, not the package catalog Cubic checked).

**Invalid (evidence):** rollback-failure discard is unreachable (no `shouldCreateTab` veto exists; `moveTab` gates are excluded at entry); divider-only first publish can't occur (init fires focus/select publications before any divider); older-host preview fallback can't render pane-map cards (hosts emitting `layout` always advertise `terminal.render_grid.v1`); per-workspace JSON round-trip remedy is a no-op (`MobileSyncFrameCoder` is stateless) and version-keyed caching would be unsafe (ratio changes don't bump the version).

**Known open finding (frame-evidenced, not a regression of this round):** the FIRST terminal→pane-map transition after a cold app open renders as a one-frame cut; every later transition in both directions animates as a continuous zoom (60fps frame diffs: smooth easing bells, no stalls or ghosts). Cause: the zoom's matched source tile lives in a UIKit collection cell that doesn't exist until the map has been visible once. Fix direction: pre-warm the map's initial cell layout at stack mount so the source geometry exists before the first unzoom.

Verified: CmuxMobileShell/CMUXMobileCore/CmuxWorkspaces/CmuxPanes package suites green on the touched areas (612+ tests, nonzero execution confirmed); pre-existing Iroh/replay suite flakes reproduce identically on origin/main (A/B on a pristine main worktree). Dogfood builds installed and paired on the isolated simulator and the physical iPhone from this exact HEAD.

azooz2003-bit:
Live dogfood findings (Aziz hit both on device, 2026-08-05 ~20:33; reproduced and diagnosed from the Mac side):

1. **Ghost screen when the connection dies during pane-map entry (P1 for this PR's bar).** With the paired Mac gone, opening workspace detail then the pane view left a black screen whose only rendered element was the alt-screen notice toolbar item — no content, no chrome, no recovery affordance. The connection-recovery overlay should own this state; a transition begun against a dead stream must not strand a bare toolbar item.
2. **Disconnected fallback is unrecognizable.** After backing out, navigation landed in `DisconnectedWorkspaceShellView` — a bare "Workspaces" list with no tab bar, no status, no reconnect affordance visible in the screenshot. Needs a visible disconnected banner/spinner.
3. **Mac-side attach retry storm from a restored hibernated tab.** After session restore, the not-yet-instantiated `alpha-notes` surface looped `surface.attach.defer reason=noWindow` several times per second indefinitely (398+ entries), flooding the debug log and plausibly starving the mobile host RPCs (phone attach timed out during the storm). Interacting with the surface once (hydrating it) stopped the loop. Likely interaction: pane-map preview replay requests repeatedly poking a surface that can never attach without a window.
4. **The tagged Mac app appears to exit after its web API becomes unreachable.** Three occurrences, each with `auth: auth.phase=access_token timed out after 60.0 seconds` in the final log window and no crash report. A dev build should degrade, not exit, when token refresh fails. Needs a targeted repro before fixing.

Environment is repaired: fresh Mac instance (no churn), web API up, phone re-paired with a readiness receipt at HEAD 7cbecf3b45.

azooz2003-bit:
Diagnosis of the on-device breakage Aziz reported (reproduced twice on an isolated simulator, crash reports in hand):

**1. P0 crash: opening a workspace fatals in SwiftUI navigation.** `SwiftUI/NavigationColumnState.swift:684: Fatal error: 'try!' ... AnyNavigationPath.Error.comparisonTypeMismatch`, faulting in `NavigationColumnState.boundPathChange(to:)` via `NavigationAuthority.flushRequestQueue()`. Two .ips reports (21:31, 21:58). The second fired seconds after a plain workspace open on a healthy connection, so this is not connection-dependent. Root-cause hypothesis: `WorkspaceDetailView` mounts the nested `PaneZoomNavigationStack` (typed path `[PaneZoomPresentationState.Endpoint]`, initial `[.terminal]`) inside a `navigationDestination` of the compact shell stack (typed path `[MobileWorkspacePreview.ID]`); the window's navigation authority routes the inner stack's bound-path change to the outer column and the element-type comparison traps. Fix direction: stop nesting two typed stacks — either fold the pane/terminal endpoints into the outer stack using a heterogeneous `NavigationPath`, or drive the zoom presentation without a second `NavigationStack`. Every crash also resets onboarding + Stack auth state, which produced the "suddenly at sign-in/onboarding/bare Workspaces" states.

**2. P1 stale iroh route after Mac app restarts.** Each tagged-Mac relaunch mints a new iroh node; when the relaunch happens while the dev web API (device registry) is down, the registry keeps advertising the dead endpoint and the phone redials it forever (`pairing route failed kind=iroh ...: requestTimedOut` every ~10s). The terminal endpoint then decays to a bare centered `exclamationmark.triangle.fill` with no recovery affordance (same placeholder Aziz screenshotted). The phone auto-recovered the moment a fresh Mac instance registered against a live registry. Fix directions: re-register on registry reachability, make the phone fall back to re-fetching the registry when all stored routes time out repeatedly, and give the decayed endpoint a visible reconnect state.

Note on evidence gap: the hosted pane-map UI tests run against the `CMUX_UITEST_PANES_PREVIEW` fixture host, which mounts `PaneZoomNavigationStack` standalone — the real nested arrangement was never exercised end-to-end, which is how the P0 crash shipped. A regression test needs to drive the real shell (mock host, no fixture bypass).

Dogfood is blocked on the P0 fix.

azooz2003-bit:
Fixed the P0 crash that killed the app seconds after opening a pane-layout workspace in the real shell (and wiped onboarding plus Stack auth on every hit).

Root cause: `PaneZoomNavigationStack` nested a second typed `NavigationStack` (`[PaneZoomPresentationState.Endpoint]`) inside the compact shell stack's pushed workspace destination (`[MobileWorkspacePreview.ID]`). Both stacks register with the window's shared `NavigationAuthority`; its flush queue reconciles one stack's bound path against the other stack's column state and force-`try!`s `SwiftUI.AnyNavigationPath.Error.comparisonTypeMismatch` in `NavigationColumnState.boundPathChange`. I verified on the sim repro that type-erasing the nested path to `NavigationPath` does not fix it: the misrouted comparison still traps on the next push, even of a layout-free workspace, against the popped pane stack's stale column. The panes fixture never crashed because it hosts the pair without an ancestor stack, which is why the hosted UI tests missed this.

Fix (afb4b6c5d1): `PaneZoomHost` now picks the navigation mechanism from an explicit hosting mode passed by each shell call site. Under pushed shells (workspaces tab, notifications, notification search) the pane map renders directly in the ancestor stack's destination and the terminal is a zoom `fullScreenCover` with its own root-only stack, so nothing joins the ancestor's authority; presentations own a separate navigation context by design. The split detail keeps the documented stack-in-a-column pattern unchanged. The zoom transition moved into the component so both hostings anchor to the same `matchedTransitionSource` tiles, and the workspace detail's sheets, dialogs, and recovery banner attach to both layers behind opposite binding gates because a covered base cannot present. `PaneZoomPresentationState`'s routing API is unchanged.

Regression coverage (b38947c07c, committed red-first): the mock sync host can now serve a two-pane layout for `workspace-main`, and `cmuxUITests/testWorkspacePaneLayoutNavigationSurvivesInRealShell` drives the production shell through every on-device trigger: initial push, pane-map round-trips, pop to list, then pushing a layout-free workspace. CI proof: [red run on the test-only commit](https://github.com/manaflow-ai/cmux/actions/runs/31147341501), [green run on head](https://github.com/manaflow-ai/cmux/actions/runs/31147351183).

Verified on the live-Mac sim repro: the unpatched build crashed within 15 s of opening "iOS panes dogfood"; this build survived the open, three zoom round-trips, and workspace switches with zero new crash reports. `CMUXMobileCore` 368/368 green; `CmuxMobileShell` green except `MobileMacConnectionPoolTests`, which fails identically on pristine `origin/main` (pre-existing flake, package untouched here).

Dictionary: NavigationAuthority is SwiftUI's private per-window bookkeeper that applies every NavigationStack path change; bound path is the array a NavigationStack mirrors through its `path:` binding; hosting mode is the new explicit switch between presentation-based and column-based pane zoom.

azooz2003-bit:
Follow-ups on the crash fix, all pushed:

- CI red/green proof, corrected run links (the links in my previous comment pointed at dispatches that failed checkout on short SHAs): the test-only commit crashed the app under the new regression test in [run 31149003065](https://github.com/manaflow-ai/cmux/actions/runs/31149003065) (`Test Case testWorkspacePaneLayoutNavigationSurvivesInRealShell failed` followed by `Restarting after unexpected exit, crash, or test timeout`). Every post-fix run shows the app surviving the full drive: open, repeated zoom round-trips, pop to list, re-push. No post-fix run contains a crash.
- 00bf15260f: with the terminal hosted as a cover, the covered pane-map bar's items stayed in the accessibility tree on the iOS 26.4 runtime, so the shared back button and title menu existed twice (VoiceOver could reach covered controls; automation failed with "multiple matching elements"). The covered bar is now hidden outright while the cover is up, centralized in PaneZoomHost; per-item accessibilityHidden was verified insufficient on that runtime and reverted.
- 61c2901725, 2d0fc11090, 7386aa181a, 51bf1abd38: test-side hardening for the CI simulator VM, where the cover re-presentation rebuilds the terminal subtree in 15-18 s (the old stack pop restored a retained view instantly). waitForPaneMap now defaults to 20 s and only requires hit-testability on the pane-map side, because toolbar items inside a presented cover never report hittable to XCUI on that runtime even though real taps land.

Open item: `testPanesPreviewCorrelatesDeckGroupsWithMapRegionsAndSwitchesTabs` (fixture host) still fails on the CI runner at the pane-map Done-button return ([run 31171860885](https://github.com/manaflow-ai/cmux/actions/runs/31171860885)); the map stays up after the Done tap there. The equivalent return path via tile taps passes on CI and both paths work on a local simulator, so this is either Done-button tap delivery on the saturated runner or a cover re-presentation edge specific to that flow; it needs one focused follow-up before merge. Dogfood note: returning from the pane map now re-presents and rebuilds the terminal instead of restoring a retained pushed view, so watch the map-to-terminal return latency on device.

Dictionary: cover means the fullScreenCover presentation hosting the zoomed terminal; re-presentation is SwiftUI presenting that cover again after it was dismissed to show the pane map; hit-testability (isHittable) is XCUI's check that an element can receive a synthesized tap at its hit point.

azooz2003-bit:
Red/green proof is complete: the regression test passes on head in [run 31176397846](https://github.com/manaflow-ai/cmux/actions/runs/31176397846) (`ios-simulator (iphone)` green, `cmuxUITests/cmuxUITests/testWorkspacePaneLayoutNavigationSurvivesInRealShell`), against the crash on the test-only commit in [run 31149003065](https://github.com/manaflow-ai/cmux/actions/runs/31149003065). The other failing jobs in these dispatch runs (package-conventions-lint, MobilePairingFailureTests in mobile-core-package) fail identically on the test-only commit and on pristine main, so they are pre-existing branch and runner state, not this change. The fixture Done-button return remains the one open CI item from my previous comment.

azooz2003-bit:
Takeover round (Claude, sole writer this session): the pane XCUITests were red on every runner because the covered pane map never left the accessibility tree. With the terminal hosted as a full-screen cover, SwiftUI's subtree `accessibilityHidden` is not honored for the mounted map underneath: the AX dump captured at the failing wait showed the overlay marker, collection, tiles, and tab strips all still queryable while the cover was up, while the bar-level `.toolbar(.hidden)` gate held. That also means assistive tech saw the covered map in production.

Fix in 38a06df046: the collection container now sets `accessibilityElementsHidden` from the map's visibility (the UIKit flag is honored across the cover boundary), and the overlay marker swaps to `MobilePaneMapOverlayCovered` while covered so presentation queries cannot match it. `waitForPaneMap` and the post-pane docs push now print the AX tree on timeout (b27745f6cb, 059ed21d08), so a future red run names the blocking element instead of only timing out.

Evidence at head 059ed21d08, iphone family, selected-test results inside runs whose overall conclusion is polluted by the pre-existing lint baseline and `MobilePairingFailureTests` flake:
- deck/map correlation passed: https://github.com/manaflow-ai/cmux/actions/runs/31698947436
- zoom both directions passed: https://github.com/manaflow-ai/cmux/actions/runs/31698950512
- real-shell pane navigation passed: https://github.com/manaflow-ai/cmux/actions/runs/31701319735
- red proof of the AX leak (pre-fix, same wait, two fleets): https://github.com/manaflow-ai/cmux/actions/runs/31690318401 and https://github.com/manaflow-ai/cmux/actions/runs/31694173117

Watch item: at 38a06df046 one real-shell run failed the trailing docs-workspace push with both row taps landing and the app alive (https://github.com/manaflow-ai/cmux/actions/runs/31698953494); it did not reproduce at 059ed21d08. The new dump will attribute the next occurrence.

Environment status: no fresh device or simulator build of this head exists yet. The cloud fleet had no free slot for over 20 minutes and the local Mac was saturated (load ~65) by an unrelated media pipeline, so the visual re-verification of the cold-open transition under the cover architecture is pending; structurally the map is now the mounted base in push hosting, so the original unloaded-root cut cannot occur there. Aziz is unreachable and the install queue is armed. The branch is currently CONFLICTING with main and needs a merge before the next dogfood build.

azooz2003-bit:
CI status on head 059ed21d08 (dispatch https://github.com/manaflow-ai/cmux/actions/runs/32057272028, filter `cmuxUITests/cmuxUITests/testPanesPreviewReordersCardsAndZoomsTerminalInBothDirections`): both ios-simulator legs (iphone, ipad) passed. The `ios-tests` gate is red only because `package-conventions-lint` fails, and that lint fails identically on current `main` (same free-function/namespace-enum offenders in files this branch does not touch; the one flagged file the branch edits, `WorkspaceShellView.swift`, carries the same flagged enum on `main`). The earlier `mobile-core-package` failures (TerminalViewportResyncTests, MobileShellForegroundResumeTests, MobileShellRenderGridInputCatchUpTests, TerminalOutputDeliveryQueueTests) reproduce locally at the merge-base in unchanged test files, so they are pre-existing too. The `CmuxMobileRPC` suite this branch adds to CI passes 3/3 full local runs.

PR #8205, opened 2026-07-16
Keep workspace activity visible across agent races

lawrencecchen:
Three independent failures could hide or delay a running workspace:

1. Delayed teardown for a replaced session cleared the panel-scoped lifecycle shared with its live same-agent replacement. PID cleanup now clears lifecycle only when no same-agent runtime remains on that panel.
2. Session-start and prompt-submit hooks captured a turn-diff baseline before publishing PID, lifecycle, and status. Hooks now publish visible runtime state and start monitoring before auxiliary diff capture.
3. Turn-diff baselines lived in one large JSON document. Every hook locked, read, decoded, re-encoded, and atomically rewrote all retained sessions. The store is now SQLite with WAL, indexed latest-baseline lookups, one-row updates, bounded retention, and automatic legacy JSON migration.

These are ownership, priority-ordering, and storage-model fixes. They do not increase timeouts or add timing sleeps.

Regression history:
- `9499f0cc37` adds the stale-cleanup regression; `8deecfc67d` fixes it.
- `f95a42d465` adds slow-baseline regressions for session start and prompt submit; `803b6833a2` fixes them.
- `917f2c8371` changes baseline regressions to require indexed storage and migration; `c42671d079` implements SQLite.
- `84c299e84c` hardens concurrent snapshot publication, downgrade migration, retention ordering, corrupt-row errors, localization, and the timing-free hook test gate.

Validation:
- Tagged cloud build: https://github.com/manaflow-ai/cmux/actions/runs/29470222808
- The real legacy store was 60 MB with 200 records. Its full optimized JSON update averaged 577.79 ms. The indexed SQLite update averaged 32.29 ms, about 18 times faster.
- Git was not the dominant measured cost: `git stash create` took 0.10 s and untracked enumeration took 0.27 s in the large worktree.
- The one-time tagged Debug migration took 4.98 s. Visible state is published first, so that migration no longer delays the sidebar indicator.
- A subsequent tagged session-start took 0.74 s including Git, socket work, and the database update.
- Two concurrent tagged session-starts persisted both rows in 1.34 s total.
- After initial migration, a newly reappearing legacy JSON record was merged into SQLite, removed, and selected successfully by `cmux diff --last-turn`.
- Tagged runtime state reported `codex=Running`, and a native window screenshot confirmed the rendered spinner and Running label.
- Swift parse, localization JSON validation, project-file checks, codesign validation, and the tagged cloud compile passed.

### DIFF
--- a/CLI/CMUXAgentTurnDiffBaselineDatabase.swift
+++ b/CLI/CMUXAgentTurnDiffBaselineDatabase.swift
@@ -4,7 +4,6 @@ import SQLite3
 
 struct CMUXAgentTurnDiffBaselineDatabaseUpdate {
     var removedRecords: [CMUXAgentTurnDiffBaselineRecord]
-    var storedRecord: Bool
 }
 
 final class CMUXAgentTurnDiffBaselineDatabase {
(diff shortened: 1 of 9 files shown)


lawrencecchen:
Superseded by the trajectory-based migration in https://github.com/manaflow-ai/cmux/pull/8255, which removes the baseline store and prompt-time Git snapshots.

PR #8254, opened 2026-07-16
Let the sizing fuzz fail on a mirror it cannot see

ejc3:
The sizing fuzz could not fail on a mirror it wasn't showing, so the operation that
reveals one was never exercised, and when it did fail it could not say why. This is
harness-only — no product file changes.

## What it could not see

The coverage census read `list-panes`, which is blind to zoom: it reports the whole
base tree, so the census compared the app against panes tmux wasn't drawing. It now
gates on `window_zoomed_flag` and, for a zoomed window, expects the pane tmux
actually draws (`select-pane` unzooms, so `zoom=1 && pane_active=1` names it
uniquely). It also reports both sides and the flag it compared, because whether a
pane is expected on screen turns entirely on zoom — a sibling of a zoomed pane is
absent by design, and naming the missing pane alone reads as an accusation either
way.

Nothing ever switched between two mirrored tabs, which is the whole reveal path.
Op 10 does, drawing targets only from mirrored windows, and it has to prove the
window it asked for is the one now shown: keeping only the window it started from
and accepting any *other* mirrored window let a third window or a concurrent
selection satisfy it without the target appearing at all.

`TMUX_TMPDIR` pointing at a directory that does not exist makes tmux fall back to
the default socket, so a run could quietly drive the wrong server. It is checked
before anything mutates.

## What a failure now says

It used to print an iteration number and 300 characters of a pretty-printed
payload — mostly indentation, cut off inside the first window, before the terms
that say which one is false. Now each iteration records the ops behind it
(`iter=10/25 ops=3 [op4,op1,op8]`), the digest keeps the unsettled windows whole,
and `why` carries the terms `settled` is made of, including the banked render frame
beside the live container and whether bonsplit's own tree agrees with the tmux
layout.

That reporting earned itself immediately. It located a real defect that two
marathons had only pointed at: the panes are portal-hosted at the window level, so
a pane's frame tracks an anchor inside the split tree; `tree_matches_layout` reports
true while a pane sits at half its planned width, which rules out every stale-tree
explanation and puts the divergence downstream of the tree.

## The judge asserts rather than filters

It opened by skipping any mirror that was not visible-for-sizing — an AND that
could only shrink the judged set, so the defect blinded the judge instead of
reddening it. It now reads view state directly rather than through the mirror's own
notion of visibility, which is the thing under test. Same fault, twice more:
derivation parity keyed off that flag and defaulted to *settled* when no grid could
be derived, so the one window that could not re-derive its claim passed while every
other term sat true on a cached plan.

The assertions are directional on purpose: a mirror can legitimately hold the flag
while off screen (a selected tab in a background window), so only the reverse is a
defect. A known gap is stated rather than implied — a *sole* stale owner is not
caught; two owners, and a shown mirror that owns nothing, are.

Zero mirror-tab switches is reported per seed but asserted across the marathon.
Whether a seed draws op 10 is the RNG's business — about one seed in thirty never
draws it — and filing that beside real defects is how a gate gets ignored.

## Status

An adversarial pass over this diff found five more ways it could report more than it
delivered, four of them the same fault above; all are fixed in the last commit and
verified by running them. Among them a real one worth naming: the op-10 target
filter used jq's `inside`, which compares strings by containment, so
`["@3"] | inside(["@30","@1"])` is true and `@1` passed as mirrored on the strength
of `@10` once tmux hands out double digits.

The marathon this improves is **not green**, and the reason is a confirmed defect
this PR neither introduces nor can fix. After a pane closes, bonsplit's logical tree
collapses correctly — `bonsplitTreeMatches` reports true — while its AppKit
container hierarchy keeps the old arrangement, so the portal syncs the surviving
pane to a stale anchor at its former width. cmux cannot detect it: every check it
owns asks the tree whether the views are right, so it cannot fail when the views
diverge from the tree. The fix belongs in bonsplit's `closePane` teardown and needs
its own change there.

Measured on this branch, 125 iterations per arm: the mirror-identity fix removes a
nine-failure wedge on one seed outright and durably, and two seeds fail on the
bonsplit divergence in every arrangement tried. The fuzz is not run in CI, so
nothing here gates the build; `RemoteTmuxSizingUITests` is 12/14, which is what
pristine main scores on the same machine (its two failures are a `lab server never
started` setup flake whose failing pair shuffles between runs).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved debug fuzz diagnostics for terminal sizing, visibility, and settlement verdicts, including stronger mismatch detection and clearer “why” reporting.
* **Testing**
  * Extended fuzzing to validate zoom-aware pane visibility per window and to exercise mirror-to-mirror tab switching.
  * Added richer per-run and marathon-wide coverage summaries, including explicit reporting and failure when required tab-switch coverage is not reached.
* **Chores**
  * Added an early check that the remote tmux runtime directory exists to prevent silent fallback behavior.

### DIFF
--- a/Sources/Debug/RemoteTmux/TerminalController+RemoteTmuxTestSupport.swift
+++ b/Sources/Debug/RemoteTmux/TerminalController+RemoteTmuxTestSupport.swift
@@ -567,6 +567,7 @@ extension TerminalController {
                                 "%\(leaf) native-geometry"
                                     + " plan=\(Int(plannedContent.width))x\(Int(plannedContent.height))"
                                     + " view=\(Int(actual.width))x\(Int(actual.height))"
+                                    + " in \(Self.ancestryDescription(of: hostedView))"
                             )
                         }
                     } else {
@@ -762,6 +763,32 @@ extension TerminalController {
                 "full_hierarchy_sync": RemoteTmuxSizingDiagnostics.fullHierarchySyncCount,
             ],
         ]
+    }
+
+    /// The hosted view's enclosing containers, innermost first, each with its own
+    /// size and — for a split — how many panes it still arranges.
+    ///
+    /// A pane that misses its planned size says nothing about WHERE the size was
+    /// lost. If the enclosing split already holds the planned width, the pane
+    /// alone failed to fill it; if the split itself is short, the miss came from
+    /// above; and an arranged count that disagrees with the tree's child count
+    /// means the split never collapsed after a sibling closed, which no divider
+    /// imposition can correct because there is no longer a divider to move.
+    nonisolated static func ancestryDescription(of view: NSView, levels: Int = 3) -> String {
+        var parts: [String] = []
+        var next = view.superview
+        while let current = next, parts.count < levels {
+            let name = String(describing: type(of: current))
+                .replacingOccurrences(of: "Bonsplit.", with: "")
+            var term = "\(name)(\(Int(current.frame.width))x\(Int(current.frame.height))"
+            if let split = current as? NSSplitView {
+                term += ",arranged=\(split.arrangedSubviews.count)"
+                term += ",vertical=\(split.isVertical ? 1 : 0)"
+            }
+            parts.append(term + ")")
+            next = current.superview
+        }
+        return parts.isEmpty ? "no-superview" : parts.joined(separator: " < ")
     }
 }
 #endif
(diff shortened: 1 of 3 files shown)


ejc3:
Superseded by #8325, which carries all six of this branch's commits (rebased onto the #8283 fixes, census race guards folded in) plus the overlay oracle and coverage reporting. Verified there with a fully green 5x25 strict marathon.

PR #8309, opened 2026-07-17
Detect coding-agent state from live terminal screens

lawrencecchen:
Adds terminal-state detection for 21 coding-agent families plus existing cmux families.

The PTY callback only advances a coalesced dirty revision. A per-surface actor waits for a quiet window, reads plain rendered text from the final 48 active-screen rows, verifies process start time and terminal generation before and after capture, and publishes effective changes only. Blocked and history evidence use the full capture; transient working evidence uses the bottom 12 physical rows so stale spinners cannot override the current composer. Complete lifecycle integrations remain authoritative.

Per-surface setup and teardown are serialized. Reinstall waits for prior cleanup, and repeated drops are idempotent. Strict blocked rules require either multiple corroborating fragments or an exact rendered-line prompt.

Verification:
- 241 `CmuxTerminalCore` tests and 90 `CmuxTerminal` tests pass
- lifecycle race, Gemini/Kimi prompts, stale Pi spinner, multiline approvals, and CRLF rendering have test-first regression commits
- focused app-host lifecycle test passes: https://github.com/manaflow-ai/cmux/actions/runs/29561150240
- final tagged macOS build `astate` from `bb89e237cb5fd5b321d74482413e6a4464a1c70e`: https://github.com/manaflow-ai/cmux/actions/runs/29561148873
- current-head Codex 0.144.3 transitioned idle → working → blocked on an approval prompt, then working → idle through `/tmp/cmux-debug-astate.sock`
- nine installed agent families were exercised in isolated tmux panes, with every captured state audited against the final classifier
- paired 15-second `sample` captures found no detector stacks; Instruments trace finalization hung, so no Instruments pass is claimed
- project consistency and `git diff --check` pass
- no user-facing strings were added


lawrencecchen:
Folded into https://github.com/manaflow-ai/cmux/pull/7970 so the live terminal observations and session lineage ship through one `cmux agents` interface.

Issue #6209, opened 2026-06-16
Session restore intermittently fails: `codex resume <id>` reports "No saved session found" and drops to bare shell


austinywang:
### Bug description

Session restore intermittently fails to bring back a **Codex** pane. The restored pane runs its `codex resume <id>` binding, but Codex reports **`No saved session found with ID <id>`** and exits, leaving the user at a bare shell prompt instead of their restored session.

Observed in a restored pane (cwd = the `cmux37` clone):

```
Last login: Mon Jun 15 10:43:37 on ttys003
ERROR: No saved session found with ID 019eca90-bc99-7293-aa1e-a604a6c6d068. Run `codex resume` without an ID to choose from existing sessions.
(base) austinwang@MacBook-Pro-189 cmux37 %
```

It is **intermittent** — most restores succeed; this one (and others like it) drop back to the shell. We really need this fixed.

### This is the Codex twin of #6194

#6194 is the identical symptom for Claude (`claude --resume <id>` → "No conversation found with session ID", drops to bare shell, intermittent). This issue is the **Codex** counterpart (`codex resume <id>` → "No saved session found with ID"). They almost certainly share a root cause and should be fixed together.

### Root cause is already diagnosed in #5393

Per #5393, the cmux agent hook **publishes a default `codex resume <sessionId>` binding even for non-restorable / reaped sessions**:

> `agentSurfaceResumeCommand(kind:, launchCommand: nil, …)` → `AgentResumeArgv().builtInKind("codex", arguments: [])` → `["codex", "resume", <sessionId>]` → a `surface.resume.set` binding is published — **regardless of whether the original invocation was restorable**.

So at restore time the pane runs `codex resume <id>` for a session Codex never actually saved (one-shot `codex exec`/`review`, a reaped PID with no captured argv, or a session reaped by Codex's own retention before restore) → `No saved session found`.

#5393's recommended **fix direction** (applies to all agents, not just codex):

> Gate at the publish / `builtInKind` layer instead of the capture layer: do not synthesize a default `<agent> resume <id>` command from a `nil` launchCommand unless there is a positive restorability signal for the session (e.g. a recorded restorable launchCommand, or `isRestorable`/transcript provenance).

### Expected behavior

1. cmux should **not** publish a `codex resume <id>` binding unless the session is actually restorable (positive restorability signal — recorded restorable launchCommand / `isRestorable` / transcript provenance), per #5393.
[...]

PR #8321, opened 2026-07-17
Restore Codex sessions through user-owned roots

### LINKED ISSUE #8890: Codex restore can resume an indexed subagent instead of its interactive parent
## Reproduction

1. Run Codex in a cmux terminal and let it spawn a subagent.
2. Let the subagent hook arrive after the interactive root session hook.
3. Relaunch cmux and restore the terminal.
4. cmux resumes the indexed subagent ID.
5. Codex opens a read-only session that says it is controlled by its parent, so the restored terminal cannot accept work.

Observed evidence:

- The saved ID existed in Codex `state_5.sqlite`, so provider-evidence validation accepted it.
- Its thread row had `thread_source = subagent` and an `agent_path`.
- Its rollout `session_meta` contained the interactive root in `parent_thread_id`.
- Resuming that parent restored the writable session and continued the original task.

## Cause

An indexed Codex thread is not necessarily a user-owned restore root. cmux currently treats provider indexing as sufficient restore authority and does not canonicalize a subagent checkpoint through Codex parent metadata.

This differs from https://github.com/manaflow-ai/cmux/issues/6209, where an internal review UUID was not indexed at all.

## Expected

- Codex restore evidence distinguishes user-owned roots from subagents and automation threads.
- A subagent checkpoint resolves through explicit `parent_thread_id`, `forked_from_id`, or nested source metadata to the user-owned root.
- Hook arrival from a child cannot replace an existing root binding with the child ID.
- Relaunch resumes the canonical root and accepts input.
- Regression coverage includes an indexed subagent whose parent is indexed and resumable.

lawrencecchen:
Fixes https://github.com/manaflow-ai/cmux/issues/8890.

Also addresses the stale restore behavior reported in https://github.com/manaflow-ai/cmux/issues/6209.

Supersedes https://github.com/manaflow-ai/cmux/pull/4543.

## Root cause

Codex hooks can report indexed and unindexed internal workers in addition to the user-owned root. An indexed child could replace the durable root identity, while truncated or nested metadata could hide an `exec` worker. Legacy kindless bindings could also skip verification. Snapshot reconciliation then erased the binding when a process scan temporarily found no agent. After resume, concurrent hook events could prefer the stored pre-resume PID over the wrapper's current PID, so a later snapshot saved the dead process generation and the second restart stopped.

## Fix

- Verify Codex candidates through provider metadata and follow parent links to the user-owned root.
- Read complete metadata lines within a 4 MiB bound and reject nested `codex exec`, automation, and subagent identities as restore roots.
- Infer and verify legacy kindless Codex bindings, then rewrite saved child checkpoints and commands to the canonical root.
- Preserve durable agent-hook bindings across empty process scans while keeping `wasAgentRunning` as the automatic-launch gate.
- Prefer the wrapper's current agent PID when hook events rebind a resumed session, retaining generation identity for stale-PID protection.

## Verification

- Added regression-test commits before each corresponding fix.
- `swift test --package-path Packages/macOS/CMUXAgentLaunch`: 249 tests in 36 suites passed.
- `./scripts/lint-pbxproj-test-wiring.sh`: all 585 test files are wired.
- Final cloud build passed: https://github.com/manaflow-ai/cmux/actions/runs/30184945977.
- Live tagged test used a root Codex session plus a nested `codex exec`, then completed two Command-Q relaunch cycles on the final head. Both relaunches restored root `019f9c3f-330e-72e2-9afd-2d45d45e5985`, preserved its transcript, accepted fresh prompts, and left `wasAgentRunning: true`.
- The tag-bound shared quit shortcut and final panel snapshot were verified. Mouse-driven Computer Use remains unverified because macOS window capture returned `cgWindowNotFound`.

Dictionary: A binding associates a cmux surface with a durable agent session. A process generation distinguishes a live process from a later process that reused its PID.


lawrencecchen:
Dogfood passed on tagged build `cdrst`.

- Started root `codex --yolo`, then had it run a nested `codex exec --yolo` that completed with `NESTED_OK`.
- Root session: `019f6f52-44c8-7f63-8d1b-ef404dab0dd0`. Nested session: `019f6f52-7a92-75e2-991e-3681267a0da5`.
- The live surface resume checkpoint remained the root session after the nested session completed.
- Killed and relaunched the tagged app twice. Both restores reopened the root conversation with `ROOT_DONE`; neither showed `No saved session found`.
- Captured and inspected debug window screenshot `2026-07-18T01-53-10Z_BFCC8E7D` after restore.

PR #8371, opened 2026-07-17
Migrate the macOS app off ObservableObject to @Observable

azooz2003-bit:
Replaces every remaining `ObservableObject` conformance in cmux-owned code with the `@Observable` macro: 42 types across the app target and the CmuxAppKitSupportUI/CmuxTerminal packages, including the `Panel` protocol (now requires `Observation.Observable`) and the `Workspace`/`TabManager` god objects. SwiftUI now invalidates views per accessed property instead of on every `objectWillChange`, which removes the whole-store redraw traffic that `@Published` fan-out caused.

Mechanics, wave by wave (one commit each):
1. 20 leaf stores (sidebar, file explorer state, shortcuts observer, todo state, hover/visibility singletons).
2. `TerminalSurface` + `SearchState` (keeps its non-MainActor isolation and nonisolated deinit).
3. `Panel` protocol + all 14 conformers and the browser stores.
4. `TerminalNotificationStore`/`SidebarUnreadModel`, `CmuxConfigStore`, `FileExplorerStore`, `SessionIndexStore`.
5. `Workspace` (39 tracked properties, 22 bridge publishers, 146 `@ObservationIgnored` internals).
6. `TabManager` + app environment plumbing, final sweep.

Behavioral-parity rules used throughout: only formerly-`@Published` properties are tracked and every other stored member is `@ObservationIgnored`, so redraw semantics and hot-path costs match the old world; remaining Combine subscribers consume explicit `CurrentValueSubject` bridges fed from `willSet` (the established `tabsPublisher` seam pattern), so emission timing and replay-on-subscribe are identical; manual `objectWillChange.send()` invalidation was replaced by tracking the announced state, and `FileExplorerStore`'s AppKit outline coordinator keeps its 50ms-debounced reload via an explicit `changePublisher` that fires on every former invalidation path. View wrappers moved `@StateObject`→`@State`, `@ObservedObject`→`let`/`@Bindable`, `@EnvironmentObject`→`@Environment(T.self)`; `.onReceive($prop)` sites became `.onChange(initial: true)` to preserve the projected publisher's replay delivery. The `TabItemView` Equatable typing-latency gate and the sidebar snapshot boundaries are untouched.

Zero `ObservableObject`/`@Published`/`@StateObject`/`@ObservedObject`/`@EnvironmentObject`/`objectWillChange` uses remain in cmux-owned macOS and iOS sources (the iOS app was already fully on `@Observable`).

[...]

PR #8390, opened 2026-07-18
Lawrence Sidebar round 2: reactivity, resize settle, selection coalescing, behavior tests

azooz2003-bit:
Round 2 of the Lawrence Sidebar (flag \`sidebar-appkit-list-experiment\`, default OFF). One PR for the whole round per owner directive.

Dogfood-reported fixes, root-caused with debug probes:
- **Rename/color reactivity (7.2s → 28ms).** Every workspace publisher observation plus the initial snapshot refresh hung off the legacy scroll area's view chain; the AppKit flag unmounts that subtree, so no workspace publisher was observed at all — renames, colors, pins, descriptions never invalidated the sidebar and painted only when an unrelated change rebuilt the rows. The observation block now lives on the shared parent, one refresh path for both implementations. Probe evidence in the commit message of 31f18efe49.
- **Continuous reflow while resizing.** Mid-drag the table kept last-width row heights until mouse-up. Visible pure-AppKit rows (plus a small buffer) now re-measure at the live width per width tick with manual frame math; off-screen and hosted rows settle in the full pass at drag end, which is forced after any live partial pass.
- **Double-click rename.** The single-click action fires for both clicks of a double-click, so click 2's queued coalesced selection landed after the rename field took the field editor, re-activated the workspace, and end-editing committed the untouched title. The double-click handler now drops the queued selection before starting the edit.
- **Fast-drag press highlight.** A drag consumes the press without any selection commit, so the optimistic press highlight lingered on the grabbed row and every other visible row stayed peeled. Drag-session begin now cancels the queued selection and restores visible cells from stored models.
- Sidebar observation hops on \`DispatchQueue.main\` (the RunLoop scheduler delivers only in the default runloop mode, stalling updates during modals/menus/drags).

Also in this round:
- \`SidebarLayoutModel\` width split (68ms → 11.3ms per divider commit), explicit resize-completion signal, selection coalescing with optimistic repaint, no-implicit-animation parity, metadata/markdown toggles, checklist popover refresh, targeted single-row updates, \`.plain\` highlight-width parity, group badge centering.
- Behavior tests: \`SidebarSelectionCoalescerTests\`, \`SidebarAppKitRowCellTests\`.
- Heals for latent main breaks merged during the CI-advisory week: 4 cmuxTests compile fixes, warning-budget violations, \`indexnow.yml\` bare runner, and \`DockPortalReconcileTests\` missing the \`instanceSerial\` argument https://github.com/manaflow-ai/cmux/pull/8310 added (all four unit-test shards failed to compile on main).

Flag OFF remains the legacy SwiftUI sidebar.


azooz2003-bit:
Review findings triage (fixes pushed in 0a97614aee):

**Fixed**
- CodeRabbit, stale group-header optimistic highlight: real bug. `previewSelection`'s peel loop only reset workspace cells; headers now clear via `clearOptimisticAnchorActive()` (re-applies the stored model), and the loop peels both cell types.
- Greptile, `RunLoop.main` on the width-settle `onReceive`: switched to `DispatchQueue.main`, same default-mode-stall rule as the rest of this PR.

**Not changing, with reasons**
- Greptile P1, `WindowTerminalPortal.usesCoalescedAnchorFailsafe` static: the feature-flag lint's one-evaluation-file rule forbids the portal reading the flag directly, so the value is pushed from the flag's single evaluation site in ContentView's dispatcher. It's written in one place, read in one place, and documented at both. A constructor-injected value would need the flag to thread through several layers that don't otherwise know about it.
- Greptile P1, `applyModelProbeForTesting`: acknowledged as a rule violation; it is `#if DEBUG`-gated and carries three behavior tests that assert apply-skip semantics (hover short-circuit, optimistic flipped-copy paint). Removing it means rewriting those tests against observable cell state; queued as a follow-up rather than rushed under the merge directive.
- Greptile P2, `Task.sleep` in `scheduleWidthRemeasure`: the trailing task is a fallback for window live-resize only; divider drags settle through the explicit `.cmuxInteractiveGeometryResizeDidEnd` signal. Clock injection queued with the seam follow-up.
- Greptile P2, `ObservableObject` on `SidebarLayoutModel`: the repo-wide `@Observable` migration is in flight on https://github.com/manaflow-ai/cmux/pull/8371; this model rides that migration rather than diverging here.
- CodeRabbit, determinism allowlist: regenerated through the repo's own `--write-allowlist` flow; the entries pre-exist on main and this PR adds no new violations.
- CodeRabbit, raw title in DEBUG log: the debug sink redacts values — the actual log emits `title=<redacted:19b>`, no raw title reaches disk.

PR #8422, opened 2026-07-18
Make terminal placement canonical and host PTYs in resilient per-terminal processes

lawrencecchen:
Depends on #8340.

## Summary

- Make the SQLite workspace registry authoritative for terminal identity, placement, and lifecycle.
- Run every terminal PTY in an independent `setsid` host process and expose authenticated, scoped frontend attachments.
- Recover crash-consistently across daemon/frontend restarts, including host adoption, stale-record handling, final-output draining, and descendant cleanup.
- Arbitrate PTY size from attached viewers while preserving canonical color, cursor, mouse, close, and topology semantics.
- Preserve the refreshed base's final-destination drag/move semantics and two-phase empty-workspace/terminal event ordering.

## Validation

- `cmux-tui`: 258 unit, 6 CLI, 3 hidden-host-mode, and 14 recovery tests passed (281 total).
- `cmux-tui-core`: 219 unit, 7 browser-runtime, 28 PTY, and 9 WebSocket tests passed (263 total).
- Full default-parallel suites passed after rebasing onto the latest #8340 head.
- All-target Clippy with `-D warnings` and formatting checks passed.
- Go, Python, and TypeScript SDK tests passed; Java was unavailable on this machine.
- Post-suite audit found zero leaked terminal hosts, daemons, PTYs/process groups, registry records, nonces, or PID files.

## Risks

- Unix-only host/runtime implementation.
- One OS host process per terminal increases process count intentionally.
- Legacy or unverifiable records fail conservatively.
- Requires the pinned manaflow Ghostty revision.

---
[...]

PR #8444, opened 2026-07-19
Make long cmux TUI session sockets portable on macOS

lawrencecchen:
Stacked on #8505, the current-main replacement for closed #8422.

## Why

Browser workspace identities produce durable session names that can exceed macOS `sockaddr_un.sun_path` when combined with the normal per-user `TMPDIR`. The daemon and clients must independently select the same reachable endpoint.

## What

- preserve the existing XDG/TMPDIR path when it fits
- fall back to `/tmp/cmux-tui-$uid/<session>.sock` when it does not
- keep the parent private (`0700`) and socket `0600`
- share the contract across the server and published Rust client

## Current stack

- Base PR: #8505 at `e4b546b2dc7484418cca012231bf4083456fc12e`
- This head: `54fa9cfd671474a8cf7d9f721b0fc8fa373d76f4`
- Current `main`: `9fb1b4b54ca215b4b77e934f8203a1841fb3ccb4`

## Verification

The final stacked head passes the full locked serial workspace suite, 302/302 core tests, a separate 20/20 terminal-host recovery rerun, formatting, and diff checks. A 39/39 range-diff preserved the stack during the current-main refresh.


lawrencecchen:

PR #8452, opened 2026-07-19
perf(tui): make terminal resize acknowledgments lightweight

lawrencecchen:
## Summary

- preserve the last authoritative terminal cell-pixel metrics when a resize ioctl reports zero pixels
- negotiate targeted `ResizeAck` responses for browser renderers while keeping protocol-v1 legacy behavior
- skip full VT replay and color serialization when a viewer resize leaves the canonical smallest grid unchanged
- order changed-grid acknowledgments after the atomic `Resized + Colors` pair without consuming the live sequence
- encode the canonical resize replay length exactly as the cross-language terminal-host protocol requires

## Compatibility

The optimization is gated by a `ClientHello` / `HostHello` feature bit. New clients fall back to the existing replay path against older hosts, and older clients continue receiving replay acknowledgments from this host.

## Validation

- `cmux-tui` unit suite: 261 passed
- integration suite: 23 passed
- terminal-host recovery suite: 15 passed
- terminal-host core suite: 223 passed
- byte-for-byte Rust/C++ resize payload golden coverage
- all-target Clippy with warnings denied
- formatting and diff checks

## Stack

- Base: #8444
- Browser companion: manaflow-ai/cmux-browser#14 (pending pin/update)


lawrencecchen:

PR #8475, opened 2026-07-19
fix(tui): release stale daemon viewer leases

lawrencecchen:
Stacked on #8460.

The durable daemon-side Admin connection used to remain a host viewer at its snapshot grid. After Chromium cut over from the compatibility control attachment to a direct renderer, the old Admin lease could keep the canonical terminal at 80x24 forever.

This PR:
- adds a red integration regression first, followed by the fix;
- auto-registers only resize-capable Renderer connections as snapshot viewers;
- explicitly releases Admin on every connect for compatibility with live protocol-v1 hosts from the parent branch;
- tracks the mux desired Admin lease across failed writes and replays it atomically during reconnect;
- propagates a missing effective mux viewer as ReleaseViewer;
- preserves smallest-viewer arbitration, same-size re-registration, and restores the direct renderer size after a TUI detaches.

Validation:
- regression repeated 20 times;
- terminal_host_recovery: 17 passed;
- cmux-tui-core: 223 unit + 44 integration/doc tests passed;
- cargo fmt --check and git diff --check passed.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8475"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787048301&installation_id=133855650&pr_number=8475&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8475&signature=0159c30d3f3362b7cab91a2c1c939a183ef436dad32704d1642dd221afd35887"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
[...]

PR #8476, opened 2026-07-19
fix(tui): make attached frontends match source state

lawrencecchen:
Stacked on #8475. Consumed by cmux Browser [#14](https://github.com/manaflow-ai/cmux-browser/pull/14), which pins this exact commit (`768c8d3f99a3b04294a6305a555cc48af638e7a1`) and Manaflow Ghostty `d6f611a3077aa12510761ca10e2a5e0a93979536`.

The complete stack contains canonical cmux main `b5631833741d0a1a2c25ede49988eadf2f412dd2`.

## Summary

- mirror authoritative source colors, resolved cursor visuals, screen/cursor activity, and exact VT state into attached frontends
- publish pane/tab/screen/workspace selection from an attached TUI back to the owner mux
- acknowledge attachment-scoped viewer sizes so the canonical PTY follows the smallest live viewer
- require canonical workspace UUIDs and preserve armed mouse releases across routing refresh/recovery
- add local-only, PID/generation-fenced daemon handoff while durable per-terminal hosts remain alive
- make hosted daemon Ghostty surfaces response-silent; the durable terminal host remains the sole authority for Kitty/DA/DSR replies
- drain the bounded Kitty graphics probe after recognition so delayed response bytes cannot leak into a shell
- retain live rendered viewport bounds after introducing the terminal-size cache, restoring mouse text selection

These focused attach, registry, state, resize, input, and handoff APIs are sufficient for cmux Browser. Another broad custom-frontends abstraction is not required.

## State and failure semantics

- cmux TUI is the only writer for canonical workspace and terminal lifecycle.
- Browser and attached TUI models are projections; Browser-only layout lives in an opaque versioned frontend projection.
- The PTY-owning host survives daemon, Browser, renderer, and attached-frontend replacement.
- Hosted mirrors parse state for projection but never write parser-generated terminal responses back into the PTY.
- Recovery retains release-only mouse tombstones to prevent a stuck button after an uncertain downstream flush. Exact-once observation would require tokenized host-side input acknowledgements/deduplication.

## Latest-main regression audit

The stacked tip was compared directly with canonical main rather than only tested in isolation. That exposed nine stack-only Clippy warnings, a stale rendered-bounds selection regression, and a smoke assertion that incorrectly assumed indexed foreground SGR would never share an escape with background state.

This tip clears every warning, restores live bounds, and makes the smoke parser recognize exact SGR commands without stripping valid combined foreground/background output. Indexed and RGB operands are consumed atomically, so an operand value such as `31` cannot falsely satisfy a foreground-red assertion.

## Validation

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo check --workspace --all-targets --locked`
- `cargo test --workspace --locked`
- `cargo build -p cmux-tui --locked`
- `python3 scripts/smoke-tui.py` (`SMOKE OK`)
- `python3 scripts/smoke-attach.py` (`ATTACH SMOKE OK`)
- TypeScript SDK build and 41 tests
- web frontend production build and 150 tests
- exact revision-stamped release artifact reports cmux `768c8d3...` and Ghostty `d6f611a...`; all 585 bundled Ghostty resource hashes verify
- consuming Browser PR: 23/23 host suites, including 149 renderer-ordering checks and 87 TUI-protocol checks


lawrencecchen:

PR #8504, opened 2026-07-20
cmuxTests: gate the keyDown-forward assertion on a live surface

ejc3:
`testTypingRepairForwardsKeyDown` asserted that a repaired keyDown reached libghostty:

```swift
XCTAssertGreaterThan(repairProbe.forwardedKeyDownCount(), 0,
                     "Typing repair should forward the keyDown into Ghostty")
```

Forwarding only happens once the runtime surface is live, and the headless xctest host does not reliably spin one up. The assertion then fails for a reason unrelated to typing repair, and the neighbouring tests in this file already respect that constraint. This gates the forward observation on a live surface the same way.

One thing a reviewer should know, because the history is confusing. This branch previously also carried a product change to `Sources/Workspace.swift` for remote-workspace restore handing the terminal a local working directory, and the commit headline described only the test half while the PR title described only the product half. That product fix is #8634, which restructures the same logic behind a named `restoresLocalTerminalSurface` predicate and gates both the local-directory and the startup-handles-working-directory paths instead of just one. Two competing changes to the same function would have conflicted, so the hunk is dropped here and #8634 owns it. What remains is the test change alone, and the title now matches it.

`cmuxTests/AppDelegateShortcutRoutingTests.swift` is also touched by #8635, which disables `isReleasedWhenClosed` on a window it closes. That is a different defect in the same file — a host kill rather than a headless assertion — so both are wanted; whichever lands second needs a trivial rebase.

### DIFF
--- a/cmuxTests/AppDelegateShortcutRoutingTests.swift
+++ b/cmuxTests/AppDelegateShortcutRoutingTests.swift
@@ -6303,18 +6303,21 @@ final class AppDelegateShortcutRoutingTests: XCTestCase {
 #if DEBUG
         XCTAssertEqual(repairProbe.repairCount(), 1, "window.sendEvent should run the focused terminal repair path")
         XCTAssertTrue(repairProbe.repairResponder() === orphanResponder, "Repair should evaluate the simulated stranded responder")
-        // Forwarding the repaired keyDown into libghostty only happens once the
-        // runtime surface is live. The headless xctest host does not always spin one
-        // up, so gate the forward observation on a live surface (the same constraint
-        // GhosttyPhysicalInputFocusReassertionTests uses). The repair routing itself
-        // is verified above regardless of surface liveness.
-        if terminalPanel.surface.hasLiveSurface {
-            XCTAssertGreaterThan(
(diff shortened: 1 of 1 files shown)


ejc3:
Replaced the `if terminalPanel.surface.hasLiveSurface { … }` wrapper around the keyDown-forward assertion with `try XCTSkipUnless(...)`, pushed as 4cf9413657.

The conditional was my own doing and it was the wrong shape. On a host without a live libghostty surface the assertion simply vanishes and the test still reports green — so nothing distinguished a run that verified the forwarding from a run that did not, and the comment right above it conceded the headless host does not always provide a surface. A skip states that; a silent `if` does not. The repair-routing assertions run before this point and are checked on every host either way.

Separately, and not fixed here: this PR's own suite is one of seven that kill their test host on current `main`. Measured one suite per host, `AppDelegateShortcutRoutingTests` shows 1 restart and a crash signature. A host death does not fail one test, it discards every verdict still pending in that host, so any count from this suite is a lower bound until that is fixed. The three timeout raises in this PR (1.0 to 5.0) wait on real predicates rather than sleeping, so they are not the cause.

PR #7768, opened 2026-07-09
Fix app-host crash-restart storm: test windows default to releasedWhenClosed = false

lawrencecchen:
Every test in `BrowserDeveloperToolsVisibilityPersistenceTests` silently crash-restarted the shared app host once per run, and main's baseline CI runs carry dozens of the same silent restarts across shards (34 on one shard of run [28984723481](https://github.com/manaflow-ai/cmux/actions/runs/28984723481)). The time-weighted resharding in https://github.com/manaflow-ai/cmux/pull/7687 made the latent bug visible by co-locating the crasher with live-WKWebView suites, which then failed collaterally (worked around there with `SEPARATED_SUITES`).

Root cause, pinned with lldb and NSZombie on the AWS M4 Pro builder: test-created `NSWindow`s keep AppKit's default `releasedWhenClosed == true`. Under ARC, teardown `close()` sends an extra release, and XCTest's post-test autorelease-pool drain then over-releases the deallocated window: `EXC_BAD_ACCESS` in `objc_release` inside `XCTMemoryChecker`, zombie message `-[NSWindow release]: message sent to deallocated instance`. The host exits with code 1 under the debugger, so no crash report is written and xcodebuild logs only 'Restarting after unexpected exit'. 269 of 299 window creations across 61 test files lack the `isReleasedWhenClosed = false` guard.

Fix: a test-bundle constructor (`cmuxTests/CmuxTestWindowReleaseGuard.m`) swizzles NSWindow's designated initializers so every window created in the test process defaults to `releasedWhenClosed = false`, the only correct value for ARC-managed windows. One hook kills all 269 latent sites plus any future ones; production code is untouched (it already sets `false` defensively at its 13 sites). Commit 1 adds the failing guard tests (red), commit 2 the fix (green), per the regression-test policy.

Verified on aws-m4pro-2: the DevTools suite went from 16 host restarts and 21 cascade failures per isolated run to 0 restarts.

---
[...]

PR #8510, opened 2026-07-20
Fix the file-open working directory keeping a trailing slash (and the test-window over-release hiding it)

ejc3:
## The bug

Open a file from a terminal and the workspace's working directory comes out with a trailing slash: `/tmp/scripts/run.command` yields `"/tmp/scripts/"` instead of `"/tmp/scripts"`. `path(percentEncoded:)` preserves a directory URL's trailing slash, and `deletingLastPathComponent()` produces a directory URL.

That value becomes the workspace working directory — which the window title renders through `{activeDirectory}`, and which directory comparisons match on — so the stray slash is user-visible. Switched to `.path`, which reports the same decoded path without it.

**Verified rather than assumed:** `.path` strips the trailing slash, still returns `"/"` for a file at the root, and decodes percent-literals identically. The test already expects `directory.path`, and the repo uses `.path` for parent directories elsewhere.

## The bigger find: the suite was reporting a fake green

`WindowKeyDownReplayGuardTests` printed `Test run with 0 tests in 1 suite passed` — **after 3 host restarts**. Nine `NSWindow` creation sites across 4 test files used AppKit's default `isReleasedWhenClosed = true` and then `close()`d under ARC, over-releasing the window; teardown crashed in `objc_release` inside `objc_autoreleasePoolPop` and took whole suites down with it while still printing a verdict. Fixed to the repo's existing convention (`isReleasedWhenClosed = false`, which `trackTestWindow` already follows).

The effect is large: one batch went from **3 tests actually running with 6 host restarts** to **53 tests across all 8 suites with 0 restarts**. An A/B (patched vs unpatched, same base) confirms this **un-hid** failures rather than causing them — the honest count went from "7 failing" to 9 once the crash stopped swallowing tests.

## One stale test assumption

Two offscreen-startup tests demanded a runtime-create attempt *inside* `TerminalPanel.init`. The Claude-shim install gate (added later) defers the first attempt through a detached Task and a main-actor continuation, so `return (false, nil)` always won — unpassable as written. Rather than assume the product regressed, this asserts the invariant still holds: `uiWindow` stays nil, and it bounded-polls for the attempt using the repo's `waitUntil` convention.

No `XCTSkip`/`XCTSkipIf` — none was warranted.

## Verified (by name, 0 host restarts everywhere)

| suite | before | after |
|---|---|---|
| `TerminalDefaultFileOpenRequestTests` | 4 tests, **2 failures** | **4 tests, 0 failures** |
| `TerminalOffscreenStartupTests` | 27 tests, **5 failures** | **27 tests, 3 failures** |
| `WindowKeyDownReplayGuardTests` | **0 tests "passed"**, 3 restarts | **6 tests**, 1 real failure |
| `WindowTitleTemplateTests` | passed (crashed others) | 12 tests passed |
| `TerminalNotificationOpenPanelFallbackTests` | passed (crashed others) | 2 tests passed |

Baseline: 55 suites in scope (31 `XCTestCase` + 24 swift-testing); 48 passed, 9 fail after the crash fix stopped hiding two of them.

## Left red on purpose (9 tests / 6 suites)

Not skipped, because none is proven unfixable: `TerminalOffscreenStartupTests` ×3 (`MobileHostService` never publishes routes in 200 polls — smells environmental, but that is an unconfirmed guess and shouldn't be trusted), `TerminalNotificationClearAllTests` ×4, `TerminalNotificationDirectInteractionTests` ×3, `TerminalSearchOverlayMouseReleaseTests` ×2, `WindowTerminalHostViewTests` ×1, `PortalHitTestingPerformanceTests` ×1, `WindowKeyDownReplayGuardTests` ×1, `TitlebarControlsSizingPolicyTests` ×1. The `TerminalOffscreenStartupTests` and `WindowKeyDownReplayGuardTests` remainders reproduce on the `main` base, so they are pre-existing.

## Harness notes for anyone running these locally

- `SWIFT_BACKTRACE=enable=no` does **not** reach the test host (`testmanagerd` spawns it, not your shell) — use **`TEST_RUNNER_SWIFT_BACKTRACE=enable=no`**, or every crash wedges ~2 minutes on the interactive backtracer.
- Several suites are declared `@Suite("display name")` and never print their type name, so a log analyzer keyed on type names will report them as "never ran" when they passed.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8510"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787114605&installation_id=133855650&pr_number=8510&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8510&signature=2a5a47ff16d0f22ef03ebbb0a3e037f9c2e0d958430b201918d17df3f3a84dc6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fix the working directory path when opening a file and stabilize terminal/window tests. This removes stray slashes in titles and stops test-host crashes that hid failures.

- **Bug Fixes**
  - File open: use `.path` for the parent directory in `TerminalDefaultFileOpenRequest` to drop trailing slashes while keeping root "/"; fixes `{activeDirectory}` titles and directory comparisons.
  - Tests: set `isReleasedWhenClosed = false` on AppKit test windows (WindowKeyDownReplayGuard, WindowTitleTemplate, Notification fallback, Ghostty config) to prevent over-releases; update offscreen-startup tests to assert `uiWindow` stays nil and poll for the first runtime creation attempt.

<sup>Written for commit a62b7599d3d3338da370718e3d3951d52c789677. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Sources/AppDelegate+CmuxSSHURL.swift
+++ b/Sources/AppDelegate+CmuxSSHURL.swift
@@ -207,7 +207,12 @@ struct TerminalDefaultFileOpenRequest: Equatable {
         }
 
         self.fileURL = standardizedURL
-        self.workingDirectory = standardizedURL.deletingLastPathComponent().path(percentEncoded: false)
+        // `path(percentEncoded:)` keeps a directory URL's trailing slash, so the parent of
+        // "/tmp/scripts/run.command" came back as "/tmp/scripts/". That string becomes the
+        // workspace's working directory, which the window title renders via {activeDirectory}
+        // and which directory comparisons match on, so the stray slash is user visible. `path`
+        // reports the same decoded path without it and still reports "/" for a file at the root.
+        self.workingDirectory = standardizedURL.deletingLastPathComponent().path
         self.initialInput = "\(Self.shellSingleQuoted(standardizedURL.path(percentEncoded: false)))\n"
     }
 
--- a/cmuxTests/GhosttyConfigTests.swift
+++ b/cmuxTests/GhosttyConfigTests.swift
@@ -2059,6 +2059,9 @@ final class BrowserPanelWebViewLifecycleTests: XCTestCase {
             backing: .buffered,
             defer: false
         )
+        // AppKit defaults to isReleasedWhenClosed, so the close() below would release a
+        // window ARC still owns and the over-release lands in a later autorelease pool drain.
+        realHostWindow.isReleasedWhenClosed = false
         defer {
             realHostWindow.contentView = nil
             realHostWindow.close()
--- a/cmuxTests/TerminalAndGhosttyTests.swift
+++ b/cmuxTests/TerminalAndGhosttyTests.swift
@@ -1041,6 +1041,22 @@ final class TerminalOffscreenStartupTests: XCTestCase {
         XCTAssertGreaterThan(panel.surface.debugRuntimeSurfaceCreateAttemptCountForTesting(), 0)
     }
 
+    /// Bounded poll on the main run loop. The first runtime-surface creation for a surface
+    /// waits on the Claude command shim install, which hops through a detached Task and a
+    /// main-actor continuation, so the attempt lands a turn or more after init returns. The
+    /// invariant these tests guard is that the attempt needs no window attach, not that it
+    /// happens synchronously inside init.
+    private func waitUntil(timeout: TimeInterval = 5.0, condition: () -> Bool) -> Bool {
+        let deadline = ProcessInfo.processInfo.systemUptime + timeout
+        while ProcessInfo.processInfo.systemUptime < deadline {
+            if condition() {
+                return true
+            }
+            RunLoop.current.run(until: Date().addingTimeInterval(0.01))
+        }
+        return condition()
+    }
+
     func testInitialInputSurfaceAttemptsRuntimeCreationBeforeWindowAttachment() {
         let panel = TerminalPanel(
             workspaceId: UUID(),
@@ -1051,9 +1067,9 @@ final class TerminalOffscreenStartupTests: XCTestCase {
             panel.surface.debugHasHeadlessStartupWindowForTesting(),
             "Restored auto-resume input should bootstrap through a hidden window rather than waiting for a user-focused portal."
         )
-        XCTAssertGreaterThan(
-            panel.surface.debugRuntimeSurfaceCreateAttemptCountForTesting(),
-            0,
+        XCTAssertNil(panel.surface.uiWindow, "The runtime must start without any real window attach.")
+        XCTAssertTrue(
+            waitUntil { panel.surface.debugRuntimeSurfaceCreateAttemptCountForTesting() > 0 },
             "Restored auto-resume input must start the terminal runtime without waiting for a window attach."
         )
     }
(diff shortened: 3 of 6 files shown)


ejc3:
Measured this on macOS at head `a62b7599d3`, one suite per app host, each with its own test host so nothing else can be blamed. `#8651` is applied on top because the `cmuxTests` target does not compile without it, so these runs bake in that one-line fixture fix.

Two suites go green and the third stops killing the host:

| suite | host restarts | tests ran | verdict |
|---|---|---|---|
| `WindowTitleTemplateTests` | 0 | 12 | `** TEST SUCCEEDED **` |
| `TerminalNotificationOpenPanelFallbackTests` | 0 | 2 | `** TEST SUCCEEDED **` |
| `WindowKeyDownReplayGuardTests` | 0 | 6 | `** TEST FAILED **` |

For the before side I ran `WindowTitleTemplateTests` alone on `f5361074f6` (main tip) with only `#8651` applied: **5 host restarts, 2 of 12 tests ran, `** TEST FAILED **`**. Same worktree and the same warm derived-data path for both arms, with this change as the only difference.

`WindowKeyDownReplayGuardTests` deserves a note rather than a green tick. It went from 0 tests to 6, so the crash is gone, and what is left is one real assertion that nobody could see before because the host died first:

```
✘ terminalHostedEditableResponderKeepsLocalUndo() at WindowKeyDownReplayGuardTests.swift:340:9
  Expectation failed: (textView.undoCallCount → 0) == 1
```

That is its own bug and wants its own change. Worth saying plainly: the log also holds two earlier `Executed 0 tests, with 0 failures` lines, so anything grepping only the last line would read this run as fine.

One thing to keep in mind while reviewing. This suite's window guard is not a style preference. The product already sets `isReleasedWhenClosed = false` everywhere it owns a window (`BrowserPanel.swift:4347`, `BrowserPrewarmedWebViewPool.swift:181`, `BrowserPopupWindowController.swift:115`, `ReleasingWindowController.swift:62`); only the tests were missing it. Counting the tests that close a self-created unguarded window predicts the host-restart count exactly across five other suites (16, 3, 1, 1, and a correct 0), which is why I would not take the process-wide swizzle approach in #7768: it would also stop the product's own windows in the host from releasing on close, changing the behavior of the code under test.

ejc3:
Closing: already on `main`. All six files, including the production fix in `Sources/AppDelegate+CmuxSSHURL.swift`, are byte-identical to main's versions by blob hash. Nothing left to merge.

PR #8549, opened 2026-07-21
Fix identify caller recovery for restored shells

### LINKED ISSUE #8362: 0.64.19: `cmux identify` returns caller:null from shells in auto-restored workspaces
Version: cmux 0.64.19 (99) [1c22c5564], macOS.

After the 0.64.19 update restart, `cmux identify` run from a shell inside a workspace that was **auto-restored** by the restart returns `"caller": null` — while shells in workspaces **freshly spawned after** the restart return the full caller block (pane_ref/surface_ref/workspace_ref).

Impact: scripted workflows that resolve their own workspace_ref (e.g. `identify` → `read-screen --workspace <ref>`) break inside restored panes; there is no other way for a shell to learn its own workspace.

Hypothesis from the pattern: restored terminal surfaces don't re-inject whatever env/socket linkage maps a shell to its surface, while newly-created surfaces do.

Repro: have open workspaces with running terminals → update/restart cmux → let it auto-restore → run `cmux identify` in a restored pane (caller:null) vs a new workspace's pane (caller populated). Expected: caller populated in both.

austinywang:
## Summary

- have caller-aware `cmux identify` requests include the CLI process's normalized controlling TTY when no explicit window target changes caller semantics
- keep a valid explicit workspace/surface caller authoritative, then recover missing or stale caller identity through an owned resolver spanning current live Ghostty PTYs and shell-reported nested-multiplexer PTYs
- prefer an unambiguous live runtime PTY match; only when no live PTY matches, accept one unique host-local shell-reported TTY binding so tmux pane TTYs remain identifiable
- fail closed with `caller: null` on ambiguity, exclude remote TTY namespaces, and never guess from the focused workspace
- share caller payload construction between explicit identity and TTY recovery

Closes #8362

## Root cause

`cmux identify` supplied caller identity only from inherited `CMUX_WORKSPACE_ID` / `CMUX_SURFACE_ID`. `system.identify` validated that caller object but had no independent signal when those variables were absent or stale, so the response necessarily contained `caller: null` even though the invoking interactive process still had a controlling TTY.

The current restore spawn path does inject managed runtime IDs, and the issue did not reproduce on a clean macOS 15 machine with either a same-version 0.64.19 restart or a v0.64.18 → v0.64.19 upgrade (including eight paced restored workspaces). The failure is therefore state-dependent rather than evidence that every restored surface skips injection. This change removes the single-signal dependency so a thin restored shell environment can still identify itself.

For ordinary shells, the server matches against current live Ghostty PTYs. A shell inside tmux sees the nested pane PTY rather than Ghostty's outer PTY, so live matching alone is insufficient; the resolver falls back to the existing shell-reported binding only when no live candidate matches and every matching report names the same host-local workspace/surface. A live match outranks stale reports, conflicting live or reported bindings return `caller: null`, and remote workspaces never participate in the local TTY namespace.

## Testing

- red proof on test-only commit `ed97f8e54`: [focused Actions run](https://github.com/manaflow-ai/cmux/actions/runs/29797273256) executed 5 tests; only `identifyWithoutCallerIdsSendsCallerTTY` failed because `caller_tty` was `nil`
- green on final HEAD `80b00e4187`: [bundled CLI regression](https://github.com/manaflow-ai/cmux/actions/runs/29799201195) executed 5 Swift tests, including `identifyWithoutCallerIdsSendsCallerTTY`
- green on final HEAD `80b00e4187`: [TTY resolver behavior](https://github.com/manaflow-ai/cmux/actions/runs/29799201198) executed 4 Swift tests covering live authority over stale reports, unique tmux fallback, duplicate-report rejection, and ambiguous-live fail-closed behavior
- `xcrun swiftc -frontend -parse` for all touched Swift files
- standalone `swiftc -warnings-as-errors -typecheck` for the new binding and resolver value types
- `./scripts/check-pbxproj.sh`
- `./scripts/lint-pbxproj-test-wiring.sh` (547 test files)
- `python3 scripts/check-workspace-package-groups.py --check`
- `python3 scripts/check-package-resolved-policy.py`
- `git diff --check`

A native live-PTY app-host test was attempted twice before the PR, but that selected CI host could not materialize a Ghostty runtime surface before the assertion. Rather than fake that coverage, the regression commit uses the real bundled CLI against a mock control socket, while the server-side resolution policy is covered independently with injected live and reported candidate snapshots.

The requested `scripts/swift_file_length_budget.py` is absent from the branch's original `origin/main`; all new Swift files are under 500 lines, and neither budget TSV changed.

## Localization audit

No user-facing strings, help text, schema descriptions, menus, settings, alerts, or tooltips changed. `Resources/Localizable.xcstrings` and both web locale catalogs are untouched.

## Demo Video

Not applicable: this is CLI/control-socket identity recovery with no visual UI change. A clean cloud-Mac GUI provisioning attempt timed out before yielding a usable session, so no full-screen video was available.

## Checklist

- [x] Docs/changelog are not needed for this internal caller-resolution repair
- [x] I requested bot reviews after my latest commit
- [x] All human review comments are resolved

## Summary by CodeRabbit

* **New Features**
  * Improved terminal caller identification using TTY information when structured caller details are unavailable.
  * Added support for resolving caller TTYs to the correct active terminal panel.
  * Added configurable port-scanner handling for improved workspace and terminal state reporting.

* **Bug Fixes**
  * CLI identification now includes the caller’s TTY when no caller IDs are provided.
  * Ambiguous or stale TTY matches are safely rejected.

* **Tests**
  * Added coverage for TTY freshness, caller resolution, ambiguity handling, and CLI identification.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -4283,6 +4283,9 @@ struct CMUXCLI {
                         params["caller"] = caller
                     }
                 }
+                if effectiveWindowRaw == nil, let callerTTY = resolveCallerTTYName() {
+                    params["caller_tty"] = callerTTY
+                }
             }
             let response = try client.sendV2(method: "system.identify", params: params)
             print(jsonString(formatIDs(response, mode: idFormat)))
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -4287,7 +4287,7 @@ struct CMUXCLI {
                 if effectiveWindowRaw == nil,
                    idWsFlag == nil,
                    idSurfaceFlag == nil,
-                   let callerTTY = resolveCallerTTYName() {
+                   let callerTTY = resolveCallerDescriptorTTYName() {
                     params["caller_tty"] = callerTTY
                 }
             }
@@ -25417,6 +25417,10 @@ struct CMUXCLI {
                 return ttyName
             }
         }
+        return resolveCallerDescriptorTTYName()
+    }
+
+    func resolveCallerDescriptorTTYName() -> String? {
         for fileDescriptor in [STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO] {
             if let rawTTYName = ttyname(fileDescriptor),
                let ttyName = normalizedTTYName(String(cString: rawTTYName)) {
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -4255,8 +4255,9 @@ struct CMUXCLI {
             let includeCaller = !hasFlag(commandArgs, name: "--no-caller")
             if includeCaller {
                 let idWsFlag = optionValue(commandArgs, name: "--workspace")
+                let idSurfaceFlag = optionValue(commandArgs, name: "--surface")
                 let workspaceArg = idWsFlag ?? (effectiveWindowRaw == nil ? ProcessInfo.processInfo.environment["CMUX_WORKSPACE_ID"] : nil)
-                let surfaceArg = optionValue(commandArgs, name: "--surface") ?? (idWsFlag == nil && effectiveWindowRaw == nil ? ProcessInfo.processInfo.environment["CMUX_SURFACE_ID"] : nil)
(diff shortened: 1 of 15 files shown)


austinywang:
@coderabbitai review

PR #8629, opened 2026-07-22
Make app-host CI classification fail closed and retire impossible waits

ejc3:
## Summary

This PR makes the macOS app-host test gate fail closed and removes two test fixtures that could hang or leak state:

- Mixed XCTest + Swift Testing runs now have explicit terminal-result classification. A missing Swift Testing phase, a Swift Testing failure, and a total deadline cannot be hidden by an earlier tolerant XCTest summary.
- The PTY wrapper supports a non-resetting total deadline while retaining the existing idle timeout and heartbeat behavior.
- App-host configuration evidence is canonicalized and rejects symlink/traversal aliases outside the isolated home.
- `TerminalOffscreenStartupTests` tear down headless surfaces and seed the public loopback route cache directly instead of waiting on the unauthenticated `mobile.host.status` probe.
- The determinism gate covers executable `Process.run`/JavaScript `exec` cases without treating rendered command text as a live operation.

The deferred-browser lifecycle fix and browser color assertions described in the original PR are already present on `main`; the effective product-facing delta here is test/CI reliability and classification.

## Verification

- Local: Python syntax, shell syntax, determinism self-test (145 positive + 87 negative fixtures), package-resolved policy, workspace grouping, pbxproj test wiring, and project normalization.
- AWS builder: the noninteractive helper harness, app-host isolation/retry harness, determinism self-test, package policy, and pbxproj wiring all passed at the pushed head.
- Hosted macOS E2E: `TerminalOffscreenStartupTests` dispatched against `refs/pull/8629/head` (run link is tracked in the handoff).

## Trade-offs

- The wrapper reserves exit codes 123/126/127 for authoritative Swift Testing failure, missing-phase, and total-timeout outcomes. This preserves the existing tolerant XCTest behavior while making mixed-framework runs auditable; callers must treat these codes as blocking.
- Six physical shards with two sequential logical batches are retained because current app-host memory/session limits are measured per worker. The total deadline bounds each invocation; the outer batch guard remains a last-resort containment boundary.
- Canonical path validation requires the app-host configuration sentinel to exist and rejects dot-segment/symlink spellings. This is stricter than accepting any path that merely resolves inside the root, by design.

No production UI/runtime behavior is changed by the current effective diff; no linked GitHub issue was attached to the PR.

## One rule for a batch

A batch whose app host died could be continued here. The per-batch check read only the exit status and the last XCTest summary, and after a relaunch that summary covers only the launch that survived. A batch that exited 0 was not graded at all. `classify-app-host-test-result.sh` now fails a batch with unreadable output, a host-restart line, or no executed tests before it looks at the status, and the step runs it for every batch. `tests/test_ci_app_host_xcodebuild_retry.sh` and `tests/test_ci_change_areas.py` cover a batch with a restart line at status 0, 65 and 125, a batch that ran nothing, and missing output. Those tests failed before the change and pass after it. The step harness now also copies the result classifier, so the multi-batch test exercises the continued path it describes.

### DIFF
--- a/Sources/Panels/BrowserPanel.swift
+++ b/Sources/Panels/BrowserPanel.swift
@@ -4290,7 +4290,13 @@ final class BrowserPanel: Panel, ObservableObject {
             hiddenWebViewDiscardManager.updateRestoredSessionRenderIntent(nil)
             currentURL = initialRequest.url
             shouldRenderWebView = renderInitialNavigation
-            guard renderInitialNavigation else { return }
+            guard renderInitialNavigation else {
+                // `shouldRenderWebView` is already false here, so the assignment above does not
+                // change it and its `didSet` never fires. Without this the panel keeps the
+                // `.newTab` state it was born with and reports a deferred URL as a new tab.
+                refreshWebViewLifecycleState()
+                return
+            }
             if let url = initialRequest.url,
                insecureHTTPBypassHostOnce == nil,
                shouldBlockInsecureHTTPNavigation(to: url) {
@@ -4309,7 +4315,10 @@ final class BrowserPanel: Panel, ObservableObject {
             hiddenWebViewDiscardManager.updateRestoredSessionRenderIntent(nil)
             currentURL = url
             shouldRenderWebView = renderInitialNavigation
-            guard renderInitialNavigation else { return }
+            guard renderInitialNavigation else {
+                refreshWebViewLifecycleState()
+                return
+            }
             if adoptedPrewarmedWebView {
                 // Already navigated while hidden; record for recovery paths.
                 navigationDelegate?.recordAttemptedRequest(URLRequest(url: url), displayURL: url)
--- a/cmuxTests/BrowserConfigTests.swift
+++ b/cmuxTests/BrowserConfigTests.swift
@@ -1879,16 +1879,30 @@ final class BrowserDeveloperToolsConfigurationTests: XCTestCase {
             ]
         )
 
+        // The panel no longer paints the translucent colour itself: it composites it over the
+        // window background and fills opaquely. Derive the expectation from the same helper the
+        // panel uses, because a hardcoded triple would encode the host's appearance -- the
+        // composite blends against `.windowBackgroundColor`, which differs in dark mode. The
+        // accuracy covers WebKit storing `underPageBackgroundColor` at 8 bits per channel.
         guard let actual = panel.webView.underPageBackgroundColor?.usingColorSpace(.sRGB),
-              let expected = updatedColor.withAlphaComponent(updatedOpacity).usingColorSpace(.sRGB) else {
+              let expected = GhosttyBackgroundTheme.color(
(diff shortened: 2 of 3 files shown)


ejc3:
Found while reconciling the open browser PRs into one tree: **this PR and #8708 fix the same bug in the same file, two different ways**, and they conflict.

Both address a render-deferred panel keeping the `.newTab` state it was born with, because `shouldRenderWebView` is assigned false over false so its `didSet` never fires.

- This PR moves `refreshWebViewLifecycleState()` *inside* the `guard renderInitialNavigation else` block, so it runs exactly on the deferred path.
- #8708 calls it *before* the guard, so it runs on every path.

For what it is worth I think this PR's placement is the better one: it refreshes precisely where the observer cannot, while #8708 also refreshes on the normal path where the subsequent navigation would refresh anyway. But #8708 is already non-draft and measured green in a combined build, so the cheap resolution is to drop the product hunk here and keep whichever of the two test changes are not duplicated elsewhere.

The test changes are where this PR still carries unique content, and they collide too: `cmuxTests/BrowserConfigTests.swift` overlaps `fix/thirteen-test-suite-repairs` and `cmuxTests/TerminalAndGhosttyTests.swift` overlaps `browser-headless-fixes`. In a tree containing #8708 plus those two, cherry-picking this commit resolves to **nothing at all** (empty diff), so as it stands it would land as a no-op once they merge.

Concretely, three PRs are editing the same browser predicate and the same settle-wait helper. Worth landing them in a deliberate order rather than letting whichever merges first decide.

austinywang:
Rechecked against current `origin/main` at `1cee402d7d`: Git reports a clean merge, and the effective merge-result diff is still exactly these three files (13 lines in `BrowserPanel.swift`, 42 in `BrowserConfigTests.swift`, and 70 in `TerminalAndGhosttyTests.swift`). In other words, the overlapping alternatives noted above have not landed on `main`, so this PR is not a no-op today. I have created an origin-only integration branch with the current-main merge so the resulting tree can be exercised by Actions; I did not push to the fork.

austinywang:
@ejc3 Resolved deliberately in merge commit `599a998d77c8`: current `main` now contains the deferred-browser lifecycle fix plus stronger independent color-compositing coverage, so all four overlapping browser hunks took the `main` side. The remaining PR diff is only `cmuxTests/TerminalAndGhosttyTests.swift` (removing the impossible public-status route wait and bounding the two started surfaces). I compared the GitHub conflict-resolution commit tree byte-for-byte with the locally reviewed integration tree; they are identical. Canonical cmux autoreview and policy checks report no actionable findings.

austinywang:

austinywang:
@ejc3 The refreshed CLA check identifies you as the PR opener/remaining committer. Please post the exact phrase `I have read the CLA Document v2.2 and I hereby sign the CLA` from your account so the required CLA check can complete.

austinywang:
recheck

ejc3:
recheck

PR #8633, opened 2026-07-22
A fake WKNavigation was killing the test host, hiding a whole suite

ejc3:
## What was happening

`BrowserDiscardRestorePolicyCancelTests` logged that it started and then produced no verdict at all. That is a dead test host, not a failing test, and it is worse than a red suite: the host takes down every suite batched with it, so any batched measurement including this one was untrustworthy. Reading the eight tests one by one finds nothing wrong with them, which is exactly what you would expect — the suite does not fail, it dies.

## Cause, and a second one behind it

The suite constructs `WKNavigation()` directly. WebKit builds that object's embedded C++ `API::Navigation` itself, so a bare instance carries unconstructed storage. Allocating one is harmless; **releasing** it is not. Reproduced outside the test bundle, deterministically: `EXC_BREAKPOINT` inside `CFRetain` from `-[WKNavigation dealloc]` with WebKit initialised, and `SIGSEGV` through `WebCoreObjCScheduleDeallocateOnMainRunLoop` from the same dealloc without it. The earliest death needs no window: the fake is stored as the pending restore navigation, the next call clears that reference, and the release traps mid-assertion. `WKNavigation()` appears nowhere else in the repo.

That also explains why no output survives: with stdout on a pipe, the crash discards the buffer, so even the line printed immediately before it never reaches the log.

The bookkeeping under test only ever compares these by identity, so the fakes are now minted through a helper that keeps them retained for the run.

Fixing that got the suite far enough to run and pass several tests, which exposed a second killer in the same file: one test builds an `NSWindow`, holds it through ARC, and closes it in a `defer` without disabling AppKit's close-time release, so the last retain goes away underneath live references. About forty other closing sites in `cmuxTests` already guard against this, and the same defect was separately confirmed in another suite.

## Verification

Measured on a macOS builder, the suite run alone in both arms:

| arm | result |
|---|---|
| before | started, then no suite verdict at all |
| this branch | **passing, 8 tests** |

No assertion changed.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8633"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787285629&installation_model_id=21920&pr_number=8633&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8633&signature=4b8584eadf1758ba00e60cb5c0f57fe6ea057cbaa932fef0eef1d98b77a92d0e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Prevents xctest host crashes by replacing fake `WKNavigation()` with a retained stub and disabling `NSWindow` close-time release in one test. The `BrowserDiscardRestorePolicyCancelTests` suite now runs and passes instead of killing the host.

- **Bug Fixes**
  - Added `BrowserDiscardRestoreNavigationStub.make()` to create and retain `WKNavigation` instances for the test run.
  - Replaced direct `WKNavigation()` calls with the stub to avoid WebKit dealloc traps.
  - Set `window.isReleasedWhenClosed = false` in the window-close test to prevent double release.
  - Result: suite passes (8 tests); no more dead host or missing logs.

<sup>Written for commit 34c246008588b63c3fa4663b565e81fcccd707ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8633?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Improved browser discard and restore test stability.
  * Added safer handling for navigation state during stale and current navigation scenarios.
  * Prevented test crashes when windows remain open during deferred closing.

### DIFF
--- a/cmuxTests/BrowserDiscardRestoreHealPredicateTests.swift
+++ b/cmuxTests/BrowserDiscardRestoreHealPredicateTests.swift
@@ -465,6 +465,11 @@ struct BrowserDiscardRestorePolicyCancelTests {
             backing: .buffered,
             defer: false
         )
+        // This test owns the window through ARC, so AppKit must not release it as well when the
+        // deferred close runs. Leaving the default on drops the last retain while other references
+        // to that address are still live, which aborts the test host rather than failing a test.
+        // cmuxTests sets this guard at about seventy other window-closing sites.
+        window.isReleasedWhenClosed = false
         defer {
             panel.resetInsecureHTTPAlertHooksForTesting()
             window.close()
--- a/cmuxTests/BrowserDiscardRestoreHealPredicateTests.swift
+++ b/cmuxTests/BrowserDiscardRestoreHealPredicateTests.swift
@@ -202,6 +202,26 @@ struct BrowserDiscardRestoreHealPredicateTests {
 
 }
 
+/// WebKit constructs a `WKNavigation`'s embedded C++ `API::Navigation` itself, so an instance
+/// made with a bare `WKNavigation()` carries unconstructed storage. Allocating one is harmless;
+/// releasing it is not — `-[WKNavigation dealloc]` traps, which kills the whole xctest host
+/// rather than failing a test, and takes every suite sharing that host down with it. Reproduced
+/// directly outside the test bundle: EXC_BREAKPOINT inside CFRetain from that dealloc with WebKit
+/// initialised, and SIGSEGV via WebCoreObjCScheduleDeallocateOnMainRunLoop without it.
+///
+/// The discard-restore bookkeeping under test only ever compares these by identity, so minting
+/// them here and holding them for the run is behaviour-preserving and keeps the host alive.
+@MainActor
+private enum BrowserDiscardRestoreNavigationStub {
+    private static var retained: [WKNavigation] = []
+
+    static func make() -> WKNavigation {
+        let navigation = WKNavigation()
+        retained.append(navigation)
+        return navigation
+    }
+}
+
 private final class BrowserDiscardRestorePolicyCancelAlert: NSAlert {
     var response: NSApplication.ModalResponse = .alertThirdButtonReturn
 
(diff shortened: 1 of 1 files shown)


ejc3:
Measured on a combined build at c59bbdf7d3 (current `main` + the build fix in #8857 + this):

```
BrowserDiscardRestorePolicyCancelTests   Test run with 8 tests in 1 suite passed
                                         ** TEST SUCCEEDED **   restarts=0  crashlines=0
```

On unfixed `main` the same suite gives 1 host restart, 3 crash lines and **zero tests executed** — no verdict at all, which is why it never showed up as a failure.

Also correcting the record on how this sat open. I re-derived the cause later in the day by running `grep WKNavigation cmuxTests/BrowserDiscardRestorePolicyCancelTests.swift`, got "no such file", and concluded the constructions lived in a sibling suite and the segfault was unexplained. Both suites live in `BrowserDiscardRestoreHealPredicateTests.swift` and all five bare constructions are inside the crashing one. A suite name is not a file name here, and a filename-derived search is not evidence that retracts a diagnosis.

PR #8619, opened 2026-07-22
Stop firing duplicate agent resumes on relaunch

austinywang:
## Summary

Closes #8446.

On app relaunch, `Workspace.createPanel` decided per-panel, independently, whether to auto-resume a restored agent session (`codex resume <id>` / `claude --resume <id>`), using only the persisted `wasAgentRunning` flag with no check of whether a process for that session was already alive. Plain (non-tmux) resumes had no liveness check at all, and nothing prevented two panels referencing the same session from both firing a resume in the same restore pass. After a crash, this could fire a resume for every restored panel simultaneously — the reported repro was 22 concurrent `codex resume` processes within a 30-second window, all idle an hour later, each holding an open handle to codex's SQLite state, causing intermittent "database is locked" failures on unrelated later Codex launches.

- Before constructing `restoredAgentResumeLaunch`, consult the same live-process index already used for "reopen closed tab" and Fork Conversation availability (`SharedLiveAgentIndex` / `RestorableAgentSessionIndex.entry(workspaceId:panelId:)`) via the new `AgentResumeLiveness.hasLiveProcess(for:kind:sessionId:)`. If a process for this exact session is already alive, skip the resume.
- Add `AgentResumeLaunchGuard`, a per-process-lifetime dedup guard: the first panel to resume a given `(kind, sessionId)` claims it; any other panel referencing the same session in the same restore pass is turned away, even before the freshly spawned process becomes visible to the live-process index (which only updates from a periodic process scan).
- Generalize `reconcileSurfaceResumeBindings`, previously tmux-only, to also drop a plain agent-hook resume binding when `AgentResumeLiveness` reports its session is no longer live, so a normal exit of a non-tmux agent doesn't leave a stale binding that gets replayed as a resume on the next relaunch.

`wasAgentRunning`'s nil-defaults-to-true backwards-compatibility behavior is intentionally untouched — a genuine crash-restart with nothing else alive still auto-resumes normally, per the issue's guidance to fix this via the liveness check rather than that default.

Scope note: this targets the `restorableAgent`-driven resume path specifically (`AgentResumeCommandBuilder` / `resumeStartupCommand()`), which is what the issue identifies as the actual unconditional/unchecked mechanism. The separate `resumeBinding`-driven (agent-hook binding) launch path already has its own tmux-specific `isProcessDetected` liveness gate at launch time; this PR generalizes only its *persist-time stale-binding cleanup*, not its launch-time gate, to keep the diff scoped to the reported bug.

## Test plan

[...]

PR #8651, opened 2026-07-22
cmuxTests: pass the liveness the Entry initializer requires

ejc3:
The `cmux-unit` test target does not compile on current main:

```
cmuxTests/AgentResumeLivenessTests.swift:25:25: error: missing argument for
parameter 'processLiveness' in call
** TEST BUILD FAILED **
```

`RestorableAgentSessionIndex.Entry` gained a `processLiveness` field with no default in #8547 (`Sources/RestorableAgentSession.swift:936`), which makes it required in the memberwise initializer. #8619 then added `cmuxTests/AgentResumeLivenessTests.swift`, whose `entry(...)` helper constructs an `Entry` without it. The file is wired into the pbxproj, so this is not a skipped file: the target fails to build, and **no suite in `cmuxTests` can run at all** until it is fixed.

`AgentResumeLiveness.hasLiveProcess` decides from the PID set alone (`guard let entry, !entry.processIDs.isEmpty`), so the value is not load-bearing — it only has to be honest. Deriving it from the PIDs each case asks for keeps the empty-PID test from claiming a running process.

Verified with the same command both ways on a pristine checkout of `main`:

```
before:  ** TEST BUILD FAILED **    (1 compile error, the one above)
after:   ** TEST BUILD SUCCEEDED ** (0 compile errors)
```

Two things that corroborate the timeline. A full 674-suite headless sweep completed on a base that is four commits behind main, and it produced verdicts for 671 suites; that base predates #8619, so the test file is simply absent there and the target builds. And the last commit to touch this test file is #8619 itself.

Worth noting how this reached main: no GitHub Actions workflow builds or tests pull requests in this repo — the only PR checks are third-party review bots — so a change that breaks the unit-test target still shows an all-green PR.

## Summary by CodeRabbit

* **Tests**
  * Improved agent resume liveness test coverage by deriving process liveness from the provided process ID set.
  * Added guidance to keep test data consistent for empty and non-empty process ID scenarios.

### DIFF
--- a/cmuxTests/AgentResumeLivenessTests.swift
+++ b/cmuxTests/AgentResumeLivenessTests.swift
@@ -23,6 +23,10 @@ struct AgentResumeLivenessTests {
             snapshot: SessionRestorableAgentSnapshot(kind: kind, sessionId: sessionId),
             lifecycle: nil,
             updatedAt: 0,
+            // `hasLiveProcess` decides from the PID set alone, so keep the recorded
+            // liveness consistent with the PIDs each case asks for instead of pinning
+            // one value that would contradict the empty-PID case.
+            processLiveness: processIDs.isEmpty ? .exited : .running,
             processIDs: processIDs,
             agentProcessIDs: processIDs,
             agentProcessIdentities: [:]


lawrencecchen:
@codex review

@coderabbitai review

Please review exact head `4bbc1ff47a3a941c8d0b9530ae37cfa25e038786`, the test-only processLiveness initializer repair. Greptile is 5/5 at this exact head and there are zero unresolved review threads.

Issue #5255, opened 2026-06-03
Markdown / file-preview text editor hard-hangs at 100% CPU on click/drag-select (TextKit 2 compatibility mode; #4576 mitigation ineffective)


austinywang:
> **Scope update (generalized):** This is **not** specific to the `note` surface. The hang lives in the **shared text editor** `SavingTextView` / `FilePreviewTextEditor`, which backs three surfaces, all reproducing the same freeze:
> 1. **Markdown** panel in text/edit mode (`MarkdownPanel.setDisplayMode(.text)`) — any `.md` document
> 2. **Note** surface — a note is just a `MarkdownPanel` (with a `noteSlug`), so it inherits the bug
> 3. **File preview** of a large text file (`FilePreviewPanel`, the original #4576 surface)
>
> A note is one *trigger*, not the cause. The fix is confined to `Sources/Panels/FilePreviewTextEditor.swift` (construct `SavingTextView` from an explicit TextKit 1 stack so `textLayoutManager == nil`) and does not depend on the note-surface branch.

## Summary

Clicking into the **`note` surface** text editor hard-hangs the app: the main thread pegs at ~100% CPU and the entire UI freezes, requiring a force-quit (`kill -9`). Observed on a DEV build of the `note` surface feature (PR #4332, branch `issue-4331-note-surface-type`).

This is the same hang signature as the now-closed #4576 (*"Drag-selecting a many-line document hangs the main thread at 100% CPU — `NSTextSelectionNavigation.textSelectionsInteractingAtPoint` O(N) per hit-test"*). The mitigation added for #4576 — `FilePreviewTextEditor.enableLargeDocumentSelectionPerformance()` — **is not actually preventing the hang in the note editor**: the live hang stack still runs through TextKit 2 selection navigation and `synchronizeTextLayoutManagers`.

## Severity

App-freezing. The hang happens inside AppKit's **modal** mouse-tracking run loop (`_bellerophonTrackMouseWithMouseDownEvent`), so the main run loop never returns — the whole app (all workspaces/surfaces) is frozen and unrecoverable without force-quit. In one capture the process had burned **23m53s of CPU at 99.9%** before it was killed.

## Repro

1. Build/run the `note` surface feature (PR #4332).
2. Open a project note with a reasonably large body (many lines — #4576 established the cost is O(N) in line-fragment count, so the threshold scales with note length).
3. Put the note panel in **text/edit mode** and **click into the body** (or start a drag-select).
[...]

Issue #4576, opened 2026-05-22
Drag-selecting a many-line document hangs the main thread at 100% CPU — `NSTextSelectionNavigation.textSelectionsInteractingAtPoint` O(N) per hit-test


nahyeongjin1:
## Summary

Click-dragging to select text over a document with a **large number of line fragments** (e.g. a big raw HTML/text file opened in File Preview, or any long `.textSelection(.enabled)` content) puts cmux into a hard hang: the main thread is pegged at ~100% CPU inside TextKit 2's selection navigation and **never returns to the run loop**, so the app stops dispatching all keyboard/mouse events (totally unresponsive). I sampled a real hang and then reproduced the identical stack **headless** (no cmux, no NSApplication) with a ~30-line TextKit 2 program — so the trigger is isolated.

- cmux **0.64.9 (89)**
- macOS **26.2 (25C56)**, Apple Silicon
- Trigger: **drag-extending** a selection (mouse-down + drag) over a many-line document. A plain click / fresh hit-test does **not** hang — only selection *extension* does.

## Real hang (sampled from a live cmux, PID pegged ~100% CPU, identical across two samples 2s apart)

```
com.apple.main-thread
 -[NSApplication run]
  @objc NSApplication.cmux_applicationSendEvent(_:)            (cmux)
   NSWindow.cmux_sendEvent(_:)                                 (cmux)
    -[NSWindow _handleMouseDownEvent:isDelayedEvent:]
     -[NSTextField mouseDown:]
      SelectionTextField.Cell._selectOrEdit(...)               (SwiftUI)   ← .textSelection(.enabled) field editor
       -[NSTextView mouseDown:]
        -[NSTextView _bellerophonTrackMouseWithMouseDownEvent:...]         ← modal drag-tracking loop
         -[NSTextSelectionNavigation textSelectionsInteractingAtPoint:inContainerAtLocation:anchors:modifiers:selecting:bounds:]  (UIFoundation)
          -[NSTextSelectionNavigation _lineFragmentInfoForPoint:...]
           -[NSTextLayoutManager enumerateSubstringsFromLocation:options:usingBlock:]
            ... NSTextContentStorage synchronizeTextLayoutManagers / enumerate churn (100% of samples)
```

[...]

PR #8719, opened 2026-07-23
tests: build the file-preview text view the way the product does

ejc3:
Seven test call sites constructed a bare `SavingTextView()`. The product never does.

`SavingTextView.makeFilePreviewTextView()` builds an explicit TextKit 1 stack, and its comment says why: a default `NSTextView` is TextKit 2, whose selection work is O(N) in line-fragment count, so clicking or drag-selecting in a large document pegged the main thread inside AppKit's modal mouse-tracking loop and froze the app (#4576, #5255). A bare init has no configured text container, and it also skips `applyFilePreviewTextEditorInsets()`, which the factory applies.

So these tests were exercising a view configuration the app deliberately refuses to ship, and failing on it. The save-shortcut tests read back an empty string instead of the saved text, and the inset test read `nil` where it expected a value. Those looked like two separate bugs and were one.

The files already disagreed with themselves, which is what settled it: `FilePreviewReviewFeedbackTests` used the bare init at line 44 and the factory at line 408, `CanvasShortcutContextTests` uses the factory at three sites, and `FilePreviewTextEditorTextKitTests` exists specifically to assert the factory yields a pure TextKit 1 view. The seven bare calls were the outliers, not the convention.

## Measured

| suite | before | after |
|---|---|---|
| `FilePreviewPanelTextSavingTests` | 27 tests, **3 failures** | 27 tests, **0 failures**, `** TEST SUCCEEDED **` |
| `FilePreviewReviewFeedbackTests` | 17 tests, **1 failure** | 17 tests, **0 failures**, `** TEST SUCCEEDED **` |

One suite per app host, same worktree and warm derived-data path for both arms.

## Depends on #8701

Both suites close a self-created `NSWindow` and need the release guard in #8701 to reach these assertions at all; without it the host dies first and neither suite reports. Both arms above were measured with that guard applied, so this change is what moves them from failing to green, and #8701 is what lets them run.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8719"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787370806&installation_model_id=21920&pr_number=8719&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8719&signature=d2c297383774daf48c3a95b96bb7193210788df7e68061ef1bd2bcca668fdced"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Build the file‑preview text view in tests the same way the app does. Replace `SavingTextView()` with `SavingTextView.makeFilePreviewTextView()` to force TextKit 1 and apply expected insets across file preview saving, theming, movement, and focus tests.

- **Bug Fixes**
  - Stops tests from creating a default TextKit 2 `NSTextView` with no container/insets, which caused empty readbacks and nil inset values.
  - Aligns tests with product behavior and other suites; resolves failures in FilePreviewPanelTextSavingTests (3 → 0) and FilePreviewReviewFeedbackTests (1 → 0).

- **Dependencies**
  - Requires the window-release guard from #8701 so these suites can reach the assertions.

<sup>Written for commit b372703acc5c67a34dad72ebd92b0c8c28beb85e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8719?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Updated file preview editor coverage to use the standard file preview text view configuration.
  * Preserved verification for saving shortcuts, window movement, theming, rendering, and text focus behavior.

### DIFF
--- a/cmuxTests/WindowAndDragTests.swift
+++ b/cmuxTests/WindowAndDragTests.swift
@@ -3113,7 +3113,7 @@ final class FilePreviewPanelTextSavingTests: XCTestCase {
         defer { panel.close() }
         await panel.loadTextContent().value
 
-        let textView = SavingTextView()
+        let textView = SavingTextView.makeFilePreviewTextView()
         textView.string = "saved by configured shortcut"
         textView.panel = panel
         panel.attachTextView(textView)
@@ -3152,7 +3152,7 @@ final class FilePreviewPanelTextSavingTests: XCTestCase {
         defer { panel.close() }
         await panel.loadTextContent().value
 
-        let textView = SavingTextView()
+        let textView = SavingTextView.makeFilePreviewTextView()
         textView.string = "should not save through command s"
         textView.panel = panel
         panel.attachTextView(textView)
@@ -3257,7 +3257,7 @@ final class FilePreviewPanelTextSavingTests: XCTestCase {
 
     func testTextEditorInsetsReapplyWhenMovedBetweenWindows() {
         _ = NSApplication.shared
-        let textView = SavingTextView()
+        let textView = SavingTextView.makeFilePreviewTextView()
         textView.textContainerInset = .zero
         textView.textContainer?.lineFragmentPadding = 5
 
@@ -3282,7 +3282,7 @@ final class FilePreviewPanelTextSavingTests: XCTestCase {
     func testTextEditorClearThemeDoesNotDrawAppKitBackgrounds() {
         _ = NSApplication.shared
         let scrollView = NSScrollView()
-        let textView = SavingTextView()
+        let textView = SavingTextView.makeFilePreviewTextView()
         scrollView.documentView = textView
 
         FilePreviewTextEditor<FilePreviewPanel>.applyTheme(
@@ -3305,7 +3305,7 @@ final class FilePreviewPanelTextSavingTests: XCTestCase {
     func testTextEditorOpaqueThemeDrawsAppKitBackgrounds() {
         _ = NSApplication.shared
         let scrollView = NSScrollView()
-        let textView = SavingTextView()
+        let textView = SavingTextView.makeFilePreviewTextView()
         let backgroundColor = NSColor(srgbRed: 0.12, green: 0.14, blue: 0.16, alpha: 1)
         scrollView.documentView = textView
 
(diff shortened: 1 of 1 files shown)


ejc3:
Measured, and this one was a stated-in-advance prediction. The three failing tests build a bare `SavingTextView()`, which is TextKit 2: `.textContainer` is nil and `.string` does not take, so the save reads empty and truncates the file, and the insets assertion reads `nil` where it expects `Optional(0.0)`. Production only ever builds the view through `makeFilePreviewTextView()`, an explicit TextKit 1 stack.

All three failed deterministically when run alone on an idle host, which ruled out the entire ordering-and-load family. I then wrote down that applying this branch would turn both suites green, and measured it on the combined tree:

```
FilePreviewPanelTextSavingTests    27 tests, 0 failures, 0 restarts   GREEN
FilePreviewReviewFeedbackTests     17 tests, 0 failures, 0 restarts   GREEN
```

Cannot be compiled in isolation until #8857 lands, like every open branch.

austinywang:
Addressing CodeRabbit’s docstring-coverage warning: the current PR diff introduces no functions, types, or other declarations. It only replaces six existing test fixture construction expressions with the product factory. Adding docstrings to unrelated pre-existing test methods would be out-of-scope churn, so no code change is appropriate for that generic warning. CodeRabbit’s 24 cmux-specific policy checks all pass.

PR #8724, opened 2026-07-23
Give the PR refresh run-loop test something real to observe

ejc3:
## Problem

`testPullRequestRefreshRepositoryDiscoveryDoesNotBlockMainRunLoop` used to observe repository discovery by counting calls to a stubbed `git remote -v` subprocess. That stopped being a real signal when #2797 moved remote discovery to in-process config parsing. The counter stayed at zero, and the remaining assertions could pass without proving that discovery had run.

## Final approach

- Add `GitRepositoryDiscovering` for the two filesystem reads that `PullRequestProbeService` already performs: resolving GitHub slugs and the checked-out branch.
- Keep `GitMetadataService` as the only production implementation.
- Widen `PullRequestProbeService` and `PullRequestPollService` from the concrete metadata service to that protocol.
- Exercise the real polling path in the test by constructing `PullRequestPollService` directly with a real `TabManager` host and a test discovery implementation. There is no `TabManager` test override and no test-only seam in app source.
- Make the run-loop proof deterministic. Discovery signals that it started and blocks on a gate; the test queues a main-run-loop turn and releases discovery only after both signals arrive. If discovery occupies the main thread, the five-second cleanup deadline opens the gate and records an explicit failure. The deadline is deadlock cleanup, not the success criterion.

The protocol and its requirements have API documentation. No user-facing strings, localization data, package lockfiles, Xcode project wiring, or Swift warning budgets change.

## Runtime impact

Production behavior is unchanged. The concrete dependency types become protocol existentials, while the app continues to pass the same `GitMetadataService` instance through the same polling and probing pipeline.

## Verification

- Targeted GitHub Actions run: https://github.com/manaflow-ai/cmux/actions/runs/30895230698
- Latest-main targeted rerun: https://github.com/manaflow-ai/cmux/actions/runs/30896696485
- `lint-pbxproj-test-wiring`: passed (651 test files checked)
- `check-package-resolved-policy.py`: passed
- Edited Swift files: parse successfully
- `git diff --check`: passed
- Swift warning-budget diff: empty

Both targeted Actions runs execute `WorkspacePullRequestSidebarTests` and pass.


austinywang:
Review resolution on the final tree:

- CodeRabbit’s production test-seam finding is fixed. `Sources/TabManager.swift` is no longer changed; the regression test constructs `PullRequestPollService` directly and attaches the real `TabManager` host.
- The earlier sleep/timer-based run-loop check is replaced by a deterministic blocking gate. Discovery must start and a queued main-run-loop turn must execute before the test releases discovery. The five-second deadline exists only to prevent a deadlock and records a test failure if reached.
- The docstring warning is addressed: `GitRepositoryDiscovering` and both public requirements are documented, as are the widened poll-service initializer parameter and existing probe API.
- Greptile’s summary reported no actionable defect, but it describes the superseded `TabManager` override and sleep-based test; neither exists in the final diff.
- The three Cursor comments are spend-limit notices and contain no code feedback to address.

Verification on the final implementation: targeted runs https://github.com/manaflow-ai/cmux/actions/runs/30895230698 and https://github.com/manaflow-ai/cmux/actions/runs/30896696485 both passed; test wiring, package-lock policy, Swift parsing, `git diff --check`, and the unchanged warning budget were also checked.

PR #8751, opened 2026-07-23
Move active surface between existing panes

### LINKED ISSUE #8752: Move the active surface between panes and create missing directional splits
## Problem or motivation

cmux supports multiple panes, with each pane containing one or more surfaces such as terminals and browsers. Surfaces can be reordered within a pane, and pane focus can be moved with configurable shortcuts, but there is no keyboard-driven way to move the active surface into another pane or create a destination pane when the requested direction is empty.

Today this requires drag and drop plus manual split management. That interrupts keyboard-only workflows and is cumbersome when panes are small or contain several surfaces.

## Proposed solution

Add configurable actions for moving the active surface:

- Move to the previous pane
- Move to the next pane
- Move to the pane on the left
- Move to the pane on the right
- Move to the pane above
- Move to the pane below

Previous/next follows spatial pane order and wraps across existing panes. Directional movement uses the existing pane-adjacency behavior when a destination exists. If that direction has no destination, split the source pane in the requested direction and move the active surface into the new pane.

Expose these actions through:

- Settings → Keyboard Shortcuts
- Command Palette
- View menu

The actions can initially be unbound so they do not conflict with existing pane-focus shortcuts.

## Expected behavior

When a surface is moved:

- Preserve the existing terminal or browser instance and its state.
- Select and focus it in the destination pane.
- When using an existing destination, insert it after that pane's selected surface.
- When creating a destination, split 50/50 along the requested axis.
- Split only the source pane's region, preserving its parent extent on the orthogonal axis:
  - up/down preserves the parent column width;
  - left/right preserves the parent row height.
- If the moved surface was the source pane's only surface, keep the source valid with a real replacement terminal rather than an unmapped placeholder.
- Safely reject unsupported layouts such as Canvas and remote-tmux mirrors.
- Do not move a background workspace surface while Dock owns keyboard focus.

## Alternatives considered

- Continue relying on drag and drop.
- Refuse directional movement when no pane exists, but this makes keyboard-only layout construction incomplete.
- Reconstruct the moved terminal or browser, but that risks losing live surface state.
- Reuse the within-pane reorder actions, but those do not transfer ownership between panes.

## Additional context

Current draft implementation: #8845

The implementation reuses the existing same-workspace surface transfer path for existing destinations and Bonsplit's moving-tab split path for missing directional destinations.

kyl33r:
## Summary

- Add configurable commands for moving the active surface to the previous/next existing pane or to the pane on the left, right, above, or below.
- Route keyboard shortcuts, Command Palette actions, and View-menu commands through `TabManager` into the existing `Workspace.moveSurface` ownership-transfer path.
- Preserve the surface instance and focus, insert after the destination pane's selected surface, wrap previous/next traversal, and let Bonsplit collapse an emptied source pane.
- Reuse the same directional helper for existing tab-context-menu moves and clarify the old in-pane actions as “Reorder Surface Left/Right”.

All six new shortcut actions intentionally start unbound because the proposed Control-Command-arrow chords already belong to pane-focus actions. The actions remain discoverable and configurable in Settings and the Command Palette.

## Scope

This is the focused existing-pane interaction slice from the movement proposal. It does not create panes or consolidate cross-workspace/window transfer.

## Verification

- [x] `./scripts/test-unit.sh -skipPackageUpdates test -only-testing:cmuxTests/MoveSurfaceBetweenPanesTests` — 10 tests in 1 suite passed
- [x] `./scripts/reload.sh --tag move-pane --launch` — tagged Debug build succeeded and launched
- [x] `CMUX_TAG=move-pane scripts/cmux-debug-cli.sh list-workspaces` — tagged socket responded
- [x] `swift build --package-path Packages/macOS/CmuxSettings`
- [x] `scripts/lint-pbxproj-test-wiring.sh`
- [x] `scripts/check-pbxproj.sh`
- [x] JSON validation for `Resources/Localizable.xcstrings`
- [x] `git diff --check`
- [x] `swiftc -parse` over changed Swift sources and the focused test file
- [ ] Complete live manual verification of terminal/browser state, pane wrapping, directional boundaries, focus, insertion order, and source-pane collapse

## Test coverage added

- terminal and browser object identity/focus preservation
- insertion after the destination selected surface
- empty-source-pane collapse
- left/right/up/down movement and previous/next wrapping
- no-op behavior for missing destinations, Canvas layout, and remote-tmux mirrors
- settings/default/Command Palette metadata alignment
- physical-arrow shortcut routing through the production event handler

Closes #8752

### DIFF
--- a/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/ShortcutAction+Defaults.swift
+++ b/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/ShortcutAction+Defaults.swift
@@ -111,6 +111,10 @@ extension ShortcutAction {
         case .prevSurface: return ShortcutStroke(key: "[", command: true, shift: true)
         case .moveSurfaceLeft: return ShortcutStroke(key: "[", command: true, shift: true, option: true)
         case .moveSurfaceRight: return ShortcutStroke(key: "]", command: true, shift: true, option: true)
+        case .moveSurfaceToPreviousPane, .moveSurfaceToNextPane,
+             .moveSurfaceToPaneLeft, .moveSurfaceToPaneRight,
+             .moveSurfaceToPaneUp, .moveSurfaceToPaneDown:
+            return nil
         case .selectSurfaceByNumber: return ShortcutStroke(key: "1", control: true)
         case .selectWorkspaceByNumber: return ShortcutStroke(key: "1", command: true)
         case .moveWorkspaceUp: return ShortcutStroke(key: "[", command: true, option: true, control: true)
--- a/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/ShortcutAction.swift
+++ b/Packages/macOS/CmuxSettings/Sources/CmuxSettings/Values/ShortcutAction.swift
@@ -48,6 +48,14 @@ public enum ShortcutAction: String, CaseIterable, Sendable, Hashable, SettingCod
     case moveSurfaceLeft
     /// Moves the selected surface one position right.
     case moveSurfaceRight
+    /// Moves the selected surface to the previous pane in spatial order.
+    case moveSurfaceToPreviousPane
+    /// Moves the selected surface to the next pane in spatial order.
+    case moveSurfaceToNextPane
+    case moveSurfaceToPaneLeft
+    case moveSurfaceToPaneRight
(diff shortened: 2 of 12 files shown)


kyl33r:
Superseded by #8845 after renaming the head branch to `feat/8752-move-surface-between-panes` as requested.

PR #8689, opened 2026-07-22
tests: stop the portal first-reveal fixture from killing the test host

ejc3:
`BrowserPortalFirstRevealScrollTests` declares 16 tests. Run alone on a macOS host it completed 10 of them and restarted the app host three times, so the suite produced no verdict, and any suite sharing that host lost its verdict too.

`makeWindowFixture` builds an `NSWindow` and three tests later call `close()` on it. AppKit releases a window when it closes unless the owner opts out, while ARC still holds a strong reference, so each of those closes over-releases and takes the process down.

The arithmetic is what convinced me this is the whole cause rather than a contributing one: exactly three tests call `close()`, and there were exactly three restarts. The same count holds across the neighbouring browser suites, where the number of tests that close a self-created unguarded window matches the measured restarts at 16, 3, 1 and 1, and a suite with none of them fails assertions without ever restarting. The one test in this file that builds its own window already sets the flag, so the shared fixture simply missed it.

The product side already does this everywhere it owns a window (`BrowserPanel.swift:4347`, `BrowserPrewarmedWebViewPool.swift:181`, `BrowserPopupWindowController.swift:115`, `ReleasingWindowController.swift:62`). Only this fixture was missing it, which is why the fix is one line in the test rather than anything in `Sources`.

## Before and after

Both arms ran on the same worktree with the same warm derived-data path, one suite per app host, with only this change between them. `#8651` is applied in both, because the `cmuxTests` target does not compile without it.

| | host restarts | tests ran | verdict |
|---|---|---|---|
| before | 3 | 10 of 16 | `** TEST FAILED **` |
| after | 0 | 16 of 16 | `** TEST SUCCEEDED **` |

```
scripts/ci/run-in-console-session.sh scripts/ci/run-app-host-xcodebuild.sh \
  -project cmux.xcodeproj -scheme cmux-unit -configuration Debug \
  -destination 'platform=macOS' \
  -only-testing:cmuxTests/BrowserPortalFirstRevealScrollTests \
  CMUX_SKIP_ZIG_BUILD=1 TEST_RUNNER_SWIFT_BACKTRACE=enable=no test
```

Read the restart count and the test count rather than the verdict line when checking this. A host that dies mid-run still prints a summary, and this suite's failing run also emits `Executed 0 tests, with 0 failures` earlier in the log.

---
[...]

PR #8755, opened 2026-07-23
Create test windows in one place, and lint that nothing else does

ejc3:
Test windows are now created in one place. `cmuxTests/TestWindowFactory.swift` is the only file in the test target that constructs an `NSWindow`, and `scripts/lint-test-window-construction.py` fails any other file that does. Everything else calls `TestWindow.make()` or `TestWindow.hosting(view)`.

The reason it needs centralizing: a window created in code has `isReleasedWhenClosed = true`, so `close()` defers a release into the autorelease pool while ARC also releases the local reference at scope exit. Two releases against one retain — the pool drain touches freed memory and the test host segfaults in `objc_release` under `objc_autoreleasePoolPop`. That kills the process instead of failing a test, and xcodebuild's summary then covers only the last host launch, so verdicts pending in the dead host go missing rather than red. Run 29684286662 on `main` finished green with 35 of these host deaths across its four app-host jobs (24, 2, 4 and 5 restart lines).

The factory sets `isReleasedWhenClosed = false`, which makes ARC the only owner, so the window deallocs once at scope exit. That is a one-line fix that has to be remembered at every window, which is why it kept being missed — the point of the factory is that there is nothing left to remember.

The lint checks two things. The source, which catches `NSWindow()`, since the argument-less form emits the selector `init` and that is too generic to look for in a binary. And the compiled object files, which catch a construction reached through an alias, a subclass, or a factory in another file, because the compiler has already resolved all of those and a construction reduces to whether the object file references `initWithContentRect:styleMask:backing:defer:`. Neither half asks whether a window is closed or whether it sets the flag, only whether a file constructs one.

94 files predate the factory and are listed by path. That list can shrink and never grow, so nothing new can construct a window while those migrate. Migrating them is a follow-up.

Those 94 files still need the flag in the meantime, so it is set at 83 places across 34 of them.

14 of those windows are also shown on screen with `makeKeyAndOrderFront`, and those get a second line, `animationBehavior = .none`. Showing a window makes AppKit run a show animation, which is a separate object from the window and does drawing work on a background thread. `animationBehavior = .none` means AppKit never creates it. That is a different problem from the crash — a leaked animation hangs a suite rather than killing the host — and it is in here because the crash report I started from named `_NSWindowTransformAnimation`, which is that object. `trackTestWindow` at `cmuxTests/TerminalAndGhosttyTests.swift:4817` already sets both properties, and its comment says leaked ones "commit CA transactions concurrently with the main thread's display cycle for the rest of the run". I am taking that comment's word for it rather than having reproduced the hang.

`RemoteTmuxRectPublicationTests` force-unwrapped the published layout and the observed pane in nine places. Each of those traps exactly when the expectation around it is already failing, so a reportable failure became a dead process. They record the nil and continue now.

The CI gate for the unit shards fails on the runner's restart line, on a shard that executed no tests, on a crash report that appeared during the run, and on the crash scan itself failing. Crash reports are attributed by listing the directory before the run and subtracting afterwards, then copying the matches somewhere macOS will not prune them. The list of tests that started and never reported now matches XCTest as well as swift-testing — it only matched swift-testing before, so it printed nothing for most shards, which reads as "nothing was lost".

## Prior fixes for this

This defect has been fixed before, repeatedly, and none of those fixes has landed:

- **#7768** — same root cause, pinned with lldb and NSZombie, and fixed at every site at once with a test-bundle constructor that swizzles NSWindow's designated initializers. Counts 269 of 299 window creations across 61 files missing the guard. Open since 2026-07-09.
- **#8477** — the DevTools window suite, 21 sites.
- **#8701** — three fixtures, 3 sites.
- **#8689** — the portal first-reveal fixture.
- **#8635** — the shortcut routing suite.
- **#8614** — twenty suites, one of which was taking the host down.
- **#8633** — a fake WKNavigation killing the host, a different cause with the same effect.

#7768 covers the same ground as the per-site half of this PR and does it with one hook instead of 83 edits. If it lands first, the 83 flag lines here become redundant and should be dropped; the factory, the lint, the CI gate and the nine unwrap fixes are independent of it. It is worth landing #7768 first for that reason.

## Why this is safe

It is all test code apart from the workflow and the two scripts.

Clearing the flag does not leak. ARC still owns the reference, so the window deallocs at scope exit instead of being released twice. Production already does this at `Sources/App/ReleasingWindowController.swift:62`, and `cmuxTests/SettingsWindowPresenterTests.swift:44` asserts it.

No test asserts that `close()` releases a window. The three in `WindowZombieRegressionTests` that assert a window deallocs use windows whose flag is already false, so they are unaffected.

`TestWindow` is a function rather than an `NSWindow` subclass. A subclass inherits the designated initializer, so every call site would emit that selector into its own object file and the object-file check could not tell a call site from a construction.

## Testing

Both scripts self-test against inputs whose answers are known ahead of time, and both self-tests run in CI. The construction check covers `NSWindow()`, `NSWindow.init(`, a typealias, a subclass, a commented-out construction, a cast, and a factory call. The crash scan covers a non-cmux crash in the same directory, a truncated report, a pre-existing crash that must not be attributed to the run, and a missing snapshot.

I confirmed each one fails before it passes. Removing the two flags this branch adds to the rename suite makes that check name those two lines and exit 1. Adding a file that constructs a window makes the construction check exit 1, and 0 once it is gone. The crash scan finds the real report from a run whose host died when given a snapshot from before it, and ignores it given one from after.

`RemoteTmuxSessionRenameTitleTests` is migrated to the factory as the first case. Its object file referenced the designated initializer before and does not now, and the factory is the only file in the target that does.

`TEST BUILD SUCCEEDED`, 0 errors. Running the suite whose crash report started this plus five others this branch edits: 68 verdicts reported, no crash report attributed to the run, one host death.

## Still broken

That host death is a different bug. `MarkdownPanelTests.testOpenMarkdownPanelReloadsWhenFileChangesOnDisk` starts and never reports, here and on `main` — in run 29684286662, job 88190103129 shows it starting once and finishing zero times, and losing five `MarkdownPanelTests` verdicts in total. None of the four windows this branch touches in that file belongs to that test.

`scripts/lint-test-window-release.py` records 210 windows that are built and never closed. They cannot double-release, so they are recorded rather than fixed, but an ordered-front window that is never closed still leaks its appearance animation.

The shard-packing comment in `scripts/ci/cmux_unit_test_shard.py` said one of these suites reliably crash-restarts the host. That is no longer true, so the comment is corrected, but the separation stays — that suite drives real WebKit inspector attach and detach and may still starve a live-navigation suite sharing its shard. Removing it needs a timing measurement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of macOS automated tests involving window creation, closing, and teardown.
  * Reduced test instability from window lifecycle and animation behavior.

* **Tests**
  * Added safeguards detecting crashes, host restarts, zero-test shards, and unsupported test window construction.
  * Added reusable test-window setup and expanded validation for window lifetime safety.
  * Added test evidence uploads, crash-report scanning, and related self-checks for CI diagnostics.


ejc3:
Closing in favour of two focused PRs.

#8832 fixes the window over-release itself — @lawrencecchen's #7768 rebased onto current main. One constructor in the test bundle swizzles NSWindow's designated initializers, so every test window defaults to `releasedWhenClosed = NO`, including subclasses and any written later. Verified on current main: `BrowserDeveloperToolsVisibilityPersistenceTests` now runs with 0 host restarts and no crash report. That makes the 83 per-site flag edits here redundant, along with the factory and the two lints built to enforce them.

#8833 keeps the part that has nothing to do with windows: a unit shard now fails when a host dies, when it executes no tests, when a crash report appears during the run, or when the crash scan itself fails. It also fixes the never-reported list, which matched only swift-testing's output shape and so printed nothing for most shards, and the nine force-unwraps in `RemoteTmuxRectPublicationTests` that trap instead of reporting.

Nothing here is lost — it is split between those two.

PR #8758, opened 2026-07-23
Perf: memoize custom-sidebar data context, cache approval-store reads on autosave hot path

luigiwidmore:
## Summary

Two independent main-thread performance fixes, both profiled on a live app (macOS, 13 workspaces / 29 surfaces, custom sidebar selected with `customSidebars.renderer = "remote"`, `sample(1)` during workspace switches):

1. **Custom-sidebar data context was rebuilt redundantly on the main thread.** `customSidebarDataContext(now:)` runs inside the `TimelineView(.periodic(by: 1))` content closure (`Sources/ContentView.swift`), so it re-executed on **every** `tabManager.objectWillChange` — a workspace switch fires several (selection willSet, window-title update, focus, unread) — **and** every 1 s tick. Each run looped all workspaces, and each workspace walked the bonsplit tree **four** separate times. This is the same unmemoized-snapshot cost class `8443096d` fixed for the default sidebar rows; the custom-sidebar path never got it. Measured: **~1 s of main-thread work per workspace switch**. The only cache that existed (`CustomSidebarPaneDataContextCache`) embedded the wall-clock second in its key, so it expired every second by design.
2. **Session autosave did Keychain + disk + per-record HMAC per terminal panel, on main.** The 8 s autosave builds a full session snapshot on the main queue; per terminal panel, `surfaceResumeStartupInput()` → `SurfaceResumeApprovalStore.matchingRecord()` → `validRecords()` did a `SecItemCopyMatching` (`defaultSigningSecret`), re-read + re-decoded the store file, and re-HMAC-verified every record (`hasValidSignature` rebuilds the base64 signing payload per record). Measured: **~0.6 s of main-thread time in this path over 25 s** of light use — and beyond the cost, each `SecItemCopyMatching` on main is a hard-hang vector if securityd wedges or an ACL prompt never presents.

## Changes (3 commits, the middle two follow the regression-test policy)

- **`d81091b8`** — failing regression tests for approval-record caching (red without the fix: cache hit proven by making the file unreadable after priming; reload on direct on-disk rewrite; invalidation on in-process write).
- **`d3603831`** — `SurfaceResumeApprovalStore` gains an NSLock-guarded cache: each `validRecords` call re-stats the file and only reloads + re-verifies when mtime/size changed, the effective signing secret changed, or `didChangeNotification` fired (every in-process write path posts it). The env-var secret is still re-read per call (tests rotate it); the Keychain/generated/file-backed resolution is memoized per process, nil results never cached. Result: per-panel cost drops from Keychain + disk + N HMACs to a stat.
- **`2860b8c9`** — new `CustomSidebarDataContextStore` (replaces `CustomSidebarPaneDataContextCache`): a two-tier memo where the expensive per-workspace interpreter input rebuilds only after invalidation, and the 1 Hz tick refreshes only the cheap clock value. Rebuilds are driven by a merged, 40 ms-coalesced stream (tabManager structure/selection, unread, per-workspace immediate+deep observation publishers, `paneLayoutVersion`, panel titles pre-coalesced at 500 ms, focus broadcasts) — the same pattern the extension sidebar already uses. A switch now costs **one** rebuild instead of ~5. Additionally, `customSidebarWorkspaceSnapshot` now walks the bonsplit tree **once** per workspace per build (was 4×). The pane variant (`CustomSidebarPanelView`) gets the identical treatment. The emitted context stays value-equal when content is unchanged, so `SwiftRenderTrigger` re-renders exactly as often as before.

## Validation status — please read

**I could not compile or run tests locally**: the authoring machine has no Xcode (CLT only, Swift 5.10 vs required 6.0) and no populated submodules. All changed files pass `swiftc -parse`, repo guards pass (`scripts/check-pbxproj.sh`, `scripts/lint-pbxproj-test-wiring.sh`, `scripts/swift_file_length_budget.py`), and every API used was cross-checked against existing call sites — but **CI is the first real build**. The regression-test commit should go red on CI before `d3603831` turns it green; if anything else fails I'm happy to iterate.

## Behavior trade-offs

- Custom-sidebar updates for workspace-local state now ride the coalesced streams: isolated changes remain effectively immediate (leading edge), bursts settle within ~40 ms, sustained panel-title churn caps at ~2 rebuilds/s. There is no longer a 1 Hz catch-all rebuild, so the invalidation trigger set is deliberately broad (it covers every field the snapshot projects); worth watching while dogfooding.
- Session autosave still runs on the main queue every 8 s; only the repeated approval lookups get cheaper (first read after an actual store change still pays full cost).

## Notes

- Follow-up hardening ideas (out of scope here): cache `ParsedProgram` in `InProcessSidebarInterpreter` (the remote worker's `RenderInterpreterRunner` already does), and add `com.cmux.sidebar` signposts around `customSidebarDataContext` / the autosave snapshot build — today nothing instruments them.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8758"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787429326&installation_model_id=21920&pr_number=8758&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8758&signature=a9865f18e5dfa6a5f67625704c7621b1927439bc48e1fe04dc7ff5750351f445"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Improves main-thread performance by memoizing the custom sidebar data context and caching surface-resume approval reads. Workspace switches trigger a single rebuild instead of many, and autosave per panel now costs a file stat instead of Keychain + disk + HMAC work.

- **Performance**
  - Custom sidebar: added a two-tier `CustomSidebarDataContextStore` with a 40 ms coalesced invalidation stream (panel titles coalesced at 500 ms). The 1 Hz tick only updates the clock; rebuilds happen on invalidation. Replaces the per-second cache and cuts redundant rebuilds during switches.
  - ContentView and custom sidebar panel now subscribe to the coalesced invalidation publisher; context is rebuilt once per switch. Snapshot reduces bonsplit walks to one per workspace per build and shares results across branches/PRs/surfaces.
  - Session autosave: `SurfaceResumeApprovalStore.validRecords` uses an NSLock-guarded cache keyed by file mtime/size + signing secret and invalidated by store writes. `defaultSigningSecret` memoizes the non-env secret for the process.

- **Tests**
  - Added regression tests for approval-record caching: serves from cache when the file becomes unreadable, reloads on direct on-disk rewrite, and invalidates on in-process writes.

<sup>Written for commit 2860b8c9876b3aca24f0a0045deb06b5d200a1ae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8758?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luigiwidmore:
Related issue for the hard-freeze (not addressed here — this PR covers the switch *lag*, not the 8-minute wedge): #8759

PR #8828, opened 2026-07-24
Update Ghostty surface config ABI lock

lawrencecchen:
PR https://github.com/manaflow-ai/cmux/pull/8717 advanced the pinned Ghostty surface config from 120 to 168 bytes, including the Windows-capable platform union and initial PTY tee callback fields. The exact ABI lock retained the previous layout and fails the speculative merge gate.

This updates every locked offset to the pinned GhosttyKit artifact and adds assertions for pty_tee_cb and pty_tee_userdata. Current main already contains the separate redundant-await warning cleanup through https://github.com/manaflow-ai/cmux/pull/8688, so this PR now has a one-file test-only diff.

Red evidence:
- https://github.com/manaflow-ai/cmux/actions/runs/30070795839/job/89411543273#step:14:1

Verification:
- Local Swift Testing ABI assertion: 1 test passed
- Prior speculative run passed every job: https://github.com/manaflow-ai/cmux/actions/runs/30072476283
- A fresh speculative run targets current main
- No runtime or user-facing behavior changes


lawrencecchen:
Superseded by https://github.com/manaflow-ai/cmux/pull/8831, which landed the same Ghostty ABI lock update and redundant-await cleanup on main.

PR #8836, opened 2026-07-24
cmuxTests: repair three red suites, one of which was killing its own test host

ejc3:
Three unrelated red suites, one commit each. All three are test-side defects; no product code changes.

Measured on `main` at `b5fac3145f`, one suite per app host, base arm and fix arm on the same base.

| suite | before | after |
|---|---|---|
| `CommandPaletteSearchEngineTests` | 2 failures, 65 tests | **0 failures**, 65 tests |
| `WorkspaceRemoteConnectionTests` | 10 failures, **38** tests executed, 2 host restarts | **0 failures**, all **83** executed, **0** restarts |
| `CMUXCLIErrorOutputRegressionTests` | 6 failing tests | 2 failing, 47 run |

## The remote-connection suite was hiding most of itself

`testPersistentReverseRelayCancelsStaleControlMasterForwardBeforeReusingRelayPort` asserts
`operations.count >= 2` and then reads `operations[0]`. A count assertion does not stop execution, so
when it fails the next line traps: `Fatal error: Index out of range`. That takes down the shared app
host rather than failing one test, and every verdict still pending in that host is absent from the
summary instead of red. It happened twice in one run. Both sites now unwrap instead of subscripting,
which is why the executed count goes from 38 to 83 — the back half of the suite had not been running
at all.

Four `@MainActor` tests waited on a `DispatchSemaphore`. `configureRemoteConnection` enqueues its
session transition as a serialized main-actor `Task`, so blocking the main actor stopped the very work
being waited on from ever being scheduled. They now use expectations, which pump the run loop. One of
those four was not in the failing set I started from and turned out to be failing too.

Fifteen fixtures passed an unresolved `%C` control path. `NativeSSHControlMasterKey` deliberately
refuses to claim lifecycle ownership of a path it cannot resolve, so no lease was taken and cleanup
could never run — and six inverted expectations were passing vacuously as a consequence, since they
assert that cleanup does *not* happen. They now use the resolved 40-hex form `ssh -G` produces, and a
new test pins the unowned-template policy so the fixtures cannot quietly regress to a template and take
the coverage with them.

Two more read `activeRemoteSessionControllerID` immediately after `configureRemoteConnection` and now
await the transition.

## The palette assertions were asserting a gate that no longer exists

#8173 replaced the fork-probe reuse gate: `!cachedResultHadFallback` became `cachedResultIsFresh`, and
the fallback case is now re-verified against `SharedLiveAgentIndex` at the call site rather than being
refused outright. The parameter stayed in both helper signatures, so two assertions still compiled while
asserting the opposite of what the product does. `WorkspaceForkConversationContextMenuTests` asserts the
new contract in both directions, so removing them loses no coverage; I ran that suite to confirm the
change does not flip it.

Removing the now-dead `cachedResultHadFallback` parameter would have turned this into a compile error at
review time instead of two silently wrong assertions. That is worth doing and is not in this PR.

## The CLI fixture broke its own invariant

The CLI derives a theme reload target from the socket file name, collapsing runs of non-alphanumerics in
the slug to dots. #6452 gave this fixture a unique socket path with a raw `UUID` to stop two runs
colliding in `/tmp`, which put the UUID's dashes into the derived identifier as dots — so the expected
literal could no longer match and the test spent its five-second budget waiting. The `stdout` assertion
kept passing because the derived id still has the expected value as a prefix, which is why this looked
like a timeout rather than a mismatch. The unique suffix is now hex-only, and the expected identifier
stays a plain template rather than a call into the CLI's own helper, which would agree by construction.

## Two commits added since the measurements above

The remote-connection suite reached zero because two of its tests were waiting on a transport that no
longer exists. #8434 moved the daemon upload off `scp` onto the ssh exec channel, streaming the binary
into `cat >`, and did not touch these tests — whose stubs only fulfilled inside an
`executable == "/usr/bin/scp"` branch. Both now capture the upload from the ssh branch, and the scp
branch is kept as a loud failure so a transport regression cannot present as a timeout again. The
reinstall test also records how many capability hellos preceded the upload and requires at least one,
because retargeting alone would have let it pass on a plain first install.

Three CLI fixtures were also wrong. Two replied to the CLI with a bareword: `SocketClient` accepts only
`OK`, `OK …`, `PONG`, `ERROR: …` or JSON as a complete single-line reply, so a bareword sends it into
the multiline drain pass, where reconfiguring the receive timeout on an already-closed socket fails
with EINVAL and the CLI prints "Invalid argument" instead of the reply it had buffered. They were the
only two barewords in the suite. Both now also assert which responder received the request, which is
the property they exist for and cannot be made vacuous by a later change to the reply.

## What is still red, and why I am not claiming otherwise

`CMUXCLIErrorOutputRegressionTests` runs 47 tests with 2 still failing, down from 6.

`testCLIErrorPathDoesNotCrashWhenStderrIsClosed` asserts a CLI that no longer exists: since #f48922aa94
an unknown command exits 2 with a one-line message and no usage dump, and that message goes to stderr,
which the test closes. The crash it was written to catch is still real, so the fix is to keep the guard
and drop the stale expectations, which needs `terminationReason` plumbed through the runner rather than
a fixture edit.

`testSessionsListDoesNotInferPiFamilyFromBasenameWhenStructuredIdentityDisagrees` passed `agent:
"project-agent"`, which is not in the CLI's catalog, so the command errored before emitting JSON and the
test had never once run its assertions. Switching to a catalog agent got it running for the first time,
and the assertions then failed with `fork_unavailable_reason = "agent_has_no_fork_command"` — the
fixture also needs an agent that has a fork command, which the one I picked does not. Still red, now for
a reason the failure output states plainly.

Two further things this suite needs that are not fixture-sized, and are not in this PR. Its shared
runner merges stderr into the stdout pipe that around thirty tests then parse as JSON, so one stray
diagnostic line fails a content test; and it reads that pipe only after waiting, so a full buffer
presents as a timeout. Separately, `CFFIXED_USER_HOME` does not isolate socket resolution, because
`readLastSocketPath` also consults the machine-global `/tmp/cmux-last-socket-path` — measured, a CLI
spawned with a pristine temp home and a scrubbed environment still reached a live running app. Three
tests also bind `/tmp/cmux.sock`, and the responder unlinks a path before binding it.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8836"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787469216&installation_model_id=21920&pr_number=8836&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8836&signature=33eabfa1bd8762c479c023323b3bef8d3e2e8986648b610fa8ba9ab961814d19"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes three red test suites and stops the remote‑connection tests from killing their host. Restores full execution of `WorkspaceRemoteConnectionTests` (38→83, 0 restarts), makes `CommandPaletteSearchEngineTests` green, and hardens CLI regression tests; no product code changes.

- **Bug Fixes**
  - Remote-connection tests: replace semaphores with expectations; await session transitions; use `XCTUnwrap` instead of indexing; switch fixtures to a resolved 40‑hex `ControlPath` and add a test asserting unowned `%C` templates request no cleanup; retarget daemon‑upload to the ssh exec channel (`cat >`), assert absolute HOME paths and destination host, fail if `scp` is used, and require a capability hello before reinstall (test renamed).
  - Command palette: remove obsolete `cachedResultHadFallback` assertions and rename the clear test to reflect stale‑cache behavior; freshness‑gate coverage lives in `WorkspaceForkConversationContextMenuTests`.
  - CLI: isolate spawns from the machine’s cmux (pin `CMUX_SOCKET_PATH` where not testing resolution, write per‑home last‑socket marker, avoid binding `/tmp/cmux.sock`); capture stderr on its own pipe with background drains and add a 60‑second hang guard; fix socket‑selection tests to use OK‑framed replies and assert which socket handled the request; sessions list diagnostics now require a completed, zero‑exit run and use catalog agent `opencode` with `omo`; the stderr‑closed guard runs under `exec`, pins `CMUX_SOCKET_PATH` and HOME, asserts exit 1, and verifies no signal; theme reload tests derive the target from a dash‑free socket suffix and scope bundle IDs (and nightly sockets) per run.

<sup>Written for commit 06cd24c4ac71b44186224284789cdb9b05771567. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8836?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Updated CLI socket-resolution regression fixtures and expectations for stricter reply framing; added deterministic environment setup and clearer assertions about which responder handled requests.
  * Strengthened command palette search engine coverage by separating stale-cache invalidation from reuse/clearing behavior, improving forkable-agent probe scenarios.
  * Improved sessions list diagnostics tests with hard failure checks, ensuring processes don’t time out and exit with the expected status.
  * Hardened remote connection tests with deterministic SSH ControlPath lifecycle handling and more reliable async synchronization; improved PTY/relay bootstrap assertions.
  * Enhanced CLI test diagnostics (stdout/stderr capture, signal classification) and added a consistent per-process hang guard timeout.

### DIFF
--- a/cmuxTests/CMUXCLIErrorOutputRegressionTests.swift
+++ b/cmuxTests/CMUXCLIErrorOutputRegressionTests.swift
@@ -72,11 +72,14 @@ import Testing
         // exited on its own terms rather than dying from a signal.
         XCTAssertFalse(result.timedOut, result.diagnostics)
         XCTAssertFalse(result.diedFromSignal, result.diagnostics)
-        // 2 is the unknown-command exit and it is deterministic now the socket is pinned. Asserting
-        // the exact code rather than merely non-zero keeps this from passing when the CLI fails for
-        // some unrelated reason. Nothing is asserted about stdout: the message goes to stderr, which
-        // this test deliberately closes.
-        XCTAssertEqual(result.status, 2, result.diagnostics)
+        // 1, measured, and deterministic for this fixture: the pinned socket has no listener, so the
+        // CLI fails at connect and the top-level handler exits 1 before the unknown-command arm (which
+        // would exit 2) is ever reached. That ordering is what makes this a good exercise of the guard
+        // rather than a weaker one — the connect error is written to the stderr this test has closed.
+        // Asserting the exact code rather than merely non-zero keeps the test from passing when the CLI
+        // fails for some unrelated reason. Nothing is asserted about stdout: every message on this path
+        // goes to the closed stderr.
+        XCTAssertEqual(result.status, 1, result.diagnostics)
     }
 
     @Test func testAgentTeamsHelpDoesNotLaunchExternalAgentCLI() throws {
--- a/cmuxTests/CMUXCLIErrorOutputRegressionTests.swift
+++ b/cmuxTests/CMUXCLIErrorOutputRegressionTests.swift
@@ -695,9 +695,15 @@ import Testing
         try fileManager.createDirectory(at: themesURL, withIntermediateDirectories: true)
         try writeTheme(named: "Theme A", background: "#101010", to: themesURL)
 
-        let socketPath = "/tmp/cmux-debug-active-theme-\(UUID().uuidString).sock"
+        // The reload target is derived from the socket file name, not from CMUX_BUNDLE_ID:
+        // `cmux-debug-<slug>.sock` becomes `com.cmuxterm.app.debug.<slug>`, where every run of
+        // non-alphanumerics in the slug collapses to a dot. A raw UUID here would put its dashes
+        // into the identifier as dots, so keep the unique part hex-only and the expected
+        // identifier stays a plain template rather than a call into the CLI's own helper.
+        let uniqueSuffix = UUID().uuidString.lowercased().replacingOccurrences(of: "-", with: "")
+        let socketPath = "/tmp/cmux-debug-active-theme-\(uniqueSuffix).sock"
         let staleBundleIdentifier = "com.cmuxterm.app.debug.stale.theme"
-        let targetBundleIdentifier = "com.cmuxterm.app.debug.active.theme"
+        let targetBundleIdentifier = "com.cmuxterm.app.debug.active.theme.\(uniqueSuffix)"
         let reloadExpectation = expectation(description: "cmux themes set targets the resolved socket bundle")
         let notificationQueue = OperationQueue()
         notificationQueue.maxConcurrentOperationCount = 1
--- a/cmuxTests/CMUXCLIErrorOutputRegressionTests.swift
+++ b/cmuxTests/CMUXCLIErrorOutputRegressionTests.swift
@@ -14,18 +14,53 @@ import Testing
         let status: Int32
         let stdout: String
         let timedOut: Bool
+        /// Defaulted so existing call sites are unaffected. A process killed by a signal reports
+        /// `.uncaughtSignal`, and its `terminationStatus` is the signal number — indistinguishable
+        /// from an ordinary non-zero exit if you only look at the status.
+        var terminationReason: Process.TerminationReason = .exit
+
+        var diedFromSignal: Bool { terminationReason == .uncaughtSignal }
+
+        var diagnostics: String {
+            "status=\(status) reason=\(diedFromSignal ? "uncaughtSignal" : "exit") "
+                + "timedOut=\(timedOut) stdout=\(stdout.isEmpty ? "<empty>" : stdout)"
+        }
     }
 
     @Test func testCLIErrorPathDoesNotCrashWhenStderrIsClosed() throws {
         let cliPath = try bundledCLIPath()
+        // Pin the socket and the home directory. Without CMUX_SOCKET_PATH the CLI's resolution can
+        // fall back to a machine-global marker file and reach whatever app happens to be running,
+        // which would make this test's exit code depend on the machine rather than on the CLI.
+        let home = FileManager.default.temporaryDirectory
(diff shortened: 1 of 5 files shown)


austinywang:
Superseded by #9572. This PR head is owned by `ejc3/cmux`, so completing it would have required pushing to an external fork. The origin-owned replacement preserved and finished the test-only fixes, passed the final focused test runs, and has now merged into `main`: https://github.com/manaflow-ai/cmux/pull/9572

PR #8857, opened 2026-07-24
Fix the main build: four errors across three files

ejc3:
Building `main` fails. Three errors, all in the two Dock session-restore files that
arrived with #8690, and the first one hides the other two:

```
Sources/DockSplitStore+RestoredAgentLifecycle.swift:13:62: error: cannot find type 'PanelShellActivityState' in scope
Sources/DockSplitStore+SessionSnapshot.swift:44:43: error: generic parameter 'Key' could not be inferred
Sources/DockSplitStore+SessionSnapshot.swift:44:43: error: generic parameter 'Value' could not be inferred
Sources/DockSplitStore+SessionSnapshot.swift:332:47: error: generic parameter 'U' could not be inferred
```

Repro on a clean checkout of `main` at 4dc00aebb8:

```
xcodebuild -project cmux.xcodeproj -scheme cmux-unit -configuration Debug \
  -destination 'platform=macOS' build
```

**The missing import.** `DockSplitStore+RestoredAgentLifecycle.swift` declares
`updatePanelShellActivityState(panelId:state:)`, whose `state` is a
`PanelShellActivityState`. That type is a `public enum` in the CmuxWorkspaces package, and
Swift resolves imports per file, so the app target depending on the package is not enough on
its own. Every other file under `Sources/` that names the type imports CmuxWorkspaces; this
one imports only Foundation.

**The two inference failures.** Both are the same shape: a multi-statement closure whose
bail-out is a bare `return nil`, handed to something generic. Line 44 gives it to
`Dictionary(uniqueKeysWithValues:)`, which then has to solve `Key` and `Value`; line 332 gives
it to `Optional.flatMap`, which has to solve `U`. A bare `return nil` carries no type, so the
closure's result type and the generic parameters each wait on the other. Naming the return
types breaks the cycle, using the types the surrounding code already fixes:
`SessionSplitContainerSnapshot` declares `sourceWorkspaceIdsByPanelId` as `[UUID: UUID]?`,
`sessionRestoreWorkspaceId` is a `UUID`, and `observation` is a
`RestorableAgentSessionIndex.Entry?`. Behaviour is unchanged.

**A fourth error, in a third file.** `surfacePromptForResumeApproval` builds an `NSAlert` over
several statements and ends with a bare `switch alert.runModal()` whose cases are `.auto`,
`.prompt` and `.manual`:

```
Sources/ControlSurfaceResumeTarget.swift:282:40: error: reference to member 'auto' cannot be resolved without a contextual type
Sources/ControlSurfaceResumeTarget.swift:283:41: error: reference to member 'prompt' cannot be resolved without a contextual type
Sources/ControlSurfaceResumeTarget.swift:284:19: error: reference to member 'manual' cannot be resolved without a contextual type
```

Swift gives a function body an implicit return only when the body is a single expression, so
in a multi-statement body that switch is an expression statement with nothing to type it, and
the leading-dot members have no `SurfaceResumeApprovalPolicy` to resolve against. Returning it
supplies the contextual type. Every other trailing switch of this shape under `Sources/` is the
single-expression body of a computed property or a one-statement function, which is why this is
the only one that fails.

Worth saying how this got in, because it will happen again otherwise: `ci.yml` is
`workflow_dispatch:`-only, so no pull request is ever built. The checks that appear on a PR
are reviewers, not a compile, so a commit that does not build can land looking green.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined how resume approval outcomes are processed for consistency.
  * Improved robustness of workspace/session snapshot mapping and related filtering behavior.
* **Tests**
  * Updated workspace inheritance test setup to explicitly omit a restore source workspace ID where appropriate.
* **No user-visible changes**
  * These changes are intended to be behavior-preserving for end-users, with improvements focused on reliability and internal consistency.

### DIFF
--- a/Sources/ControlSurfaceResumeTarget.swift
+++ b/Sources/ControlSurfaceResumeTarget.swift
@@ -278,7 +278,7 @@ extension TerminalController {
         )
         content.apply(to: alert, presentingWindow: nil)
 
-        switch alert.runModal() {
+        return switch alert.runModal() {
         case .alertFirstButtonReturn: .auto
         case .alertSecondButtonReturn: .prompt
         default: .manual
--- a/Sources/DockSplitStore+RestoredAgentLifecycle.swift
+++ b/Sources/DockSplitStore+RestoredAgentLifecycle.swift
@@ -1,3 +1,4 @@
+import CmuxWorkspaces
 import Foundation
 
 extension DockSplitStore {
--- a/Sources/DockSplitStore+SessionSnapshot.swift
+++ b/Sources/DockSplitStore+SessionSnapshot.swift
@@ -41,11 +41,12 @@ extension DockSplitStore {
                 )
             }
         let persistedPanelIds = Set(panelSnapshots.map(\.id))
-        let sourceWorkspaceIdsByPanelId = Dictionary(uniqueKeysWithValues: panelSnapshots.compactMap {
-            panel in
-            guard let transfer = detachedSurfaceTransfersByPanelId[panel.id] else { return nil }
-            return (panel.id, transfer.sessionRestoreWorkspaceId)
-        })
+        let sourceWorkspaceIdsByPanelId: [UUID: UUID] = Dictionary(
+            uniqueKeysWithValues: panelSnapshots.compactMap { panel -> (UUID, UUID)? in
+                guard let transfer = detachedSurfaceTransfersByPanelId[panel.id] else { return nil }
+                return (panel.id, transfer.sessionRestoreWorkspaceId)
+            }
+        )
         let layout = layoutCodec.pruned(
             rawLayout,
             keeping: persistedPanelIds
@@ -329,7 +330,7 @@ extension DockSplitStore {
         let expectedSessionId = resumeBinding?.isAgentHookBinding == true
             ? resumeBinding?.checkpointId
             : restorableAgent?.sessionId
-        let relevantObservation = observation.flatMap { entry in
+        let relevantObservation = observation.flatMap { entry -> RestorableAgentSessionIndex.Entry? in
             guard entry.snapshot.kind == expectedKind, entry.snapshot.sessionId == expectedSessionId else {
                 return nil
             }
--- a/cmuxTests/DockWorkingDirectoryInheritanceTests.swift
+++ b/cmuxTests/DockWorkingDirectoryInheritanceTests.swift
@@ -214,6 +214,7 @@ struct DockWorkingDirectoryInheritanceTests {
     ) -> Workspace.DetachedSurfaceTransfer {
         Workspace.DetachedSurfaceTransfer(
             sourceWorkspaceId: sourceWorkspaceId,
+            sessionRestoreSourceWorkspaceId: nil,
             panelId: panel.id,
             panel: panel,
             title: panel.displayTitle,


ejc3:
Verified on a clean worktree at 94a329e24e: `** BUILD SUCCEEDED **`, 0 errors, building the `cmux-unit` scheme for macOS with `SWIFT_COMPILATION_MODE=wholemodule` so every remaining type error in the module would surface in one pass rather than one file per build. The same worktree at `origin/main` fails on the first of the four.

ejc3:
Correction to my previous comment, and a fourth commit.

That `** BUILD SUCCEEDED **` came from `xcodebuild build`, which for the `cmux-unit` scheme compiles the app target but **not** `cmuxTests`. So it proved less than I implied. The test target was still broken by the same change:

```
cmuxTests/DockWorkingDirectoryInheritanceTests.swift:216:49: error: missing argument for parameter 'sessionRestoreSourceWorkspaceId' in call
```

`DetachedSurfaceTransfer` gained a `sessionRestoreSourceWorkspaceId: UUID?` stored property, and since the struct has no explicit init that became a required argument of the memberwise one. Four places under `cmuxTests` construct it; three were updated and this one was missed. It now passes `nil`, which is what the other three pass and is behaviour-preserving, because `sessionRestoreWorkspaceId` resolves to `sessionRestoreSourceWorkspaceId ?? sourceWorkspaceId`.

Giving the property a default would not work: a `let` with an initial value is excluded from the memberwise init, which would break the five call sites under `Sources` that pass it explicitly.

Re-verifying with `build-for-testing` so the check covers both targets.

ejc3:
Rebased onto current `main` (`69b2c55e27`), which had moved 15 commits ahead. Now 4 commits, 4 files, +10/-7.

I checked whether any of it had become unnecessary rather than assuming. **All four errors are still present on `main`**, so every commit still earns its place:

| | on current main |
|---|---|
| missing `import CmuxWorkspaces` | still broken |
| two bare `return nil` closures | still broken |
| trailing `switch` with no `return` | still broken |
| missing memberwise arg in the test | still broken |

None of the four files this touches were modified by those 15 commits, so every line number cited in the commits and in the description above still points at the right line.

Re-verifying the build on the rebased tip, since my earlier `** TEST BUILD SUCCEEDED **` was measured against the older base and a rebase invalidates it. I will post the result either way.

ejc3:
Re-verified after the rebase onto current `main` (`69b2c55e27`):

```
** TEST BUILD SUCCEEDED **   errors=0   sha=62066a0bd6
```

`build-for-testing` with `SWIFT_COMPILATION_MODE=wholemodule`, so both the app and `cmuxTests` are covered and any remaining type error in the module would have surfaced in the same pass. The claim I withdrew when rebasing is now re-earned on the current base.

Incidental evidence of why this one matters: I tried to build an unrelated one-file test fix (#8867) on plain `main` an hour ago and it came back with 7 errors, every one of them from this list and none from the change itself. Nothing can be verified in isolation until this lands.

azooz2003-bit:
Heads up: the three app-target fixes here (the `CmuxWorkspaces` import, the `return switch` in `ControlSurfaceResumeTarget.swift`, and the closure annotations in `DockSplitStore+SessionSnapshot.swift`) just landed via https://github.com/manaflow-ai/cmux/pull/8869 (merged as 34e7fa5df274) — I hit the same broken main from the reload builder and didn't find this PR until CodeRabbit cross-referenced it after mine was open, sorry for the duplication. The `cmuxTests/DockWorkingDirectoryInheritanceTests.swift` fix here is still needed (my reload builds only compile the app target, so I never hit it). After a rebase this PR should shrink to that one file.

ejc3:
Closing: `main` has since fixed all four of these independently, so this PR is redundant.

- The missing `import CmuxWorkspaces`: fixed by #8873 (`518822336a`).
- The two generic-inference failures in `DockSplitStore+SessionSnapshot.swift` and the missing memberwise argument in the test: fixed by #8869 (`34e7fa5df2`).
- The trailing `switch` in `surfacePromptForResumeApproval`: fixed on `main` with explicit `return`s per case rather than this PR's `return switch` — same effect, different spelling, which is why a search for my own fix text misses it.

Verified by reading the current code at all four sites rather than pattern-matching for this branch's text. Nothing left here to merge.

PR #8883, opened 2026-07-24
iroh: re-key binding slot to (user, device, tag), newest-auth-wins

azooz2003-bit:
## Problem

The active-binding slot was keyed on `app_instance_id` (partial-unique index `active_app_instance_unique` where `revoked_at is null`). A reinstall, sign-out/in, or key rotation mints a fresh `app_instance_id`, so a returning computer collided with its own past self and got `409 binding_replacement_requires_revocation`. There was no client-side recovery, so the App Store review Mac was stranded behind a stale non-revoked binding for 17h (see [cmux App Store review rig notes]).

## Fix

Re-key the binding slot to `(user_id, device_uuid, tag)`, partial-unique where `revoked_at is null`. A registration for an existing slot overwrites it **in place** (newest authenticated registration wins) and preserves the binding row id, so existing pair grants keep resolving. There is deliberately **no generation gate**: a reinstall resets `identity_generation` to 1, and gating on it would reintroduce the same wedge.

- The endpoint id stays globally unique, re-checked excluding self so a slot can rotate its own key (`endpoint_already_bound` still thrown for a foreign claim).
- Advisory lock moves from `iroh:app:<appInstance>` to `iroh:slot:<user>:<device>:<tag>` so same-slot registrations serialize; the new slot unique index is defense behind the lock.
- Drops the per-device (8) and per-account (32) binding caps, the stale-binding recycler, and the `bindingQuota` plumbing. The **challenge-issuance quota is kept** (unchanged).

## Migration

`20260724140000_iroh_rekey_bindings` collapses any duplicate active `(user, device, tag)` rows the old key allowed: keep the most-recently-seen row per slot, soft-revoke the rest with reason `slot_rekey_collapsed`, revoke their pair grants, and bump each affected account's LAN discovery generation. Then drops `active_app_instance_unique` and creates `active_slot_unique`.

## Tests

`web/tests/iroh-db-behavior.test.ts` and `web/tests/iroh-trust-broker.test.ts`: 57 pass, 0 fail (DB-gated + unit). New coverage:
- rejects a second active binding for the same `(user, device, tag)` slot (23505 on the new index)
- re-keys a reinstalled slot in place (same row id, rotated endpoint, generation resets to 1) and frees its old endpoint; a second tag creates a distinct slot
- overwrites the slot in place when the endpoint or generation changes (memory repository)

`bunx drizzle-kit check` clean; web typecheck clean.

## Notes

Broker schema change only; no prod migration is applied by this PR.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/8883"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787521234&installation_model_id=21920&pr_number=8883&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F8883&signature=34f5b108e54b2e0d53d615e5e25c603340f0ccacffd51089e9c4b71255127ff9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Re-keyed Iroh binding slots to `(user, device, tag)` so reinstalls and sign‑ins reuse the same slot without 409s. Adds safe reincarnation, a 512 active‑binding cap, and a strict challenge freshness gate that uses mint time + issuance sequence to prevent out‑of‑order overwrites.

- **New Features**
  - Slot is unique on `(user_id, device_uuid, tag)` where `revoked_at is null`; newest authenticated registration wins with no generation gate.
  - Safe reincarnation: if endpoint, platform, or identity generation changes, mint a new binding id, retire the old via `revokeActiveBindings` with `slot_reincarnated`, revoke its pair grants, and rotate LAN discovery generation.
  - Challenge freshness/ordering: use the `(registeredAt, registeredMintSeq)` high‑water mark from the landing challenge; reject stale and same‑millisecond older challenges as `challenge_superseded`; map unique‑slot races to `slot_registration_superseded`.
  - Enforces global `endpointId` uniqueness excluding self; advisory lock moved to `iroh:slot:<user>:<device>:<tag>` to serialize same‑slot registrations.
  - Removed per‑device/account caps and the stale‑binding recycler; added a 512 active‑binding cap that rejects genuinely new slots at the cap (`active_binding_limit`).

- **Migration**
  - Collapses duplicate active `(user, device, tag)` rows: keep most‑recently‑seen, soft‑revoke others with `slot_rekey_collapsed`, revoke their pair grants, and bump LAN discovery generation.
  - Drops `iroh_endpoint_bindings_active_app_instance_unique`; adds `iroh_endpoint_bindings_active_slot_unique`; adds `mint_seq` to `iroh_registration_challenges` and `registered_mint_seq` to `iroh_endpoint_bindings`.

<sup>Written for commit bd64f42c0c0c7d73ab321315f56e61595f089d25. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/8883?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * Reinstalling or re-registering for the same user/device/tag now reuses the active binding (no duplicates).
  * Slot re-keying and platform/identity changes now soft-retire the prior binding and create a new active one as needed, while keeping pairing continuity.
  * Active endpoint IDs are now unique across all active bindings using the user/device/tag slot model.
  * Registration enforces an active-binding sanity cap: adding at the cap succeeds; adding beyond it is rejected.
  * Developer/registration quota behavior was updated to align with the new flow.
* **Tests**
  * Updated trust-broker and database behavior coverage for the revised slot uniqueness, reincarnation behavior, direct-port handling, and limit enforcement.

## Device-id stability dependency

The `(user_id, device_uuid, tag)` slot key assumes `device_uuid` survives an app reinstall; the database cannot enforce that, the client owns it (documented on the slot index in `web/db/schema.ts`). The iOS half of that contract ships in https://github.com/manaflow-ai/cmux/pull/8888: `DeviceRegistryService.durableDeviceID(defaults:)` reads a device-only Keychain identity (`KeychainDeviceIdentityStore`, `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`) that survives reinstall, and binding registration defers while the Keychain is unreadable instead of registering a throwaway id. The legacy UserDefaults-only `deviceID(defaults:)` remains only on the registry read path, which never registers bindings. Merge order: https://github.com/manaflow-ai/cmux/pull/8888 lands first so the runtime device-id source in `main` is the Keychain-backed one before this slot key relies on it. Old clients that predate 8888 behave exactly as they do today (a reinstall mints a new id and a fresh slot; the old row ages out), so the interim is not a regression.


azooz2003-bit:
Superseded by https://github.com/manaflow-ai/cmux/pull/9196, which fixes the same same-millisecond registration-ordering bug at the mint source (per-slot strict total order on challenge createdAt under the per-user advisory lock) with no schema change, and ships the equivalent regression test. Merging this PR would only add a redundant mint_seq column and prod migration. Note: staging DB already has the unused mint_seq column from this PR's staging validation; it is harmless (default-filled, unreferenced) and can be dropped in a future cleanup migration.

Issue #8564, opened 2026-07-21
300% CPU usage when cmux is idle


bningdd:
### cmux version

0.64.20 build 100

### macOS version

26.5.2 (25F84)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Direct download (DMG)

### Can you reproduce this on cmux NIGHTLY?

I could not test NIGHTLY

### Bug description

window process has 300% CPU usage while all sessions are idle
<img width="930" height="337" alt="Image" src="https://github.com/user-attachments/assets/e5e4effc-34a0-4854-97c2-dc86f8f3735d" />

### Expected behavior

Window process should use less than 50% CPU 

### Steps to reproduce

Not sure how to reproduce. 

### Shell and environment

oh-my-zsh and starship. 

### Relevant logs or crash reports

```text

```

Digman:
I reproduced the same sustained high-CPU behavior on cmux 0.64.20 (build 100, commit `14e3400b9`) and captured what appears to be the root cause.

### Environment

- macOS 26.5.2 (25F84)
- Apple Silicon (`arm64`)
- cmux 0.64.20 (100)

### Observations

- The cmux app process continuously used approximately 107–133% CPU while the terminals were otherwise idle.
- Its memory footprint was approximately 1.4 GB, with 85–86 threads.
- In a one-second unified-log capture, I parsed 46,702 messages. 46,194 of them (98.9%) were the same message:

```text
[com.mitchellh.ghostty:os-open] os-open: open stderr=
```

- A separate three-second capture received approximately 186,635 messages (about 62,000 messages/second), while macOS reported that additional messages had been dropped.
- An eight-second `sample` of the cmux app process showed two background threads spending most of their time in:

```text
zig_os_log_with_type
→ os_log
→ firehose_buffer_tracepoint_reserve_slow
→ __FIREHOSE_CLIENT_THROTTLED_DUE_TO_HEAVY_LOGGING__
```

- The main AppKit thread was mostly waiting, so this does not look like the sidebar layout livelock fixed by #8532.
- There was no remaining `/usr/bin/open` child process when I inspected the process tree.

### Suspected cause

The embedded Ghostty opener drains the spawned `open` command's stderr in `src/os/open.zig` using:

```zig
reader.takeDelimiterExclusive('\n')
```

and logs every returned value unconditionally:

```zig
log.warn("open stderr={s}", .{line});
```

[...]

Issue #145, opened 2026-02-20
Browser panel: back/forward/refresh button click targets too small


lawrencecchen:
The back, forward, and refresh buttons in the browser panel toolbar have small click targets, making them hard to hit accurately. The clickable area should be larger to improve usability, especially for quick navigation.

## Expected behavior
- Buttons should have generous padding / hit areas
- Should be easy to click without precise aiming
- Consider making the entire button area (including padding) clickable, not just the icon

## Current behavior
- Click targets are small, requiring precise mouse placement
- Easy to miss the button and click empty toolbar space instead

austinywang:
Resolved in #227. Should be fixed alongside the next release (v0.60.0, not released yet). Keep an eye out and reopen the issue if the issue is not fully addressed.

Issue #143, opened 2026-02-20
Session persistence (layouts, scrollback)


mathieubellon:
Hey guys!

Closing or reopening cmux deletes all my workspaces, even the pinned ones. Is this a bug or a feature? Am I missing something? I read the documentation but couldn't find anything about how to save this state.

lawrencecchen:
We can just track with this issue!

vitto32:
Session persistence is also blocking us from fully switching. We're evaluating workarounds like dtach to keep agent processes alive across restarts, but workspace layout and tab names are still lost. Even "just reopening structure" would be enough for us to seriously consider making it our daily driver.

lawrencecchen:
Makes sense. Will prioritize session persistence.

lawrencecchen:
Sorry! Will land the PR soon

PR #8884, opened 2026-07-24
Fix io-gather spin and reap dead pane children

### LINKED ISSUE #8872: io-gather threads spin on poll() after pane child exits — defunct children never reaped, ~0.7 core leaked per dead pane (475% CPU after 5 days)
### cmux version

0.64.20 (build 100)

### macOS version

26.2 (25C56)

### Mac chip

Apple Silicon (ARM64)

### Installation method

Direct download (Sparkle auto-update)

### Bug description

After ~5 days of uptime, the main cmux process sits at a sustained **~475% CPU** (181h cumulative CPU time since launch on 2026-07-19). The machine also shows elevated `logd` (~40-70% CPU) and high system CPU (~27% sys), suggesting heavy syscall churn coming from cmux.

Profiling with `sample` shows the **main thread is essentially idle** (blocked in `mach_msg2_trap` for 1214/1932 samples). The CPU is burned by ~7 background threads, each pinned at ~65-68% CPU. Per-thread CPU time (`ps -M`) shows the worst offender has accumulated **36 hours of CPU time on a single thread**:

```
USER  PID    %CPU STAT       STIME      UTIME
...   60654  68.1 R      509:16.36 2170:03.55   <- 36h+ CPU on one thread
...   60654  68.1 R      272:16.33  804:15.05
...   60654  68.0 R      253:56.40  733:44.91
...   60654  67.9 R      238:25.42  688:00.06
...   60654  ...  R      324:46.64  979:33.19
(3 more similar threads; remaining ~195 threads are normal, <2min each)
```

Every stack capture of the hot threads lands in the same place — `io-gather` threads inside `poll()`:

```
2019 Thread_XXXXXX: io-gather
+ 2019 thread_start  (in libsystem_pthread.dylib) + 8
+   2019 ???  (in cmux)  load address 0x104958000 + 0x3dbcea4
+     2019 poll  (in libsystem_kernel.dylib) + 8
```

Since `poll()` is nominally a blocking call but these threads each consume a full ~68% of a core with a very high sys-time share, the signature is a **tight `poll()` loop that returns immediately** — i.e. polling an fd that is permanently ready (typically `POLLHUP` on a closed/dead pty) and never removing it from the set.

Consistent with that: the cmux process has **2 `<defunct>` zombie children that were never reaped**, dating from several days before the capture:

```
  PID  PPID STARTED                      COMMAND
36038 60654 Mon Jul 20 11:31:19 2026     <defunct>
36619 60654 Mon Jul 20 11:31:28 2026     <defunct>
```

Hypothesis: when a pane's child process exits (or crashes), cmux (a) doesn't `wait()` on the child, leaving zombies, and (b) leaves the pane's `io-gather` thread polling the now-dead pty fd, which returns `POLLHUP` instantly and spins forever. Each dead pane then permanently burns ~0.7 core. Over days of opening/closing panes this accumulates — in my case to ~5 cores.

Session context at capture time: 35 child processes, 202 threads, 1389 open fds, ~20 open panes (heavy agent/teammate usage). Thread IDs of the `io-gather` pool range from ~7.4M to ~37M, so a large number of these threads have been created over the process lifetime.

Possibly related: #8564 (same version, "300% CPU usage when cmux is idle", no diagnosis attached — the signature there looks compatible with this root cause).

### Expected behavior

When a pane's child process exits, cmux should reap it (`waitpid`) and tear down / unblock the pane's io threads instead of leaving them spinning on a dead pty fd. CPU usage should return to baseline after panes close.

### Steps to reproduce

Not deterministic, but the pattern that got me here:

1. Run cmux for several days without restarting (heavy use: many panes, agent teams, dev servers streaming output).
2. Have some pane child processes exit or crash over time (zombies under the cmux PID are the telltale).
3. Watch main-process CPU ratchet upward in steps of roughly ~0.7 core and never come back down.
4. Confirm with `sample <pid>`: hot threads are all `io-gather` in `poll()`, while the main thread is idle.

Workaround: restarting cmux clears it completely.

austinywang:
## Summary

- stop POSIX `io-gather` threads when a PTY reaches EOF or reports a terminal
  poll condition
- reap an already-exited pane child when teardown wins the normal
  process-watcher race
- retain deterministic red/green Zig coverage for both the persistent-EOF
  spin and Darwin zombie case
- document the lifecycle ownership invariant and retain the checksum for the
  scoped ReleaseFast GhosttyKit artifact

Closes #8872.

This may also explain the compatible idle-CPU symptoms reported in #8564, but
this PR does not claim to close that issue without matching process/thread
evidence.

## Root cause

Ghostty's POSIX gather loop treated `POLLHUP` as terminal but treated a
zero-byte `read()` as a reason to return to `poll()`. An fd that remains
immediately readable at EOF without the expected HUP bit therefore cycles
`read(0) -> poll(readable) -> read(0)` forever on its dedicated `io-gather`
thread. The loop also ignored `POLLERR` and `POLLNVAL`.

Separately, pane teardown called `getpgid()` before signaling and waiting for
the child. Darwin stops exposing the process group after the child exits even
while its wait status is still pending. If teardown reached that state before
the libxev process watcher callback, `killPid` returned without `waitpid`,
leaving the child as a zombie.

The fix lives in Ghostty `Exec`, the layer that owns the PID, PTY fd, and
reader/gather threads. Swift teardown already calls this owner and joins the
native IO lifecycle, so no second app-layer cleanup path was added.

## Regression proof

Ghostty commit `8bf503f98` contains tests only and fails on the old
implementation:

- `io-gather exits on persistent EOF readiness` uses nonblocking `/dev/null`,
  whose `read()` always returns zero while `poll()` reports immediate
  readability. The old gather thread fails to return within the one-second
  assertion window.
- `subprocess stop reaps an already-exited child on Darwin` forks a short-lived
  session child, waits until Darwin reports `ESRCH` from `getpgid`, invokes
  teardown, then proves no wait status remains.

Ghostty commit `5ef5cba63` makes both tests pass:

- zero-byte reads are authoritative EOF
- `POLLHUP | POLLERR | POLLNVAL` terminate polling after readable tail bytes
  are drained
- teardown performs a nonblocking raw `waitpid` fallback and treats `ECHILD`
  as success when the normal process watcher won the reaping race

## Updated branch state

Merged cmux `main` at `67d8479e05` in `26ee8f65f5`. Current `main` now pins
Ghostty `50ad1963d`, which contains the scoped lifecycle fix through Ghostty PR
#143 and the later serial frame-lease/typing regression fix through Ghostty PR
#145. The effective parent-repository diff is now the detailed lifecycle fork
documentation plus retention of the scoped `5ef5cba63` archive checksum; the
runtime fix is already an ancestor of the newer pin.

## Verification

- `zig build test -Dtest-filter='termio.Exec'`
- `zig build test -Dtest-filter='frame lease pool rotates slots for serial frames'`
- `./tests/test_ci_ghosttykit_checksum_present.sh` (current pin:
  `50ad1963d9c73ee957932ccb4d26bf6d15575ee7`)
- `./tests/test_ci_ghosttykit_checksum_verification.sh`
- `python3 scripts/check-package-resolved-policy.py`
- `python3 scripts/check-workspace-package-groups.py --check`
- Ghostty PR: https://github.com/manaflow-ai/ghostty/pull/143
- scoped ReleaseFast GhosttyKit build:
  https://github.com/manaflow-ai/cmux/actions/runs/30129390408
- scoped published archive:
  https://github.com/manaflow-ai/ghostty/releases/tag/xcframework-5ef5cba6351dbdf315307fc7efd1aaa4a36ab31c-crashsubdir-cmux-crash-v1
- downloaded scoped archive SHA-256:
  `4f74c4cb69fe576a2b597df7d84513440494389a20b60324844a758f8fabd70b`

No local Xcode build, app launch, or XCUITest was run. The PR introduces no
Swift changes, warnings, user-facing strings, localization changes, or budget
TSV changes. `scripts/swift_file_length_budget.py` is absent on current
`main`.

## Summary by CodeRabbit

* **Documentation**
  * Added documentation covering PTY reader shutdown, child-process cleanup, EOF handling, and terminal-condition behavior.
  * Documented constraints for maintaining reliable PTY teardown in future changes.

* **Chores**
  * Added a pinned checksum for an additional GhosttyKit build.


austinywang:
Manual CI workflow run
[30130011335](https://github.com/manaflow-ai/cmux/actions/runs/30130011335)
was dispatched because PR-triggered CI is currently paused on `main`.

The task-owned validation is green, but the full workflow's fail-open
`workflow_dispatch` routing also exercised unrelated web and TestFlight lanes.
Two current-`main` failures prevented the aggregate `ci-status` job from going
green:

- `workflow-guard-tests` expects
  `IOS_BETA_BUNDLE_ID: dev.cmux.app.internal`, while the current checked-in
  TestFlight workflow uses the demo identity.
- `web-typecheck` has ten founders-welcome route tests returning 503 instead of
  their expected responses.

This PR has no changes in either area (`git diff origin/main...HEAD` contains
only `ghostty`, `docs/ghostty-fork.md`, and
`scripts/ghosttykit-checksums.txt`). A scoped rerun reproduced both failures
unchanged, with `linux-preflight`, `tests`, and `ci-status` failing only as
downstream aggregators. No unrelated repair was made.

Relevant gates remain green:

- deterministic Ghostty red/green regression tests and the full
  `termio.Exec` filtered slice
- ReleaseFast GhosttyKit build and archive validation
- checksum presence/verification guards
- Package.resolved policy and workspace grouping
- canonical autoreview merge-conflict gate, structured Codex review, and cmux
  policy check
- CodeRabbit: no actionable comments
- Greptile: safe to merge, 5/5 confidence

austinywang:
## Typing-regression warning and reversion control

A user reports an immediate typing regression in a build containing the new Ghostty pin. I inspected both Ghostty commits and built/launched a pre-bump control.

### Attribution and mechanism

- `8bf503f98` is test-only (`test` blocks in `src/termio/Exec.zig`) and cannot change the shipped runtime.
- `5ef5cba63` is the only production commit. The typing-capable regression path is its POSIX reader shutdown change, principally the promotion of `read(...) == 0` from `break :gather`/re-poll to `fatal = true`, plus the new treatment of `POLLERR | POLLNVAL` as terminal alongside `POLLHUP`.
- On either condition, `gatherMainPosix` returns; its defer sets `pipeline.done = true`; `threadMainPosix` drains the ring and exits. This is permanent for that surface.
- That does **not** synchronously stop the independent write path. cmux still sends `ghostty_surface_key`; Ghostty queues the resulting bytes through `Termio` to `Exec.queueWrite` and the xev `write_stream`. If the PTY is still live, the shell can receive those bytes and echo/output them, but the exited reader no longer calls `processOutput`, so the pane appears to drop/mangle typing or stop accepting input. If the fd is genuinely invalid, queued writes fail/disarm instead. This is a half-dead terminal: writer/process state and reader state can disagree.
- The `waitpid` change is not on the normal live-input path. `reapExitedChild` is reached from `Subprocess.stop`/`killPid` during IO-thread teardown, targets cmux's direct child PID, and does not reap the shell's job children. It is therefore not the plausible live-pane typing mechanism.

The exact Darwin readiness event seen by the reporter is not available, so I am not claiming whether the live trigger was a transient zero-byte read or a newly-fatal `POLLERR`. Both converge on the same new irreversible reader shutdown. The primary suspect is the `n == 0 => fatal` hunk because it deliberately changes a longstanding cautious behavior, and it is also the hunk that fixes the persistent-EOF spin.

### Reversion viability and cost

A full control revert is mechanically safe:

- cmux `main` is still at `3eeed22dec` and pins Ghostty `c55514dd52`; no Swift code or build script depends on the new behavior.
- `scripts/ensure-ghosttykit.sh` derives its cache/download key from the checked-out Ghostty SHA and build flavor. `c55514dd52` already has a checksum-pinned ReleaseFast `xcframework-c55514dd52...-crashsubdir-cmux-crash-v1` archive, so the control cannot accidentally link the `5ef5cba63` kit.
- The cost is real: reverting the EOF shutdown restores #8872's `read(0) -> immediately-readable poll -> read(0)` loop, approximately **0.7 core per dead pane**, accumulating over days. Reverting all of `5ef5cba63` also removes the Darwin teardown reaping fallback, so zombies can recur. That long-uptime leak is preferable to shipping an immediate typing showstopper while the lifecycle fix is corrected.

I did not make a speculative partial revert because the suspected typing hunk and the CPU fix are the same hunk. I kept this branch exactly at the pre-bump `c55514dd52` pin and built/launched tag `revert-ghostty-dead-pty-bump` as the A/B control. Cloud build: https://github.com/manaflow-ai/cmux/actions/runs/30131295136

### Forward-fix direction

The durable invariant should be: **only definitive direct-child/PTY lifecycle state owned by `Exec` may permanently terminate the reader; an ambiguous readiness/read result cannot leave the writer alive while the reader is dead.** A small forward fix can gate EOF/terminal-event teardown on a child-exit/liveness signal from the process owner (for Darwin, the same definitive `getpgid(...)=ESRCH` condition already used during teardown), while retaining tail draining and the nonblocking `waitpid` fallback. Coverage should combine the existing persistent-EOF test with a live-PTY test that injects the suspect zero/error condition, then proves input is written and echo/output still reaches `processOutput`.

Recommendation: hold PR #8884 until the control A/B result is known; do not merge the current `5ef5cba63` pin as-is.

austinywang:
## Post-`main` merge verification

Merged `origin/main` at `67d8479e05` in branch commit `26ee8f65f5` and
re-ran the scoped lifecycle checks plus a full manually dispatched CI:

- Full run: https://github.com/manaflow-ai/cmux/actions/runs/30136993633
- `zig build test -Dtest-filter='termio.Exec'`: pass
- `zig build test -Dtest-filter='frame lease pool rotates slots for serial frames'`: pass
- all four app-host unit-test shards: pass
- aggregate `tests`: pass
- Swift package tests: pass
- universal Release build: pass
- workflow guards, including the repaired TestFlight guard: pass
- `AppDelegateTerminalTypingShortcutFastPathTests`: executed in shard 2 and
  passed in 6.253 seconds

The updated Ghostty pin is `50ad1963d`; `5ef5cba63` (the #8872 lifecycle fix)
is an ancestor of that pin, and the later serial frame-lease/typing work is
also present.

The full workflow concluded red only because `tests-build-and-lag` built the
test target successfully and then found three warning-budget entries in files
inherited unchanged from current `main`:

- `Sources/AppDelegate+WindowDock.swift`
- `Sources/ContentView.swift`
- `Sources/Sidebar/AppKitList/Cells/SidebarGroupHeaderRowView.swift`

This PR's effective diff against current `main` is limited to
`docs/ghostty-fork.md` and the retained scoped GhosttyKit checksum, so no
unrelated warning repair or warning-budget update was added here.

PR #8978, opened 2026-07-27
Use minimal pre-layout terminal bounds

usr-bin-roygbiv:
## Purpose

Initialize terminal views with the smallest nonzero pre-layout bounds instead of allocating an 800×600 backing layer before the pane has a real layout. Headless startup still expands its hidden window to at least 800×600 before runtime creation.

## Tests

- Added focused `TerminalOffscreenStartupTests` contracts for 1×1 normal pre-layout bounds and 800×600 headless startup bounds.
- Current-base red: the immutable test-only commit expected 1×1 and observed 800×600 in the focused hosted method ([run 30323866703](https://github.com/usr-bin-roygbiv/cmux/actions/runs/30323866703)).
- Current-head green: the identical focused hosted method is pinned to `6ac55b13ad31ade5ed77749011b1d849d7594e5d` and is pending ([run 30326409977](https://github.com/usr-bin-roygbiv/cmux/actions/runs/30326409977)).

## Metrics

On the deterministic hosted macOS 15 snapshot workload (12 workspaces, 66 terminals, five timed samples per variant across three iterations), the adjacent baseline/treatment runs measured:

- peak physical footprint: 343,407,168 → 343,275,968 bytes (−131,200 bytes; about 128.1 KiB)
- post-snapshot physical footprint: 335,247,936 → 334,707,136 bytes (−540,800 bytes)
- peak/post-snapshot RSS: 582,746,112 → 575,143,936 bytes (−7,602,176 bytes)

The recorded pair is [baseline run 30174286497](https://github.com/usr-bin-roygbiv/cmux/actions/runs/30174286497) and [treatment run 30176222098](https://github.com/usr-bin-roygbiv/cmux/actions/runs/30176222098). A repetition pinned to the current head was also dispatched as [run 30228114688](https://github.com/usr-bin-roygbiv/cmux/actions/runs/30228114688).

**Claim boundary:** these are directional workload-level measurements from adjacent hosted runs. They do not establish per-terminal savings, a general runtime-memory guarantee, or a post-layout performance change.

## Safety

The bootstrap frame remains nonzero. Headless startup retains the 800×600 minimum through its existing `max` sizing, and post-layout sizing behavior is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS terminal startup rendering by initializing the surface with minimal non-zero bounds to prevent blank or stretched intermediate frames.
  * Refined headless startup behavior so the hosted view stays at the placeholder size until layout, and expands to the expected dimensions when an initial command is provided.

* **Tests**
  * Added regression tests covering both scenarios: placeholder bounds retention without an initial command, and hosted view expansion (including window/content bounds) when an initial command is used.


usr-bin-roygbiv:
@lawrencecchen, this pre-layout terminal-bounds optimization is ready for review when you have a chance.

Issue #8919, opened 2026-07-25
`0.64.20`: uncached `os_log_create` burns a full core even with logging disabled; `renderer_thread` (`error.Defunct`) and `os-open` error floods throttle the firehose and pin `logd` at 60%


jtsternberg:
### cmux version

0.64.20 (build 100) [`14e3400b9`]

### macOS version

26.5.2 (25F84)

### Mac chip

Apple Silicon (ARM64) — Mac14,5, 12 cores, 64 GB

### Bug description

Two related but independently-fixable logging bugs. Together they cost **~1.6 cores sustained** on an otherwise-healthy machine, and — more importantly — the first one is a **trap for anyone trying to diagnose the second**.

Session at capture: 2 windows, 8 workspaces, 14 panes, 20 surfaces, 30 `claude.exe` agent sessions, 137 child processes, 128 threads, 444 open fds. cmux process age 2d 02h; cumulative CPU 303 min, so the runaway started roughly **5 hours** before capture, not at launch.

---

## Bug A — `os_log_create()` is called per-iteration, so disabling the log does not stop the CPU cost

A single thread (`Thread_19132165` = tid `0x123EF05`) sits at a **hard 100% of a core, zero wait time** — 3030 of 3030 sample ticks burning, no blocking syscall anywhere in its stack:

```
    3030 Thread_19132165
    + 3030 thread_start  (in libsystem_pthread.dylib) + 8
    +   1332 ???  (in cmux)  load address 0x100230000 + 0x3d66354
    +   ! 332 os_log_create  (in libsystem_trace.dylib) + 280
    +   ! : 190 _platform_strlcpy  (in libsystem_platform.dylib) + 68
    +   ! 202 zig_os_log_with_type  (in cmux) + 36
    +   ! 140 os_log_type_enabled  (in libsystem_trace.dylib) + 32
```

Aggregated leaf frames for the process (`sample`'s own `Sort by top of stack` section):

```
_platform_memmove        333      os_simple_hash_with_seed  189
_xzm_free                292      os_log_type_enabled       186
_platform_strlen         263      os_log_create             181
_malloc_zone_memalign    214      _os_log_find               71
_platform_memset         209      DYLD-STUB$$os_log_create   32
```

The critical detail: **at the time of this capture the relevant log category was fully disabled at the OS level** —

```console
$ sudo log config --subsystem com.mitchellh.ghostty --category os-open --mode "level:off"
[...]

PR #8981, opened 2026-07-27
Display-link occlusion replay (unproven; parity with main — do not merge for perf)

lawrencecchen:
## What I measured

Profiled the running `cmux.app` 0.64.20 with `sample` (370 threads) while the session was locked, so nothing was on screen:

| | |
|---|---|
| Ghostty surfaces | 64 |
| CVDisplayLinks **running** | **25** |
| Renderer threads awake | **24** |
| Correct number with nothing visible | **0** |

A CoreVideo display-link IO thread exists only while the link is created, and a running one parks in `CVDisplayLink::waitUntil` at the display refresh rate, so thread count is an exact proxy (`created == running` in every sample). Awake renderer threads correlated one-to-one with running links; the other 37 renderer threads sat idle in `kevent64`.

Each of those threads wakes ~120x/sec for a pane nobody can see, and the wakeup work is dominated by `zig_os_log_with_type` -> `_os_log_impl_flatten_and_send` -> `voucher_activity_trace_v_2` -> `_dlock_wait`, so they also serialize on the libtrace lock and feed `logd` (measured at 26% CPU). That amplification is https://github.com/manaflow-ai/cmux/issues/8919; this PR removes the wakeups driving it. Related: https://github.com/manaflow-ai/cmux/issues/8564.

## Root cause

Ghostty's renderer starts every surface at `focused = true` / `visible = true`, starts the CVDisplayLink unconditionally in `loopEnter`, and afterwards only reacts to *transitions*. The link runs while `visible && focused`.

cmux drives both, but only occlusion was lossy:

- `createSurface` replays `desiredFocusState` but **never replayed occlusion**.
- Surfaces for background workspaces are created lazily, so the portal's `setOcclusion(false)` typically lands while `surface` is still `nil`, and `guard let surface else { return }` drops it.
- `GhosttySurfaceScrollView.setVisibleInUI` then cached in `lastRequestedPortalOcclusionVisible` that it had already asked for "hidden", so nothing re-sent it.

Ghostty therefore believed the pane was visible for the rest of the surface's life; once anything focused it, the display link ran forever.

## Fix

Mirrors the existing focus-replay pattern:

- `TerminalSurface.desiredOcclusionVisible` records the wanted state even with no runtime surface.
- `TerminalSurface.appliedOcclusionVisible` dedupes against what the *live* surface was actually told, cleared on runtime teardown so a recreated surface relearns it.
- `createSurface` calls `replayDesiredOcclusionToRuntime()` right after the focus replay.
- `GhosttySurfaceScrollView` drops its duplicate cache and defers deduping to the surface.

## A regression I introduced and backed out

My first attempt also made `setVisibleInUI` push occlusion unconditionally, on the theory that the view layer held authoritative state. It does not: `visibleInUI` starts `true` on every `GhosttyNSView` and a pane that is never mounted is never corrected. Measured with an identical scripted workload (launch, create 4 workspaces focused, cycle, send input):

| build | surfaces | display links running |
|---|---|---|
| `main` (baseline) | 6 | **0** |
| first attempt | 10 | **9** |
| this PR (corrected) | 9 | **0** |

Second commit restores the transition guard. The lost-hide fix does not depend on the push being unconditional.

## Verification status — read before merging

- **Proven:** the lost pre-creation occlusion push is fixed (unit tests below), and this branch does not start links that `main` leaves stopped (table above).
- **Not yet proven end-to-end:** I could not reproduce the original 25-link leak in a controlled build. The leak needs a surface to become first responder, and this Mac's session was locked for the whole session, so no surface ever took interactive focus and both `main` and this branch measured 0. The real confirmation is running this build in a normal multi-workspace session and re-counting.

Count on any build with:

```
sample <pid> 3 -file /tmp/s.txt && grep -c 'CVDisplayLink' /tmp/s.txt
```

Expected: roughly the number of panes actually on screen and focused, not the number of workspaces ever visited.

## Tests

`Packages/macOS/CmuxTerminal/Tests/CmuxTerminalTests/TerminalSurfaceOcclusionReplayTests.swift` asserts the real `ghostty_surface_set_occlusion` value through the existing runtime stub (new `cmux_test_ghostty_occlusion_visible()` accessor):

- pre-creation `setOcclusion(false)` reaches Ghostty when the runtime surface appears
- a visible pane is not occluded by the replay
- a recreated runtime surface relearns that it is hidden

Red/green verified locally by neutering the implementation: tests 1 and 3 fail, test 2 (the no-regression case) stays green. Kept as one commit because the test needs the new seam to compile, so a test-only commit would fail to build rather than fail its assertion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal visibility and occlusion state becoming out of sync after runtime surfaces are recreated or resumed.
  * Ensured hidden terminal panes remain hidden and visible panes are correctly restored.
  * Prevented stale visibility state from suppressing necessary updates.

* **Tests**
  * Added coverage for occlusion replay during surface creation, transitions, and recreation.

### DIFF
--- a/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface+Renderer.swift
+++ b/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface+Renderer.swift
@@ -43,17 +43,22 @@ extension TerminalSurface {
         }
     }
 
-    /// Applies the occlusion state to the runtime surface (deduplicated).
+    /// Applies the occlusion state to the runtime surface.
     ///
     /// The desired value is recorded even when there is no runtime surface yet,
     /// so `createSurface` can replay it. Without that, a `false` pushed while the
     /// surface was still nil is lost and Ghostty keeps its default
     /// `visible = true`, leaving the surface's `CVDisplayLink` free to run at the
     /// display refresh rate for a pane that is not on screen.
+    ///
+    /// Deliberately NOT deduplicated. Ghostty's renderer starts the display link
+    /// from `setFocus(true)` without consulting its own visibility, so a repeated
+    /// `false` here is what stops a link that a focus event restarted behind our
+    /// back. Suppressing those repeats measurably leaks links (7 running vs 1 on
+    /// main for the same workload).
     public func setOcclusion(_ visible: Bool) {
         desiredOcclusionVisible = visible
         guard let surface = surface else { return }
-        guard appliedOcclusionVisible != visible else { return }
         appliedOcclusionVisible = visible
         ghostty_surface_set_occlusion(surface, visible)
     }
--- a/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface+Renderer.swift
+++ b/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface+Renderer.swift
@@ -43,10 +43,29 @@ extension TerminalSurface {
         }
     }
 
-    /// Applies the occlusion state to the runtime surface.
+    /// Applies the occlusion state to the runtime surface (deduplicated).
+    ///
+    /// The desired value is recorded even when there is no runtime surface yet,
+    /// so `createSurface` can replay it. Without that, a `false` pushed while the
+    /// surface was still nil is lost and Ghostty keeps its default
+    /// `visible = true`, leaving the surface's `CVDisplayLink` free to run at the
+    /// display refresh rate for a pane that is not on screen.
     public func setOcclusion(_ visible: Bool) {
+        desiredOcclusionVisible = visible
         guard let surface = surface else { return }
+        guard appliedOcclusionVisible != visible else { return }
+        appliedOcclusionVisible = visible
         ghostty_surface_set_occlusion(surface, visible)
+    }
+
+    /// Pushes `desiredOcclusionVisible` onto a freshly created runtime surface,
+    /// bypassing the dedupe cache (a new runtime surface starts from Ghostty's
+    /// own `visible = true` default, so the cache from the previous one cannot be
+    /// trusted). Called from `createSurface`.
+    @MainActor
+    func replayDesiredOcclusionToRuntime() {
+        appliedOcclusionVisible = nil
+        setOcclusion(desiredOcclusionVisible)
     }
 
     /// Whether this surface currently holds realized GPU renderer resources.
--- a/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface+RuntimeLifecycle.swift
+++ b/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface+RuntimeLifecycle.swift
@@ -328,6 +328,9 @@ extension TerminalSurface {
         pendingSocketInputQueue.removeAll(keepingCapacity: false)
         pendingSocketInputBytes = 0
         desiredFocusState = false
+        // The next runtime surface starts from Ghostty's own default, so drop the
+        // applied cache and let `createSurface` replay `desiredOcclusionVisible`.
+        appliedOcclusionVisible = nil
 
         guard let surfaceToFree else {
             callbackContext?.release()
@@ -629,6 +632,14 @@ extension TerminalSurface {
         // surface converges with any focus changes that happened while the
         // surface was being initialized.
         ghostty_surface_set_focus(createdSurface, desiredFocusState)
+
+        // Same convergence for occlusion. A hidden pane's `setOcclusion(false)`
+        // usually lands before its runtime surface exists (background workspaces
+        // create surfaces lazily), and Ghostty's renderer defaults to
+        // `visible = true` and only reacts to transitions, so without this replay
+        // the surface stays "visible" to Ghostty forever and keeps its
+        // CVDisplayLink ticking at the display refresh rate off screen.
+        replayDesiredOcclusionToRuntime()
 
         flushPendingSocketInputIfNeeded()
 
--- a/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface.swift
+++ b/Packages/macOS/CmuxTerminal/Sources/CmuxTerminal/Surface/TerminalSurface.swift
@@ -286,6 +286,24 @@ public final class TerminalSurface: Identifiable, ObservableObject {
     /// state unless the workspace focus path requests it.
     var desiredFocusState: Bool = false
 
+    /// Portal-visibility state cmux wants Ghostty to have, tracked independently
+    /// of whether a runtime surface exists yet.
+    ///
+    /// Surfaces for background workspaces are created lazily, long after the
+    /// portal has already marked the view hidden, so the `false` push lands on a
+    /// nil runtime surface and is lost. Ghostty's renderer thread defaults to
+    /// `visible = true` and only reacts to transitions, so a lost `false` leaves
+    /// it believing the surface is on screen forever. Combined with a focused
+    /// surface that keeps its `CVDisplayLink` running, that is a display-refresh
+    /// -rate wakeup for a surface nobody can see. `createSurface` replays this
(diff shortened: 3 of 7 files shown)


lawrencecchen:
## Correction: this fix is unproven. Do not merge on the strength of the original description.

I went back and did the baseline stress comparison I had skipped. Results below.

### The pathology is real (measured on the running 0.64.20)

Monitored over ~12 minutes of normal use:

- idle CPU 200-509%, sustained
- running CVDisplayLinks: 25 -> 31 -> 25 -> 25 across samples, against 64-65 surfaces
- 89,173 cumulative CPU-seconds (~97% of a core averaged over the whole process lifetime)

### But I could not reproduce it, and this branch shows no measured improvement

Identical scripted workloads, `main` vs this branch:

| workload | main | this branch |
|---|---|---|
| 4 workspaces, 1 switch cycle | 0 links | 0 links |
| 14 workspaces + splits, 3 focus-churn rounds | 1 link, 14% idle | 1 link, 13% idle |
| restart/session-restore, run 1 | 4 links, 10% | 1 link, 8% |
| restart/session-restore, run 2 | 0 links, 3% | 0 links, 5% |
| restart/session-restore, run 3 | 0 links, 4% | 0 links, 9% |

The one favourable data point (4 vs 1) did not replicate. The restore runs also came back with `surfaces=1`, so that test was not exercising the multi-surface restore path it was meant to — the persisted workspaces did not come back.

Net: **parity with main on every workload I can drive. No demonstrated improvement.**

### Two regressions I introduced and backed out

Both were caught only by this baseline comparison, which is why the comparison mattered:

1. Making `setVisibleInUI` push occlusion unconditionally. `visibleInUI` starts `true` on every view and an unmounted pane is never corrected, so this told Ghostty that off-screen panes were visible and *started* their links: **9 running of 10 surfaces vs 0 on main**.
2. Deduplicating `setOcclusion`. Ghostty's renderer starts the display link from `setFocus(true)` without consulting its own visibility flag, so a repeated `setOcclusion(false)` is load-bearing — it stops a link a focus event restarted. Deduping leaked **7 running vs 1 on main**.

### What is still standing

Only the record-and-replay half: a `setOcclusion(false)` issued before the runtime surface exists is recorded and replayed in `createSurface`, next to the existing `desiredFocusState` replay. That is a genuine hole (`createSurface` replayed focus but never occlusion) and the unit tests pin it. It is additive — it only makes a push `main` never made at all — and it measures at parity. But I have no evidence it is what causes the 25+ leaked links in practice.

### Most likely remaining lead

`becomeFirstResponder` calls raw `ghostty_surface_set_focus(surface, true)`, bypassing `TerminalSurface.setFocus`'s dedupe, and Ghostty starts the link on focus without checking visibility. If a view is unmounted while first responder, no `resignFirstResponder` fires, so Ghostty keeps `focused = true`. The only thing that then stops the link is a later `setOcclusion(false)` — which `setVisibleInUI` skips when `wasVisible == visible`, i.e. when the view was already marked hidden before the focus arrived.

That suggests guarding only the *show* direction and always re-asserting a *hide*:

```swift
if !visible || wasVisible != visible {
    surfaceView.terminalSurface?.setOcclusion(visible)
}
```

Untested — I cannot validate it without a reproduction, and against my synthetic workloads it would also just show parity.

### Suggested disposition

Do not merge for the battery claim. Either park this until there is a reproduction, or land only the replay + tests as a correctness fix with no performance claim attached.

lawrencecchen:
## Found the actual cause. It is not the display links.

I switched from `sample` (wall-clock stack presence) to true per-thread CPU accounting via `proc_pidinfo(PROC_PIDTHREADINFO)` deltas. That immediately overturned my earlier attribution.

Live `cmux.app` 0.64.20, 374 threads, 148% CPU:

| thread group | CPU | count |
|---|---|---|
| **(unnamed)** | **132.8%** | 26 |
| renderer | 7.6% | 64 |
| CVDisplayLink | 3.5% | 16 |
| io-gather | 2.5% | 64 |
| io-reader / io | 1.5% | 128 |

So the entire display-link/renderer story I spent the first pass on is worth **at most ~11%**. 132.8% — **90% of the process's CPU** — is in unnamed threads.

### What those threads are

Exactly **11 threads**, each **94-97% of its stack inside `zig_os_log_with_type`**, and 13 threads hitting `__FIREHOSE_CLIENT_THROTTLED_DUE_TO_HEAVY_LOGGING__`. The process also has exactly **11 zombie children**. Stable across samples: 11 zombies <-> 11 spinning threads <-> ~12.4% CPU each.

Stack:

```
thread_start -> (cmux) -> zig_os_log_with_type -> _os_log_impl
  -> _os_log_impl_flatten_and_send -> voucher_activity_trace_v_2
    -> firehose_buffer_tracepoint_reserve_slow
      -> __FIREHOSE_CLIENT_THROTTLED_DUE_TO_HEAVY_LOGGING__
```

### Root cause: `ghostty/src/os/open.zig`, `openThread`

```zig
if (exe.stderr) |stderr| {
    while (true) {
        const line = reader.takeDelimiterExclusive('\n') catch ... break;
        log.warn("open stderr={s}", .{line});   // every line, unbounded
    }
}
_ = exe.wait() catch {};                        // only reachable by exiting the loop
```

`takeDelimiterExclusive` can hand back empty slices without advancing or reporting `EndOfStream`. When that happens the loop spins forever: it emits `open stderr=` (empty payload - matches what landed in the log) as fast as it can until macOS throttles the process firehose, and it never reaches `wait()`, so the child stays a zombie. One leaked thread + one zombie per stuck `open`, permanently, at ~12.4% CPU each.

Note this also makes *every other* `os_log` call in the process expensive, since firehose throttling is process-wide. That is the mechanism behind https://github.com/manaflow-ai/cmux/issues/8919, and it plausibly contributes to https://github.com/manaflow-ai/cmux/issues/8564 and the never-reaped children in https://github.com/manaflow-ai/cmux/issues/8872.

### Fix

manaflow-ai/ghostty `fix-open-stderr-spin` (673d52df14), pointer bumped here:

- `defer _ = exe.wait() catch {}` so the child is reaped unconditionally
- break out after 64 consecutive zero-length reads (reader not progressing)
- cap reported lines at 32, but keep draining so a child blocked on a full pipe can still exit

### Verification status

- GhosttyKit rebuilt for the new SHA (had to dispatch `build-ghosttykit.yml`, which runs on macos-15; the macos-26 path fails to link zig 0.15.2 against the Xcode 26 SDK).
- `battfix` build runs, panes render, **0 zombies** after the stress workload.
- Idle CPU on the synthetic workload: main 14%, this branch 13%, ghostty-fixed 11% - **within noise, and not a valid measure**, because the synthetic workload never reproduces the leak. No other cmux instance on this machine has a single zombie; it takes a long-running real session.

**The before/after that matters can only be taken in a real session:** run `battfix` as a daily driver, then

```
ps -axo ppid=,stat= | awk -v P=<pid> '$1==P && $2 ~ /Z/' | wc -l   # expect 0, was 11
threadcpu <pid> 15                                                 # expect no ~12.4% unnamed cluster
```

### Disposition

The display-link commits (1-3) are at parity with main and worth <=11% at best; the ghostty bump is the one carrying the ~90%. Happy to split them if you would rather land only the ghostty fix.

lawrencecchen:
## Split done — the ghostty fix moved out of this PR

The submodule bump that carries the real win is now standalone in https://github.com/manaflow-ai/cmux/pull/9000 (leaked `openThread` loops, ~90% of idle CPU). This branch has been force-pushed to drop that commit and is now **only** the display-link occlusion work.

What remains here, honestly:

- True per-thread CPU accounting puts the whole renderer + CVDisplayLink path at **≤11%** of the process (renderer 7.6%, CVDisplayLink 3.5%), not the dominant cost I originally claimed.
- All three commits measure at **parity with main** on every workload I could drive (1 link vs 1 link; 0 vs 0; idle 13% vs 14%).
- Two of the three commits exist only to back out regressions I introduced in the first one.
- The genuine defect it fixes is real but narrow: `createSurface` replayed `desiredFocusState` and never replayed occlusion, so a `setOcclusion(false)` issued before the runtime surface existed was dropped on a nil optional. Unit tests pin it.

Moved to draft. It is a small correctness fix with tests and no measured performance benefit — land it on that basis or not at all. Do not merge it for battery.

PR #9087, opened 2026-07-28
Fix terminal Find Next navigation

### LINKED ISSUE #8971: Cmd-G (Find Next) searches for the literal word "next" — findNext() sends Ghostty `search:next` instead of `navigate_search:next`
### cmux version

0.64.20 (100) `[14e3400b9]` — bundled Ghostty 1.3.2-HEAD

### macOS version

macOS 26.5.2 (25F84)

### Mac chip

Apple Silicon (M1 Max)

### Installation method

Direct download (DMG)

I could not test NIGHTLY — but the cause is visible in `main` at the time of writing (see below), so it should still reproduce.

### Bug description

Pressing **⌘G (Find Next)** in a terminal pane does not advance to the next match. Instead it starts a brand-new search for the literal word **`next`**, discarding whatever query was typed. **⌥⌘G (Find Previous)** does the same with `previous`.

Because `search:` *replaces* the active search, the original query is destroyed — there's no "press it again" recovery. The find bar's own ↑/↓ buttons and its Return handler work correctly, so only the keyboard shortcuts are affected.

**Root cause.** Ghostty parses a binding action as `action:parameter`, and its two search actions are not interchangeable — from `ghostty +list-actions --docs` on the Ghostty bundled with cmux:

```
search:
  Start a search for the given text. If the text is empty, then
  the search is canceled. …
  If a previous search is active, it is replaced.

navigate_search:
  Navigate the search results. If there is no active search, this
  is not performed.

  Valid values: `previous`, `next`.
```

`TabManager` sends the wrong one:

```swift
// Sources/TabManager.swift:837
_ = panel.performBindingAction("search:next")
// Sources/TabManager.swift:846
_ = panel.performBindingAction("search:previous")
```

So `search:next` means *"start a new search for the text `next`"*. Ghostty is doing exactly what it was asked.

Meanwhile `Sources/Find/SurfaceSearchOverlay.swift` already uses the correct action in four places (lines 55, 56, 88, 99):

```swift
"navigate_search:previous" / "navigate_search:next"
```

That divergence looks like the real root cause: the same action string is duplicated across two files and the two spellings drifted.

### Expected behavior

⌘G advances to the next match of the current search; ⌥⌘G goes back one — matching what the find bar's ↑/↓ buttons already do.

### Steps to reproduce

1. Open a terminal pane with some scrollback.
2. Press ⌘F, type any query (e.g. `error`), press Return — matches highlight correctly.
3. Press ⌘G.

**Actual:** the search is replaced with a search for the word `next`, and the original query is gone.
**Expected:** jump to the next match of `error`.

### Shell and environment

zsh 5.9, no tmux. Not shell-dependent — the keystroke never reaches the shell.

### Additional context

**Suggested fix** — point both call sites at the navigation action:

```diff
 func findNext() {
     if let panel = selectedTerminalPanel {
-        _ = panel.performBindingAction("search:next")
+        _ = panel.performBindingAction("navigate_search:next")

 func findPrevious() {
     if let panel = selectedTerminalPanel {
-        _ = panel.performBindingAction("search:previous")
+        _ = panel.performBindingAction("navigate_search:previous")
```

Given the drift described above, it may be worth hoisting the two strings into a single shared constant used by both `TabManager` and `SurfaceSearchOverlay.swift` so they can't diverge again.

**One behavior change worth deciding on:** `navigate_search` is documented as *"If there is no active search, this is not performed."* After the fix, ⌘G with no active search would do nothing, where today it starts a bogus search. That's strictly better, but you may want an explicit "no active search → open the find bar" fallback. The ⌘F path is unaffected either way, since `startSearch()` already recovers `surface.lastSearchNeedle`.

**Workaround for anyone hitting this:** unbind the shortcuts in cmux so they fall through to Ghostty, which already binds ⌘G to `navigate_search:next` by default.

```jsonc
// ~/.config/cmux/cmux.json
"shortcuts": { "bindings": { "findNext": null, "findPrevious": null } }
```

```
# ~/.config/ghostty/config
keybind = cmd+g=navigate_search:next
keybind = cmd+opt+g=navigate_search:previous   # Ghostty's own default for previous is shift+cmd+g
```

Then `cmux reload-config`. Note that rebinding ⌘G *within* cmux does not help — it just re-triggers the same `findNext()`.

austinywang:
## Summary

- route terminal Find Next and Find Previous through one typed navigation action
- keep the active search query intact while moving between matches
- cover both directions with a regression test that rejects literal `search:` actions

## Behavior with no active search

Ghostty documents `navigate_search` as a no-op when no search is active. This PR preserves that contract instead of making Cmd-G implicitly open the find bar; Cmd-F remains the explicit entry point for starting or restoring search.

## Testing

- two-commit red/green regression sequence
- project/test wiring guards
- CI (local Xcode builds and tests intentionally skipped per task instructions)

## Localization audit

No user-facing strings were added or changed.

Closes #8971

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9087"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787871580&installation_model_id=21920&pr_number=9087&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9087&signature=ffa20ba0bb74a5ea2c9a7debb791d80c53186e6705577c0893d8e2b8c008322c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fix terminal Find Next/Previous by routing navigation through `navigate_search` via a typed enum, preserving the active query. Closes #8971.

- **Bug Fixes**
  - Keep Cmd-G/Cmd-Shift-G as no-ops when no search is active; Cmd-F starts search.
  - Replace string commands with a typed navigation API across the find bar and keyboard; add tests to cover next/previous and reject `search:` actions.

<sup>Written for commit 48d13aacef918c86fd771b05eb8d47cd65846b43. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9087?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Sources/Find/SurfaceSearchOverlay.swift
+++ b/Sources/Find/SurfaceSearchOverlay.swift
@@ -22,7 +22,7 @@ struct SurfaceSearchOverlay: View {
     let surfaceId: UUID
     @ObservedObject var searchState: TerminalSurface.SearchState
     let canApplyFocusRequest: () -> Bool
-    let onNavigateSearch: (_ action: String) -> Void
+    let onNavigateSearch: (_ direction: TerminalSearchNavigation) -> Void
     let onSearchTextChanged: () -> Void
     let onFieldDidFocus: () -> Void
     let onClose: () -> Void
@@ -51,10 +51,7 @@ struct SurfaceSearchOverlay: View {
                         onClose()
                     },
                     onReturn: { isShift in
-                        let action = isShift
-                            ? "navigate_search:previous"
-                            : "navigate_search:next"
-                        onNavigateSearch(action)
+                        onNavigateSearch(isShift ? .previous : .next)
                     }
                 )
                 .accessibilityIdentifier("TerminalFindSearchTextField")
@@ -85,7 +82,7 @@ struct SurfaceSearchOverlay: View {
                     #if DEBUG
                     cmuxDebugLog("findbar.next surface=\(surfaceId.uuidString.prefix(5))")
                     #endif
-                    onNavigateSearch("navigate_search:next")
+                    onNavigateSearch(.next)
                 }) {
                     Image(systemName: "chevron.up")
                 }
@@ -96,7 +93,7 @@ struct SurfaceSearchOverlay: View {
                     #if DEBUG
                     cmuxDebugLog("findbar.prev surface=\(surfaceId.uuidString.prefix(5))")
                     #endif
-                    onNavigateSearch("navigate_search:previous")
+                    onNavigateSearch(.previous)
                 }) {
                     Image(systemName: "chevron.down")
                 }
--- a/Sources/Find/TerminalSearchNavigation.swift
+++ b/Sources/Find/TerminalSearchNavigation.swift
@@ -4,6 +4,6 @@ enum TerminalSearchNavigation: String, CaseIterable, Sendable {
 
     @discardableResult
     func perform(_ performBindingAction: (String) -> Bool) -> Bool {
-        performBindingAction("search:\(rawValue)")
+        performBindingAction("navigate_search:\(rawValue)")
     }
 }
--- a/Sources/Find/TerminalSearchNavigation.swift
+++ b/Sources/Find/TerminalSearchNavigation.swift
@@ -1,0 +1,9 @@
+enum TerminalSearchNavigation: String, CaseIterable, Sendable {
+    case next
+    case previous
+
+    @discardableResult
+    func perform(_ performBindingAction: (String) -> Bool) -> Bool {
+        performBindingAction("search:\(rawValue)")
+    }
+}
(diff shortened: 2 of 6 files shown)


austinywang:
@coderabbitai review

PR #9225, opened 2026-07-30
Replay sidebar clicks after reveal actions return

### LINKED ISSUE #9221: Sidebar workspace click intermittently does not register (AppKit sidebar): click paints, selection never applies
## Symptom

Clicking a workspace row in the sidebar sometimes does nothing — the row does not become active and the terminal content does not swap. Clicking again usually works. It feels timing/load dependent: it happens more often shortly after the sidebar appears / the window becomes active / workspaces are still settling, which suggests a race between the click and the sidebar's row-loading and presentation lifecycle rather than a hit-testing bug.

This is a regression relative to the pre-AppKit sidebar. The AppKit `NSTableView` sidebar landed in #8270, with selection coalescing (#8366, #8390) and the down-then-up optimistic highlight (#8461) layered on top; the dropped-click behavior is consistent with those paths.

## Where the click goes

`NSTableView` owns the mouse tracking loop, so row clicks route through the table's target/action, not a gesture recognizer:

- `Sources/Sidebar/AppKitList/SidebarWorkspaceTableController.swift:79` — `table.action = #selector(didClickTableRow)`
- `Sources/Sidebar/AppKitList/SidebarWorkspaceTableController.swift:443` — `didClickTableRow()`
- `Sources/Sidebar/AppKitList/SidebarWorkspaceTableViewImpl.swift:50` — `mouseDown`, deliberately paints nothing on press; the action fires on mouse-up
- `tableView(_:shouldSelectRow:)` returns `false` (`SidebarWorkspaceTableController.swift:569`) — selection is fully custom, so a dropped action means a fully dropped click

## Suspects (each independently able to swallow a click)

**1. Rows can be visible but action-less (strongest match for the loading race).**
`presentationSnapshot()` (`Sources/Sidebar/AppKitList/SidebarWorkspaceTableRowConfiguration.swift:168`) intentionally keeps `appKitWorkspaceRowModel` / `appKitGroupHeaderModel` — so the row still paints completely, looking normal and clickable — while nil-ing out `appKitWorkspaceRowActions` and `appKitGroupHeaderActions`. `suspendPresentation()` (`SidebarWorkspaceTableController.swift:195`) replaces every row with that snapshot and sets `actions = nil`.

`didClickTableRow()` then has no fallback: both branches are `if let ... actions`, so with nil actions the click paints an optimistic highlight via `previewSelection` (line 460) and **returns having done nothing** — no selection, no log, no recovery. The highlight then snaps back on the next authoritative apply, which is exactly the "I clicked it and it flickered but stayed put" feel.

`setPresentationActive(true)` (line 163) only calls `mutationScheduler.stageViewportChange()`; live actions are restored solely by the next `apply(...)` from `updateNSView` (`Sources/Sidebar/AppKitList/SidebarWorkspaceTableView.swift:28-36`). Any window where the table is on screen and hit-testable but has not yet received that apply is a window where every click is silently discarded.

**2. `table.clickedRow` is an index resolved at action time, not at press time.**
`didClickTableRow()` reads `table.clickedRow` and indexes `rows[row]` (lines 445-449). Rows are identity-keyed (`SidebarWorkspaceRenderItemID`) everywhere else, but the click resolves by position. A reload/mutation landing between mouse-down and the action — `SidebarWorkspaceTableMutationScheduler` defers applies — either drops the click (`guard rows.indices.contains(row)`) or, worse, activates a different workspace than the one clicked.

**3. The 100 ms selection coalescer can drop a pending click entirely.**
`Sources/Sidebar/AppKitList/SidebarSelectionCoalescer.swift` applies the leading click immediately, but a click landing inside the 100 ms window only becomes `pendingApply` with a trailing task. Four sites call `cancel()`, which discards that pending apply outright — including `workspaceDragSessionDidBegin()` (line 685) and `suspendPresentation()` (line 210). A press with a few pixels of pointer travel makes `NSTableView` ask for a pasteboard writer (`tableView(_:pasteboardWriterForRow:)`, line 573), which calls `beginWorkspaceDrag` + `workspaceDragSessionDidBegin()` and cancels the queued selection. If that press does not become a real drag, the user made a click that is now permanently gone.

## Repro / debugging notes

Not deterministic by hand; reproduce it under load rather than by clicking carefully. Suggested angles:
- Click a row immediately after toggling the sidebar open, after `cmd-tab` back to the window, or while workspaces are still being restored/created.
- Rapid alternating clicks between two rows (exercises the coalescing window plus the cancel sites).
- Click with 1-3 px of pointer drift (exercises the drag-session cancel path).

There is already a DEBUG probe at `SidebarWorkspaceTableController.swift:447`: `cmuxDebugLog("sidebar.table.click row=\(row) rows=\(rows.count)")`. It fires **before** the nil-actions guard, so a debug log showing `sidebar.table.click` with no following selection change confirms suspect 1 or 2. It is `#if DEBUG` only, so it does not help on a release build a user is running.

## Asks

- Reproduce and confirm the actual trigger before fixing — do not assume any one of the three suspects above. They are plausible, not proven, and there may be a fourth.
- Add a regression test that fails on current `main` and passes after the fix. The existing `cmuxTests/SidebarSelectionCoalescerTests.swift` drives the coalescer with an injected `Clock`, so the coalescer path is deterministically testable; `cmuxTests/SidebarWorkspaceTableSuspensionTests.swift` covers the suspension path.
- A click on a visible, hit-testable sidebar row should never be silently discarded. Whatever the specific race, the fix should close the class: either the row is not clickable, or the click is honored (queued and replayed once actions are live). A no-op fall-through in `didClickTableRow()` is not an acceptable resting state.

austinywang:
## Summary

- reproduce the dropped click on a visible, hit-testable AppKit sidebar row while its presentation snapshot has released live actions
- retain a completed click by stable render-item identity until the authoritative reveal apply restores actions
- route immediate and replayed clicks through the same selection/coalescing path, and retire pending intent on teardown or a consuming gesture
- add behavior-level coverage through the real `NSTableView` target/action path

## Proof

The DEBUG trace showed a valid row click with no following selection action; the live reveal apply arrived about 2 ms later:

```text
sidebar.table.issue9221Repro row=1 resolved=1 frame=(0.0, 0.0, 240.0, 612.0) hidden=0
sidebar.table.click row=1 rows=2
```

The test-only commit fails on the original implementation at the expected selection assertion. The fix commit makes the same test pass.

## Tests

- `SidebarWorkspaceTableSuspensionTests/visibleRowClickWhileRevealApplyIsPendingReplaysWhenActionsReturn()`
- `./scripts/lint-pbxproj-test-wiring.sh`
- `git diff --check`
- `./scripts/reload.sh --tag issue-9221 --launch`

Fixes #9221

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9225"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788005925&installation_model_id=21920&pr_number=9225&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9225&signature=6c76c156b11ce3f2dc596efea1f0c790e6229b24604eac02bead5f697c4396e5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Replays valid sidebar row clicks that occur during a reveal when live actions are temporarily unavailable, so selection is applied once actions return. Routes both immediate and delayed clicks through the same coalesced selection path to prevent dropped clicks. Fixes #9221.

- **Bug Fixes**
  - Defers a completed click by storing row ID and modifiers, and replays it after the reveal apply restores actions.
  - Adds a unified click dispatch with coalesced selection using the original event modifiers; cancels pending intent via `cancelSelectionIntent` on teardown, drag, or rename.
  - Adds a regression test that exercises the real `NSTableView` target/action path to verify replay behavior.

<sup>Written for commit 2e78c84f42c96526a8e64e1fd7bf624772b3bf4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace row selection while rows are loading or becoming visible.
  * Clicks are now replayed automatically once the row’s actions become available.
  * Prevented pending selection updates from interfering with renaming or drag-and-drop interactions.

* **Tests**
  * Added coverage for row-click replay during suspended and resumed sidebar states.

### DIFF
--- a/Sources/Sidebar/AppKitList/SidebarWorkspaceTableController.swift
+++ b/Sources/Sidebar/AppKitList/SidebarWorkspaceTableController.swift
@@ -7,10 +7,22 @@ import SwiftUI
 /// Main-actor owner of the default sidebar table lifecycle and its AppKit interactions.
 @MainActor
 final class SidebarWorkspaceTableController: NSObject, NSTableViewDataSource, NSTableViewDelegate {
+    private struct DeferredRowClick {
+        let rowId: SidebarWorkspaceRenderItemID
+        let modifiers: NSEvent.ModifierFlags
+    }
+
+    private enum RowClickDispatchOutcome {
+        case dispatched
+        case awaitingActions
+        case invalid
+    }
+
     private weak var containerView: SidebarWorkspaceTableContainerView?
     private let createdCellViews = NSHashTable<NSView>.weakObjects()
     private var rows: [SidebarWorkspaceTableRowConfiguration] = []
     private var actions: SidebarWorkspaceTableActions?
+    private var deferredRowClick: DeferredRowClick?
     private var hoveredRowId: SidebarWorkspaceRenderItemID?
     private var contextMenuRowId: SidebarWorkspaceRenderItemID?
     private var workspaceIds: [UUID] = []
@@ -149,7 +161,7 @@ final class SidebarWorkspaceTableController: NSObject, NSTableViewDataSource, NS
         selectedScrollTargetWorkspaceId = nil
         hoveredRowId = nil
         contextMenuRowId = nil
-        selectionCoalescer.cancel()
+        cancelSelectionIntent()
         clearDropViewActions(in: container)
         setAppKitDropIndicator(nil, scope: .raw, includeRowTargets: false)
         container.tableView.workspaceController = nil
@@ -207,7 +219,7 @@ final class SidebarWorkspaceTableController: NSObject, NSTableViewDataSource, NS
         contextMenuRowId = nil
         optimisticallyPaintedRowIds.removeAll(keepingCapacity: true)
         pumpHeightOverrides.removeAll(keepingCapacity: true)
-        selectionCoalescer.cancel()
+        cancelSelectionIntent()
         rowHeightCache.suspendPresentation(retaining: Set(rows.map(\.id)))
         if let containerView {
             clearDropViewActions(in: containerView)
@@ -436,6 +448,7 @@ final class SidebarWorkspaceTableController: NSObject, NSTableViewDataSource, NS
         enforceHoverOnVisibleCells()
         updateDropTargets()
         replanReorderDragIfActive()
+        replayDeferredRowClickIfPossible()
     }
 
     /// Row clicks route through the table's action (NSTableView owns the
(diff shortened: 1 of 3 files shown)


austinywang:
TDD red proof for commit c81abab968d105eb73320ef2f452a1ce0a7c3c8e: https://github.com/manaflow-ai/cmux/actions/runs/30545532439

The focused macOS CI run executed visibleRowClickWhileRevealApplyIsPendingReplaysWhenActionsReturn() exactly once and failed at SidebarWorkspaceTableSuspensionTests.swift:178 with the intended assertion:

> A completed click on a visible reveal-time row must replay once live row actions return.

The log records: Test run with 1 test in 1 suite failed ... with 1 issue.

austinywang:
Focused green proof for the fix HEAD `2e78c84f42`:

- Run: https://github.com/manaflow-ai/cmux/actions/runs/30546655884
- Exact selector: `cmuxTests/SidebarWorkspaceTableSuspensionTests/visibleRowClickWhileRevealApplyIsPendingReplaysWhenActionsReturn()`
- The selected test passed once: `Test run with 1 test in 1 suite passed after 0.053 seconds.`

This is the green counterpart to the preserved test-only red proof in https://github.com/manaflow-ai/cmux/pull/9225#issuecomment-5131379690.

PR #9259, opened 2026-07-31
Classify transient token misses as connectivity so launch activation stops failing closed

azooz2003-bit:
## Problem

Every iOS launch and foreground return kicks a network `/users/me` session revalidation, and `sessionTokenTransitionIsActive` is true for its whole round trip (up to 8s). `authenticatedSessionSnapshot()` threw `.unauthorized` for that window, and the iroh activation token source collapsed every error to nil with `try?`. The broker then reported `missingAuthentication`, classified `authorizationFailed`, and endpoint activation failed closed with `endpointFailed` on every attempt until the revalidation settled and exponential backoff (2s/4s/8s) happened to retry after it. Net effect: ~15-20s stuck on "disconnected" at every app launch, visible in cmuxdiag exports from build 1.0.4 (20260731034828) as three consecutive `relayPolicyRefreshFailed`/`endpointFailed(authorizationFailed)` events 10ms after their start events (no network round trip involved).

The same window also mattered offline: an expired stored access token whose re-mint could not complete read as "signed out" and skipped the cached-offline-policy bootstrap that exists precisely for broker-unreachable activation.

## Mechanism

Three layers, one contract: **throw means "cannot read credentials right now" (transient), nil means "definitively signed out" (fail closed).**

1. `AuthCoordinator.authenticatedSessionSnapshot()` throws `.networkError` while a session transition owns the token store, matching `accessToken()`'s existing classification of the same state. `.unauthorized` remains for signed-out, account-changed, and generation-moved.
2. `CmxIrohBrokerTokenSource.credentialPair` is now throwing. `CmxIrohTrustBrokerClient.performRequest` maps a thrown token-source error to `.connectivity`, which every existing caller policy already treats as retryable: `retriesInitialActivation`, `preservesVerifiedPolicyDuringRefresh`, and `resolvePolicy`'s cached offline-policy bootstrap. nil still maps to `.missingAuthentication`.
3. The iOS activation token source and the Mac host activation token source both map `AuthError.unauthorized` to nil and rethrow everything else, instead of `try?`-collapsing both classes.

With this, an activation that races the revalidation window proceeds through the cached verified policy (the same path as an offline launch) instead of tearing down as unauthorized, and converges to fresh registration when the revalidation completes.

First commit adds the failing auth regression test only (CI red), second commit the fix (CI green), per regression policy.

## Tests

- `AuthCoordinatorSessionSnapshotRevalidationTests` (new, red then green): snapshot mid-revalidation throws `.networkError`, not `.unauthorized`.
- `CmxIrohTrustBrokerClientNetworkTests.throwingTokenSourceClassifiesAsConnectivityBeforeAnyNetworkRequest` (new): throwing source classifies `.connectivity` with zero network requests; existing nil-source test still pins `.missingAuthentication`.
- `MobileIrohRuntimeCompositionTests.activationBrokerCredentialsRethrowTransientTokenMiss` (new): the activation-pinned source rethrows a transient miss (refresh survives, access unresolvable) and still fails closed on account switch.
- Full local runs: CmuxAuthRuntime 174/174, CmuxIrohTransport 499/499.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9259"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788065793&installation_model_id=21920&pr_number=9259&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9259&signature=db21291e454823aaeb2be36c5064cc4e1e5976819142f120752af2a59d7c0dbd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Stop activation from failing closed at app launch by treating transient token misses as connectivity and by serving stored token pairs from the keychain immediately; also propagate cancellation from broker token reads so cancelled requests don’t retry. Merges connectivity v2 from `main` and keeps `authenticatedRefreshTokenSnapshot` for browser handoff, so activation uses cached policy and retries correctly.

- **Bug Fixes**
  - `AuthCoordinator.authenticatedSessionSnapshot()` now serves a persisted token pair directly from the keychain when available; otherwise it throws `.networkError` while `sessionTokenTransitionIsActive` and `.unauthorized` only when truly signed out or moved.
  - `CmxIrohBrokerTokenSource.credentialPair` is now throwing; `CmxIrohTrustBrokerClient` maps thrown reads to `.connectivity`, `nil` to `.missingAuthentication`, and rethrows `CancellationError`.
  - iOS and Mac activation token sources now map `AuthError.unauthorized` to `nil` and rethrow all other errors so transient misses classify as connectivity (no teardown).

- **Migration**
  - If you construct `CmxIrohBrokerTokenSource`, update `credentialPair` (and derived access/refresh reads) to be throwing.
  - Call sites must use `try` with `credentialPair()` and handle errors: map `.unauthorized` to `nil`, and rethrow others (let `CancellationError` propagate) so the broker reports `.connectivity` or cancellation.

<sup>Written for commit 7a3979f74ffcde86ede9e520012fe2721206283d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication recovery during session refresh and launch restore by serving valid persisted credentials without unnecessary delays.
  * Treated temporary token unavailability as a retryable connectivity issue instead of a sign-out.
  * Prevented broker requests when credentials cannot be retrieved due to connectivity failures.
  * Preserved secure handling for unauthorized sessions and account mismatches.
* **Tests**
  * Added regression coverage for session revalidation, transient token failures, and broker connectivity behavior.


azooz2003-bit:
Per Aziz: the snapshot now loads the token pair from the keychain immediately before any connection attempt, instead of queueing behind launch restore / foreground revalidation and classifying the window as retryable. `4ddeedf19d`: `authenticatedSessionSnapshot()` tries a keychain-only fast path first (stored access + refresh, coherence-bracketed by a refresh re-read, generation and account id pinned across the reads). Backend calls send both tokens and the server refreshes a stale access token itself, so the stored pair is sufficient to dial with. The fast path declines and falls back to the existing full path on an auth-environment-switch launch, while a sign-in exchange or sign-out capture owns the store, or when no complete pair exists — the transient classification from the earlier commits still covers those cases. CmuxAuthRuntime suite: 177/177 pass, including new coverage for launch-restore-parked, revalidation-parked, empty-store, and sign-in-owns-store paths. Note for the owning session: this changes runtime behavior, so the tokhyd tag needs a rebuild before its next dogfood round.

azooz2003-bit:
Heads-up: https://github.com/manaflow-ai/cmux/pull/9344 (credential authority) touches the same seam (`AuthCoordinator+Tokens`, broker token source). Semantically complementary — 9259 classifies the transition-owned store as retryable, 9344 adds single-flight rejection recovery — but whichever merges second needs a small textual rebase.

PR #9300, opened 2026-07-31
ci: reload-build.yml accepts iOS dev build origins

azooz2003-bit:
Adds `api_base_url` and `iroh_broker_base_url` inputs to reload-build.yml (iOS only). They export as `CMUX_IOS_API_BASE_URL` / `CMUX_IOS_IROH_BROKER_BASE_URL` and flow into the archive via `CMUX_API_BASE_URL` / `CMUX_IROH_BROKER_BASE_URL` build settings, which `ios/Config/Info.plist` bakes into the app. Both default to `https://cmux-staging.vercel.app`, the same default the build step already used, so dispatches that omit them are byte-identical to today.

Why: the iOS device dev-build cloud path was fleet-only because the Blacksmith workflow could not take dynamic origins. With these inputs, staging-default builds (the normal case) and explicit-origin builds both run on Blacksmith. The companion cmuxterm-hq PR with the dispatch plumbing and origin routing policy is https://github.com/manaflow-ai/cmuxterm-hq/pull/246. Merge order is safe in both directions: a dispatch naming a missing input fails and the dispatcher falls back to the fleet.

The simulator leg this PR originally added landed on `main` separately (5d0cba06c7), so the remaining diff is workflow inputs only: no app runtime code, no build link.


azooz2003-bit:
Verified the origin inputs end to end on this head (615f936b00): dispatched `platform=ios` with `api_base_url=https://cmux-origin-check.manaflow.dev` and `iroh_broker_base_url=https://cmux-broker-check.manaflow.dev` (run https://github.com/manaflow-ai/cmux/actions/runs/32055430480, success). Downloaded the artifact: both the device archive app and `simulator/cmux.app` bake `CMUXApiBaseURL` and `CMUXIrohBrokerBaseURL` to exactly those values in Info.plist. Omitted inputs fall back to the same staging default as before.

PR #9256, opened 2026-07-31
Guarantee bounded iOS foreground reconnect

azooz2003-bit:
## Summary
- Rebased and combined the reconnect stack from https://github.com/manaflow-ai/cmux/pull/9190, https://github.com/manaflow-ai/cmux/pull/9188, https://github.com/manaflow-ai/cmux/pull/9186, and close-attribution coverage from https://github.com/manaflow-ai/cmux/pull/9192.
- Validates pooled Iroh sessions during same-generation foreground activation, evicting closed or all-paths-unavailable sessions before the first post-wake reuse.
- Defers recovery triggers while the iOS scene is inactive and replays one pending recovery only after the app is active, which avoids dials launched into the inactive/background bounce.
- Preserves the dial gate and recovery throttle behavior from the stack, including cancelled dial outcomes, route-gated classification, presence recovery throttling, and all-paths-closed eviction.

## Regression structure
- Commit 1 adds the failing `CmuxIrohTransport` regressions only.
- Commit 2 lands the fix plus the rebased reconnect stack.

## Test plan
- `swift test --package-path Packages/Shared/CmuxIrohTransport`
- `swift test --package-path Packages/iOS/CmuxMobileRPC --filter MobileCoreRPCAbandonedConnectTests`
- `swift test --package-path Packages/iOS/CmuxMobileRPC --filter MobileRPCTransportConnectEventTests`
- `swift test --package-path Packages/Shared/CMUXMobileCore --filter DiagnosticLogTests`
- `swift test --package-path Packages/iOS/CmuxMobileShell --filter MobileShellForegroundResumeTests`
- `swift test --package-path Packages/iOS/CmuxMobileShell --filter MobilePresencePushRecoveryThrottleTests`
- `swift test --package-path Packages/iOS/CmuxMobileShell --filter MobileShellWorkspaceCreateTests`

---
[...]

PR #9301, opened 2026-07-31
Bound iOS iroh client retries with one shared backoff policy

azooz2003-bit:
## Why

https://github.com/manaflow-ai/cmux/pull/9269 removed every server-side quota and rate limit from the iroh broker, so nothing bounds a runaway iOS client anymore. Field evidence from the last two days shows both failure directions:

- Phones wedged in failure loops retried registration, discovery, and relay-policy every 2-10 seconds for 40+ hours (thousands of requests per day). That traffic previously self-sustained quota lockouts and kept the Vercel firewall tripping 429s for the whole account.
- The existing ladders overshoot on single transient blips: observed `retryScheduled` naps of 32-36 s and idle gaps of 33-65 s, because the default `CmxIrohRetrySchedule` is a host profile (first retry 30 s, cap 1 h). Recovery feels dead in the foreground.

Mechanism of the storm: a failed activation leaves `runtime == nil`, and every dial or `prepareForConnection` re-triggers a full reconcile, which re-runs registration + relay-policy + discovery against the broker with no client-side spacing. `CmxIrohBrokerBackpressureGate` arms floors only from server 429/Retry-After responses, which no longer exist.

## What

One shared, injectable backoff policy in `CmuxIrohTransport`:

- `CmxIrohReconnectBackoff`: decorrelated jitter (each delay drawn uniformly from `[floor, min(cap, previous * 3)]`), foreground profile floor 1 s / cap 30 s, `reset()` back to floor, server Retry-After honored as a lower bound (clamped to 24 h). Deterministic via a seedable SplitMix64 stream, so tests assert exact schedules.

Wired through the existing ladders without changing their success paths:

1. **Activation ladder** (`MobileIrohRuntimeComposition.reconcile`): a failed broker-bound activation arms the ladder and emits `retryScheduled` (code 43); reconciles inside the armed window skip broker work entirely and dials keep the existing `.inactive` error shape. The window clears on scenePhase-active, on network-path-change (new hook threaded through `startNetworkPathObservation`), on account switch, and on success, so a real state change always retries at the floor. Local precondition deferrals (`CmxIrohClientRuntimeError.inactive`) never arm it because they made no broker request.
2. **Relay-policy refresh loop** (same file): the failure delay now comes from the shared ladder instead of a fresh `CmxIrohRetrySchedule()` (first retry ~1-3 s instead of 30-37.5 s, capped at 30 s instead of 1 h).
3. **Relay credential coordinator**: the client runtime now constructs it with `CmxIrohRetrySchedule.foregroundClient` (1 s floor / 30 s cap, derived from the same foreground bounds); the coordinator's expiry-aware retry logic is untouched.
4. **Broker mutation pacing**: `CmxIrohBrokerBackpressureGate` reserves 2 s-spaced slots for `.registration` (challenge + signed register) through an injected sleep. It waits instead of failing, so a legitimate back-to-back re-activation still succeeds, and a wedged phone can never exceed ~30 registration attempts per minute even if every other guard fails.

Foreground + reachable can never see a locally scheduled nap beyond 30 s; only an explicit server Retry-After directive may exceed it.

## Tests

Two-commit regression: commit 1 adds tests that fail on `main`, commit 2 adds the fix.

- `cmuxFeatureTests.MobileIrohRuntimeCompositionCooldownTests`: frozen-clock proof that a failing activation makes broker-silent re-drives inside the window, `retryScheduled` ms ≤ 30 000, exactly one attempt after the cap elapses, scenePhase-active and network-path-change resets, and an exact seeded 10-minute schedule (attempt count and every scheduled nap match a same-seed twin ladder).
- `CmuxIrohTransportTests.CmxIrohReconnectBackoffTests`: identical schedules per seed, jitter bounds inside `[floor, cap]` and the decorrelated growth window, cap never exceeded without a server directive, reset returns to the floor window, Retry-After wins and is bounded.
- `CmuxIrohTransportTests.CmxIrohBrokerBackpressureGateTests`: registration slots spaced 2 s (globally, across accounts), a spinning wedge needs ≥ 60 simulated seconds for 31 attempts, read operations never paced.

```
swift test --package-path Packages/Shared/CmuxIrohTransport   # full suite: 507 tests, 63 suites, all passed
```

`cmuxFeatureTests` is iOS-only (its dependency graph does not build for macOS), so the composition tests run through the hosted simulator workflow with `test_filter=cmuxFeatureTests/MobileIrohRuntimeCompositionCooldownTests`:

- Red on the tests-only commit (737703dafe): https://github.com/manaflow-ai/cmux/actions/runs/30624142535 — `ios-simulator (iphone)` fails exactly at the new assertions (broker-silent re-drives, `retryScheduled` presence, one-attempt-after-cap, scenePhase reset).
- Green on head: https://github.com/manaflow-ai/cmux/actions/runs/30621750934 — the suite's 9 tests pass, including the exact seeded 10-minute schedule.

The `package-conventions-lint` and `mobile-core-package` job failures in those runs reproduce identically on plain `main` dispatches (https://github.com/manaflow-ai/cmux/actions/runs/30621623466, https://github.com/manaflow-ai/cmux/actions/runs/30621302620): a free function in `Packages/iOS/CmuxMobileShell/Sources/CmuxMobileShell/SecondaryControlAttemptPolicy.swift` and CmuxMobileShell test failures that predate this branch. Nothing in this PR touches those files.

## Deliberately left alone (PR 9256 seams and host side)

- `CmxIrohClientSessionPool`, `MobileShellComposite+ConnectionRecovery`, `MobileShellComposite+ReconnectRoutes`, `MobileCoreRPCSession`: wake-time session eviction and recovery coalescing belong to https://github.com/manaflow-ai/cmux/pull/9256; this PR bounds broker traffic underneath whatever cadence those loops choose.
- Host-side ladders (`CmxIrohHostRuntime`, `MobileHostIrohRuntime` in the mac app target) keep the 30 s / 1 h profile; the mac is not the runaway client.
- Residual: the relay-policy refresh loop still sleeps on wall-clock `Task.sleep`, so a scenePhase reset affects the next computed delay rather than interrupting an in-flight nap; that nap is now bounded at 30 s.

Dictionary: broker (the web API that registers device bindings and issues relay credentials), ladder (one independent retry loop with its own delay schedule), decorrelated jitter (backoff where each delay is drawn randomly from a window based on the previous delay, avoiding synchronized retry bursts), floor/cap (smallest and largest delay a ladder may schedule), Retry-After (server header naming the earliest allowed retry time), reconcile (the composition pass that tears down and re-activates the account runtime).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconnection behavior after connectivity subscription failures using bounded, jittered retry delays.
  * Successful connections now reset the retry delay.
  * Cancellation reliably stops reconnection attempts, including during delays.

* **Tests**
  * Added coverage verifying retry timing follows the expected backoff sequence and remains within the configured limit.


azooz2003-bit:
## Connectivity v2 re-map

`main` merged back into this branch cleanly (one conflict, in the cooldown test file, resolved by taking `main`'s superset). While this PR sat open, the connectivity v2 branch cherry-picked its fix commit verbatim (f9aa3d2a7b on `main`, same message/content as 7f4384acdd here) and adapted it to the new engine, so after the merge this branch's tree matched `main` exactly. The whole original contract survived v2:

- `CmxIrohReconnectBackoff`, `CmxIrohBrokerBackpressureGate` 2 s registration pacing, `CmxIrohRetrySchedule.foregroundClient`, and all their tests are byte-identical on `main`.
- The activation arming/skip in `MobileIrohRuntimeComposition.reconcile`, the scenePhase-active / network-path-change / account-switch / success resets, and the relay-policy refresh ladder wiring all live on in v2's reworked composition.
- The credential coordinator kept `.foregroundClient` at both client construction sites; v2 only retyped its supervisor to `any CmxIrohRelayEndpointControlling`.
- The old `CmxIrohClientSessionPool` retired-dial settle wait died with the pool; its replacement in `CmxConnectivityPeerSession.waitForRetiredDials` is a bounded drain deadline, not a retry ladder, and stays untouched.

One genuine reconciliation remained, and it is the new head commit here: v2 added its own private ladder in `CmxConnectivityInvalidationSubscriber.run()` (unjittered exponential 1, 2, 4 ... 60 s, hardcoded `ContinuousClock`, no reset semantics, cap above the 30 s foreground bound). The shared policy won: failed subscribes now draw decorrelated-jittered delays from an injected `CmxIrohReconnectBackoff` (1 s floor / 30 s cap), a served stream resets the ladder to its floor window (the jittered draw also spreads a fleet's re-subscribes when a service deploy closes every socket at once), and both the backoff and the sleep are injectable with source-compatible defaults, so neither the iOS composition nor the mac `ConnectivityInvalidationSubscriberCoordinator` call site changed. A seeded twin-ladder test pins the exact schedule.

Verified post-merge: `swift test --package-path Packages/Shared/CmuxIrohTransport` — 523 tests in 62 suites, all passed, including the new "failed subscribes follow the shared seeded reconnect ladder" test and all previously landed backoff/gate/composition suites.

Dictionary: ladder (one independent retry loop with its own delay schedule), decorrelated jitter (each delay drawn randomly from a window based on the previous delay, avoiding synchronized retry bursts), served stream (a WebSocket subscribe that delivered a frame or stayed healthy for a minute before closing), drain deadline (a bounded wait for in-flight work to settle, not a retry).

PR #9325, opened 2026-08-01
Fix iOS reload Simulator architecture

azooz2003-bit:
The iOS reload workflow archived the phone app successfully, then failed while linking an x86_64 Simulator binary because GhosttyKit only ships an arm64 Simulator slice.

Constrain the packaged Simulator build to arm64. The hosted runner and supported verification Simulators are Apple Silicon.

Verification:
- `node --test scripts/lib/mobile-attach.test.mjs` (37 passed)
- exact hosted phone archive plus Simulator artifact: https://github.com/manaflow-ai/cmux/actions/runs/30675611310 (passed)

### DIFF
--- a/.github/workflows/reload-build.yml
+++ b/.github/workflows/reload-build.yml
@@ -188,6 +188,7 @@ jobs:
             -configuration Debug \
             -destination 'generic/platform=iOS Simulator' \
             -derivedDataPath "$RUNNER_TEMP/cmux-ios-dd" \
+            ARCHS=arm64 \
             PRODUCT_BUNDLE_IDENTIFIER="$bundle_id" \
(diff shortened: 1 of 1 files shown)


azooz2003-bit:
Build-lane coordination for tag `simarc` (Claude, main pane): I am building the simarc macOS app on the fleet builder and will run the iOS leg (`ios/scripts/reload-cloud.sh --tag simarc`) right after. reload-build.yml cancels in-flight runs per tag+platform, so our earlier Blacksmith dispatches kept cancelling each other (runs 30683344800/30683397144/30683449914/30683552368). Codex: please do not dispatch further `simarc` reload builds; adopt the installed app at the local tag path once I post build evidence here.

PR #9336, opened 2026-08-01
Commit sidebar selection synchronously so rename shortcut targets the focused workspace

### LINKED ISSUE #9199: Cmd+Shift+R rename shortcut targets workspace group, not the focused workspace
## Bug

When a workspace inside a group is focused (e.g. "Swappa" group), pressing Cmd+Shift+R to rename opens the rename prompt for the parent workspace **group** (e.g. renaming "Group 3" / the group folder) instead of renaming the currently focused workspace itself.

## Steps to reproduce

1. Open a workspace that belongs to a workspace group in the sidebar (e.g. a workspace under the "Swappa" group).
2. Press Cmd+Shift+R.
3. Observe the rename prompt is scoped to the workspace group, not the individual workspace that's focused.

## Expected behavior

Cmd+Shift+R should rename the currently focused workspace, matching what happens when right-clicking that workspace and choosing "Rename" from the context menu.

## Actual behavior

The shortcut renames nothing useful for the workspace itself; the only way to rename the focused workspace is to right-click it directly and select "Rename" from the context menu. Right-clicking the workspace *group* folder and choosing "Rename" is a separate, correctly-working action.

## Screenshot

Screenshot attached shows the rename prompt ("Group 3") appearing over the terminal for the focused workspace, when Cmd+Shift+R was pressed while that workspace (not its group) was focused.

austinywang:
Fixes https://github.com/manaflow-ai/cmux/issues/9199

## Bug 1: rename shortcut targeted the group

With a workspace inside a group focused, Cmd+Shift+R opened the rename prompt for the parent workspace **group** instead of the focused workspace.

**Root cause:** plain sidebar clicks routed through `SidebarSelectionCoalescer`, which deferred the selection commit by up to 100 ms after a preceding click. Clicking a grouped child right after its group header left the header as the committed command target during that window, so the rename shortcut resolved against the group.

**Fix:** remove the coalescer entirely — plain clicks commit selection synchronously in `SidebarWorkspaceTableController`, the same path modifier clicks already used.

## Bug 2 (dogfood follow-up): group title disappeared after header → child clicks

Clicking the group header and then a grouped child left the group's title invisible (white-on-white) until an unrelated repaint.

**Root cause:** `NSColor.labelColor.withAlphaComponent(0.9)` resolves the dynamic color against the **ambient drawing appearance at call time** and returns a static color. With selection now committing synchronously inside the click dispatch, the header's authoritative reconfigure can run while a dark appearance (the swapped-in terminal view's) is current — the title snapshots the dark variant (white) and keeps it in the light sidebar. Verified with a standalone AppKit probe: `labelColor.withAlphaComponent(0.9)` called under darkAqua stays brightness 1.0 when later resolved under aqua, while plain `labelColor` resolves to 0.0.

**Fix:** the header title keeps the dynamic `labelColor` and dims via `alphaValue` (no appearance resolution at configure time); all layer-color conversions in `SidebarGroupHeaderTableCellView` and `SidebarWorkspaceRowCellView` resolve through the cell's own `effectiveAppearance`; the header re-resolves on `viewDidChangeEffectiveAppearance`. A DEBUG-only probe logs the resolved title paint to the debug event log for live diagnosis.

## Tests

Failing tests commit first, fix after, per the regression-test commit policy:

- `rapidGroupHeaderThenChildClickCommitsChildSelectionSynchronously` / `rapidChildThenGroupHeaderClickCommitsAnchorSelectionSynchronously` — real `SidebarWorkspaceTableController` row clicks asserting the command target synchronously after the click (bug 1's exact repro path).
- `groupHeaderTitleIgnoresAmbientAppearanceAtConfigureTime` / `anchorActiveGroupHeaderTintResolvesAgainstCellAppearance` — configure the header cell under a darkAqua ambient appearance and assert the title/tint still resolve dark in a light appearance (bug 2's mechanism).

## Notes

- Localization audit: no user-facing strings added or changed (behavioral + comment changes only).
- Rebased onto main after #9334/#9338 fixed the main compile break.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar selection when rapidly switching between workspace groups and child rows.
  * Prevented temporary text-color flicker during selection updates.
  * Cleared deferred clicks during double-click and drag operations.
  * Improved sidebar colors and text appearance across different interface themes.

* **Tests**
  * Added coverage for selection commits and appearance-aware sidebar rendering.


austinywang:
Live verification of the disappearing-title mechanism and fix, from the tagged dev build's debug probe (`sidebar.groupHeader.titlePaint`), while switching between the group anchor and a grouped child:

```
22:37:25.981 name=Issue 9199 G active=0 ambientLabelBrightness=1.00 effective=NSAppearanceNameAqua brightness=0.00 colorAlpha=0.85 fieldAlpha=0.90 frameW=160
22:37:26.551 name=Issue 9199 G active=1 ambientLabelBrightness=1.00 effective=NSAppearanceNameAqua brightness=0.00 colorAlpha=0.85 fieldAlpha=1.00 frameW=160
22:37:27.121 name=Issue 9199 G active=0 ambientLabelBrightness=1.00 effective=NSAppearanceNameAqua brightness=0.00 colorAlpha=0.85 fieldAlpha=0.90 frameW=160
```

`ambientLabelBrightness=1.00` shows the ambient drawing appearance is genuinely **dark** at header configure time on every selection transition (the synchronously swapped-in terminal's context) — with the old `labelColor.withAlphaComponent(0.9)` this snapshotted a static white title on a light sidebar. With the fix the drawn title resolves through the cell's own appearance (`brightness=0.00`, black, visible) and dims via `fieldAlpha`.

Startup entries also caught the cell's own `effectiveAppearance` transiently reporting `darkAqua` before settling to `aqua`; the `viewDidChangeEffectiveAppearance` re-apply covers that window.

PR #9405, opened 2026-08-02
Restore Markdown scroll per generated projection

JunfengJia:
## Problem

A live Markdown panel can display multiple logical documents through one watched file path. cmux currently carries the outgoing document's scroll state into the incoming document, so returning to a tab can show a later section instead of that tab's prior viewport.

## Fix

- accept an invisible `cmux-scroll-key` on generated Markdown projections
- retain a bounded LRU of viewport snapshots by key
- keep existing scroll preservation for ordinary edits and unkeyed Markdown
- start a never-seen keyed projection at the top and restore its own viewport when revisited

`agent-state` emits a stable key from the selected surface and task in dotfiles commit `e99e546`.

## Red / green proof

The commits intentionally follow the regression policy:

1. `28353f6f96` adds the WebKit behavior test only.
2. `f97da3f109` adds the fix.

On the same Mac and viewer assets, an off-screen WebKit harness produced:

```text
RED: surface B inherited surface A scroll (6805.0)
GREEN: surface A restored 6805.0 after A -> B -> A
```

Additional verification:

- Markdown shell JavaScript syntax check passed.
- Dotfiles adapter frozen suite passed 51/51 tests.
- Independent cross-repo review was clean.
- `git diff --check` passed.

Full `cmux-unit` compilation and the required tagged `reload.sh` build are not available on this Mac because it currently has Command Line Tools but no full Xcode installation. The PR remains draft until a tagged build can be produced and dogfooded.

## Localization

No user-facing strings were added or changed.

## Summary by CodeRabbit

* **New Features**
  * Markdown documents can now preserve scroll positions independently across different views or projection keys.
  * Returning to a previously viewed key restores its saved position, while new keys begin at the top.
  * Up to 32 recent scroll positions are retained automatically.

* **Bug Fixes**
  * Improved scroll-position handling when document content or the active view changes.

### DIFF
--- a/Resources/markdown-viewer/shell.html
+++ b/Resources/markdown-viewer/shell.html
@@ -1259,6 +1259,58 @@ html { scroll-behavior: smooth; }
     });
   }
 
+  var activeMarkdownScrollKey = null;
+  var markdownScrollStatesByKey = new Map();
+  var maxRememberedMarkdownScrollKeys = 32;
+
+  function scrollKeyForMarkdown(md) {
+    var match = String(md || '').match(/^\s*<!--\s*cmux-scroll-key:\s*([^\r\n]*?)\s*-->/i);
+    if (!match) { return null; }
+    var key = String(match[1] || '').trim();
+    return /^[A-Za-z0-9._:/-]{1,256}$/.test(key) ? key : null;
+  }
+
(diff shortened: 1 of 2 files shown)


JunfengJia:
@coderabbitai review

PR #9425, opened 2026-08-03
Fix cmux ssh startup script syntax error in no-progress retry loop

### LINKED ISSUE #9423: cmux ssh: shell syntax error in startup script on 0.64.21 (env prefix before `if`)
### cmux version

0.64.21 (101) [33ac210ab]

### macOS version

macOS 26.1 (Build 25B78)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Direct download (DMG)

I could not test NIGHTLY

### Bug description

`cmux ssh <host>` fails immediately on **0.64.21** with a POSIX shell syntax error while executing the generated startup script under `/var/folders/.../T/cmux-ssh-startup.*`.

```text
/var/folders/.../T//cmux-ssh-startup.XXXXXX: line 291: syntax error near unexpected token `then'
/var/folders/.../T//cmux-ssh-startup.XXXXXX: line 291: `  CMUX_SSH_PTY_ATTACH_NO_PROGRESS_RETRY="$cmux_ssh_attach_no_progress_retry" CMUX_SSH_PTY_ATTACH_NO_PROGRESS_LIMIT="$cmux_ssh_attach_no_progress_limit" if [ "$cmux_ssh_attach_no_progress_retry" -gt 0 ]; then cmux_ssh_begin_attempt || exit 1; fi'
```

This is a local script-generation bug (not a remote host / PTY allocation failure). Downgrading to **0.64.20** avoids the bad path.

### Expected behavior

`cmux ssh <host>` should start the SSH / persistent PTY attach session without a shell parse error.

### Steps to reproduce

1. Install cmux **0.64.21**.
2. Run `cmux ssh <host>` against a host that uses the foreground-auth + SSH PTY attach startup path (e.g. a host that goes through `buildReusableForegroundAuthThenSSHPTYAttachStartupCommand`).
3. Observe the startup script fail with `syntax error near unexpected token 'then'`.

### Shell and environment

zsh on macOS; error is from `/bin/sh` executing the generated `cmux-ssh-startup.*` script.

Generated line (invalid POSIX):

```sh
CMUX_SSH_PTY_ATTACH_NO_PROGRESS_RETRY="$cmux_ssh_attach_no_progress_retry" CMUX_SSH_PTY_ATTACH_NO_PROGRESS_LIMIT="$cmux_ssh_attach_no_progress_limit" if [ "$cmux_ssh_attach_no_progress_retry" -gt 0 ]; then cmux_ssh_begin_attempt || exit 1; fi
```

Environment-variable prefixes may only precede a **simple command**. Prefixing a compound `if` is a syntax error.

### Root cause (source)

In `CLI/cmux.swift` (`buildReusableForegroundAuthThenSSHPTYAttachStartupCommand`):

```swift
let registeredAttachAttemptCommand = [
    "if [ \"$cmux_ssh_attach_no_progress_retry\" -gt 0 ]; then cmux_ssh_begin_attempt || exit 1; fi",
    attachAttemptCommand,
].joined(separator: "\n")
let attachScript = SSHPTYAttachExitCode.noProgressRetryLoopLines(
    command: registeredAttachAttemptCommand
).joined(separator: "\n")
```

`SSHPTYAttachExitCode.noProgressRetryLoopLines(command:)` interpolates `command` onto the same line as env assignments:

```swift
"  CMUX_SSH_PTY_ATTACH_NO_PROGRESS_RETRY=\"...\" CMUX_SSH_PTY_ATTACH_NO_PROGRESS_LIMIT=\"...\" \(command)"
```

Because `registeredAttachAttemptCommand` starts with an `if`, the first generated line becomes `VAR=... VAR=... if ...; then ...; fi`, which `/bin/sh` rejects.

This appears new in the **0.64.21** release (related to the no-progress PTY attach retry work).

### Suggested fix

Any of:

1. Wrap the multi-line attach attempt in a shell function / simple command name, then pass that name to `noProgressRetryLoopLines` (same pattern as `SSHPTYAttachStartupCommandBuilder`, which uses `command: "cmux_ssh_attach_attempt"`).
2. Change `noProgressRetryLoopLines` to `export` the env vars on a separate line before running `command`, so compound commands are legal.
3. At minimum, do not pass a compound `if` as the first line of `command` when it will be prefixed with `VAR=value`.

### Additional context

- Broken: 0.64.21 (101) [33ac210ab]
- Works again after downgrade: 0.64.20 (100) [14e3400b9]
- Workaround: stay on 0.64.20, or use plain `ssh` until fixed.

austinywang:
`cmux ssh <host>` failed immediately on 0.64.21 with `syntax error near unexpected token 'then'` from the generated `cmux-ssh-startup.*` script.

`SSHPTYAttachExitCode.noProgressRetryLoopLines` prefixed the caller's command with environment assignments on the same line. POSIX only allows a variable-assignment prefix before a simple command, and the reusable foreground-auth attach path passes a compound `if ...; then ...; fi`, so the generated line was a parse error under /bin/sh.

The fix assigns and exports `CMUX_SSH_PTY_ATTACH_NO_PROGRESS_RETRY`/`_LIMIT` on their own lines before running the command, so any command shape is legal and child processes still see the budget. That fixes every caller of the loop rather than reshaping one call site.

Fixes #9423

## Testing

- New Swift Testing case `noProgressLoopAcceptsCompoundCommands` in `cmuxTests/SSHPTYAttachNoProgressRetryTests.swift`: generates the loop around a compound command, asserts `/bin/sh -n` accepts the script, runs it, and asserts the command still observes `0/3` for retry/limit.
- Red then green in the Commits tab: the test lands before the generator fix.
- Shell semantics verified locally: `X=1 if [ 1 -gt 0 ]; then :; fi` fails `/bin/sh -n` with exactly the reported error; the new form passes `/bin/sh -n` and exports the budget into an inner `/bin/sh -c` under both bash-as-sh and dash.
- `swift build` clean for `Packages/macOS/CmuxFoundation`; local Aziz policy check clean.
- Tagged dev build `sym9423` built from this branch head. An end-to-end `cmux ssh` run needs a reachable SSH host, which this machine has no sshd for; the script-generation path is covered by the test above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSH attach retry handling for compound commands.
  * Ensured retry state is correctly exported and preserved across attempts.
  * Added validation to prevent shell syntax issues in generated retry commands.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -10483,13 +10483,15 @@ struct CMUXCLI {
             remoteShellCommand: remoteShellCommand
         )
         let attachAttemptCommand = "/bin/sh -c \(shellQuote(attachAttemptScript))"
-        let registeredAttachAttemptCommand = [
-            "if [ \"$cmux_ssh_attach_no_progress_retry\" -gt 0 ]; then cmux_ssh_begin_attempt || exit 1; fi",
-            attachAttemptCommand,
-        ].joined(separator: "\n")
-        let attachScript = SSHPTYAttachExitCode.noProgressRetryLoopLines(
-            command: registeredAttachAttemptCommand
-        ).joined(separator: "\n")
+        // The no-progress retry loop prefixes this command with env assignments, and a
+        // variable-assignment prefix is only legal before a simple command. Wrap the
+        // registration plus the attach attempt in a shell function and pass its name,
+        // matching SSHPTYAttachStartupCommandBuilder.
+        let registeredAttachAttemptFunction = "cmux_ssh_attach_no_progress_attempt"
+        let registeredAttachAttemptDefinition = "\(registeredAttachAttemptFunction)() { if [ \"$cmux_ssh_attach_no_progress_retry\" -gt 0 ]; then cmux_ssh_begin_attempt || return 1; fi; \(attachAttemptCommand); }"
+        let attachScript = ([registeredAttachAttemptDefinition] + SSHPTYAttachExitCode.noProgressRetryLoopLines(
+            command: registeredAttachAttemptFunction
+        )).joined(separator: "\n")
         var authScriptLines: [String] = []
         let sharingOptions = SSHConnectionSharingOptions()
         let authenticationLockPath = sharingOptions.foregroundAuthenticationLockPath(
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -10483,15 +10483,13 @@ struct CMUXCLI {
             remoteShellCommand: remoteShellCommand
         )
         let attachAttemptCommand = "/bin/sh -c \(shellQuote(attachAttemptScript))"
-        // The no-progress retry loop prefixes this command with env assignments, and a
-        // variable-assignment prefix is only legal before a simple command. Wrap the
-        // registration plus the attach attempt in a shell function and pass its name,
-        // matching SSHPTYAttachStartupCommandBuilder.
-        let registeredAttachAttemptFunction = "cmux_ssh_attach_no_progress_attempt"
-        let registeredAttachAttemptDefinition = "\(registeredAttachAttemptFunction)() { if [ \"$cmux_ssh_attach_no_progress_retry\" -gt 0 ]; then cmux_ssh_begin_attempt || return 1; fi; \(attachAttemptCommand); }"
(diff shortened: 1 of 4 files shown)


austinywang:
Addressing CodeRabbit's description-template warning: this is a non-visual CLI shell-generation fix, so a demo video is not applicable. The PR body documents the red/green regression, local shell verification, and the generator-level invariant. Automated review bots completed, and CodeRabbit reported no actionable code findings.

PR #9426, opened 2026-08-03
Add Code workspace surface

lawrencecchen:
Adds a bundled Code workspace surface to the Plus menu.

- Loads only the actual bundled client. There is no placeholder shell, fake composer, or renderer handoff.
- Starts the TypeScript sidecar when the webview mounts.
- Routes local fetch and socket traffic through the native WebKit message bridge. The renderer receives no loopback address or credential.
- Keeps Cmd T, Cmd D, Cmd Shift D, Cmd N, and Cmd Shift N in Code context.
- Prewarms four fully connected actual clients in the current window, withdraws any client that reconnects, then moves the claimed client into the destination pane.
- Applies the live Ghostty palette, transparency, 14 px SF Pro Text, and SF Mono, including configuration reloads.
- Starts the internal sidebar closed while retaining its toggle.
- Persists Code surface identity across sessions and Dock layouts.

Verified in the cw32 tagged app: Plus presented in 16 ms, and Cmd D, Cmd T, and Cmd N each presented in 1 ms. Every inspected surface had zero placeholder-shell nodes and zero connection alerts. The packaged HTML body contains only the actual root. All 238 TypeScript tests, type checking, lint, Swift parsing, and the cloud app build pass. The focused Swift test run was infrastructure-blocked before execution because no fleet Mac passed the GUI/toolchain preflight.


lawrencecchen:
Latest startup pass:

- The final cw20 binary painted the complete inert shell in 85 ms. The runtime remained inactive and the sidecar process remained absent.
- The renderer now contains empty hosted-login and tracing values. Native fetch and socket bridges retain the real loopback bearer token outside WebKit.
- 241 webview tests passed, with TypeScript, lint, Swift parsing, generated assets, packaged-assets security checks, localization JSON, project-file validation, and the cloud app build passing.
- The focused Swift test request could not start. Three cloud Macs failed their environment preflight and the remaining Macs stayed occupied.
- Final Plus-menu presentation is unverified because Computer Use returned cgWindowNotFound for the tagged app and unrelated apps. Earlier menu and shortcut coverage predates the final portal activation patch.

Dictionary: first paint is the first rendered non-background content; preflight is the runner environment check before project tests start.

Issue #4529, opened 2026-05-22
[Bug] Severe memory leak leading to OOM (Out of Memory) on macOS


hangvlog:
### cmux version

latest

### macOS version

14.6.1 (23G93)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Can you reproduce this on cmux NIGHTLY?

Yes, it still reproduces on NIGHTLY

### Bug description

I have noticed a severe memory leak issue while using cmux on macOS. The application's memory usage continuously increases over time without releasing it. Eventually, it consumes all available system RAM (reaching tens of gigabytes), causing the system to freeze, swap heavily, or trigger an Out of Memory (OOM) crash.

### Expected behavior

The memory footprint of cmux should remain relatively stable during normal usage and properly release memory when it's no longer needed. It should not infinitely consume RAM.

### Steps to reproduce

The memory footprint of cmux should remain relatively stable during normal usage and properly release memory when it's no longer needed. It should not infinitely consume RAM.

### Relevant logs or crash reports

```text

```

frinzekt:
Duplicate of #4520  ?

liudp1988:
<img width="1920" height="1280" alt="Image" src="https://github.com/user-attachments/assets/6db5146e-787b-4267-a9b9-1864e97115c0" /> Memory keeps growing, reaching 80G — the system goes OOM.

liudp1988:
Memory keeps growing, reaching 80G — the system goes OOM.

wuyang630:
I am experiencing the same issue. 

**cmux version:** 0.64.8
**macOS version:** 15.7.5 (24G624)
**Mac chip:** Apple Silicon M3 (24GB RAM)

YHM404:
same issue on my mac.

hyqCrystal:
<img width="390" height="179" alt="Image" src="https://github.com/user-attachments/assets/45c5e8dd-c50b-43be-958f-4b885164c6a5" />

lifejwang11:
me too

Orchardxyz:
The same issue on my Mac.

- cmux version: 0.64.8
- macOS version: 15.6.1
- Mac chip: Apple M3 Pro (arm64)

yuanyuan06:
me too

Yaozhuwa:
same, M4 MacBook Pro

Lambor-Chai:
+1

lkangd:
[...]

PR #9541, opened 2026-08-04
Fix the Sleepy Mode lock button on macOS 26

gianlocc:
## The bug

Sleepy Mode's **Lock Mac** button does nothing on macOS 26. No lock, no error, nothing in the logs.

## Root cause

`SleepyPowerControls.lockMacNow()` shelled out to:

```
/System/Library/CoreServices/Menu Extras/User.menu/Contents/Resources/CGSession -suspend
```

Apple removed `User.menu` from `Menu Extras`. On macOS 26.5.2 it is absent — the folder still holds `AirPort.menu`, `Eject.menu` and friends, but there is no `CGSession` anywhere under `CoreServices`.

It failed *silently* because `SystemCommandRunner.run()` used `try? process.run()` with stdout and stderr both routed to `/dev/null` and no exit-status check. A missing executable throws, `try?` discards it, and the function returns as though the lock happened.

Per the macOS-version note in `CLAUDE.md`, the AWS builders run macOS 15.7.4, where `User.menu` still exists — so this works on CI and on maintainer machines and breaks for users on 26. Same shape as #4529.

## The fix

Choose the mechanism by **capability, not OS version**: use the supported `CGSession` tool wherever it still ships, and fall back to `login.framework`'s `SACLockScreenImmediate` only when it is absent. I verified the symbol resolves on 26.5.2 (lookup only, not called), using the same `dlsym` approach `SystemCommandRunner` already uses for `AuthorizationExecuteWithPrivileges`.

Probing beats a version threshold: the exact release Apple dropped the tool in isn't something to guess at, and a wrong threshold would break every version in between.

Also:
- `run()` now reports whether the launch succeeded and logs failures, so a missing system tool can't present as a working button again.
- `lockMacNow()` returns `Bool`, and the scene shows "Couldn't lock the Mac" (localized en + ja) instead of looking inert.

## Tests

`cmuxTests/SleepyLockMechanismTests.swift` covers: prefers the supported tool when present, falls back when the tool is missing, falls back when it exists but won't launch, and reports failure when nothing is available. Wired into the pbxproj (`lint-pbxproj-test-wiring.sh` passes).

Deviates from the two-commit regression policy: the fix changes the `SleepyCommandRunning` protocol the test drives, so a test-only first commit wouldn't compile.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9541"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788419377&installation_model_id=21920&pr_number=9541&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9541&signature=e0f6fb2778e4191125d4a5abd6ad49c5df0e563091b5b3b1456c3e50f3c25e2a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes the Sleepy Mode "Lock Mac" button on macOS 26 by preferring `CGSession -suspend` when available, awaiting a zero exit, and falling back to the system login lock; the lock attempt is now owned and cancellable to avoid stale UI after teardown.

- **Bug Fixes**
  - Prefer `CGSession -suspend` when present and require exit status 0; otherwise call `SACLockScreenImmediate` on the main actor.
  - `SystemCommandRunner`: added `canRun()`, `runAwaitingExit()`, `lockScreen()`; `run()` returns Bool and logs via `OSLog`; uses `Process.terminationHandler` (no blocked threads); `SleepyCommandRunning` methods are `nonisolated`.
  - Shared `SleepyPowerUIState` owns a single in-flight lock across displays (`isLocking`, `lockFailed`), disables the button, and cancels on controller teardown so late completions don’t surface in a new session; shows "Couldn't lock the Mac — use the system menu › Lock Screen" localized in 20 locales.
  - Added `SleepyLockMechanismTests` for prefer/fallback behavior, non-zero exit handling, and `-suspend` arg assertion.

<sup>Written for commit ae1e563ab6741e07f5d7b17cea9b7a49d1db3307. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9541?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Lock Mac now uses an available fallback when the primary locking method is unavailable.
  * Sleepy Mode reports successful and unsuccessful lock attempts.
  * Overlapping lock attempts are prevented.

* **Bug Fixes**
  * Lock failures are no longer silently ignored.
  * Added a localized warning with Apple menu → Lock Screen instructions in 20 languages.
  * Improved handling of unavailable or unsuccessful locking methods.

### DIFF
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -229001,6 +229001,23 @@
         }
       }
     },
+    "sleepyMode.lockFailed": {
+      "extractionState": "manual",
+      "localizations": {
+        "en": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Couldn't lock the Mac \u2014 use the Apple menu \u203a Lock Screen"
+          }
+        },
+        "ja": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Mac \u3092\u30ed\u30c3\u30af\u3067\u304d\u307e\u305b\u3093\u3067\u3057\u305f \u2014 Apple \u30e1\u30cb\u30e5\u30fc \u203a \u300c\u753b\u9762\u3092\u30ed\u30c3\u30af\u300d\u3092\u4f7f\u3063\u3066\u304f\u3060\u3055\u3044"
+          }
+        }
+      }
+    },
     "sleepyMode.mascot.cat": {
       "extractionState": "manual",
       "localizations": {
--- a/Resources/Localizable.xcstrings
+++ b/Resources/Localizable.xcstrings
@@ -229004,16 +229004,124 @@
     "sleepyMode.lockFailed": {
       "extractionState": "manual",
       "localizations": {
+        "ar": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "تعذّر قفل الـ Mac — استخدم قائمة النظام › قفل الشاشة"
+          }
+        },
+        "bs": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Nije moguće zaključati Mac — koristite sistemski meni › Zaključaj ekran"
+          }
+        },
+        "da": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Kunne ikke låse din Mac — brug systemmenuen › Lås skærm"
+          }
+        },
+        "de": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Mac konnte nicht gesperrt werden — verwende das Systemmenü › Bildschirm sperren"
+          }
+        },
         "en": {
           "stringUnit": {
             "state": "translated",
-            "value": "Couldn't lock the Mac \u2014 use the Apple menu \u203a Lock Screen"
+            "value": "Couldn't lock the Mac — use the system menu › Lock Screen"
+          }
+        },
+        "es": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "No se pudo bloquear el Mac — usa el menú del sistema › Bloquear pantalla"
+          }
+        },
+        "fr": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Impossible de verrouiller le Mac — utilisez le menu système › Verrouiller l'écran"
+          }
+        },
+        "it": {
+          "stringUnit": {
+            "state": "translated",
+            "value": "Impossibile bloccare il Mac — usa il menu di sistema › Blocca schermo"
           }
         },
         "ja": {
           "stringUnit": {
             "state": "translated",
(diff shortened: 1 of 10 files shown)


gianlocc:
**Wait for `CGSession` to exit (Major)** — fixed. Added `runAwaitingExit` and used it for the legacy tool, so `lockMacNow()` only reports success on exit status zero and otherwise falls through to `SACLockScreenImmediate`. You're right that launch success wasn't evidence of a lock; that was a real hole in the fallback.

**Shared Lock Mac state** — fixed. Moved `isLocking` / `lockFailed` onto the shared `SleepyPowerUIState`, which already exists for exactly this reason on the Low Power button, and disabled every Lock Mac button while an attempt is in flight. This also brings it in line with the shared-behavior rule about not letting each surface keep its own optimistic copy.

**Assert the `CGSession` arguments** — fixed. The test double now records tool *and* args, both the success and fallback paths assert `["-suspend"]`, and there's a new case for the tool existing but exiting non-zero. Good catch that launching `CGSession` bare would exit cleanly while locking nothing.

**Add 18 more locales for `sleepyMode.lockFailed`** — skipping, I think this one's wrong for this repo. All 63 existing `sleepyMode.*` keys carry exactly `en` + `ja`, as do 2855 of the catalog's 4914 keys; the wider locale sets are on older strings. The localization skill also states the supported set is "currently English and Japanese". Adding 18 machine translations for this one key would make it inconsistent with every neighbouring string. Happy to add them if the policy has actually changed.

Issue #9615, opened 2026-08-05
events.stream replay is boot-scoped while the event log is durable: a subscriber that reconnects across a restart silently loses the gap


austinywang:
Split out of https://github.com/manaflow-ai/cmux/issues/9611, which deliberately left this alone.

Events are appended durably to `~/.cmuxterm/events.jsonl` by `CmuxEventLogWriter`, capped at 16 MB (`Sources/CmuxEventBus.swift:118`, `:338`). What `events.stream` serves for replay is not that file: it is the in-memory `retained` buffer of 4,096 events (`:116`), keyed to a per-launch `bootId` (`:134`).

Two consequences.

Sequence-based resume cannot cross a restart. A subscriber that reconnects with `after_seq` after cmux has relaunched is talking to a fresh `bootId` and a sequence counter that restarted at 1. The gap-detection in `subscribe` compares against the new in-memory window, so the subscriber is not told that everything between its last sequence and the restart is unavailable. It silently loses that span.

The durable record and the replayable record disagree. Events the log file still holds are unreplayable, which is exactly backwards for a record whose purpose is answering "was this recorded". https://github.com/manaflow-ai/cmux/issues/9611 has now put a `caller` identity on every socket-sourced event so automated input is attributable; that attribution is only worth as much as the ability to read the record back after a restart.

Worth deciding two things:

1. Should replay be backed by `events.jsonl` rather than the in-memory buffer, with `boot_id` becoming a segment marker in a continuous stream instead of a hard resume boundary? At minimum, a reconnect that crosses a boot should report the gap explicitly rather than resuming as if nothing was missed.
2. Is a 16 MB rolling cap the right retention? On one developer machine that file is 13 MB with 13,378 lines spanning many boots, so the cap is close to actively discarding history.

WTF-Am-ID:
Third independent confirmation of the 4,096 window, and a note on what it costs
a long-running subscriber, from 0.64.22 (102) on macOS.

Two acks taken minutes apart today:

```
--after 1     → gap:true,  gap_reason:"requested sequence is older than the
                 retained in-memory event log", oldest_seq:41227, latest_seq:45322
--after 45000 → gap:false, oldest_seq:41233, latest_seq:45328, replay_count:328
```

`latest_seq − oldest_seq` is 4095 in both, matching `Sources/CmuxEventBus.swift:116`.

The practical consequence, which I think strengthens the case for question 2 in
your body: the window is measured in events, so its duration collapses exactly
[...]

PR #9616, opened 2026-08-05
Record a caller identity on socket-sourced events

### LINKED ISSUE #9611: Record a caller identity on surface.input_sent (and other socket-sourced events): today only the transport is logged, so cmux send is unattributable
`surface.input_sent` records that text was injected into a surface, but not who injected it:

```json
{"name": "surface.input_sent",
 "payload": {"method": "surface.send_text",
             "params": {"surface_id": "...", "text": null,
                        "text_length": 22, "redacted_fields": ["text"]}},
 "source": "socket.v2"}
```

`source` is the transport, not a caller. No pid, no process name, no originating surface. Compare `agent.hook.PreToolUse`, which carries `session_id`, `cwd`, `tool_name`, and the originating `surface_id`: agent actions are attributable, socket actions are not.

That matters because `cmux send` aimed at a coding-agent session is indistinguishable from the human having typed it, both to the agent at the time and to anyone reading the record afterwards. A rule like "automated input is permitted, provided it is recorded" is not expressible today, because the record cannot answer who sent it.

**Ask:** put a caller identity on `surface.input_sent` (peer pid, process name, and the originating surface id when the caller is itself running inside a cmux surface).

The plumbing is already most of the way there. The control socket captures the peer pid at accept time via `LOCAL_PEERPID` (`Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Server/ControlConnection.swift:15`, `Transport/SocketTransport+Peer.swift:22`), and that pid is in scope in `TerminalController.swift` at the exact point it calls `publishSocketEvents(command:response:)` (`Sources/TerminalController.swift:1651`, pid resolved at `:1576`). What is missing is that `publishSocketEvents` (`Sources/CmuxEventStream.swift:143`) and `CmuxSocketEventMapper.publish(command:response:)` (`Sources/CmuxSocketEventMapper.swift:4`) drop it: neither takes a caller parameter, so `CmuxEventBus.publish` never sees one. Attribution should land on every socket-sourced event, not just `surface.input_sent`, since the same gap covers `surface.key_sent`, `notification.requested`, and the sidebar mutations.

Process name should be resolved from the pid at publish time, not trusted from the client, and the originating surface id should come from the caller's own connection context rather than a self-reported field, or the identity is spoofable by the thing it is meant to attribute.

**Two smaller notes from the same investigation.**

Keystroke injection is more legible than text injection. `surface.send_key` maps with `params: .unchanged` (`Sources/CmuxSocketEventMapper.swift:100`) so the key is recorded in full, while `surface.send_text` maps with `.redactedInput` (`:98`) and nulls the text down to a `text_length`. Whatever the right redaction policy is, the two should agree.

Durability is partly better than it looks. Events are appended to `~/.cmuxterm/events.jsonl` by `CmuxEventLogWriter`, capped at 16 MB (`Sources/CmuxEventBus.swift:118`, `:338`); on this machine that file is 13 MB with 13,378 lines spanning many boots. What is boot-scoped is the in-memory replay buffer that `events.stream` serves: 4,096 retained events (`:116`) keyed to a per-launch `bootId` (`:134`), so sequence-based resume cannot cross a restart and a subscriber that reconnects after one silently loses the gap. Worth deciding whether replay should be backed by the log file, and separately whether a 16 MB rolling cap is the right retention for a record meant to answer "was this recorded".

austinywang:
## Summary
- Every `socket.v1` and `socket.v2` event now carries a `caller` object: `pid`, `process_name`, and `surface_id`. Before, `source` recorded only the transport, so `cmux send` aimed at a coding-agent session was indistinguishable from the human having typed it.
- All three fields are resolved server-side from kernel state: `LOCAL_PEERPID` on the accepted connection, `proc_pidpath` for the name, and the caller's controlling terminal matched against live Ghostty PTYs (the same `.controllingTTY` resolution agent hooks use) for the surface. Nothing is read from the request, since a self-reported caller field would be spoofable by exactly the automation the identity exists to attribute.
- Unresolved fields are `null`, never absent and never guessed, so "we looked and could not tell" is distinguishable from "we did not look". Non-socket producers omit `caller` entirely rather than emitting an empty one.
- Attribution covers all socket-sourced events, not just `surface.input_sent`: `surface.key_sent`, `notification.requested`, and the sidebar mutations included.

## Hot path cost
The identity is passed as a provider, not a value, and memoized per connection. A command that maps to no event resolves nothing; a connection that publishes resolves once regardless of how many events it emits, because the peer pid is fixed for the connection's lifetime. `cmux send` therefore pays one bounded-cache pid lookup plus one main hop, on top of the main hop the command itself already takes.

The pid to name lookup is `proc_pidinfo` + `proc_pidpath`, never a subprocess. Results are memoized in a cache bounded at 512 entries and keyed by pid **and** process start time, so a recycled pid cannot inherit a previous process's name and a machine with churning callers cannot grow the socket path's memory.

## Redaction asymmetry: direction chosen
The issue notes `surface.send_key` mapped `.unchanged` and logged the key in full while `surface.send_text` mapped `.redactedInput` and nulled the text. **I kept the key and kept the text redacted**, under one stated rule that decides both: free-form caller text is redacted to a length, bounded control vocabulary is recorded.

`send_text`'s `text` is arbitrary user content and can carry a secret, so it stays `.redactedInput`. `send_key`'s `key` is drawn from a fixed named-key table that the server validates before the command succeeds (an unknown key returns an error, and events are only published on success), so a `key` on an emitted event is always something like `ctrl-c`. It cannot contain a secret, and it is the entire diagnostic value of the record: redacting it would make "did an automation send ctrl-c to my agent?" unanswerable, which is the same class of question this PR exists to make answerable.

To make the agreement visible rather than accidental, `send_key` now uses a new `.boundedInput` mapping that emits `redacted_fields: []`. A reader can tell "redaction was considered and nothing needed it" apart from a method that never considered redaction at all.

## Out of scope
The replay-durability question from the issue (boot-scoped in-memory replay vs the durable `~/.cmuxterm/events.jsonl`, and whether a 16 MB rolling cap is right retention) is deliberately not touched here. Filed separately as https://github.com/manaflow-ai/cmux/issues/9615.

## Testing
- `cmuxTests/CmuxSocketCallerIdentityTests.swift`, 11 behavior-level tests that publish through the real `CmuxSocketEventMapper` and read the event the bus actually emitted: caller fields land on v1 and v2 events; a failed pid lookup serializes as JSON `null` (asserted on the encoded line, not just the dictionary, since absent and null are the same in a Swift dictionary but not to a consumer); the real resolver returns nil for a dead pid and does not cache the failure; the provider is not invoked for unmapped or failed commands and is invoked once for mapped ones; non-socket events grow no `caller` key; and the two redaction mappings agree.
- `./scripts/lint-pbxproj-test-wiring.sh`: ok (checked 653 test files).
- Tagged cloud build `i9611` green, then driven live against `/tmp/cmux-debug-i9611.sock`.

Emitted `surface.input_sent` from `cmux send` run inside a pane of the tagged instance, showing both the resolved process name and the caller's own surface:

```json
{"boot_id":"19BC7D5B-5C4B-4D97-8F12-A9F3A1F49EFF","caller":{"pid":13850,"process_name":"cmux","surface_id":"13FAA162-9EF6-42D8-B8D2-6C04128FA426"},"category":"surface","id":"19BC7D5B-5C4B-4D97-8F12-A9F3A1F49EFF-25","name":"surface.input_sent","occurred_at":"2026-08-05T01:11:00.185Z","pane_id":null,"payload":{"method":"surface.send_text","params":{"redacted_fields":["text"],"surface_id":"surface:1","text":null,"text_length":20},"result":{"queued":false,"surface_id":"13FAA162-9EF6-42D8-B8D2-6C04128FA426","surface_ref":"surface:1","window_id":"55E0FE66-43EC-45B1-B1DD-4540D9CABDDF","window_ref":"window:1","workspace_id":"9FD013D4-26FA-4D2A-A6D5-3E20E12773BF","workspace_ref":"workspace:1"}},"protocol":"cmux-events","seq":25,"source":"socket.v2","surface_id":"13FAA162-9EF6-42D8-B8D2-6C04128FA426","type":"event","version":1,"window_id":"55E0FE66-43EC-45B1-B1DD-4540D9CABDDF","workspace_id":"9FD013D4-26FA-4D2A-A6D5-3E20E12773BF"}
```

The same command run from a shell outside the tagged app resolves `pid` and `process_name` but `"surface_id":null`, which is correct: that shell's controlling terminal belongs to a different cmux instance, and the resolver refuses to guess.

And `surface.key_sent` under the new policy:

```json
{"caller":{"pid":8418,"process_name":"cmux","surface_id":null},"name":"surface.key_sent","payload":{"method":"surface.send_key","params":{"key":"ctrl-c","redacted_fields":[],"surface_id":"surface:1"}},"source":"socket.v2","seq":18}
```

## Issues
- Closes https://github.com/manaflow-ai/cmux/issues/9611

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9616"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788484358&installation_model_id=21920&pr_number=9616&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9616&signature=67b32cf1d8d922be44948f2f070de687a1dbc4672aa6d99d3c11cd297c850b03"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Add caller attribution to all socket-sourced events so automation (e.g. cmux send) is distinguishable from human input. Identity is resolved server-side (pid, process_name, surface_id) with low overhead. Closes #9611.

- **New Features**
  - Every `socket.v1` and `socket.v2` event now includes `caller`: `pid`, `process_name`, `surface_id` (unresolved fields are `null`; non-socket events omit `caller`).
  - Unified input redaction: `surface.send_text` stays redacted to `text_length`; `surface.send_key` records `key` and emits `redacted_fields: []` (now also on v1).

- **Refactors**
  - Moved caller identity and resolver into `CmuxControlSocket`; resolver is instance-owned by `TerminalController`, with a bounded cache keyed by process generation (pid + start time) and generation revalidation before publish.
  - Attribution is primed before command execution only for requests that can publish (via `mapsToEvent`), avoiding nulls from short-lived callers; only TTY-owning callers incur one main-queue hop per connection.
  - Added `v1PublishingCommands`/`mapsToEvent`, tests to guard mapping drift, and updated docs with a “Caller attribution” section.

<sup>Written for commit 7ca3c8ae8a3d0dbb30e877d373dc9597a4362f06. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9616?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Socket events now include caller attribution, such as process ID, process name, and associated surface when available.
  * Caller details are validated to prevent stale process information and are omitted or marked unavailable when unresolved.
  * Input events now report redacted text lengths and validated named keys with explicit redaction metadata.

* **Documentation**
  * Updated event documentation to describe caller attribution, unresolved values, input redaction, and key event behavior.


austinywang:
Addressed all six pre-merge check failures and the four inline findings. Details are in the updated PR description; summary of what changed in code:

- **Ambient global state + production test seams** (`Cmux No Ambient Global State`, `Cmux No Test Or Debug Seam In Production Source`, inline on `CmuxSocketCallerIdentity.swift:61`): the resolver is now a constructable `final class` owning its lock and cache, held by `TerminalController` at the socket seam and injected into the publish path. `resetCacheForTesting()` and `cachedProcessNameCountForTesting` are gone; tests construct their own resolver. `NSLock` over an actor is documented on the type: every caller is a synchronous `nonisolated` function on the control-socket read loop, which has no async context to await from.

- **Package boundaries** (`Cmux Swift Package Boundaries`): moved to `Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Caller/`, next to the `LOCAL_PEERPID` transport code that produces the pid.

- **Pid generation binding** (inline on `CmuxSocketCallerIdentity.swift:64-82`, `Cmux Cache Substitution Correctness`): correct and now fixed. Identity reads pid, start time, and controlling terminal in one `proc_pidinfo` call, keys the name cache by that generation instead of by pid, and revalidates before publishing any derived field.

  This one found a real bug in my original design. Adding the revalidation made `process_name` come back `null` on live dogfood, because a short-lived `cmux send` exits between the response being written and the event being published, so lazy-at-publish resolution was already resolving a dead pid and getting nothing. Attribution is now primed **before the command runs**, while the client is still blocked on its response and provably alive, and skipped entirely for commands that publish no event via `CmuxSocketEventMapper.mapsToEvent`. A test drives every v1 command name through the mapper and asserts the predicate matches what `publishV1` actually publishes, so the two cannot drift.

- **Blocking main hop** (`Cmux Swift Blocking Runtime`, inline on `CmuxEventStream.swift:171`): reduced rather than removed, and I want to be explicit about that. It is now skipped outright when the caller has no controlling terminal, since such a process cannot be inside a pane, and runs at most once per connection for the rest. I did not convert publication to async: the control-socket read loop is synchronous, and making it async would reorder event emission relative to the socket response the event reports on. `v2MainSync` is the same bridge every socket command on this path already uses. Rationale is documented at the call site.

- **O(S) TTY scan** (`Cmux Algorithmic Complexity`): the same scan agent hook delivery already performs per hook event, now gated behind the tty check and run at most once per connection instead of per event.

- **v1 redaction metadata** (outside-diff comment on `CmuxSocketEventMapper.swift:196-212`): correct, `docs/events.md` documents `redacted_fields` as present on `surface.key_sent` and the v1 path emitted nothing. v1 key events now carry the empty marker and v1 send events name their redacted field. Covered by `v1InputEventsCarryRedactionMarkers`.

- **Test counter `@unchecked Sendable`** (inline on the test file): removed. The provider closure is non-escaping and runs synchronously on the calling thread, so a plain box needs no conformance.

Rebuilt tag `i9611` and re-drove all four socket paths live. `process_name` now resolves on every call where it previously came back null; the in-pane call resolves `surface_id` and the outside-the-app calls correctly leave it null. Test count went from 11 to 16.

PR #9627, opened 2026-08-05
Fix startup-race crash in v2RefreshKnownRefs against a half-restored session (#2751)

### LINKED ISSUE #2751: 0.63.1: PAC failure in TerminalController.v2EnsureHandleRef when a v2 socket command races startup state restore
### cmux version

0.63.1 (build 78)

### macOS version

macOS 26.4 (25E246)

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Homebrew

### Bug description

cmux 0.63.1 (build 78) crashed with `EXC_BAD_ACCESS (SIGSEGV)` and a pointer-authentication failure inside `TerminalController.v2EnsureHandleRef(kind:uuid:)`, called from `TerminalController.v2RefreshKnownRefs()`. The crash happened ~1.13s after process launch while a CLI client was processing a v2 command on the cmux Unix socket — Thread 23 had `dispatch.sync`-ed onto the main thread to run the refresh, and the main thread faulted while iterating handle UUIDs. A second client (Thread 22) was already connected and idle in `read()` at the same time, so this is at minimum a 2-client scenario during state restore.

Apple's crash reporter annotates the fault as `KERN_INVALID_ADDRESS at 0x8000000000000010 -> 0x0000000000000010 (possible pointer authentication failure)`. The faulting instruction is `LDR x8, [x25, #0x10]` (`28 0b 40 f9`); `x25` was `0x8000000000000000` (a pointer with all PAC bits set after authentication failure) and `far = x25 + 0x10` is the dereference target. Stripping the high bit gives `0x10`, i.e. cmux is reading field-at-offset-0x10 of an object whose pointer failed PAC verification — almost always either a use-after-free where the freed slot was reused with garbage, or an uninitialized pointer being read out of `v2RefByUUID` / `v2UUIDByRef` before the handle table was fully populated.

I have one crash report with this exact signature so far. The 1.1s launch-to-crash delta plus the dispatch chain (`processV2Command → dispatch.sync → v2RefreshKnownRefs`) make a startup-race interpretation more plausible than a deterministic bug, but a one-off transient corruption isn't ruled out.

### Expected behavior

A v2 socket command arriving immediately after cmux launch should not be able to fault the main thread.

Two parallel-fix patterns already exist in this code area and either would address it:

1. Wait for handle-table readiness, the way `send_key` does for missing surfaces (see #2006 — `send_key` calls `waitForTerminalSurface()` while `read_text` was crashing/erroring instead).
2. Catch and gracefully degrade, the way #1935 made `claude-hook stop` survive a nil `tabManager` during teardown.

A heavier fix would be to defer `acceptLoop` (or just v2 command processing) until `v2RefByUUID` / `v2UUIDByRef` are fully populated from persisted session state, so no v2 command can be dispatched against a half-built handle table.

### Steps to reproduce

I have not been able to deterministically reproduce this on demand — it's happened once for me in a week of regular use. The best repro pattern I can describe:

1. Have any shell init script call a v2 cmux command immediately on startup. Minimal example for `~/.zshrc`:
   ```bash
   if [[ -n "$CMUX_WORKSPACE_ID" ]]; then
     cmux list-workspaces 2>/dev/null
   fi
   ```
2. Quit cmux fully (`Cmd+Q`) so the next launch performs full state restore.
3. Relaunch cmux. The bundled shell sources `.zshrc`, and `cmux list-workspaces` hits the socket within the first ~1s of cmux being alive.
4. Repeat across many launches over several days. The race surfaces intermittently.

What likely matters more than the exact CLI command is **(a)** there is persisted session state to restore on launch, **(b)** at least one v2 socket client connects within the first second, and **(c)** the v2 handle table is being built up at the same time the v2 command is running.

### Shell and environment

zsh (the bundled shell that cmux launches via its `command =` config), oh-my-zsh, starship prompt, zsh-autosuggestions and fast-syntax-highlighting loaded. The trigger client is `cmux list-workspaces` invoked from `.zshrc`.

```text
### Build / binary identity

So you can resolve symbols against the matching dSYM:

- cmux short version: `0.63.1`
- cmux build version: `78`
- cmux Mach-O UUID: `e9384773-36fe-3706-9f29-aa90e319e3c6`
- cmux load address: `0x102200000`

### Exception


Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x8000000000000010
                   -> 0x0000000000000010 (possible pointer authentication failure)
Termination Reason: Namespace SIGNAL, Code 11, Segmentation fault: 11


### Faulting thread (main thread)


Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   cmux                   TerminalController.v2EnsureHandleRef(kind:uuid:) + 124
1   cmux                   TerminalController.v2RefreshKnownRefs() + 1088
2   libswiftDispatch.dylib partial apply for thunk for @callee_guaranteed () -> (@out A, @error @owned Error) + 28
3   libswiftDispatch.dylib partial apply for thunk for @callee_guaranteed () -> (@out A, @error @owned Error) + 16
4   libswiftDispatch.dylib closure #1 in closure #1 in OS_dispatch_queue._syncHelper<A>(fn:execute:rescue:) + 192
5   libswiftDispatch.dylib partial apply for thunk for @callee_guaranteed () -> () + 28
6   libswiftDispatch.dylib thunk for @escaping @callee_guaranteed () -> () + 28
7   libdispatch.dylib      _dispatch_client_callout + 16
8   libdispatch.dylib      _dispatch_async_and_wait_invoke + 84
9   libdispatch.dylib      _dispatch_client_callout + 16
10  libdispatch.dylib      _dispatch_main_queue_drain.cold.6 + 832
11  libdispatch.dylib      _dispatch_main_queue_drain + 176
12  libdispatch.dylib      _dispatch_main_queue_callback_4CF + 44
... [Apple framework runloop frames]


### Faulting instruction


PC:   0x102555d34 = TerminalController.v2EnsureHandleRef(kind:uuid:) + 124
LR:   0x102555d30 = TerminalController.v2EnsureHandleRef(kind:uuid:) + 120
Bytes at PC: 28 0b 40 f9   →   LDR x8, [x25, #0x10]


### Register state at fault


x25 = 0x8000000000000000   ← high bit set; PAC verify failed
x24 = 0x00000001fc7faf20   value witness table for UUID
x26 = 0x00000001fc7faf88
x27 = 0x0000000b38de4920
x28 = 0x0000000b3bab7c60
x10 = 0x0000000103aa7b08   value witness table for PaneID
x11 = x12 = 0x0000000000185093
far = 0x8000000000000010   ← x25 + 0x10
esr = 0x92000006           Data Abort, byte read, Translation fault


So the function is loading 8 bytes from a `[reference + 0x10]` field of an object whose pointer came out of memory with PAC bits already corrupted. The two value witness tables in the surrounding registers (`UUID` and `PaneID`) suggest the iteration is over `[PaneID: <something>]` or `[UUID: <something>]` — i.e. the v2 handle dictionaries directly. This lines up with #2192's description of `v2RefByUUID[.surface]` / `v2UUIDByRef[.surface]` being the long-lived state in this area.

### Triggering CLI client (Thread 23)


Thread 23
0   libsystem_kernel.dylib __ulock_wait + 8
1   libdispatch.dylib      _dispatch_thread_main_event_wait_slow + 76
2   libdispatch.dylib      __DISPATCH_WAIT_FOR_QUEUE__ + 464
3   libdispatch.dylib      _dispatch_sync_f_slow + 140
4   libswiftDispatch.dylib OS_dispatch_queue.asyncAndWait<A>(execute:) + 144
5   libswiftDispatch.dylib OS_dispatch_queue.sync<A>(execute:) + 64
6   cmux                   TerminalController.processV2Command(_:) + 2320
7   cmux                   TerminalController.processCommand(_:) + 228
8   cmux                   TerminalController.handleClient(_:peerPid:) + 1008
9   cmux                   closure #5 in TerminalController.acceptLoop(listenerSocket:generation:) + 92


### Second client connected at the same time (Thread 22)


Thread 22
0   libsystem_kernel.dylib read + 8
1   cmux                   TerminalController.handleClient(_:peerPid:) + 768
2   cmux                   closure #5 in TerminalController.acceptLoop(listenerSocket:generation:) + 92


So at the moment of crash, two socket clients are alive: Thread 22 idle in `read()` waiting for the next command from one client, Thread 23 deep in `processV2Command` for another. Both arrived within the first ~1s of cmux launch. The race is at minimum a multi-client one.

### Crash timing across the three reports I have locally

| date | launch +N | top frame | likely separate bug? |
|---|---|---|---|
| 2026-04-04 | +0.33s | `NSFileHandleOperationException` from `CMUXTermMain.main()` | yes — different signature |
| 2026-04-05 | +0.35s | same `NSFileHandleOperationException` | yes — same as above |
| **2026-04-09** | **+1.13s** | `TerminalController.v2EnsureHandleRef + 124` | **this report** |

All three are within ~1.2s of process launch, but only the Apr 9 one is the v2EnsureHandleRef PAC failure being reported here. The Apr 4/5 NSFileHandle crashes look like a separate bug class that probably warrants its own report.

The full `.ips` for the Apr 9 crash is attached to this issue (sanitized — per-Mac identifiers, per-boot UUIDs, and Apple submission IDs stripped; all stack frames, register state, instruction bytes, and image UUIDs preserved).
```

_No response_

### Closely related issues — same bug class

This isn't a one-off; cmux has had a recurring pattern of two adjacent bug classes that converge here.

**(A) CLI commands racing app-state lifecycle:**

- **#1935** — `claude-hook stop` failing because `TerminalController.tabManager` was set to nil during teardown before the stop hook fired. Same shape: a CLI command hits a v2 handler whose backing state isn't there. Was fixed by catching/logging instead of propagating. Our crash is the inverse-time variant: the backing state isn't there *yet* (during startup), and the handler doesn't catch — it dereferences and PACs.
- **#2006** — `readTerminalTextBase64` crashing/erroring on a nil surface during display-sleep reparenting, while `send_key` already handled this correctly via `waitForTerminalSurface()`. The same asymmetry probably exists between v2 commands that wait for handle-table readiness and the ones that don't. `v2RefreshKnownRefs` clearly doesn't.

**(B) PAC failures in workspace/tab object lifecycle:**

- **#2131** — `Cmd+N` PAC failure in `TabManager.newTabInsertIndex`, fixed by **#2133** (snapshot value-only state, insert into the live tabs array). 0.62.2-nightly.
- **#2157** + **#2169** + **#2178** — successive variants in `TabManager.workspaceCreationSnapshot` and the `ghostty_surface_config_s` C struct, all PAC/null-deref crashes during workspace creation.
- **#2180** — *still crashing post-#2178*: PAC failure in `swift_retain` called from `TabManager.addWorkspace + 760`, in 0.62.2-nightly. Confirms this code area keeps regressing.

**(C) The data structure being faulted on:**

- **#2192** — open issue/PR identifying that `v2RefByUUID[.surface]` and `v2UUIDByRef[.surface]` are *only added to* by `v2RefreshKnownRefs()` and never cleaned up on surface close. The crashing function in this report is `v2EnsureHandleRef` called from `v2RefreshKnownRefs`, accessing those exact dictionaries. #2192 is a *leak* and ours is a *use-after-free / uninitialized read*, but they live in the same dictionary, and the cleanup #2192 proposes might also remove the stale-pointer condition we're hitting.

The pattern: cmux has had repeated PAC failures whenever workspace/tab state lifecycle races a UI or socket access. UI-trigger paths got fixes; the `TerminalController.v2*` socket path is the next manifestation.

### Workaround on my side

I deferred the offending CLI call into a backgrounded subshell with a 2s sleep, so cmux has time to populate its v2 handle table before any commands arrive. That eliminates the trigger from my side, but the underlying issue is still there for any third-party shell init, AI agent, status-bar tool, or automation script that talks to the cmux socket immediately on launch.

### Why I can't test NIGHTLY

I'm pinned to 0.63.1 because of the drag-select auto-scroll regression in 0.63.2 (tracked separately). Installing nightly would re-introduce that bug for me. If nightly already has a fix for this race — point me at the commit and I'll re-symbolicate this `.ips` against it and confirm.

--- 

#### Full sanitized crash report: 

[cmux-2026-04-09-023322.redacted.ips.txt](https://github.com/user-attachments/files/26600463/cmux-2026-04-09-023322.redacted.ips.txt)

hdimer:
Fixes #2751.

## Problem

`EXC_BAD_ACCESS (SIGSEGV)` with an arm64e pointer-authentication failure in `TerminalController.v2RefreshKnownRefs()`, called from the control-command resolution hop (`controlResolveOnMain` → `v2MainSync { v2RefreshKnownRefs() }`). A v2 socket command (e.g. `cmux list-workspaces` from a shell rc) arriving within ~1s of launch runs the handle-ref **pre-mint pass** on the main actor while startup session restore is still building/mutating the window/tab tree, so it enumerates a half-built object and faults. Two users reported the identical signature on macOS 26.4/26.5, and it reproduces on nightly.

## Root cause

`v2RefreshKnownRefs()` (and its `drag_surface_to_split` twin `controlSidebarRefreshKnownRefs()`) iterate `listMainWindowSummaries()` / `tm.tabs` / `tm.workspaceGroups` unconditionally, with no gate on restore completion. Because a v2 command can be serviced during the restore window, the pass can run against a tree that `restoreSessionSnapshot` is populating in place.

The pass is only an **optimization** — it pre-mints handle refs so callers can pass `workspace_group:N` immediately after restore; otherwise refs mint lazily on the first `list`/`create` (as the function's own comment notes). So skipping it until restore settles is behavior-preserving.

## Fix

Gate both functions on a new readiness signal on `AppDelegate`:

```swift
var didCompleteInitialSessionRestore: Bool {
    didAttemptStartupSessionRestore && !isApplyingSessionRestore
}
```

It is **false** during the pre-attempt window, the **deferred-signing-secret** window (cold launch where the Keychain secret isn't ready yet, so `attemptStartupSessionRestore` defers before setting `didAttemptStartupSessionRestore`), and while a restore is applying (`isApplyingSessionRestore` brackets `restoreSessionSnapshot`, including the async additional-windows span and manual reopen). It is **true** only once the tree is stable. Once restore settles the guard is a no-op, so steady-state behavior is unchanged.

The deferred-signing-secret path is the important one: a naive "did we attempt restore" flag flips true too early there and leaves the crash window open (likely why dev builds, which set `environmentSigningSecret()`, never hit it). The composite signal stays false through that whole window.

## Testing

Full Debug build succeeds. Added `cmuxTests/AppDelegateSessionRestoreReadinessTests.swift`, which locks `didCompleteInitialSessionRestore` across the four flag combinations (verified running under the `cmux-unit` scheme). A deterministic red/green crash test isn't feasible — the fault needs a transiently half-built tree that can't be reproduced in a unit test — so the test guards the invariant that governs the skip (this is exactly the invariant an earlier draft got wrong on the deferred-secret path).

---

Used AI assistance on this; I reviewed, built, and tested it.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9627"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788496866&installation_model_id=21920&pr_number=9627&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9627&signature=da239aaa2a2e0a81a3b4b37f2887e2826e112c29735fb75c34ec410e1712c877"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Prevents a launch-time crash by skipping the v2 handle-ref pre-mint until session restore finishes. Behavior is unchanged because refs mint lazily. Fixes #2751.

- **Bug Fixes**
  - Added `AppDelegate.didCompleteInitialSessionRestore` (`didAttemptStartupSessionRestore && !isApplyingSessionRestore`) to signal when the tree is stable, including the deferred-signing-secret path.
  - Guarded `TerminalController.v2RefreshKnownRefs()` and `controlSidebarRefreshKnownRefs()` to no-op until the signal is true, avoiding iteration of a half-built window/tab tree.

<sup>Written for commit 5c60ef923da1260447965b95d008e98d4b11a4ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9627?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reference refreshes from running prematurely during startup session restoration.
  * Reduced unnecessary topology processing while restored sessions are still being applied.
  * Ensured reference refresh behavior resumes once restoration has settled.

* **Tests**
  * Added coverage for session restoration readiness across pre-restore, active restore, and completed states.

### DIFF
--- a/Sources/AppDelegate.swift
+++ b/Sources/AppDelegate.swift
@@ -1039,6 +1039,18 @@ final class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCent
     private var didPrepareStartupSessionSnapshot = false
     var didAttemptStartupSessionRestore = false
     var isApplyingSessionRestore = false
+    /// True only when it is safe to enumerate the window/tab tree for the v2
+    /// control pre-mint pass (#2751): the initial session-restore decision has
+    /// resolved (`didAttemptStartupSessionRestore`) AND no restore pass is
+    /// currently populating a tab manager in place (`!isApplyingSessionRestore`).
+    /// Deliberately FALSE during the pre-attempt window, the deferred
+    /// signing-secret window (where `didAttemptStartupSessionRestore` stays
+    /// false until the secret arrives), and while `restoreSessionSnapshot` runs
+    /// — the exact windows in which iterating a half-built tree faults. Refs
+    /// mint lazily meanwhile, so skipping the pre-mint is safe.
+    var didCompleteInitialSessionRestore: Bool {
+        didAttemptStartupSessionRestore && !isApplyingSessionRestore
+    }
     /// Durable navigation links that arrived before startup restore registered
     /// their target workspaces.
     var pendingStartupNavigationURLRequests: [CmuxNavigationURLRequest] = []
--- a/Sources/TerminalController+ControlSidebarContext3.swift
+++ b/Sources/TerminalController+ControlSidebarContext3.swift
@@ -110,6 +110,12 @@ extension TerminalController {
         // pre-pass), minting into the same coordinator-owned registry.
         guard let app = AppDelegate.shared else { return }
 
+        // #2751: same guard as the v2 twin. This runs on the
+        // `drag_surface_to_split` v1 path and iterates the same structures, so
+        // skip the pre-mint pass while session restore is pending or in flight
+        // to avoid faulting on a half-built tree; refs mint lazily otherwise.
+        guard app.didCompleteInitialSessionRestore else { return }
+
         let windows = app.listMainWindowSummaries()
         for item in windows {
             _ = controlCommandCoordinator.ensureRef(kind: .window, uuid: item.windowId)
--- a/Sources/TerminalController.swift
+++ b/Sources/TerminalController.swift
@@ -3698,6 +3698,14 @@ class TerminalController {
     func v2RefreshKnownRefs() {
         guard let app = AppDelegate.shared else { return }
 
+        // #2751: skip the pre-mint pass until session restore has settled.
+        // Refreshing here is only an optimization (refs otherwise mint lazily
+        // on the first list/create); iterating the half-built window/tab tree
+        // while restore is pending or in flight faults (EXC_BAD_ACCESS / arm64e
+        // ptrauth). A v2 socket command arriving within ~1s of launch can
+        // re-enter this on the main actor mid-restore, so degrade gracefully.
+        guard app.didCompleteInitialSessionRestore else { return }
+
         let windows = app.listMainWindowSummaries()
         for item in windows {
             _ = v2EnsureHandleRef(kind: .window, uuid: item.windowId)
--- a/cmux.xcodeproj/project.pbxproj
+++ b/cmux.xcodeproj/project.pbxproj
@@ -231,6 +231,7 @@ C0DE71B10000000000000001 /* AppDelegate+AgentChatNotifications.swift in Sources 
 		79380000000000000000000A /* AppDelegateOversizedFrameRestoreTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 793800000000000000000009 /* AppDelegateOversizedFrameRestoreTests.swift */; };
 		D37777050000000000000001 /* AppDelegatePresentPreferencesWindowTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D37777050000000000000002 /* AppDelegatePresentPreferencesWindowTests.swift */; };
 		C34670010000000000000001 /* AppDelegateRenameShortcutContextTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C34670010000000000000002 /* AppDelegateRenameShortcutContextTests.swift */; };
+		89550000000000000000A751 /* AppDelegateSessionRestoreReadinessTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 89550000000000000000A752 /* AppDelegateSessionRestoreReadinessTests.swift */; };
 		D37777040000000000000001 /* AppDelegateSettingsPresentation.swift in Sources */ = {isa = PBXBuildFile; fileRef = D37777040000000000000002 /* AppDelegateSettingsPresentation.swift */; };
 		6419B0016419B0016419B001 /* AppDelegateShortcutRoutingRepairProbe.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6419B0026419B0026419B002 /* AppDelegateShortcutRoutingRepairProbe.swift */; };
 		F6000000A1B2C3D4E5F60718 /* AppDelegateShortcutRoutingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F6000001A1B2C3D4E5F60718 /* AppDelegateShortcutRoutingTests.swift */; };
@@ -2912,6 +2913,7 @@ C0DE71B10000000000000002 /* AppDelegate+AgentChatNotifications.swift */ = {isa =
 		793800000000000000000009 /* AppDelegateOversizedFrameRestoreTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegateOversizedFrameRestoreTests.swift; sourceTree = "<group>"; };
 		D37777050000000000000002 /* AppDelegatePresentPreferencesWindowTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegatePresentPreferencesWindowTests.swift; sourceTree = "<group>"; };
 		C34670010000000000000002 /* AppDelegateRenameShortcutContextTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegateRenameShortcutContextTests.swift; sourceTree = "<group>"; };
+		89550000000000000000A752 /* AppDelegateSessionRestoreReadinessTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegateSessionRestoreReadinessTests.swift; sourceTree = "<group>"; };
 		D37777040000000000000002 /* AppDelegateSettingsPresentation.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = App/AppDelegateSettingsPresentation.swift; sourceTree = "<group>"; };
 		6419B0026419B0026419B002 /* AppDelegateShortcutRoutingRepairProbe.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegateShortcutRoutingRepairProbe.swift; sourceTree = "<group>"; };
 		F6000001A1B2C3D4E5F60718 /* AppDelegateShortcutRoutingTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegateShortcutRoutingTests.swift; sourceTree = "<group>"; };
@@ -7331,6 +7333,7 @@ C0DE71B10000000000000002 /* AppDelegate+AgentChatNotifications.swift */ = {isa =
 					F17F00000000000000000002 /* FullWidthTabPersistenceTests.swift */,
 					FD21AD844E5D4D50A6DFC442 /* AppDelegateWindowFrameReconcileTests.swift */,
 					895500000000000000000002 /* AppDelegateFullScreenFrameRestoreTests.swift */,
+					89550000000000000000A752 /* AppDelegateSessionRestoreReadinessTests.swift */,
 					A1B2C3D4E5F600000000CF02 /* AppDelegateDisplayConfigRestoreTests.swift */,
 					793800000000000000000009 /* AppDelegateOversizedFrameRestoreTests.swift */,
 				B35750000000000000000009 /* PiVaultAgentPersistenceTests.swift */,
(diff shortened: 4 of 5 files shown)


hdimer:
Re CodeRabbit's note about wanting a single authoritative readiness signal: that's already what this does. `didCompleteInitialSessionRestore` is that one signal, and both `v2RefreshKnownRefs()` and the `drag_surface_to_split` twin `controlSidebarRefreshKnownRefs()` gate on it before traversing the tree.

It's `didAttemptStartupSessionRestore && !isApplyingSessionRestore`, so it stays false through the pre-attempt window, the deferred signing-secret window (on cold launch `didAttemptStartupSessionRestore` isn't set until the secret arrives and restore actually runs), and while a restore is applying, and true only once the tree has settled. That's the case the readiness gate is specifically designed to cover.

PR #9650, opened 2026-08-05
Fix ControlHandleRegistry data race (SIGSEGV/SIGABRT in ensureRef under concurrent socket access)

cicerohellmann:
## Summary

**What:** Fix a data race in `ControlHandleRegistry` (the `kind:N` handle-ref mint behind the v2 control socket) that corrupts its backing dictionaries and crashes the app with `SIGSEGV`/`SIGABRT` inside `ensureRef`.

**Why:** The registry holds three `Dictionary`s guarded only by an unenforced "owner provides isolation" convention (documented in its own header). Callers reach it from two lanes — the main-actor dispatch path and the socket worker's resolution hop (`v2MainSync { v2RefreshKnownRefs() }`). When a caller mutates it without the main-actor hop, two threads mutate the same dictionary concurrently and corrupt its buffer. A dictionary undergoing a concurrent `insert` mid-resize hands back a freed/half-written bucket → crash in `ensureRef`.

This reproduces under load: an external poller walking the tree every few seconds (repeated `system.tree` / ref-minting) races the main lane and takes the app down.

## Root cause

`ensureRef` is check-then-mint on shared, unsynchronized state:

```swift
if let existing = refByUUID[kind]?[uuid] { return existing }   // read
let next = nextOrdinal[kind] ?? 1
refByUUID[kind, default: [:]][uuid] = ref                      // write
uuidByRef[kind, default: [:]][ref] = uuid                      // write
nextOrdinal[kind] = next + 1                                   // write
```

Two threads here either (a) mutate the same `Dictionary` storage concurrently → buffer corruption → `SIGSEGV`/`SIGABRT`, or (b) both miss the `existing` entry and mint two different ordinals for one identity → duplicate/colliding refs.

## Fix

- Make `ControlHandleRegistry` a shared reference type (`final class`, was a `struct`).
- Serialize all three access points (`ensureRef`, `removeRef`, `uuid(forRef:)`) behind a single `NSLock`.

The lock is the enforcement the old convention relied on by hand. This is defensive at the type level, so it holds regardless of which lane forgets to hop — no need to chase every call site. The public API and the `kind:N` ref semantics are unchanged.

## Scope / risk

- **App target is unaffected by the struct→class change.** Every app-target use goes through `controlCommandCoordinator.ensureRef / removeRef / resolveRef` (11 sites in `Sources/`); nothing in the app references the `ControlHandleRegistry` type or a `handles` value directly. The coordinator (the only direct user) lives in this package and its method signatures are unchanged.
- **Hot path:** `ensureRef` runs on nearly every socket response. An uncontended `NSLock` is a few ns; contention here is rare and short. No measurable cost expected.

## Testing

Two commits, so CI proves the test catches the bug (per the repo's regression policy):

1. `test:` — registry made shareable, two concurrency tests added, **no lock** → CI red (crash).
2. `fix:` — add the lock → CI green.

Run locally:

```bash
cd Packages/macOS/CmuxControlSocket

# On commit 1 (no lock): aborts with signal 11/6 — the bug, reproduced.
swift test --filter concurrentEnsureRefForDistinctIdentities

# On commit 2 (lock): passes, repeatedly and under Thread Sanitizer.
swift test --filter ControlHandleRegistry
swift test --filter ControlHandleRegistry --sanitize=thread
```

Verified locally on the fix commit: 6/6 plain runs green, and all 8 registry tests pass under `--sanitize=thread` (no data race reported). Full `CmuxControlSocket` suite green.

### Scenarios covered

| Test | Without lock | With lock |
|---|---|---|
| `concurrentEnsureRefForSameIdentityMintsExactlyOneRef` — 4,000 threads mint the *same* identity | can mint multiple refs for one identity | resolves to exactly one ref (`surface:1`) |
| `concurrentEnsureRefForDistinctIdentitiesStaysConsistent` — 2,000 threads mint *distinct* identities | corrupts the dictionaries → `SIGSEGV`/`SIGABRT` | every identity keeps a unique, round-tripping ref |

## Demo Video

N/A — non-visual crash fix. Evidence is the RED→GREEN test split above (the Commits tab shows the test failing without the fix).

```text
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review
```

## Checklist

- [ ] I updated docs/changelog if needed (n/a)
- [ ] I requested bot reviews after my latest commit (copy/paste block above or equivalent)

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9650"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788536598&installation_model_id=21920&pr_number=9650&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9650&signature=fb125b5b08f217f2fdbc189a84f26165aed5a89e4452b3d394bb643c3527877c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes a data race in `ControlHandleRegistry` that corrupted its dictionaries and crashed with SIGSEGV/SIGABRT under concurrent v2 control socket access. The registry is now synchronized without changing public behavior or `kind:N` ref semantics.

- **Bug Fixes**
  - Serialized access with a single `NSLock` across `ensureRef`, `removeRef`, and `uuid(forRef:)` to prevent concurrent dictionary mutation; added inline docs explaining why a lock is used over an actor/`@MainActor` for these synchronous call sites.
  - Converted `ControlHandleRegistry` from a `struct` to a `final class` (`@unchecked Sendable`) so it can be safely shared across the main-actor path and the socket worker, eliminating duplicate refs and crashes in `ensureRef`.

<sup>Written for commit 66215744cc37c5fe1dd663344a94fa33e79638b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9650?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved control-handle registry reliability during concurrent operations.
  - Prevented duplicate references and lost or colliding entries when handles are registered simultaneously.
  - Preserved accurate reference removal and alias resolution.

- **Tests**
  - Added coverage for concurrent registration and round-trip reference handling.

### DIFF
--- a/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Wire/ControlHandleRegistry.swift
+++ b/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Wire/ControlHandleRegistry.swift
@@ -5,9 +5,12 @@ public import Foundation
 /// dictionaries + `v2EnsureHandleRef`/`v2ResolveHandleRef` on
 /// `TerminalController`).
 ///
-/// A plain value type; the owner provides isolation (legacy: main-actor
-/// state on the controller).
-public struct ControlHandleRegistry: Sendable {
+/// A shared reference type. Callers reach it from more than one lane (the
+/// main-actor dispatch path and the socket worker's resolution hop) under an
+/// unenforced "owner provides isolation" invariant; a missed hop races the
+/// three dictionaries and corrupts their buffers (SIGSEGV/SIGABRT in
+/// `ensureRef`).
+public final class ControlHandleRegistry: @unchecked Sendable {
     private var nextOrdinal: [ControlHandleKind: Int]
     private var refByUUID: [ControlHandleKind: [UUID: String]]
     private var uuidByRef: [ControlHandleKind: [String: UUID]]
@@ -34,7 +37,7 @@ public struct ControlHandleRegistry: Sendable {
     ///   - kind: The handle kind.
     ///   - uuid: The object identity.
     /// - Returns: The stable ref string.
-    public mutating func ensureRef(kind: ControlHandleKind, uuid: UUID) -> String {
+    public func ensureRef(kind: ControlHandleKind, uuid: UUID) -> String {
         if let existing = refByUUID[kind]?[uuid] {
             return existing
         }
@@ -54,7 +57,7 @@ public struct ControlHandleRegistry: Sendable {
     /// - Parameters:
     ///   - kind: The handle kind.
     ///   - uuid: The object identity to forget.
-    public mutating func removeRef(kind: ControlHandleKind, uuid: UUID) {
+    public func removeRef(kind: ControlHandleKind, uuid: UUID) {
         if let ref = refByUUID[kind]?[uuid] {
             uuidByRef[kind]?.removeValue(forKey: ref)
         }
--- a/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Wire/ControlHandleRegistry.swift
+++ b/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Wire/ControlHandleRegistry.swift
@@ -12,9 +12,26 @@ public import Foundation
 /// `ensureRef`).
 public final class ControlHandleRegistry: @unchecked Sendable {
     /// Serializes all access to the three dictionaries so concurrent callers
-    /// (a missed main-actor hop from the socket worker) can never corrupt a
-    /// dictionary buffer mid-mutation. This is the enforcement the old
-    /// "owner provides isolation" invariant relied on by convention.
+    /// can never corrupt a dictionary buffer mid-mutation.
+    ///
+    /// A lock, not an actor or `@MainActor`, is deliberate here — two concrete
+    /// reasons, neither of which is "simpler":
+    ///
+    /// 1. `ensureRef` / `removeRef` / `uuid(forRef:)` are **synchronous** and
+    ///    called from synchronous, non-`async` contexts: the coordinator mints
+    ///    refs inline while building socket responses (`ref()` runs on nearly
+    ///    every response), and the legacy path reaches them through
+    ///    `v2MainSync`. An `actor` exposes only `async` members to outside
+    ///    callers, so adopting one would force `await` across the entire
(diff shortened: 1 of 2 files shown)


cicerohellmann:
Addressed the **Cmux Swift Blocking Runtime** check in 6621574 (docs-only; no behavior change).

`swift-blocking-runtime.md` allows a lock around non-async access "when an actor cannot be used and the code documents the reason." The two concrete reasons are now documented on the lock:

1. `ensureRef` / `removeRef` / `uuid(forRef:)` are **synchronous** and called from synchronous, non-`async` socket-dispatch code — the coordinator mints refs inline while building responses (`ref()` on nearly every response), and the legacy path reaches them via `v2MainSync`. An `actor` exposes only `async` members, so adopting one forces `await` across the whole synchronous dispatch chain.
2. `@MainActor` isolation is the design that just failed: this state was already main-actor-"owned" by convention, and the crash is a caller reaching it **without** the hop (the socket worker's resolution lane). A lock is safe regardless of a missed hop; re-asserting `@MainActor` reinstates the same unenforced invariant.

Tests unchanged and still green, including under `swift test --sanitize=thread`.

PR #9662, opened 2026-08-05
Run Cloud VM attach on the Mac when the CLI is relayed from a remote session

### LINKED ISSUE #9657: Remote (SSH) cmux sessions can't spawn Cloud-VM-hosted workspaces, forcing raw tmux fallback
A remote (SSH) cmux session can't spawn a new Cloud-VM-hosted workspace from inside it, so the only option from a remote session is the raw `ssh-tmux` mirror fallback instead of a first-class remote workspace. Reported by a user on 0.64.22 as one of a batch of follow-ups that hadn't landed yet:

> I'm still on the raw tmux fallback because a remote session can't spawn box-hosted workspaces

Not yet root-caused against a specific line — filing from the reproducible symptom so it's tracked, since I couldn't find an existing issue covering it. Repro: from inside a cmux workspace that is itself SSH/remote (or a Cloud VM), try to open a new Cloud-VM-hosted workspace the normal way; it's unavailable and the only remote path offered is `cmux ssh-tmux` (raw tmux mirror), not a real Cloud VM-backed workspace.

Starting points for whoever picks this up: the "New Workspace" gating logic in `Sources/AppDelegate+NewWorkspaceMenuRendering.swift` and `Sources/AppDelegate+NewWorkspaceContextMenu.swift`, Cloud VM launch in `Sources/CloudVMActionLauncher.swift`, and workspace-creation flow in `Sources/TerminalController+WorkspaceCreate.swift`. Determine whether this is a deliberate restriction (and if so, whether it's still the right one) or an oversight, then either lift the restriction or give a clear reason/error instead of a silent fallback to raw tmux.

Positive signal from the same report, for context: Mosh for remote workspaces (0.64.21, #8442) is landing well — "sessions surviving network changes is a real improvement."

austinywang:
Fixes https://github.com/manaflow-ai/cmux/issues/9657.

## What was actually broken

Not the menu. The Cloud VM section in New Workspace is gated only on `CmuxFeatureFlags.shared.isCloudVMUIEnabled`, and `CloudVMActionLauncher` always spawns the bundled CLI on the Mac against the local socket, so the GUI entrypoint works from any workspace, remote or not.

The broken entrypoint is the `cmux` CLI typed inside a remote workspace. It reaches the app through the remote CLI relay, so its socket RPCs land on the Mac while the process itself runs on the remote host. Every Cloud VM attach funnels through `vmOpenShell`, which mints a credential and builds a shell command out of process-local values, then hands that command to a Mac-hosted pane through `workspace.create`. Those values name the wrong machine.

Reproduced against the shipped 0.64.x CLI driven over a relay socket. It sends `workspace.create` with an `initial_command` that opens with:

```
cmux_freestyle_cli=/Applications/cmux.app/Contents/Resources/bin/cmux
```

That is the *calling* machine's CLI path. From a real remote host it is `/home/dev/.local/bin/cmux` or nothing at all, and the Mac pane runs it, so the workspace opens onto a pane that dies immediately. `client.socketPath` is wrong the same way. That is the reported symptom: no working Cloud-VM-hosted workspace, so `cmux ssh-tmux` stays the only remote option.

## Deliberate constraint or missing plumbing

Missing plumbing. There is no remote-related gate anywhere in the Cloud VM path: no denylist on the relay, no remoteness check in `NewWorkspaceMenuModel`, `AppDelegate+NewWorkspaceContextMenu`, or `TerminalController+WorkspaceCreate`. Nothing intended to block this. The attach step simply assumes the CLI process and the pane share a filesystem, which stops being true the moment the relay is in play.

There is no nested SSH hop to worry about here: the Mac reaches the Cloud VM gateway directly, exactly as it does for a local `cmux vm base open`. The hop count is unchanged.

## Fix

A relayed CLI stops building a Mac command and instead names which attach it wants. New control-socket method `vm.open_local`; the app runs the attach on its own machine.

- Base routes through `AppDelegate.performCloudVMAction`, the same workspace-owning action the New Workspace menu and titlebar button already use, so the loading panel, pinning, and reattach-to-existing behavior are identical whichever entrypoint asked.
- Per-VM attaches (`vm shell`, `vm ssh`, and the tail of `vm new` / `fork` / `restore`) spawn the bundled CLI locally through `CloudVMActionLauncher`.
- The guard sits in `vmOpenShell`, the single chokepoint all six of those paths already share, so no surface keeps a private copy of the decision.
- The relayed caller never supplies argv. It sends at most an id; the app builds the argument vector from a fixed set of shapes (`CloudVMHostAppOpenCommand`) and rejects any id that is not an opaque identifier, including a leading `-` that would land in argv as a flag.
- An app too old to know `vm.open_local` produces an explicit "update the app, or open Base from the Mac" error rather than the old dead pane.

Principled, not a workaround: the pre-existing bug was that a machine-local value escaped to a different machine, and the fix moves the whole path-sensitive step onto the machine that executes it, rather than trying to translate paths across the relay.

Scope note: plain `cmux ssh <destination>` from a remote session is left alone. Delegating it would make the *Mac* SSH to that destination, which may be unreachable from the Mac; that one is a real constraint, unlike a Cloud VM.

## Tests

`CloudVMHostAppOpenCommandTests` (package) pins the trust boundary: parameter round-trip, and rejection of shell-significant, flag-shaped, non-string, overlong, and empty ids.

`CLIRemoteRelayCloudVMOpenTests` (app target) runs the real bundled CLI against a relay socket and asserts wire behavior, which is the boundary the remote-only paths used to escape through:
- relayed `vm base open` sends `vm.base_open` then `vm.open_local`, and never `workspace.create` or `vm.attach_info`
- relayed `vm ssh <id>` sends `vm.open_local` with the id and `force_ssh`
- `method_not_found` produces the explicit error, with no `ssh-tmux` suggestion

Two commits so CI shows red then green.

## Verified on the tagged build

- Relayed `vm base open` through the fixed bundled CLI: `vm.base_open` then `vm.open_local`, exit 0.
- Relayed `vm shell <id>` proxied into the real tagged app socket: app logged `cloudVM.launch pid=1149`, i.e. it spawned the attach on the Mac.
- `vm.open_local` with `vm; rm -rf /` and `--workspace` both answered `invalid_params`, not `method_not_found` and not a spawn.
- Same command against the shipped stable CLI still emits the remote-path `workspace.create` shown above.

Not verified: a full authenticated Base open end to end. The tagged dev build has no dogfood sign-in on this machine (`~/.secrets/cmuxterm-dev.env` lacks the Stack account), and `scripts/setup-team-dev.sh` prompts interactively.

Dictionary:
- **relay**: the authenticated TCP channel that forwards a remote workspace's `cmux` CLI calls back to the Mac app's control socket.
- **attach**: the step that connects a pane to a running Cloud VM, as opposed to the control-plane call that creates it.
- **control plane**: the web API that creates, lists, and hands out credentials for Cloud VMs.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9662"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788550815&installation_model_id=21920&pr_number=9662&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9662&signature=efb060f1ab84558dea6505a83f248b839321deebe02381e262900bfc7e2b82f2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Run Cloud VM attaches on the Mac when the CLI is relayed from a remote session. Fixes #9657 by delegating attach requests to the host app so Cloud‑VM‑hosted workspaces open reliably instead of launching a dead pane.

- **Bug Fixes**
  - Added `vm.open_local` control-socket method; a relayed CLI names the VM (or Base) and the app runs the attach locally.
  - Base opens route through `performCloudVMAction`; per‑VM attaches launch via `CloudVMActionLauncher` with consistent UI and reattach behavior.
  - Single guard in `vmOpenShell` so all CLI paths share the fix; non‑relay behavior is unchanged (e.g., `cmux ssh <destination>`).
  - Relay policy now allows only the typed cloud open actions with allowlisted parameters; credential‑returning `vm.*` methods stay denied.
  - Tight VM id validation; the relay never supplies argv and malformed/flag-like ids are rejected, and older apps get a clear update error instead of a dead pane.

<sup>Written for commit e768e6048b1acff16eff88ed0c9117aa8bb90580. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Cloud VM workspace opening can now be handled through the Mac host app.
  - Supports Base, managed shell, and SSH workspace flows.
  - CLI output includes JSON and localized human-readable responses.
- **Bug Fixes**
  - Relay-backed connections avoid unnecessary local SSH setup.
  - Invalid VM identifiers and unsupported host capabilities now produce clear, actionable errors.
- **Documentation**
  - Added English and Japanese messaging for host-app workspace opening and window switching.
- **Tests**
  - Added coverage for command validation, relay requests, responses, and failure scenarios.

### DIFF
--- a/CLI/CMUXCLI+CloudVMHostAppOpen.swift
+++ b/CLI/CMUXCLI+CloudVMHostAppOpen.swift
@@ -1,0 +1,63 @@
+import CmuxFoundation
+import Foundation
+
+extension CMUXCLI {
+    /// Asks the Mac app to open a Cloud VM workspace on its own machine.
+    ///
+    /// Used when this CLI is talking to cmux through the remote CLI relay: the
+    /// socket RPCs land on the Mac but this process, its executable path, and
+    /// anything it writes live on the remote host, so it cannot build the
+    /// startup command the Mac-side pane has to run.
+    func openCloudVMOnHostApp(
+        id: String,
+        workspaceName: String?,
+        forceSSH: Bool,
+        client: SocketClient,
+        jsonOutput: Bool
+    ) throws {
+        let command = hostAppCloudVMOpenCommand(
+            id: id,
+            workspaceName: workspaceName,
+            forceSSH: forceSSH
+        )
+        do {
+            let response = try client.sendV2(
+                method: CloudVMHostAppOpenCommand.method,
+                params: command.socketParameters,
+                responseTimeout: Self.vmAttachResponseTimeoutSeconds
+            )
+            if jsonOutput {
+                print(jsonString(response))
+                return
+            }
+            print(String(
+                localized: "cli.vm.open.delegatedToHostApp",
+                defaultValue: "Opening the Cloud VM workspace on your Mac. Switch to the cmux window to use it."
+            ))
+        } catch let error as CLIError {
+            guard error.v2Code == "method_not_found" else { throw error }
+            throw CLIError(message: """
+                This cmux app is too old to open a Cloud VM workspace from inside a remote session.
+
+                What to do:
+                  Update the cmux app on your Mac, then retry this command.
+                  Until then, open the workspace from the Mac: New Workspace ▸ Open Base.
+
+                Details:
+                  The app does not support `\(CloudVMHostAppOpenCommand.method)`.
+                """)
+        }
+    }
+
+    /// A relayed caller names which attach it wants; the app owns argv.
+    func hostAppCloudVMOpenCommand(
+        id: String,
+        workspaceName: String?,
+        forceSSH: Bool
+    ) -> CloudVMHostAppOpenCommand {
+        if workspaceName == Self.persistentCloudVMWorkspaceName {
+            return .base
+        }
+        return forceSSH ? .ssh(vmID: id) : .shell(vmID: id)
+    }
+}
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -3064,7 +3064,9 @@ struct CMUXCLI {
 
     private static let vmCreateIdempotencyTTLSeconds: TimeInterval = 10 * 60
     private static let vmCreateResponseTimeoutSeconds: TimeInterval = 16 * 60
-    private static let vmAttachResponseTimeoutSeconds: TimeInterval = 16 * 60
+    // Internal (not private): the relay-delegation path in
+    // CMUXCLI+CloudVMHostAppOpen.swift reuses this attach timeout.
+    static let vmAttachResponseTimeoutSeconds: TimeInterval = 16 * 60
     private static let sshPTYTerminalConnectedResponseTimeoutSeconds: TimeInterval = 0.5
     private static let sshPTYTerminalConnectedRetryDelaySeconds: TimeInterval = 0.1
     private static let sshPTYTerminalConnectedMaximumRetryDelaySeconds: TimeInterval = 2
@@ -3073,7 +3075,9 @@ struct CMUXCLI {
     // reconnect, session restore, and mobile attach targets the same provider VM once
     // creation succeeds. Do not rotate it without a migration.
     private static let persistentCloudVMSlotID = "cmux-default-freestyle-sshd-v1"
-    private static let persistentCloudVMWorkspaceName = "sshd"
+    // Internal (not private): CMUXCLI+CloudVMHostAppOpen.swift compares
+    // against it to tell a Base open from a per-VM attach.
+    static let persistentCloudVMWorkspaceName = "sshd"
     private static let claudeCodeStatusKey = "claude_code"
 
     private static var allowedAgentLifecycleStatusKeys: Set<String> {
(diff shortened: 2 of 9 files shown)


lawrencecchen:
Mac fleet instructions for head `e768e6048b1acff16eff88ed0c9117aa8bb90580`. Planned tag: `pr-9662-e768e604`; this is not yet a published build.

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-9662-e768e604 /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git e768e6048b1acff16eff88ed0c9117aa8bb90580' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/9662 --source-digest e768e6048b1acff16eff88ed0c9117aa8bb90580 --cache-key cmux:pr-9662 --min-free-bytes 268435456000 --label cmux --label ram48)
```

PR #9699, opened 2026-08-06
Gate hosted coderouter on Pro

lawrencecchen:
## Summary
- gates hosted coderouter session issuance on local authoritative cmux Pro state
- principal-scopes route tokens and revokes them on all inactive subscription update paths
- leaves self-hosted deployments explicitly billing-independent
- adds coderouter-only, privacy-scrubbed Sentry with manaflow source-map uploads

## Testing
- `bun run typecheck`
- `bun run db:check`
- 47 focused billing/coderouter/Sentry tests
- production free-user rejection and self-host integration coverage
- production Sentry ingest, grouping, environment/release tags, and 2,796 uploaded source-map files

## Demo
- Ratatui Ctrl-C flow and production CLI were verified in cmux; no video is needed for this backend-focused change.

## Review trigger
@coderabbitai review

## Checklist
- [x] No Stripe query on model traffic hot path
- [x] Billing/storage failures fail closed
- [x] Secrets and PII are excluded from Sentry
- [x] Migrations applied using expand/deploy/contract ordering
- [x] Living architecture document updated

### DIFF
--- a/web/app/api/coderouter/session/route.ts
+++ b/web/app/api/coderouter/session/route.ts
@@ -1,25 +1,99 @@
+import { env } from "../../../env";
+import { hasActiveStripeProSubscription } from "../../../../services/billing/pro";
 import {
+  authenticateRouteToken,
   issueRouteToken,
   revokeRouteToken,
 } from "../../../../services/coderouter/repository";
 import { resolveCodeRouterRequestContext } from "../../../../services/coderouter/requestContext";
+import { captureCoderouterError } from "../../../../services/errors";
 
 export const runtime = "nodejs";
 export const dynamic = "force-dynamic";
 
-export async function POST(request: Request): Promise<Response> {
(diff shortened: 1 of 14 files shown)


lawrencecchen:
Addressed the review findings: removed fixed Retry-After and API prose, made production entitlement configuration mandatory, centralized inactive subscription revocation so invoice paths are covered, removed all Sentry request headers and user identity, added opaque API-key and invoice regression tests, and used a validated CHECK before SET NOT NULL. The hard-coded landing additions were reverted rather than adding untranslated copy.

The non-concurrent index recommendation is intentionally not applied: this project’s Drizzle migrator runs migrations transactionally, where `CREATE INDEX CONCURRENTLY` is invalid. The expand migration was applied before deployment while this private-beta table contained one token, then the principal-aware app was deployed before the contract migration; there was no meaningful write-availability risk.

PR #9717, opened 2026-08-06
CI: serialize Swift Testing inside app-host shards

lawrencecchen:
## Summary

- disable in-process Swift Testing parallelism in the shared app-host `xcodebuild` wrapper
- keep four app-host shards running on separate Macs
- run a focused parameterized behavior test before every broad shard
- exclude that focused policy test from ordinary shard allocation

## Why

App-host tests share process-global AppKit, account, socket, and store state. Xcode's in-process concurrency let unrelated tests overlap inside one host and produced nondeterministic crashes and hangs. Job-level environment variables and scheme variables were disproven by real CI runs. The supported `xcodebuild -parallel-testing-enabled NO` option controls the test runner directly.

This is a principled CI ownership fix. One test host runs one case at a time, while independent machines retain shard-level throughput.

## Validation

- Exact head: `fb589e7765c4a2cf86b3c1c2b01854e2993a2cb2`
- red control without the supported option: both parameterized cases overlapped and the focused test failed
- strict determinism scan: zero findings
- workflow routing, app-host retry, attempt-limit, sharder, and self-hosted guards: passed
- canonical `$autoreview`: clean, confidence 0.9
- CodeRabbit: success
- exact speculative merge run: https://github.com/manaflow-ai/cmux/actions/runs/31128817575

## Merge state

The exact speculative merge run is in progress. This PR changes CI only, so it does not require a tagged runtime build.

## Dictionary

- App-host test: A test loaded into the cmux application process, where UI and account state is shared.
- Shard: One independent subset of app-host tests assigned to a separate CI machine.
- Speculative merge: A temporary commit combining the PR head with current `main` before required checks run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -478,10 +478,6 @@ jobs:
       CMUX_CI_XCODE_APP: ${{ vars.CMUX_CI_XCODE_APP_MACOS_15 }}
       CMUX_CI_REQUIRED_MACOS_SDK_MAJOR: "26"
       CMUX_SKIP_ZIG_BUILD: "1"
-      # cmuxTests share process-global AppKit, account, socket, and store state.
-      # Xcode copies TEST_RUNNER_-prefixed values into the app test host with
-      # the prefix removed. Keep each shard serial while four jobs stay parallel.
-      TEST_RUNNER_SWT_EXPERIMENTAL_MAXIMUM_PARALLELIZATION_WIDTH: "1"
       # Keep one-off focused gates on the lightest measured shard so shard 1 no
       # longer carries the full shard plus every extra regression guard.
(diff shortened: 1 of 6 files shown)


lawrencecchen:
Closing because the exact speculative merge disproved the approach: disabling Xcode parallel testing collapsed each shard into one long-lived app host and exposed process-order dependence instead of stabilizing the suite. Shard 2 reported 127 failures and shard 1 reported 55 failures in https://github.com/manaflow-ai/cmux/actions/runs/31128817575. The failures span account state, shell configuration, session persistence, hibernation, and WebKit lifecycle, so the focused two-case probe was insufficient. I am keeping process isolation and validating the Sentry stack against the existing sharded runner instead.

PR #9765, opened 2026-08-07
Fix horizontal tab strip insertion positions

### LINKED ISSUE #9746: Horizontal tab strip: dragged tab cannot be inserted between two tabs — drop always lands at the end
## Summary

When rearranging surface tabs by dragging within a pane's horizontal tab strip, the dragged tab cannot be dropped between two existing tabs. The drop always lands at the end of the strip, making precise reordering impossible.

## Repro

1. Open a workspace with a pane that has 3+ surface tabs in the horizontal tab strip.
2. Drag the last tab and try to drop it between the first and second tabs.
3. The tab lands at the end of the strip instead of the intended position (no insertion indicator / drop slot between tabs).

## Expected

- Dragging a tab over the strip should show an insertion point between any two tabs, and dropping should insert it exactly there.
- This should work at any index inside the tab group, and consistently across different panes/splits (including dragging within the strip of any split pane).

## Out of scope

Tab sizing/clipping in overflowed strips (tab not fully visible before cutoff) is tracked separately in https://github.com/manaflow-ai/cmux/issues/8069 — this issue is only about drag-reorder insertion position.

austinywang:
## Summary

- update bonsplit to use one pane-local drop destination across the horizontal tab strip
- compute insertion indices from live tab-frame midpoints so every between-tab slot is reachable
- preserve exact insertion behavior in translated split panes and for cross-pane moves
- keep tab sizing and overflow clipping unchanged

Upstream implementation: https://github.com/manaflow-ai/bonsplit/pull/202

## Regression coverage

The bonsplit PR preserves the required red/green history:

1. failing pointer-to-index regression tests
2. the AppKit drop destination, shared mutation path, and behavior tests

Follow-up review coverage rejects stale pasteboard-only transfers that cannot complete, validates model/visual frame order, and exercises translated AppKit hit testing.

## Testing

- `arch -arm64 "$(xcrun --find swift)" test` in `vendor/bonsplit`
- 216 XCTest cases and 7 Swift Testing cases passed
- no new Swift warnings observed
- cloud visual proof was attempted twice, but neither pool assigned a runner within 15 minutes:
  - macOS 26: https://github.com/manaflow-ai/cmux-loader/actions/runs/31148006838
  - macOS 15: https://github.com/manaflow-ai/cmux-loader/actions/runs/31148877428

## Review follow-up

- removed sorting from the drag-update hot path; insertion lookup is linear over already pane-ordered frames
- changed the resolver from static helper APIs to an instance collaborator
- require an actionable cross-pane fallback handler before advertising a pasteboard-only move

Issue: https://github.com/manaflow-ai/cmux/issues/9746

Closes #9746


austinywang:
Review follow-up on the current bonsplit head `4da56ab`:

- Algorithmic complexity: removed the per-`draggingUpdated` sort. The destination supplies frames in pane-model order, the resolver validates model/visual ordering in O(n), then performs a linear midpoint scan.
- Ambient global state: `TabDropIndexResolver` is now an instance collaborator owned by `TabBarDropDestinationNSView`; its insertion APIs are no longer static.
- Pasteboard fallback: a no-live-state transfer advertises `.move` only for a cross-pane source with an installed external handler.

The follow-up behavior tests cover unordered frame rejection, same-pane stale transfers, and cross-pane transfers without a handler. Local validation is 216 XCTest cases plus 7 Swift Testing cases, all passing with no warnings observed.

PR #9796, opened 2026-08-07
Make cold PTY spawn unstarvable: reclaim bootstrap-window custody, bound the shim gate, unpin prime slots, surface queued sends (#9769)

### LINKED ISSUE #9769: CLI dispatch burst wedges PTY spawn app-wide: new workspaces never get a shell, sends silently dropped/delayed, socket refuses connections (main-thread starvation)
## Summary

A burst of CLI-driven workspace automation (bulk agent dispatch: `workspace create` + `new-pane --type browser` + `send` across ~10 workspaces from 3 concurrent shells) drives the app into a degraded state that persists for 30+ minutes after the burst stops:

1. **New workspaces never get a shell.** Every `cmux new-workspace --cwd <dir> --focus false` afterwards creates a workspace whose terminal surface shows title "Terminal" and never acquires a tty. Measured: some surfaces got a PTY after ~10 minutes; three workspaces created 30+ minutes ago (during quiet load) still have none. `cmux tree` shows the surface with no `tty=`.
2. **`--command` payloads do not run** until/unless the PTY eventually spawns (`new-workspace --cwd <dir> --command 'touch /tmp/probe; exec zsh'` produced no probe file for ~10 minutes).
3. **`cmux send` keystrokes to LIVE surfaces are silently queued or dropped.** During the burst, `git fetch`/`git checkout -b` sends took minutes to execute (branch appeared long after a 36s verification window expired), and some sends apparently never landed. No error is returned; the CLI prints `OK surface:N workspace:M`.
4. **The control socket intermittently refuses connections** during the burst: `Error: Failed to connect to socket at ~/.local/state/cmux/cmux.sock (Connection refused, errno 61)`, `Error: Socket closed before reply`, and `Error: Command timed out` from plain `cmux tree` calls.

The app process did NOT crash (same PID throughout, 8h+ uptime). Existing PTYs keep working normally the whole time — only NEW surface/PTY creation and command delivery degrade.

## Environment / measurements (2026-08-06, 0.64.22 dev machine)

- App fd count during wedge: ~1080 (358 PIPE, 267 KQUEUE) — not at any obvious limit
- User procs 1207 / 6000 limit; disk was 99% full (9-14GB free) during part of the window, but the wedge persisted with 31GB free too
- ~130 workspaces, ~20 concurrent codex TUIs in other workspaces
- Reproduced socket refusals with zero parallelism afterwards (single sequential CLI calls)

## Likely mechanism

Known architecture issue: CLI socket commands execute on the app main thread. A dispatch burst piles up main-thread work (workspace creation, WKWebView browser-pane loads, O(N) sidebar updates per event, PTY spawn completions). PTY spawn appears to be queued behind this backlog and starves indefinitely, while the socket listener itself stalls long enough to refuse/drop connections. Related: the pipe-fd leak family (#9175) and CLI-on-main-thread hangs; also #9649 (SwiftUI reentrant layout crash under high-frequency browser automation) came out of the same burst.

## Impact

Bulk agent dispatch — the flagship cmux workflow — cannot create workspaces at all once the wedge sets in; the only recovery appears to be restarting the app, which kills every running agent PTY. Dispatcher had to fall back to hijacking shells of pre-existing workspaces.

## Repro sketch

1. From 2-3 shells concurrently, loop ~10x: `cmux new-workspace --cwd <clone> --focus false`, `cmux new-pane --type browser --direction right --workspace <ws> --url <github-issue> --focus false`, several `cmux send` git commands with 3s sleeps.
2. Observe socket refusals/timeouts begin; observe late/dropped keystroke delivery.
3. After the burst: create a single workspace and watch `cmux tree` — the terminal surface stays without `tty=` for many minutes or forever.

## Expected

- PTY spawn for a new surface must not be starvable by CLI/browser/sidebar load; move socket command handling and PTY spawn off the main thread (or give spawn its own queue with priority).
- `cmux send` must fail loudly (or retry/ack) when delivery cannot be confirmed, never silently drop or delay keystrokes by minutes.
- The control socket should keep accepting connections under main-thread load.

austinywang:
Fixes #9769.

## What the investigation found

The issue's live data has two distinct signatures: transient socket refusals/timeouts during the burst, and a **permanent** PTY-spawn wedge that persists on a healthy socket (the follow-up comment shows `workspace:161`/`surface:380` never spawning while `tree` answers instantly). Tracing the full path (`handleSocketWorkerV2` → `v2MainSync` → `TabManager.addWorkspace` → `BackgroundWorkspacePrimeCoordinator` → `TerminalSurface` cold start → `ghostty_surface_new`) turned up a class of **silent one-shot drops** in the cold-start path, each with no re-arm:

1. **Bootstrap-window custody loss (the permanent wedge).** A cold surface starts its runtime inside a hidden borderless "bootstrap" `NSWindow` that hosts the pane host (`GhosttySurfaceScrollView`). Window-portal churn (`TerminalWindowPortal.detachHostedView` → `removeFromSuperview`) can reparent the pane host out of that window and park it with **no window at all** — while `headlessStartupWindow` stays recorded. From then on every cold start hits `ensureHeadlessStartupWindowIfNeeded`'s `guard headlessStartupWindow == nil else { return }`, and the follow-up `attachToView` defers on `noWindow` and returns. Nothing ever re-arms: input-demand starts (`cmux send`, `new-workspace --command`), background priming, everything funnels through the same dead end. The surface never acquires a PTY, and its queued socket input (the `--command` payload) never flushes — exactly symptoms (1) and (2).
2. **Claude command-shim gate.** `createSurface` returns early until a detached filesystem task installs the optional `claude` wrapper shim. If that task never completes (the machine had a 99%-full disk during part of the window), PTY spawn is starved forever by a PATH convenience.
3. **Background-prime mount-slot pinning (the app-wide amplifier).** `BackgroundWorkspacePrimeCoordinator`'s timeout path deliberately keeps a workspace's hidden mount retained. A surface that can never become ready (custody-wedged as above, closing panel, agent-hibernation suspension) therefore pins one of the **two** global mount slots forever, and every later workspace's prime pass starves behind it.
4. **Silent queued sends.** `surface.send_text` to a PTY-less surface queues the bytes and replies `ok` with `queued: true` — but the CLI's human output (`v2OKSummary`) dropped the flag and printed a bare `OK surface:N workspace:M`, indistinguishable from a delivered send. Symptom (3).

## The fix (removing the class, not patching a symptom)

- **Custody reclamation** (`TerminalSurface+RuntimeLifecycle.swift`): `ensureHeadlessStartupWindowIfNeeded` now (a) re-adopts a parked, windowless pane host into the existing bootstrap window, and (b) discards a stale bootstrap window once the pane host lives in a real window. The bootstrap path is now level-triggered: any cold-start request on a surface whose views are windowless deterministically ends with the pane host hosted in a window and `createSurface` attempted. This closes the drop for input-demand, queued-send, and background-prime starts in one place.
- **Shim deadline** (`TerminalSurface+ClaudeCommandShimLifecycle.swift`): a bounded, cancellable deadline (injected `Clock`, default 5s, stored task, cancelled on completion/teardown/hibernation) lets spawn proceed without the shim when the install hangs. A late install result still serves future runtime creations. (Clock-sleep carve-out per `cmux-architecture`: a genuine deadline, driven by an injected clock.)
- **Prime work excludes non-startable surfaces** (`Workspace.swift` + new `TerminalSurface.canCreateRuntimeSurface`): a surface whose lifecycle forbids runtime creation no longer counts as background-prime work, so the coordinator completes the workspace and releases its mount slot instead of pinning it forever.
- **Queued sends are visible** (`CLI/cmux.swift`): `send`, `send-key`, `send-panel`, `send-key-panel` append a localized `queued (terminal starting; input delivers when its PTY spawns)` marker when the reply says `queued: true`. Combined with the spawn fixes, `queued` is now a delivery guarantee rather than a silent maybe; overflow and dead-process cases already return errors (`input_queue_full`, `process_exited`). The v1 socket lane's bare-`OK` reply contract predates the queued flag and is left unchanged for legacy parsers — delivery itself is fixed at the surface level for every entrypoint.

## Regression tests (commit 1, red → commit 2, green)

- `TerminalSurfaceBootstrapCustodyTests` (CmuxTerminal package, runs in the `swift-package-tests` CI job): reproduces the parked-pane-host wedge for input-demand starts, queued sends, and background-prime starts, plus the hung-shim starvation. All four fail on the test-only commit (verified locally: the post-park start attempts stay at 0/1 forever) and pass with the fix.
- `BackgroundPrimeStartableSurfaceTests` (app-host): a closing surface must not count as background-prime work / must let priming complete.
- `CLISendQueuedOutputTests` (app-host, bundled CLI against a mock socket): a `queued: true` reply must print the queued marker; a plain reply must keep printing exactly `OK surface:N workspace:M`.

## What this PR does not claim to fix

The transient burst-time socket refusals/timeouts (symptom 4) are the #5757 family: every `.mainActor` command serializes through a timeout-less `DispatchQueue.main.sync` (`v2MainSync`), and connection threads pile up behind it until the accept-side limits (32 preauthorization slots, 32-connection stream buffer, 128 listen backlog) spill. Fixing that means moving CLI command execution off the main thread wholesale, which is an orthogonal, much larger change tracked in #5757 — this PR removes the part of #9769 that never recovers (PTY spawn and silent sends), which is also the part the issue's own follow-up data shows is independent of socket health. A true CI integration repro of the full multi-shell burst (3 concurrent CLI loops against a live app) is not cleanly testable in the app-host harness; the scheduling logic that wedged is covered by the deterministic unit/behavior tests above instead.

## Localization audit

New user-facing string: the CLI queued-send suffix, added as `cli.send.queuedSuffix` to `Resources/Localizable.xcstrings` with `en` and `ja` translations. No web UI, settings, menu, shortcut, schema, or docs text changed; no web message catalogs affected.

## CI evidence (dispatch-triggered; this repo's CI does not auto-run on PRs)

**Red on the test-only commit `6d4a2933a5`** ([run 31159985469](https://github.com/manaflow-ai/cmux/actions/runs/31159985469)): all four `TerminalSurfaceBootstrapCustodyTests` failed in `swift-package-tests` on exactly the intended assertions (post-park create attempts stuck at 0/1), and `BackgroundPrimeStartableSurfaceTests` failed in app-host shard 3 on its two assertions.

**Green on the head** ([run 31173407217](https://github.com/manaflow-ai/cmux/actions/runs/31173407217) + rerun): `swift-package-tests` green (custody, shim-deadline, and shim-re-gate tests), shard-4 focused gate green (`BackgroundPrimeStartableSurfaceTests`, `CLISendQueuedOutputTests` against the real bundled CLI), shards 1/2 green, `release-build` green on the head SHA, zero new Swift warnings.

**Jobs still red are the pre-existing main baseline**, verified fingerprint-identical against main's latest run ([31150302964](https://github.com/manaflow-ai/cmux/actions/runs/31150302964)):
- `tests-build-and-lag`: inherited warning-budget violation in `Sources/NotificationsPage.swift` (untouched here; main additionally has a second violation this branch predates).
- app-host shards 3/4: pre-existing environment-dependent suites (`CLINotifyProcessIntegrationRegressionTests` ssh/hostname expectations, `AgentSessionAutoResumeSettingsTests` `/usr/bin/ssh` normalization, `BrowserPanelRemoteStoreTests`, …) and the `RemoteTmuxMirrorCloseDetachTests` focused step exiting 65 after its suite passes — all reproduced on main's runner class. No `cmuxTests` failure on the head run's shard 4 at all.
- `tests` / `ci-status` aggregate the above. No required status checks are configured on `main`; the PR is `MERGEABLE`/`CLEAN`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -671,6 +671,27 @@ jobs:
             -only-testing:cmuxTests/GhosttySurfaceOverlayTests/testFiveTabRendererFootprintReturnsToOneRendererTargetAcrossHideRevealCycles \
             test
 
+      - name: Run PTY spawn starvation regressions
+        if: ${{ matrix.shard == fromJSON(env.CMUX_APP_HOST_FOCUSED_REGRESSION_SHARD) }}
+        run: |
+          # The tolerant full-suite step can accept ordinary assertion
+          # failures. Keep the issue #9769 coverage — background-prime work
+          # must exclude never-startable surfaces, and `cmux send` must
+          # surface queued delivery — on a non-tolerant focused invocation.
+          set -euo pipefail
+          SOURCE_PACKAGES_DIR="$PWD/.ci-source-packages"
+          scripts/ci/run-in-console-session.sh \
+            scripts/ci/run-app-host-xcodebuild.sh \
+            -project cmux.xcodeproj -scheme cmux-unit -configuration Debug \
+            -derivedDataPath "$CMUX_DERIVED_DATA_PATH" \
+            -clonedSourcePackagesDirPath "$SOURCE_PACKAGES_DIR" \
+            -disableAutomaticPackageResolution \
+            -destination "platform=macOS" \
+            CMUX_SKIP_ZIG_BUILD=1 \
+            -only-testing:cmuxTests/BackgroundPrimeStartableSurfaceTests \
+            -only-testing:cmuxTests/CLISendQueuedOutputTests \
+            test
+
       - name: Run notification routing regressions
         if: ${{ matrix.shard == fromJSON(env.CMUX_APP_HOST_FOCUSED_REGRESSION_SHARD) }}
         run: |
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -18313,7 +18313,7 @@ struct CMUXCLI {
         guard (payload["queued"] as? Bool) == true else { return summary }
         let suffix = String(
             localized: "cli.send.queuedSuffix",
-            defaultValue: "queued (terminal starting; input delivers when its PTY spawns)"
+            defaultValue: "queued (terminal starting; input will be sent when its PTY is ready)"
         )
         return "\(summary) \(suffix)"
     }
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -5317,7 +5317,7 @@ struct CMUXCLI {
             let sfId = try normalizeSurfaceHandle(surfaceArg, client: client, workspaceHandle: wsId, windowHandle: winId)
             if let sfId { params["surface_id"] = sfId }
             let payload = try client.sendV2(method: "surface.send_text", params: params)
-            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2OKSummary(payload, idFormat: idFormat))
+            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2SendSummary(payload, idFormat: idFormat))
 
         case "send-key":
             let (wsArg, rem0) = parseOption(commandArgs, name: "--workspace")
@@ -5336,7 +5336,7 @@ struct CMUXCLI {
             let sfId = try normalizeSurfaceHandle(surfaceArg, client: client, workspaceHandle: wsId, windowHandle: winId)
             if let sfId { params["surface_id"] = sfId }
             let payload = try client.sendV2(method: "surface.send_key", params: params)
-            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2OKSummary(payload, idFormat: idFormat))
+            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2SendSummary(payload, idFormat: idFormat))
 
         case "send-panel":
             let (wsArg, rem0) = parseOption(commandArgs, name: "--workspace")
@@ -5358,7 +5358,7 @@ struct CMUXCLI {
             let sfId = try normalizeSurfaceHandle(panelArg, client: client, workspaceHandle: wsId, windowHandle: winId)
             if let sfId { params["surface_id"] = sfId }
             let payload = try client.sendV2(method: "surface.send_text", params: params)
-            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2OKSummary(payload, idFormat: idFormat))
+            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2SendSummary(payload, idFormat: idFormat))
 
         case "send-key-panel":
             let (wsArg, rem0) = parseOption(commandArgs, name: "--workspace")
@@ -5380,7 +5380,7 @@ struct CMUXCLI {
             let sfId = try normalizeSurfaceHandle(panelArg, client: client, workspaceHandle: wsId, windowHandle: winId)
             if let sfId { params["surface_id"] = sfId }
             let payload = try client.sendV2(method: "surface.send_key", params: params)
-            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2OKSummary(payload, idFormat: idFormat))
+            printV2Payload(payload, jsonOutput: jsonOutput, idFormat: idFormat, fallbackText: v2SendSummary(payload, idFormat: idFormat))
 
         case "notify":
             let title = optionValue(commandArgs, name: "--title") ?? "Notification"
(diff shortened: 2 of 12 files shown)


austinywang:
Review findings addressed in 869da76ed3:

- **Bugbot (shim deadline blocks later reinstall)** — fixed. `cancelClaudeCommandShimInstallLifecycle` now reopens the install gate when no shim materialized, so a deadline-released spawn plus a lifecycle cancel (teardown / agent-hibernation suspend) can no longer lock a surface into shim-less mode. The hung-shim regression test now also covers the post-cancel re-gate.
- **Bugbot (hung shim test leaks continuation)** — fixed earlier in c723de4faf; the test resumes the parked installer continuations at the end.
- **CodeRabbit (locale-dependent "queued" assertion)** — fixed; the child CLI is pinned to an English locale.
- **CodeRabbit (mock-server timeout swallowed)** — fixed; the test now throws when the server does not complete.
- **CodeRabbit (grammar of the queued marker)** — reworded to "queued (terminal starting; input will be sent when its PTY is ready)" in the catalog and CLI default.
- **CodeRabbit (xcstrings locale coverage)** — declined: `en` + `ja` is the dominant convention for `cli.*` keys in this catalog (250 of ~407 entries, including the sibling `cli.creation.remoteTmux.*` keys this entry was modeled on).
- **CodeRabbit (inject a virtual clock into the shim deadline test)** — declined as disproportionate: the deadline duration and clock are injectable (the test injects a 100 ms deadline), and the bounded polling wait matches the existing pattern in `TerminalSurfaceRestoreSpawnSchedulerTests`.

PR #9844, opened 2026-08-08
fix(shell-integration): back off PR polling for branches with no PR

### LINKED ISSUE #9843: Shell integration polls gh pr view every 45s even for branches with no PR, draining the GitHub GraphQL rate limit
## Summary

The shell integration's per-panel PR poll loop (`_CMUX_PR_POLL_INTERVAL=45` in `Resources/shell-integration/cmux-zsh-integration.zsh`, same in the bash variant) runs `gh pr view "$branch" --json number,state,url` every interval unconditionally. The probe writes a per-panel cache including a `.no-pr-branch` marker when a branch has no PR — but that cache is never consulted to skip the next poll.

Net effect: a panel whose cwd sits on a branch that will never have a PR (most commonly `main`) burns one GitHub **GraphQL** call every 45 seconds, forever. `gh pr view` uses GraphQL, and GitHub's GraphQL quota (5,000/hr) is pooled per user account across all machines and tools. Measured on a machine with six workspaces open (each on `main`): ~750 GraphQL calls/hour of pure no-op polling — enough to exhaust the shared quota when combined with normal `gh` usage, causing unrelated tooling to hit rate-limit errors.

## Reproduction

1. Open a few cmux workspaces with shells sitting in git repos on `main` (no open PR for `main`).
2. Watch `gh api rate_limit --jq .resources.graphql.used` climb steadily (~80/hr per panel), or shim `gh` to log invocations — every ~45s per panel: `pr view main --repo <slug> --json number,state,url`.
3. Note the probe's own debug log records `cache-refresh`/`cache-miss` reads of the no-PR cache, then calls `gh` regardless.

## Secondary defect found while fixing

The probe stamps its cache with `now="${EPOCHSECONDS:-$SECONDS}"`. `EPOCHSECONDS` requires `zsh/datetime` (not loaded here), so this silently falls back to `$SECONDS` — seconds since shell start, not epoch. The timestamps written to `.timestamp` are therefore not comparable across shells or restarts. This has gone unnoticed because nothing ever reads them — but any fix that starts consulting the cache (like the backoff below) trips over it immediately.

## Proposed fix

PR incoming from this account: consult the existing `.no-pr-branch` cache and skip the `gh` call for 900s per no-PR branch (the existing force-refresh signal still bypasses immediately; branches *with* PRs keep the 45s cadence), and normalize `now` to a real epoch via `date +%s` so the timestamp comparison is coherent. Reduces the no-PR baseline by ~95% while still noticing a PR opened later within 15 minutes.

MrPicklesRocks:
Fixes #9843.

## Problem

The per-panel PR poll loop calls `gh pr view` every `_CMUX_PR_POLL_INTERVAL` (45s) even when the probe's own cache already recorded that the current branch has no PR — the `.no-pr-branch` marker is written and read into a variable, but never used to skip polling. A panel sitting on `main` therefore burns one GitHub GraphQL call every 45s forever; with several workspaces open this consumes a large share of the user's account-wide 5,000/hr GraphQL quota (measured ~750 calls/hr with six workspaces).

## Fix

- In `_cmux_report_pr_for_path`, when not force-probed and the cached no-PR branch equals the current branch and the cache is fresher than 900s, skip the `gh` call (debug-logged as `no-pr-backoff`). The force-refresh signal still bypasses immediately; branches with PRs keep the 45s cadence; a PR opened later on a quiet branch is noticed within 15 minutes.
- Normalize `now` to a real epoch via `date +%s`: `EPOCHSECONDS` requires `zsh/datetime` (not loaded), so the existing `${EPOCHSECONDS:-$SECONDS}` fallback wrote shell-relative values to `.timestamp` — harmless while unread, but incorrect the moment anything compares them. This also makes the function's existing timestamp writes epoch-based.
- Same change applied to the bash integration (with a bash-compatible numeric check).

## Testing

Verified with a stub `gh` on PATH, a seeded per-panel cache, and a throwaway repo, across three cases: fresh no-PR cache + unforced probe → 0 gh calls; fresh cache + forced probe → 1 call; stale (>900s) cache + unforced → 1 call. `zsh -n` / `bash -n` clean on both files. Also running patched in production across four machines since 2026-08-07.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9844"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788758620&installation_model_id=21920&pr_number=9844&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9844&signature=f7e892dbaa053c1e9c299eec172e45a2114eb43f44d817b907ca3a73de582ea9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---

> [!NOTE]
> **Low Risk**
> Shell-side polling and cache timing only; reduces GitHub API usage without changing auth or app server behavior.
> 
> **Overview**
> Stops **repeated `gh pr view` calls** when the per-panel PR cache already marks the current branch as having no PR (e.g. `main`). In `_cmux_report_pr_for_path`, unforced probes now **return early** if the cached no-PR branch matches, the timestamp is younger than **900s**, and log `no-pr-backoff`; **force refresh** still runs `gh` immediately, and branches with PRs keep the normal ~45s poll.
> 
> Also **normalizes time to wall-clock epoch** via `date +%s` before comparing cache timestamps and writing them, so backoff works when `EPOCHSECONDS` is missing (bash/zsh) and avoids shell-relative `$SECONDS` values breaking comparisons. Same logic in **bash** and **zsh** integration files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38b3176d798ee6c1d88bb2df33e39eceac1e8261. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Backs off PR polling for branches with no PR to cut unnecessary `gh` calls and guard against bad cache data. Also normalizes timestamp handling in `zsh`/`bash` for correct cache age checks (fixes #9843).

- **Bug Fixes**
  - Skip `gh pr view` for 900s when the cache says "no PR" for the same branch and repo with a valid, non-future timestamp; forced probes still bypass, and branches with PRs keep the 45s cadence.
  - Normalize timestamps to epoch via `date +%s`; if epoch derivation fails, skip the backoff for that check to avoid mixed time bases; apply to both shell integrations.

<sup>Written for commit 1dac9715069cb7e1ce6a5dd8211b1e97c330f7d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9844?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Performance**
  * Reduced repeated pull request checks for branches recently confirmed to have no pull request.
  * Added a 15-minute cache backoff for matching non-forced checks.
  * Forced checks continue to bypass the backoff for immediate updates.
  * Checks proceed normally when cache timestamps are invalid, unavailable, or cannot be safely validated.

### DIFF
--- a/Resources/shell-integration/cmux-bash-integration.bash
+++ b/Resources/shell-integration/cmux-bash-integration.bash
@@ -1276,6 +1276,24 @@ _cmux_report_pr_for_path() {
         _cmux_pr_debug_log "$branch" "cache-miss"
     fi
 
+    # A branch already known to have no PR is re-checked at most every 900s
+    # (the force signal still bypasses immediately). Without this, a panel
+    # sitting on a branch that will never have a PR (e.g. main) burns a
+    # GitHub GraphQL API call every poll interval forever — and the GraphQL
+    # quota (5,000/hr) is shared across the user's whole account.
+    # EPOCHSECONDS may be unavailable in older bash and the $SECONDS
+    # fallback is shell-relative, so normalize `now` to a real epoch here;
+    # this also makes the timestamp writes below epoch-based so the
+    # comparison stays coherent across shells.
+    now="$(command date +%s 2>/dev/null || echo "$now")"
+    local cache_timestamp=""
+    [[ -r "$timestamp_file" ]] && cache_timestamp="$(<"$timestamp_file")"
+    if [[ "${2:-0}" != "1" && -n "$branch" && "$cache_no_pr_branch" == "$branch" ]] \
+        && [[ "$cache_timestamp" =~ ^[0-9]+$ ]] && (( now - cache_timestamp < 900 )); then
+        _cmux_pr_debug_log "$branch" "no-pr-backoff"
+        return 0
+    fi
+
     repo_slug="$(_cmux_github_repo_slug_for_path "$repo_path")"
     if [[ -n "$repo_slug" ]]; then
         gh_repo_args=(--repo "$repo_slug")
--- a/Resources/shell-integration/cmux-bash-integration.bash
+++ b/Resources/shell-integration/cmux-bash-integration.bash
@@ -1281,17 +1281,28 @@ _cmux_report_pr_for_path() {
     # sitting on a branch that will never have a PR (e.g. main) burns a
     # GitHub GraphQL API call every poll interval forever — and the GraphQL
     # quota (5,000/hr) is shared across the user's whole account.
-    # EPOCHSECONDS may be unavailable in older bash and the $SECONDS
-    # fallback is shell-relative, so normalize `now` to a real epoch here;
-    # this also makes the timestamp writes below epoch-based so the
-    # comparison stays coherent across shells.
-    now="$(command date +%s 2>/dev/null || echo "$now")"
-    local cache_timestamp=""
-    [[ -r "$timestamp_file" ]] && cache_timestamp="$(<"$timestamp_file")"
-    if [[ "${2:-0}" != "1" && -n "$branch" && "$cache_no_pr_branch" == "$branch" ]] \
-        && [[ "$cache_timestamp" =~ ^[0-9]+$ ]] && (( now - cache_timestamp < 900 )); then
-        _cmux_pr_debug_log "$branch" "no-pr-backoff"
-        return 0
+    # The guard only trusts a cache record written for this exact repo with
+    # an explicit "none" result and a plausible (not future) epoch age.
+    # EPOCHSECONDS may be unavailable in older bash and the $SECONDS fallback is
+    # shell-relative, so `now` is normalized to a real epoch first; if that
+    # normalization fails the backoff is skipped for this invocation and
+    # `now` is left untouched.
+    local epoch_now="" cache_repo="" cache_timestamp=""
+    epoch_now="$(command date +%s 2>/dev/null || true)"
+    [[ "$epoch_now" =~ ^[0-9]+$ ]] || epoch_now=""
+    if [[ -n "$epoch_now" ]]; then
+        now="$epoch_now"
+        [[ -r "$timestamp_file" ]] && cache_timestamp="$(<"$timestamp_file")"
+        [[ -r "$repo_file" ]] && cache_repo="$(<"$repo_file")"
+        if [[ "${2:-0}" != "1" && -n "$branch" \
+            && "$cache_no_pr_branch" == "$branch" \
+            && "$cache_repo" == "$repo_path" \
+            && "$cache_result" == "none" ]] \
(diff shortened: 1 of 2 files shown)


MrPicklesRocks:
Thanks @coderabbitai — all three findings were valid; addressed in the latest commits:

1. **Future timestamps**: the guard now requires `now >= cache_timestamp` in addition to the 900s upper bound, so a corrupt/future value can never suppress polling indefinitely.
2. **Per-panel cache vs repo identity**: the guard now also requires the cached `.repo` to equal the current `repo_path` and the cached `.result` to be exactly `none` — a panel that moves between repos with same-named branches (e.g. `main`) no longer inherits the previous repo's no-PR record.
3. **Non-epoch fallback removed**: `now` is only overwritten when `date +%s` yields a valid integer; on failure the backoff is skipped for that invocation and `now` (and therefore the function's existing timestamp writes) keep their legacy behavior, so no mixed-time-base value is ever compared. I intentionally did not touch the existing write paths beyond that — making writes conditional on epoch validity would grow the diff into code this PR otherwise doesn't change; happy to do that as a follow-up if maintainers prefer.

Re-verified with a stubbed `gh`, seeded per-panel cache, and a throwaway repo across six cases: fresh-cache unforced → 0 calls; forced → 1; stale (>900s) → 1; **future timestamp → 1**; **different cached repo → 1**; **cached result ≠ none → 1**. `zsh -n`/`bash -n` clean on both files.

PR #9884, opened 2026-08-10
iOS: restore the search session after popping a search-opened workspace

azooz2003-bit:
Follow-up to https://github.com/manaflow-ai/cmux/pull/9820, from device dogfood direction: returning from a workspace opened via search must land back in the search state that was left behind, the detail must be a normal navigation push, and the tab bar must not be visible on the detail.

**Design.** Hosted runs measured a hard platform constraint: with a pushed detail whose tab bar is hidden, every programmatic re-presentation of the search field (isPresented re-activation in six shapes, the searchFocused focus engine, onAppear/onDisappear timings) hosts it in the top navigation drawer; the bottom morph is granted exclusively to the user's own activation of the search control. A fullScreenCover preserved everything but was rejected as not-a-push. The shipped design keeps the push and stops fighting the platform: selecting a result pushes the detail inside the search tab's NavigationStack, the stack's path drives `toolbarVisibility(.hidden, for: .tabBar)`, and the session ends from the destination's `onAppear` (ending it at select-time swallows the push, measured twice). Popping lands on the still-filtered results with the query preserved in the bottom search control; one tap re-expands the field at the bottom with the query intact.

Three structural fixes came out of the failing notification leg, diagnosed from CI xcresult screen recordings: the search tab now hosts ONE NavigationStack with ONE searchable mounted from launch and only the root content switches with the coordinator scope (the platform adopts the search-role tab's searchable once per TabView, so a scope-switched second stack's searchable was never adopted); `searchToolbarBehavior(.minimize)` keeps the inactive field out of the top chrome (a List root otherwise pins the deactivated field full-width under the nav bar, measured at y=117); and NotificationFeedView suppresses its large title in the search destination so the search stack stays title-less like the workspace one. Shell and both preview fixtures share the shape, and the two per-scope paths and restore flags collapsed into one.

**Evidence** (hosted iphone legs on the final head, isolated runner simulators, all green): run 31547578400 covers the push inside the search tab, hidden tab bar on the detail, settled no-top-field invariant after pop, bottom control frame, one-tap re-expansion carrying "Docs", and typing/clearing in the re-expanded field; run 31547625513 covers the pop-minimized state plus explicit submit with the committed filter surviving a refresh; run 31547674703 covers control geometry across primary roots; run 31546817747 covers the notification analog including read-row/unread-filter semantics after opening a notification.

Known pre-existing issues, untouched here: the unread-filter chip inside the notification search destination does not respond to taps; testNotificationFeedPreviewSupportsTriageInteractions fails identically on plain main (run 31475643831).

Localization audit: no new user-facing strings (the notification title string moved into a conditional modifier; prompts unchanged).


azooz2003-bit:
Closing unmerged by dogfood decision: after measuring every restore variant (top-hosted programmatic re-presentation, cover, one-tap bottom re-expansion), the simple checkpointed behavior from https://github.com/manaflow-ai/cmux/pull/9820 wins: popping the search-opened workspace detail returns to the Workspaces list with no search active. Main already ships that, so this PR's restore design is retired. The measured platform constraints live in the PR history for the next attempt.

Issue #9348, opened 2026-08-01
Flaky CI: app-host shard 4 test host crashes during RemoteTmuxMirrorCloseDetachTests


austinywang:
CI job `app-host unit tests (4/4)`, step "Run remote tmux mirror detach and placement regressions", fails intermittently because the test-host app process crashes mid-suite in `RemoteTmuxMirrorCloseDetachTests`. xcodebuild logs `Restarting after unexpected exit, crash, or test timeout`, the relaunched host passes every remaining test, and the session still exits 65.

Evidence it is repo-wide, not branch-specific:

- `issue-9231-workspace-switch-latency` run 30682396883: host crashed while `ordinaryCloseOfLiveMirrorDetachesWithoutKillingSession()` was running; on rerun the crash moved to `dedicatedWindowSocketDefaultsToFocusNeutral()` (that test's window teardown). The crash point is nondeterministic.
- `issue-9330-stale-surface-socket-binding` run 30680581107: same step failed.
- `feat-ios-dogfood-ready` run 30679561158: same step failed.
- Same branch, earlier HEAD (run 30680051894): same step failed.

Each crash writes a sentry ghostty crash report on the runner (e.g. `~/.local/state/cmux/crash/<uuid>.ghosttycrash`) that is not uploaded as an artifact, so the faulting stack is currently unrecoverable from CI. Suggest (1) uploading `~/.local/state/cmux/crash` as an artifact on failure so the next occurrence yields a stack, and (2) auditing the dedicated-window create/close path these tests exercise for the crash itself.

azooz2003-bit:
More data: the crashing launch dies in -[NSThemeFrame addSubview:] under installFileDropOverlay(on:tabManager:) → AppDelegate.createMainWindow → bootstrapInitialMainWindowIfNeeded, i.e. the test host's async initial-window bootstrap racing the test bundle. Reproduced 3/3 on gate runs today, including directly on main-tip 5a8a973ebc (https://github.com/manaflow-ai/cmux/actions/runs/30684026978, shard 4), so it predates today's commits and is not PR-correlated. Suites pass after the harness restart; xcodebuild still exits 65.

austinywang:
[...]

PR #9921, opened 2026-08-10
Prevent recursive deferred-action release chains

### LINKED ISSUE #9920: Recurrence on 0.64.20 (10-day uptime, Intel): main-thread stack overflow releasing chained DispatchWorkItems — chain node is a god-object deinit; the real producer from #9173 was never identified
Another main-thread stack overflow from a recursive `DispatchWorkItem` release chain, same signature as https://github.com/manaflow-ai/cmux/issues/8606 and https://github.com/manaflow-ai/cmux/issues/9173 — but this report adds binary forensics that narrow the chain node to a specific shape of class, which the https://github.com/manaflow-ai/cmux/pull/9179 audit explicitly never identified ("Static analysis therefore did not prove that one of these post-#8615 sites was the released-build chain producer").

## Crash

Crash report: `cmux-2026-08-10-160412.ips` (local copy on the triage machine at `/tmp/cmux-2026-08-10-160412.ips`).

- cmux **0.64.20** release (build 100), slice UUID `07374F7E-E17C-3814-81E5-A8CB5947F49E` (x86_64)
- Macmini8,1 (Intel), macOS 15.7.8 (24G824)
- Process launched **2026-07-31 18:46**, crashed **2026-08-10 16:03** — **~10 days of uptime**, foreground
- `EXC_BAD_ACCESS` / `KERN_PROTECTION_FAILURE` at `0x7ff7b6675ff8`, which vmRegionInfo places **7 bytes from the end of "STACK GUARD for thread 0"** — classic main-thread stack overflow
- Faulting thread 0 (`com.apple.main-thread`), 512 recorded frames (truncated), ~46 repetitions of this cycle:

```
cmux 0x9f5c54                                  <- __deallocating_deinit of a cmux class (see below)
swift_release_dealloc / doDecrementSlow
DispatchWorkItem.__deallocating_deinit
_Block_release -> _call_dispose_helpers_excp
__destroy_helper_block_... dispatch_block_private_data_s
_Block_release -> _call_dispose_helpers_excp
swift_release_dealloc / doDecrementSlow
cmux 0x9f5c54                                  <- next node, same deinit
```

Innermost frames: the recursion bottoms out in cmux offset `0x9eada1` calling `0x3ff947`, whose prologue `___chkstk_darwin` stack probe hits the guard page.

## What the binary says (new evidence)

Symbolicated against the official `v0.64.20` `cmux-macos.dmg` (UUID matches). The binary is stripped, but function boundaries and disassembly show:

- Both cmux frames in the cycle (`0x9eada1` and `0x9f5c54`) live inside **one enormous function starting at `0x9e0100`** — over **88 KB of code**, releasing hundreds of stored properties, with field offsets loaded indirectly from runtime field-offset vectors. This is the `__deallocating_deinit` of a **god-object class with a very large number of stored properties** (pure Swift — it goes through `swift_release_dealloc`, not objc `-dealloc`).
- The cycle is: god-object deinit → releases last reference to a stored `DispatchWorkItem` → its block's dispose helpers release a strong capture of **another instance of the same class** → its deinit runs → repeat. A linked chain of `<big object> -> work item -> <big object> -> ...` torn down recursively on the 8 MB main stack.
- ~10 days of uptime to build the chain implies a **slow producer** (and a matching unbounded memory leak while the chain lives), unlike the high-frequency sidebar-resizer path fixed in https://github.com/manaflow-ai/cmux/pull/8615.

Symbolication recipe for whoever picks this up:

```
gh release download v0.64.20 --repo manaflow-ai/cmux --pattern cmux-macos.dmg
# mount, extract cmux binary; dwarfdump --uuid confirms 07374F7E-... (x86_64)
# crash base 0x10908a000; file-address = 0x100000000 + offset
# lldb: disassemble --start-address 0x1009e0100 ... (giant deinit)
#       frames of interest: 0x1009eada1, 0x1009f5c54 (both inside the giant deinit), 0x1003ff947
```

## Relationship to prior fixes

- https://github.com/manaflow-ai/cmux/pull/8615 (two `ContentView` sites) and https://github.com/manaflow-ai/cmux/pull/9179 (`MainActorDeferredActionScheduler` / `MainActorCoalescingDeadlineTimer` migration + parent-repo guard) are both in **v0.64.21+**; this crash binary is 0.64.20 and predates them, so it does not by itself prove main is still vulnerable.
- However, #9179's stored-work audit covered small owners (`ContentView`, `SidebarResizerOcclusionResolver`, `GhosttyTerminalView` slots, `Workspace` layout follow-up, `BrowserPanel` slots) and **never proved which site produced the released-build crashes** with this exact ~46-cycle signature. The chain node here is a god-object with hundreds of stored properties — that should be matchable against a small set of candidate classes, and the producing edge may or may not be one of the migrated sites.

## Task

1. Identify the god-object class whose deallocating deinit is at v0.64.20 x86_64 offset `0x9e0100` (Swift metadata in the stripped binary, arm64 slice comparison, or careful source/layout matching), and the exact site that stored a `DispatchWorkItem` whose closure strongly captured the next instance of that class.
2. Verify whether that producing edge still exists on current `main`, or is representable despite the https://github.com/manaflow-ai/cmux/pull/9179 guard (e.g. work items stored in arrays/dictionaries, chained `asyncAfter` captures, or code paths outside the guard's audited scope).
3. If a gap exists, fix it at the root (scheduler primitive migration or ownership redesign) with a regression test; also consider an iterative-teardown backstop for long work-item chains so any future producer degrades to a leak instead of a crash.
4. If current main is genuinely covered, document the producer identification here and close.

austinywang:
Closes https://github.com/manaflow-ai/cmux/issues/9920

## Summary

- identify the v0.64.20 crash producer as `ContentView.scheduleSidebarResizerCursorRelease`
- correct the apparent “god-object class” finding: the 91 KB destroy function tears down a heap capture box containing a 130-field `ContentView` value
- remove every representable closure-bearing deferred handle stored inline in macOS SwiftUI `@State`, including 11 sites outside the earlier `ContentView` audit
- add reference-owned scheduler primitives and a repository guard that prevent successor operations from retaining predecessor view snapshots
- close inferred array/dictionary and shared-package escape hatches in that ownership guard
- cover pinned Bonsplit in the ownership audit and package-test lane

## Binary evidence

The official v0.64.20 DMG has SHA-256 `a6ae6ab2f98173e5bfd4dcf46b4529fe23a9ca75fdc879310d7e6de826693e75`; its x86_64 slice UUID is `07374F7E-E17C-3814-81E5-A8CB5947F49E`, matching the crash.

The function at file address `0x1009e0100`:

- reserves `0x5ba8` bytes of stack per invocation, explaining overflow after roughly 46 recursive nodes
- calls metadata accessor `0x100909080`, destroys the reflected value payload after the heap-object header, and tail-calls `swift_deallocObject`
- corresponds to a capture-box destroy helper, not a class destructor

The metadata descriptor at `0x1054c2938` identifies the captured type as the `ContentView` struct and reports 130 fields. At the repeating return address `0x1009f5c54`, the preceding instructions load field-offset-vector entry `+0xd0`. The vector starts at `+0x10`, so this is field `(0xd0 - 0x10) / 4 = 48`; reflected field record 48 at `0x1056fe738` is `_sidebarResizerCursorReleaseWorkItem`.

The exact v0.64.20 producer was:

```swift
sidebarResizerCursorReleaseWorkItem?.cancel()
let workItem = DispatchWorkItem {
    sidebarResizerCursorReleaseWorkItem = nil
    releaseSidebarResizerCursorIfNeeded(force: force)
}
sidebarResizerCursorReleaseWorkItem = workItem
```

The closure captured a `ContentView` snapshot before replacement. Its stale inline `State._value` retained the predecessor work item, so every successor linked another capture box into the chain. https://github.com/manaflow-ai/cmux/pull/8615 removed this exact edge; https://github.com/manaflow-ai/cmux/pull/9179 later migrated the known work-item sites but could not identify the released-build producer.

## Current-main gaps and fix

The binary evidence disproves the earlier assumption that cancelling and clearing a live SwiftUI `State` location before replacement is sufficient: a captured view snapshot can still carry the wrapper's stale inline value. The same shape was representable with `Task`, `DispatchSourceTimer`, and Foundation `Timer`, not only `DispatchWorkItem`.

The first audit found five directly stored `ContentView` handles plus pinned Bonsplit's `TabItemView.globeFallbackWorkItem`. A repository-wide macOS SwiftUI audit then found 11 more:

- `SidebarSection.fontSaveTask`
- `TerminalSection.fontSaveTask`
- `SectionPopoverView.loadTask`
- `BrowserPanelView.omnibarSuggestionRefreshConsumerTask`
- `BrowserPanelView.suggestionTask`
- `BrowserPanelView.emptyStateImportBrowserRefreshTask`
- `BrowserPanelView.screenshotPageCopiedTimer`
- `GlobalSearchPaletteView.searchDebounceTimer`
- `GlobalSearchPaletteView.searchTask`
- `GlobalSearchPaletteView.refreshTask`
- `TextBoxInputContainer.pendingProviderLaunchTimeoutTimer`

The fix:

- adds `MainActorTaskStore`, where tasks strongly own private handle owners while the store indexes those owners only weakly; a lock-backed nonisolated registry makes teardown cancellation synchronous without creating a store → task → captured-view edge
- cancels predecessors before constructing successors and uses generations so stale completion cannot clear current work
- uses `MainActorDeferredActionScheduler` for one-shot delays and adds `MainActorRepeatingActionScheduler`, backed by one persistent timer handle, for the existing 16 ms cursor stabilizer
- migrates all 16 cmux task/timer state sites and the Bonsplit icon fallback away from inline closure-bearing handles
- keeps the global `DispatchWorkItem` audit and rejects `Task`, `DispatchSourceTimer`, or `Timer` stored in macOS SwiftUI `@State` under `Sources/`, `Packages/Shared/`, `Packages/macOS/`, and `vendor/bonsplit/Sources/`, including direct bracket, `Array`, and `Dictionary` initializers
- fails closed when Bonsplit is not initialized and runs Bonsplit package tests in CI

The Bonsplit gitlink is `1031b68621`, merged to `manaflow-ai/bonsplit` `main` by https://github.com/manaflow-ai/bonsplit/pull/212.

## Regression evidence

The three guard expansions were committed before their corresponding fixes:

- first red proof (the five `ContentView` handles plus Bonsplit): https://github.com/manaflow-ai/cmux/actions/runs/31448260575
- second red proof (exactly the 11 additional macOS SwiftUI handles): https://github.com/manaflow-ai/cmux/actions/runs/31451553647
- third red proof (inferred handle containers plus `Packages/Shared/`): https://github.com/manaflow-ai/cmux/actions/runs/31454601203
- Xcode 26.3 proof for the crash fix: [swift-package-tests](https://github.com/manaflow-ai/cmux/actions/runs/31483936845/job/93766606498) passed Bonsplit and every new `CmuxFoundation` scheduler/task-store regression, including both 5,000-replacement bounded-stack tests and synchronous store-teardown cancellation. The lane later failed only in the unrelated `SSHPTYAttachRetryScriptBuilderTests` fixture added by https://github.com/manaflow-ai/cmux/pull/9966 (`expected 23 auth attempts, observed 1`): its zero-count expression combines `grep -c` output with `|| printf 0`, yielding `0\n0` and bypassing the intended retry branch. Automatic PR checks and reviewers are green; the advisory full workflow also reproduced the pre-existing app-host failures tracked by https://github.com/manaflow-ai/cmux/issues/8565 and https://github.com/manaflow-ai/cmux/issues/9348.

Runtime coverage includes:

- 5,000 Bonsplit scheduler replacements with bounded teardown stack
- 5,000 main-actor keyed task replacements, requiring all capture releases on the main thread with bounded stack span
- synchronous cancellation when the last external task-store reference is released
- stale task completion cannot clear a successor
- repeating-action cancellation and non-replacement behavior
- deferred-scheduler cancellation and self-rescheduling behavior

## Iterative-teardown assessment

No generic iterative teardown backstop is included. Dispatch block captures are opaque to application code, so cmux cannot safely peel an arbitrary ownership chain one node at a time. Moving the final release to another queue only moves the overflow, while quarantining arbitrary cancelled work creates an unbounded leak. Keeping predecessor handles out of captured value snapshots is the reliable root fix. The new reference-owned primitives isolate any stranded operation as an independent leak rather than linking it to predecessors, so eventual teardown cannot recurse through the replacement history.

## Validation

- `python3 scripts/lint-stored-dispatch-work-items.py`
- `python3 -m py_compile scripts/lint-stored-dispatch-work-items.py tests/test_lint_stored_dispatch_work_items.py`
- `git diff --check`
- no local app build, `xcodebuild`, test suite, launch, or dogfood run, per issue instructions
- no user-facing strings changed; localization audit found nothing to localize

---

> [!NOTE]
> **High Risk**
> Changes core main-thread concurrency and lifecycle patterns across many SwiftUI surfaces and tightens CI gates on deferred-work ownership; regressions could affect cancellation timing, search debouncing, or crash on teardown, though the direction reduces known stack-overflow risk.
> 
> **Overview**
> Addresses the v0.64.20 stack-overflow class of crashes where **successor** deferred work (tasks, timers, `DispatchWorkItem`) could retain **predecessor** SwiftUI `@State` snapshots and form deep recursive teardown chains.
> 
> **New primitives in `CmuxFoundation`:** `MainActorTaskStore` (keyed replaceable async work with weak task-owner indexing so captured view state cannot chain through task handles), `MainActorRepeatingActionScheduler` (single persistent timer for repeating main-actor work, e.g. the sidebar resizer cursor stabilizer).
> 
> **App migrations:** Inline `@State` `Task`, `DispatchSourceTimer`, and `Timer` handles are replaced across `ContentView` (command palette search/index/fork probes, cursor stabilizer), browser panel omnibar work, global search, session index pagination, settings font saves, and text-box provider launch timeouts—mostly via task stores and existing deferred schedulers.
> 
> **CI and guards:** Bonsplit submodule init runs before the ownership lint; Bonsplit package tests run in the Swift package lane (with the same startup crash retry as other packages). `lint-stored-dispatch-work-items.py` now audits macOS SwiftUI `@State` for closure-bearing deferred handles under `Sources/`, `Packages/Shared/`, `Packages/macOS/`, and pinned `vendor/bonsplit/Sources/`, fails closed if Bonsplit sources are missing, and tightens SwiftPM test-runner flake detection.
> 
> Tests cover burst replacement stack bounds, cancellation on store release, stale completion, and repeating-scheduler behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit baa99df7cc4f6041406443f9ba7d19ad2a97310b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


austinywang:
Addressed all three CodeRabbit findings in `553a53af4f`:

- Both SwiftPM retry paths now require `Build complete!`, signal 5 or 6, and no XCTest/Swift Testing output, so post-start crashes are not retried as infrastructure flakes.
- `MainActorTaskStore` now constructs tasks internally. Each task strongly owns a private handle owner, the store holds that owner weakly for cancellation, and completion captures the store weakly. This removes the store → task → captured owner cycle while preserving real `Task.cancel()` propagation and generation-safe stale completion.
- Added a behavioral regression that releases the last external store reference while an operation is suspended and verifies cancellation runs. The README now demonstrates the direct operation API and no longer captures the store from a task body.

The deferred-ownership guard and `git diff --check` pass locally. Per the issue instructions, Swift compilation and tests are delegated to the current CI run: https://github.com/manaflow-ai/cmux/actions/runs/31449551482

austinywang:
Follow-up on current review/CI findings:

- The Swift 6.3 compile failure was valid: an ordinary global-actor `deinit` is nonisolated. `d201ef3412` changes `MainActorTaskStore` to `isolated deinit`, which is the repo’s shipping-toolchain pattern and preserves cancellation of suspended operations without an unsafe isolation escape. The replacement CI run is https://github.com/manaflow-ai/cmux/actions/runs/31450286433.
- I do not accept the “new 16 ms polling” pre-merge finding as introduced behavior. The sidebar cursor stabilizer has used that exact 16 ms timer since https://github.com/manaflow-ai/cmux/pull/255; later code added the physical-button check because a cancelled SwiftUI drag can omit `onEnded`. This PR changes only ownership: a replaceable timer stored inline in `ContentView` becomes one persistent reference-owned handle, so restarting it cannot link predecessor view snapshots. It remains active only while the pointer/drag state requires stabilization and is cancelled at the existing lifecycle boundaries. Replacing the established cross-portal cursor behavior with an unproven event-only path in this crash-forensics PR would risk restoring the cursor/drag failures that necessitated the timer.

austinywang:
Correction to the compiler-fix note above: before treating `isolated deinit` as final, I checked the supported-runtime history and found that macOS code in this repo deliberately avoids that runtime dependency. Final HEAD `c5541b0b67` instead snapshots the private weak task owners during ordinary deinit and enqueues one main-actor cancellation task. That cleanup captures neither the store nor a SwiftUI view snapshot, so it preserves cancellation and the no-predecessor-edge invariant while remaining compatible with the supported macOS range. Replacement CI: https://github.com/manaflow-ai/cmux/actions/runs/31450652555

austinywang:
Addressed the two remaining Cursor findings on current HEAD:

- `ab3a92cc3b` migrates all 11 declarations rejected by the broadened macOS SwiftUI ownership guard. That failure was the intentional second red proof (https://github.com/manaflow-ai/cmux/actions/runs/31451553647); the same guard passes after the migrations.
- `84ad00f525` removes deferred teardown from `MainActorTaskStore`. Task handles now live behind lock-protected private owners, a nonisolated registry indexes those owners only weakly, and `deinit` synchronously cancels all live owners. The release regression now asserts the cancellation handler has run before store release returns.

The final replacement CI run is https://github.com/manaflow-ai/cmux/actions/runs/31452588170.

austinywang:
Addressed the detached install/completion race in `814ebd6e0c`. The lock state now records `isFinished` before clearing the handle, so a late `install` cannot reattach a completed task. It also records cancellation and immediately cancels a late-installed handle, making both orderings explicit. Final CI replacement: https://github.com/manaflow-ai/cmux/actions/runs/31453053055.

austinywang:
Addressed both current Cursor findings in `45287586b1`:

- the ownership scanner now recognizes bracket, `Array<...>()`, and `Dictionary<..., ...>()` initializers containing `DispatchWorkItem`, `Task`, `DispatchSourceTimer`, or `Timer`
- macOS-capable SwiftUI `@State` auditing now includes `Packages/Shared/` as well as `Sources/`, `Packages/macOS/`, and pinned Bonsplit

The failing-test commit is independently proven red at https://github.com/manaflow-ai/cmux/actions/runs/31454601203, where all three new expectations failed in the ownership-lint step. On the fix commit, `python3 scripts/lint-stored-dispatch-work-items.py`, Python compilation, and `git diff --check` pass. Full replacement CI: https://github.com/manaflow-ai/cmux/actions/runs/31455160243.

PR #9980, opened 2026-08-11
fix(tui): restore native Windows GNU support

### LINKED ISSUE #8904: Native Windows (x86_64-pc-windows-gnu): three small platform-guarded fixes (build.rs / graphics.rs / file_url)
Building and running `cmux-tui` natively on Windows (`x86_64-pc-windows-gnu`, **not** WSL) surfaced three independent problems. Each has a small, platform-guarded fix that does not affect macOS/Linux (all changes are behind `#[cfg(windows)]` / `#[cfg(unix)]`). Details and proposed patches below — happy to split these into separate PRs.

**Environment**

- Target/host: `x86_64-pc-windows-gnu` (rustup GNU host, target == host)
- Toolchain: zig 0.15.2, LLVM 20.1.8 (libclang), MSYS2 mingw64 (binutils 2.46.1, gcc 16.1.0)
- Terminal: Windows Terminal
- No MSVC Build Tools / no Windows SDK installed

---

## 1. `ghostty-vt-sys/build.rs`: `WindowsSdkNotFound` when target == host on a gnu-hosted Windows toolchain

**Symptom**

`cargo build -p cmux-tui --target x86_64-pc-windows-gnu` fails while building the vendored `ghostty-vt` static lib. The zig build aborts with `WindowsSdkNotFound` even though a full MinGW/GNU toolchain is present.

**Root cause**

`crates/ghostty-vt-sys/build.rs` only passes `-Dtarget=<zig-target>` to the zig build when `target != host`:

```rust
if target != host
    && let Some(zig_target) = zig_target_for_rust_target(&target)
{
    command.arg(format!("-Dtarget={zig_target}"));
}
```

On a gnu-hosted Windows toolchain the Rust `target` (`x86_64-pc-windows-gnu`) equals the `host`, so `-Dtarget` is skipped and zig falls back to its **native default ABI, which is msvc**. That path requires a Windows SDK, which isn't installed in a pure MinGW setup, hence `WindowsSdkNotFound`.

**Fix (proposed)** — always pass `-Dtarget` when a mapping exists, regardless of target vs host:

```rust
let _ = &host; // otherwise unused now; keep reading it for the log
if let Some(zig_target) = zig_target_for_rust_target(&target) {
    command.arg(format!("-Dtarget={zig_target}"));
}
```

This keeps behavior identical on cross builds (where `target != host` already holds) and fixes the same-target gnu case by pinning the gnu ABI explicitly.

**Note:** a second, environment-side workaround is also required in this setup — `BINDGEN_EXTRA_CLANG_ARGS` must pass `--target` / `--sysroot` / `-resource-dir` so bindgen's libclang finds MinGW's builtin headers (`limits.h` etc.). That is a local env setting, not a source change, so it is not part of this patch — but it may be worth documenting for gnu-host contributors.

---

## 2. `ui/graphics.rs`: terminal capability queries (DA1 / `CSI 14 t`) leak into the pane prompt on Windows

**Symptom**

On startup, stray bytes such as `[?61;...c` and `[4;...t` appear in the active pane's shell prompt on Windows.

**Root cause**

`probe_kitty_graphics()` and `query_cell_pixels()` write terminal queries to stdout, then call `read_stdin_for()` to read the reply. `read_stdin_for` only has a real (poll(2)-based) implementation under `#[cfg(unix)]`; on Windows the `#[cfg(not(unix))]` stub just returns an empty `Vec`, so the reply is never read. The terminal's `CSI ? ... c` / `CSI 4 ; ... t` responses therefore go unconsumed and end up echoed into the pane.

Since kitty-graphics probing already resolves to "unsupported" on Windows (the reply can't be read anyway), emitting the query has no upside and only produces the leak.

**Fix (proposed)** — guard the query writes with `#[cfg(unix)]`. The subsequent empty read then yields "unsupported" / default cell size, which is the intended Windows behavior (no in-TUI kitty graphics):

```rust
pub fn probe_kitty_graphics() -> bool {
    #[cfg(unix)]
    {
        let mut stdout = std::io::stdout();
        let _ = write!(stdout, "\x1b_Gi=31,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\\x1b[c");
        let _ = stdout.flush();
    }
    let bytes = read_stdin_for(Duration::from_millis(180));
    // ...
}
```

The same guard is applied to the `\x1b[14t` write in `query_cell_pixels()`.

---

## 3. `sidebar_files/mod.rs` `file_url()`: Windows paths produce a broken `file://` URL → blank browser preview + wedged CDP page target

**Symptom**

Opening a local `.html`/`.md` from the file sidebar with `o` shows a blank page in the browser, and the CDP page target wedges (`Page.enable` hangs). Opening a fresh `about:blank` browser tab works fine — only file URLs are affected.

**Root cause**

`file_url()` percent-encodes everything that isn't an unreserved char. For a Windows path like `C:\a\b` this turns the drive letter and backslashes into percent-escapes **inside the URL authority**:

```
C:\a\b  →  file://C%3A%5Ca%5Cb
```

Chrome parses `C%3A%5C...` as a bogus host, so the page never loads and the page target wedges.

**Fix (proposed)** — on Windows, emit an empty authority (`file:///`), keep the drive colon, and map backslashes to forward slashes, so `C:\a\b` becomes `file:///C:/a/b`:

```rust
#[cfg(windows)]
if text.as_bytes().first().is_some_and(|b| b.is_ascii_alphabetic())
    && text.as_bytes().get(1) == Some(&b':')
{
    url.push('/'); // -> file:///
}
for byte in text.bytes() {
    match byte {
        b'A'..=b'Z' | b'a'..=b'z' | b'0'..=b'9' | b'-' | b'.' | b'_' | b'~' | b'/' => {
            url.push(char::from(byte));
        }
        #[cfg(windows)] b'\\' => url.push('/'),
        #[cfg(windows)] b':'  => url.push(':'),
        _ => url.push_str(&format!("%{byte:02X}")),
    }
}
```

---

## Suggested filing / merge order

1. **#1 (build.rs)** is the highest-value fix — it unblocks the native gnu-hosted Windows build for everyone and is a strict superset of the current condition.
2. **#2** and **#3** are Windows-only behavior fixes, each self-contained and guarded so they cannot cause regressions on Unix.

I have these as three separate commits on a local branch and can open them as individual PRs if that's easier to review.

blackdurumi:
Closes #8904

## Summary

- pass the explicit `x86_64-windows-gnu` Zig target for same-host MinGW builds
- suppress terminal capability queries on non-Unix platforms where replies cannot be consumed
- encode Windows drive paths as valid `file:///C:/...` URLs
- add platform-specific regression coverage

The regression test commit is intentionally separate from the fix commit so CI can demonstrate the pre-fix failure.

## Validation

- `git diff --check`
- local Cargo/Rust/Zig execution was not run because `cmux-tui/AGENTS.md` requires hosted verification
- hosted Rust verification and PR checks are pending

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/9980"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789025157&installation_model_id=21920&pr_number=9980&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F9980&signature=241161367ec81fea84095a0aaecdd92111a985543ebb57154420fa2003fa8820"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---

> [!NOTE]
> **Low Risk**
> Platform-scoped fixes for Windows GNU builds, non-Unix startup probing, and file URL encoding; no changes to auth, security, or core session logic on Unix.
> 
> **Overview**
> Restores **native MinGW (`x86_64-pc-windows-gnu`)** builds and fixes Windows-specific sidebar/browser behavior.
> 
> **Ghostty VT Zig build:** Zig target selection moves to shared `build_support` so same-host MinGW builds always pass `-Dtarget=x86_64-windows-gnu` (Zig otherwise picks MSVC and needs a Windows SDK). Other native same-host builds still omit `-Dtarget` so Zig keeps its default.
> 
> **Startup terminal probe:** Kitty/window-size escape queries are emitted only on Unix, where stdin replies can be read. On non-Unix, `write_terminal_probe_queries` is a no-op so unread sequences are not written to stdout.
> 
> **File URLs:** `file_url` on Windows drive paths produces `file:///C:/...` (extra slash after `file://`, backslashes as `/`, drive colon preserved).
> 
> Regression tests cover Windows file URLs, non-Unix probe no-op, and Zig target args for native GNU vs non-Windows hosts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02ac6e1fe94af3760714f301c81e6c7e2184e1e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Restores native Windows GNU support in the TUI by forcing the correct Zig target on MinGW hosts, skipping unreadable terminal probes on non-Unix, and producing valid `file:///C:/...` URLs for Windows paths. Extracts Zig target selection to `ghostty-vt-sys::build_support` with tests covering cross-target mappings and these behaviors.

- **Bug Fixes**
  - Use `build_support::zig_target_arg` in `ghostty-vt-sys` to pass the Zig target for same-host MinGW and cross builds; keep Zig’s native target for other native builds.
  - Suppress terminal capability queries on non-Unix; `write_terminal_probe_queries` is a no-op off Unix to avoid unread replies.
  - Encode Windows drive paths and backslashes into valid file URLs (e.g., `file:///C:/a/b`).

<sup>Written for commit fa295c5c5ebaf2f5c733fed06d8523763c444aa2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/9980?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows file links so drive-letter paths open correctly with consistent URL formatting.
  * Improved terminal detection on non-Unix platforms by avoiding unsupported terminal queries.

* **Compatibility**
  * Improved cross-platform build support across Windows, macOS, and Linux, including GNU and musl environments.
  * Preserved native platform behavior while ensuring Windows GNU builds use the correct target configuration.

* **Tests**
  * Added coverage for Windows file links, platform-specific terminal detection, and supported build targets.

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/sidebar_files/mod.rs
+++ b/cmux-tui/crates/cmux-tui/src/sidebar_files/mod.rs
@@ -473,4 +473,10 @@ mod tests {
     fn creates_percent_encoded_file_url() {
         assert_eq!(file_url(Path::new("/tmp/a file#1.md")), "file:///tmp/a%20file%231.md");
     }
+
+    #[cfg(windows)]
+    #[test]
+    fn creates_windows_drive_file_url() {
+        assert_eq!(file_url(Path::new(r"C:\a\b")), "file:///C:/a/b");
+    }
 }
--- a/cmux-tui/crates/cmux-tui/src/sidebar_files/mod.rs
+++ b/cmux-tui/crates/cmux-tui/src/sidebar_files/mod.rs
@@ -337,11 +337,22 @@ pub fn shell_single_quote(value: &str) -> String {
 pub fn file_url(path: &Path) -> String {
     let text = path.to_string_lossy();
     let mut url = String::from("file://");
+    #[cfg(windows)]
+    let windows_drive_path = text.as_bytes().first().is_some_and(|b| b.is_ascii_alphabetic())
+        && text.as_bytes().get(1) == Some(&b':');
+    #[cfg(not(windows))]
+    let windows_drive_path = false;
+    if windows_drive_path {
+        url.push('/');
+    }
     for byte in text.bytes() {
         match byte {
             b'A'..=b'Z' | b'a'..=b'z' | b'0'..=b'9' | b'-' | b'.' | b'_' | b'~' | b'/' => {
                 url.push(char::from(byte));
             }
+            #[cfg(windows)]
+            b'\\' => url.push('/'),
+            b':' if windows_drive_path => url.push(':'),
             _ => url.push_str(&format!("%{byte:02X}")),
         }
     }
--- a/cmux-tui/crates/cmux-tui/src/ui/graphics.rs
+++ b/cmux-tui/crates/cmux-tui/src/ui/graphics.rs
@@ -956,6 +956,7 @@ struct ParsedTerminalProbe {
     pending_input: Vec<u8>,
 }
 
+#[cfg(unix)]
 fn write_terminal_probe_queries(stdout: &mut impl Write, query_window_pixels: bool) {
     if query_window_pixels {
         let _ = write!(stdout, "\x1b[14t");
@@ -963,6 +964,9 @@ fn write_terminal_probe_queries(stdout: &mut impl Write, query_window_pixels: bo
     let _ = write!(stdout, "\x1b_Gi=31,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\\x1b[c");
     let _ = stdout.flush();
 }
+
+#[cfg(not(unix))]
+fn write_terminal_probe_queries(_stdout: &mut impl Write, _query_window_pixels: bool) {}
 
 /// Probe terminal capabilities in one exchange and return any user input read
 /// alongside the replies. The final DA1 request acts as an ordering marker:
--- a/cmux-tui/crates/cmux-tui/src/ui/graphics.rs
+++ b/cmux-tui/crates/cmux-tui/src/ui/graphics.rs
@@ -956,6 +956,14 @@ struct ParsedTerminalProbe {
     pending_input: Vec<u8>,
 }
 
+fn write_terminal_probe_queries(stdout: &mut impl Write, query_window_pixels: bool) {
+    if query_window_pixels {
+        let _ = write!(stdout, "\x1b[14t");
+    }
+    let _ = write!(stdout, "\x1b_Gi=31,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\\x1b[c");
+    let _ = stdout.flush();
+}
+
 /// Probe terminal capabilities in one exchange and return any user input read
 /// alongside the replies. The final DA1 request acts as an ordering marker:
 /// any preceding Kitty reply advertises support, while its absence does not
@@ -967,11 +975,7 @@ pub fn probe_terminal(known_cell_pixels: Option<(u16, u16)>) -> StartupTerminalP
         ioctl_pixels.is_none() && terminal_size.is_some_and(|(cols, rows)| cols > 0 && rows > 0);
 
     let mut stdout = std::io::stdout();
-    if query_window_pixels {
-        let _ = write!(stdout, "\x1b[14t");
-    }
-    let _ = write!(stdout, "\x1b_Gi=31,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\\x1b[c");
-    let _ = stdout.flush();
+    write_terminal_probe_queries(&mut stdout, query_window_pixels);
 
     let bytes = read_stdin_until(TERMINAL_PROBE_TIMEOUT, terminal_probe_complete);
     let parsed = parse_terminal_probe(&bytes);
@@ -1275,6 +1279,16 @@ mod tests {
     #[test]
     fn missing_initial_metrics_use_synthetic_fallback() {
         assert_eq!(resolve_cell_pixels(None, None), FALLBACK_CELL_PIXELS);
+    }
+
+    #[cfg(not(unix))]
+    #[test]
+    fn terminal_probe_does_not_write_queries_without_a_reply_reader() {
+        let mut output = Vec::new();
+
+        write_terminal_probe_queries(&mut output, true);
+
+        assert!(output.is_empty(), "unread terminal queries were written: {output:?}");
     }
 
     #[test]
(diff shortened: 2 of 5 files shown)


blackdurumi:
Maintainers, could an upstream admin approve the two `cmux-tui` workflow runs so hosted validation can start?

- [cmux-tui spec inventory](https://github.com/manaflow-ai/cmux/actions/runs/31502999881)
- [cmux-tui SDKs](https://github.com/manaflow-ai/cmux/actions/runs/31502999742)

Both runs are currently `action_required` with no jobs because this is a fork PR. The PR is limited to five Rust files, and the CodeRabbit and Socket checks pass. Local Cargo/Rust/Zig execution is prohibited by `cmux-tui/AGENTS.md`, so hosted verification is the required validation path. Please approve these runs; we will review the hosted results before merging. If the full hosted gate must be dispatched from an upstream ref, please also run it against commit `fa295c5c5ebaf2f5c733fed06d8523763c444aa2`.

Issue #7989, opened 2026-07-13
Add remote-aware resume bindings for agents in SSH workspaces


austinywang:
## Context

Split from #7978. The sleep/reconnect fix preserves and reattaches a live persistent remote PTY, but `surface.resume.get` can still report `resume_binding: null` for agent sessions launched inside SSH workspace terminals. This issue tracks remote-agent restore after the remote PTY itself is genuinely gone.

## Findings

- Local process discovery (`ProcessDetectedResumeIndexes`) scans only processes on the Mac, so it cannot discover agents running on the SSH host.
- Agent SessionStart hooks can publish through `surface.resume.set`: remote shells carry `CMUX_WORKSPACE_ID` / `CMUX_SURFACE_ID`, and the reverse relay rewrites restored workspace/surface aliases.
- `SurfaceResumeBindingSnapshot` currently stores a command, cwd, environment, and checkpoint metadata without execution-location/remote-workspace context.
- Existing restore paths can feed a resume command into a non-persistent remote shell, but persistent-PTY restore prioritizes the SSH PTY attach command. If the remote PTY is gone, treating the binding as an ordinary local startup command would execute a remote path/cwd on the Mac or skip the binding behind the attach path.

## Required design

Add an explicit remote-aware binding/launch flavor rather than inferring locality from paths or terminal titles. The binding must retain the owning remote-workspace context and restore by entering through the SSH persistent-session machinery, then execute the resume command on the remote host. Preserve relay ID aliasing and remote cwd semantics, and keep local approval/sanitization guarantees intact.

## Acceptance criteria

- An agent SessionStart hook inside an SSH workspace produces a non-null `surface.resume.get.resume_binding` tied to the correct workspace/surface.
- If the persistent remote PTY is still alive, reconnect reattaches it and does not run a duplicate agent.
- If the persistent remote PTY is genuinely gone, restore creates/enters the remote session and executes the resume command on the remote host, never locally.
- Tests cover relayed registration, restored ID aliasing, remote cwd/environment handling, and live-PTY-vs-gone-PTY behavior.

Related: #7978 and PR #7987.

austinywang:
Scope retest from remote-tmux dogfood on current main `c9f2d8c4`.

Remote-tmux persistence does not satisfy this issue's remote-aware resume-binding design:

[...]

PR #10049, opened 2026-08-12
resume: hookless directory-scoped continue bindings for remote agents (#7989)

alloevil:
Implements the Tier-1 flavor proposed in [#7989's discussion](https://github.com/manaflow-ai/cmux/issues/7989#issuecomment-5263284712): synthesize a directory-scoped continue binding for remote agents that never published a hook binding, so a persistent-SSH restore can resume the agent after the remote PTY is genuinely gone — without requiring `cmux hooks setup` on every remote host.

## Problem

Local process discovery can't see agents on the SSH host, and hook-published bindings require per-host, per-agent installation — a real adoption cliff for multi-host users. Result: restore reattaches the PTY, but a genuinely-gone PTY leaves the agent unresumed (`resume_binding: null`).

## Approach

When the snapshot still records the agent kind + remote working directory (and `wasAgentRunning`), `RemoteAgentContinueSynthesizer` builds a binding from inline templates:

| kind | command |
|---|---|
| claude | `cd -- '<dir>' … && claude --continue \|\| claude` |
| codex | `cd -- '<dir>' … && codex resume --last \|\| codex` |
| all others | nothing (no trustworthy sessionless continue invocation) |

Design points, mapped to the issue's acceptance criteria:

- **Live PTY → attach-only, never a duplicate agent** (criterion 2): rides the existing `requireExisting` pipeline; the synthesized command is additionally hard-gated in `reattachPersistentRemotePTYPanels` to inject only once the PTY is confirmed ended. A directory-scoped continue has no session checkpoint the remote once-guard could reconcile against, so this gate is stricter than for hook bindings.
- **Gone PTY → command executes on the remote host, never locally** (criterion 3): binding carries `.persistentSSH(SurfaceResumeRemoteContext)` — the existing flavor, no new Codable case — and enters through the SSH persistent-session machinery via `remotePTYAttachStartupCommand`. No wrapper-resolver tokens (they resolve local Mac paths), mirroring `remoteStartupInput()`'s `repairPortableAgentExecutable: false` convention.
- **Trust model**: new source `remote-synthesized` shares the process-detected tier (bypasses the signed approval store) because the command is built exclusively from cmux's own inline templates with no caller-supplied arguments — an observation-grade artifact, not a proposal from an arbitrary process. The gate for every other source is unchanged (covered by a regression test).
- **Precedence**: agent-hook / cli / process-detected bindings always win; synthesis only fills the gap, and respects the auto-resume setting.

## Known limitations (documented in code + tests)

- Directory-scoped continue resumes the *most recent* session in that directory: a directory shared by same-kind agents can continue the wrong conversation. Hook bindings (Tier 2, this issue's original design) remain the precise path and always take precedence — the two tiers compose rather than compete.
- Persistent-SSH workspaces only; plain SSH panes have no reattach seam to hang the liveness gate on.
- Dock restore path has no remote-PTY seam today; left as follow-up rather than duplicating workspace machinery.

## Tests

`RemoteAgentContinueSynthesizerTests` (15 cases, Swift Testing): exact command form per kind, nil for the 16 uncovered kinds + custom (parameterized), cwd guards, single-quote splice injection safety, non-ASCII printf-octal quoting, `isRemoteSynthesized` predicate + mutual exclusion across all four sources (parameterized), trust-tier resolution with a `.pending` signing secret, cli-source demotion regression guard, reconcile pass-through, `remoteStartupInput()` verbatim replay, Codable round-trip with quote-bearing cwd.

## Caveats

- Developed on Linux against `main`; every call was verified against the callee's source, but **I could not compile or run the app locally**. Happy to iterate on anything the macOS build or the full restore matrix surfaces.
- `xcodeproj` wiring for the 2 new files is included (lint-pbxproj-test-wiring passes).

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10049"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789144407&installation_model_id=21920&pr_number=10049&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10049&signature=42b7711328a3c19cf98f2632f56a421f8a5f981167bccaddbb2890b052b60be5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Resumes hookless remote agents in persistent-SSH workspaces after a lost session by synthesizing a directory-scoped continue command from daemon-sampled foreground “tombstones.” Previously, restore reattached the PTY but left the agent unresumed once the PTY had ended.

- Daemon: `cmuxd-remote` samples the PTY foreground process (Linux: TIOCGPGRP → `/proc/<pgid>/{comm,cwd}`; macOS dev: `ps` for command only), annotates live `pty.list` entries with `foreground_command`/`foreground_cwd`, and returns a bounded `ended_sessions` FIFO (64) for finished persistent sessions.
- App/CLI path: new RPC `workspace.remote.pty_session_lost_resume` (non–main-actor lane) waits for the daemon link, reads tombstones, applies policy, and returns a remote command or null. The CLI `ssh-pty-attach` wrapper calls it when it confirms session loss and injects the payload by replacing `--command-b64`; otherwise it degrades to the plain replacement shell.
- Synthesis: supports only `claude` and `codex`
  - `claude`: “claude --continue || claude”
  - `codex`: “codex resume --last || codex”
  The binding uses `.persistentSSH(SurfaceResumeRemoteContext)` and `.direct` mode, runs on the remote host, and is stored with source `remote-synthesized`.
- Precedence and liveness: an existing agent-hook/CLI/process-detected binding always wins. Inject a synthesized command only after the PTY is confirmed ended; a live PTY is attach-only. If an ended session has no injectable binding, the replacement attach keeps `--require-existing` so the wrapper owns tombstone-backed synthesis.
- Policy: honors the auto-resume setting and does not replay previously persisted synthesized bindings when auto-resume is off.
- Trust: `remote-synthesized` shares the process-detected tier and bypasses the signed approval store; other sources are unchanged.
- Limits and compatibility: directory-scoped continue may pick the wrong conversation when multiple same-kind agents share a cwd. Works in persistent-SSH workspaces only (plain SSH panes and Dock restore unchanged). Older daemons omit `ended_sessions`; the path degrades to a plain shell.

<sup>Written for commit 97647cde9fcfe01743d1fc6ffdaf0c94adfeacad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10049?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Remote terminal sessions can automatically resume supported Claude and Codex sessions after SSH or PTY loss.
  * Resume commands remain scoped to the correct working directory and persistent SSH session.
  * Recent ended-session details support reliable restoration.
  * Live sessions attach without injecting duplicate resume commands.

* **Bug Fixes**
  * Automatic resume settings are consistently honored.
  * Improved handling for ended sessions, unavailable directories, and unsupported agents.
  * Remote-synthesized resume bindings are trusted and restored consistently.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -4794,8 +4794,27 @@ struct CMUXCLI {
                     defaultValue: "[cmux] remote session was lost; starting a new shell."
                 )
                 cliWriteStderr(Data((notice + "\n").utf8))
+                var respawnArgs = stableAttachArgs.filter { $0 != "--require-existing" }
+                // Hookless remote resume (#7989): the app can synthesize a
+                // directory-scoped continue command from the daemon's
+                // foreground tombstone for the session that just died. Any
+                // failure degrades to the plain replacement shell.
+                if let resumeCommand = sessionLostResumeRemoteCommand(
+                    client: client,
+                    attachArgs: stableAttachArgs
+                ) {
+                    let resumeNotice = String(
+                        localized: "cli.sshPtyAttach.remoteSessionLostResume",
+                        defaultValue: "[cmux] resuming the agent that was running in the lost session."
+                    )
+                    cliWriteStderr(Data((resumeNotice + "\n").utf8))
+                    respawnArgs = replacingSSHPTYAttachCommandB64(
+                        in: respawnArgs,
+                        with: Data(resumeCommand.utf8).base64EncodedString()
+                    )
+                }
                 try runSSHPTYAttach(
-                    commandArgs: stableAttachArgs.filter { $0 != "--require-existing" },
+                    commandArgs: respawnArgs,
                     client: client,
                     explicitPassword: socketPasswordArg
                 )
@@ -12799,6 +12818,67 @@ struct CMUXCLI {
             throw CLIError(message: "\(commandName): workspace not found")
         }
         return ["workspace_id": workspaceId]
+    }
+
+    /// Asks the app for a hookless remote continue command (#7989) after the
+    /// wrapper confirmed the persistent session was lost. The daemon's
+    /// foreground tombstone supplies the facts; binding precedence,
+    /// auto-resume, and approval policy stay in the app. Every failure —
+    /// older app, no tombstone, policy says attach plain — returns nil so the
+    /// respawn degrades to the existing replacement-shell behavior.
+    private func sessionLostResumeRemoteCommand(
+        client: SocketClient,
+        attachArgs: [String]
+    ) -> String? {
+        let (workspaceOpt, _) = parseOption(attachArgs, name: "--workspace")
+        let (sessionIDOpt, _) = parseOption(attachArgs, name: "--session-id")
+        let (attachmentIDOpt, _) = parseOption(attachArgs, name: "--attachment-id")
+        let environmentSurfaceID = Self.normalizedEnvValue(
+            ProcessInfo.processInfo.environment["CMUX_SURFACE_ID"]
+        )
+        let surfaceID = environmentSurfaceID
+            ?? Self.normalizedEnvValue(attachmentIDOpt).flatMap { UUID(uuidString: $0) == nil ? nil : $0 }
+        guard let workspaceID = Self.normalizedEnvValue(workspaceOpt)
+            ?? Self.normalizedEnvValue(ProcessInfo.processInfo.environment["CMUX_WORKSPACE_ID"]),
+            let sessionID = Self.normalizedEnvValue(sessionIDOpt),
+            let surfaceID else {
+            return nil
+        }
+        // The app-side handler deliberately outwaits the coordinator's daemon
+        // bootstrap (bounded at 20s), so this call must outlast it.
+        guard let payload = try? client.sendV2(
+            method: "workspace.remote.pty_session_lost_resume",
+            params: [
+                "workspace_id": workspaceID,
+                "surface_id": surfaceID,
+                "session_id": sessionID,
+            ],
+            responseTimeout: 30
+        ) else {
+            return nil
+        }
+        guard let command = payload["command"] as? String,
+              !command.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
+            return nil
+        }
+        return command
+    }
+
+    /// Returns `args` with any `--command-b64 <value>` pair replaced by the
+    /// synthesized resume payload.
+    private func replacingSSHPTYAttachCommandB64(in args: [String], with base64: String) -> [String] {
+        var result: [String] = []
+        var index = 0
+        while index < args.count {
+            if args[index] == "--command-b64", index + 1 < args.count {
+                index += 2
+                continue
+            }
+            result.append(args[index])
+            index += 1
+        }
+        result.append(contentsOf: ["--command-b64", base64])
+        return result
     }
 
     private func runSSHPTYAttach(commandArgs: [String], client: SocketClient, explicitPassword: String?) throws {
--- a/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Wire/ControlCommandExecutionPolicy.swift
+++ b/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Wire/ControlCommandExecutionPolicy.swift
@@ -125,6 +125,14 @@ public enum ControlCommandExecutionPolicy: Sendable, Equatable {
         // lane like the other workspace reads below.
         "workspace.env",
         "workspace.remote.pty_sessions",
+        // `workspace.remote.pty_session_lost_resume` deliberately outwaits the
+        // workspace coordinator's daemon bootstrap (bounded, signal-driven) and
+        // then queries the daemon's ended-session tombstones over the tunnel
+        // (#7989), so it must never hold the main actor; policy and binding
+        // registration take one `v2MainSync` hop. Without this entry the
+        // policy routes it to the main-actor processV2Command switch, which
+        // lacks the case, and the control socket returns method_not_found.
+        "workspace.remote.pty_session_lost_resume",
         "workspace.remote.pty_close",
         "workspace.remote.pty_detach",
         "workspace.remote.pty_bridge",
--- a/Packages/macOS/CmuxControlSocket/Tests/CmuxControlSocketTests/ControlCommandExecutionPolicyTests.swift
+++ b/Packages/macOS/CmuxControlSocket/Tests/CmuxControlSocketTests/ControlCommandExecutionPolicyTests.swift
@@ -32,6 +32,9 @@ struct ControlCommandExecutionPolicyTests {
             "feed.push", "browser.download.wait", "system.top", "system.memory",
             "workspace.remote.pty_bridge", "workspace.env", "sidebar.custom.reload",
             "sidebar.custom.open",
+            // Tombstone-backed hookless resume outwaits daemon bootstrap and
+            // queries the tunnel; it must never hold the main actor (#7989).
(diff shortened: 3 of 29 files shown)


alloevil:
End-to-end verification on a real rig (macOS 26.5 client, Ubuntu 24.04 remote, custom build of this branch + #10048) showed the Tier-1 gate never fires in practice, so this branch now carries a redesign of where the facts come from — verified through to a live FAIL→PASS replay on the same rig. Full narrative below; the observable failure it fixes:

## What the rig showed

Checklist scenario (claude conversing in a persistent-SSH workspace → quit cmux → kill the remote agent **and** its PTY holder → relaunch):

| expected | observed |
|---|---|
| synthesized `cd -- '<dir>' … && claude --continue \|\| claude` injected | `[cmux] remote session was lost; starting a new shell.` → bare shell in `$HOME` |
| `surface.resume.get` → `remote-synthesized` binding | `resume_binding: null` (the exact #7989 symptom) |

**Root cause, pinned with artifacts:** in the very same session snapshot, a *local* claude panel carries `terminal.agent {kind, launchCommand}` + `wasAgentRunning: true` (written by local shim/launch tracking), while the *remote* persistent-SSH panel captured moments before quit — with claude demonstrably running (67k-token session) — has only `workingDirectory`. Local process discovery cannot see the SSH host (this PR's own problem statement), so `guard let restorableAgent` in the synthesis gate is unsatisfiable for exactly the hookless remote agents the PR targets. Deterministic; reproduced across three quit/kill/relaunch cycles.

The command template itself is correct: running the synthesized command manually in the dead session's replacement shell resumed the conversation and answered the planted codeword.

## The fix: facts from the component that watches the agent die

The daemon owns the PTY and outlives every cmux quit — it is the only component that actually witnesses the agent's death, which makes it the honest source for both `agent kind` and `wasAgentRunning`:

- **cmuxd-remote** samples the PTY foreground process (existing `TIOCGPGRP` reader + `/proc/<pgid>/{comm,cwd}`, platform-split per repo convention) inside the output pump, throttled to one probe per 5 s of active output — no timers, no polling. `finishSession` retains a bounded FIFO (64) of ended-session tombstones. `pty.list` now reports `foreground_command`/`foreground_cwd` on live sessions plus a top-level `ended_sessions`; older daemons omit the keys and every decoder treats absence as "no tombstones".
- **A new v2 method** `workspace.remote.pty_session_lost_resume` converts a tombstone into policy on the app side: an existing binding (agent-hook / cli / process-detected) always wins, the executable must map to a known agent kind via the same `TerminalForegroundCommandCapture` table process detection uses (shells never resume), and the synthesized binding flows through the unchanged `approvedPersistentSSHResumeCommand` gates (approval store, auto-resume setting, `allowsAutomaticResume`).
- **The ssh-pty-attach wrapper** asks the app at the exact choke point where it concludes "remote session was lost" — the path a real restore takes (rig traces showed the wrapper's internal respawn, not `reattachPersistentRemotePTYPanels`, handles this case) — and splices the returned command into the replacement session's `--command-b64`. Any failure (older app, no tombstone, policy says plain shell) degrades byte-for-byte to today's behavior.

The snapshot-based synthesis introduced earlier in this branch stays intact for snapshots that do carry agent identity (hibernation, future serializers); the two tiers compose.

## Tests

- Go: foreground sampling (injected lookups, stale-session guard, last-known retention), tombstone exactly-once across both `finishSession` branches, FIFO eviction, restart-clears-tombstone, `pty.list` shape incl. nil-hub and RPC-level assertions.
- Swift: `RemoteResumeBindingTests` gains six cases for `remoteContinueResumeCommandAfterSessionLoss` — claude + codex synthesis (asserting the decoded initial command and the registered `remote-synthesized` binding with a matching `persistentSSH` context), shell/unknown/empty foregrounds parameterized-never-synthesize, auto-resume off, non-persistent config, and existing-binding precedence.

## Related findings filed while verifying

- #10159 — persistent-SSH reattach reseed replays terminal query responses into the live remote shell (prompt pollution + `>|` clobber risk). The wrapper-side injection here lands on a *fresh* session (nothing to reseed), so it does not race that storm, but the fix for #10159 should keep it that way.
- #10160 — `workspace disconnect` → `reconnect` wedges the connection until app restart.

## Verification

- **Fork CI green**: `go vet` + daemon tests + linux-amd64 cross-compile, and the Swift suites (`RemoteAgentContinueSynthesizerTests` + the new hermetic `RemoteSessionLossResumeTests`) — alloevil/cmux actions run 31988555960.
- **Live rig FAIL→PASS replay** (same machine pair, same checklist scenario — agent conversing → quit cmux → kill the agent and its PTY holder → relaunch):
  - before this change: `remote session was lost; starting a new shell` → bare shell in `$HOME`, `resume_binding: null`;
  - with this change: the relaunch resumed Claude Code in `~/resume_lab` with the full prior conversation restored (planted-codeword probe answered correctly), remote host shows exactly one agent process, and `surface.resume.get` returns the synthesized binding: `source: remote-synthesized`, `command: cd -- '/home/gaoruilin/resume_lab' … && claude --continue || claude`.
- The rig loop also flushed out two integration bugs that are part of this push:
  1. `ControlCommandExecutionPolicy` lane registration — without the worker-lane entry the dispatcher never reaches the handler and the socket returns `method_not_found` (the policy file documents this exact failure mode for `mobile.terminal.set_font`); covered by `ControlCommandExecutionPolicyTests`.
  2. The wrapper's session-lost verdict travels the shared tunnel's lifecycle channel and can settle before the workspace coordinator's daemon link is ready, so the handler resolves its target with the waiting resolver and retries the tombstone query bounded and signal-driven on the shared availability condition instead of degrading to a plain shell on a transient `remote daemon is not ready`.
- A negative path fell out of the rig loop for free: when the agent died minutes *before* the session (leaving a shell as the last sampled foreground), the tombstone honestly reports the shell and the respawn stays a plain shell — synthesis only fires when an agent was genuinely the session's last foreground.

alloevil:
Follow-up push: independent re-verification on the rig hit the *second* session-gone path — when the wrapper's first attach dies on the transient daemon-not-ready race, the app-side `reattachPersistentRemotePTYPanels` respawn used to create the replacement session directly (plain shell, no synthesis, no notice; which path wins is millisecond timing). The respawn now keeps `--require-existing` when the session is ended with no injectable binding, so the replacement wrapper confirms the loss itself and both paths converge on the same tombstone-backed synthesis exit. A binding the app can already inject still skips the round trip unchanged. Two new `RemoteSessionLossResumeTests` cases pin both shapes; CI green at alloevil/cmux run 32010588910.

PR #10187, opened 2026-08-15
Recover fresh hook-bound agents after crash restore

krandder:
## Summary

- preserve the existing `terminal.autoResumeAgentSessions` setting
- restore fresh local `agent-hook` bindings for Codex, Claude, and Kimi even when the last periodic snapshot recorded `autoResume:false` / `wasAgentRunning:false`
- require a complete managed identity, matching persisted agent kind, and a binding no older than two hours
- reject future timestamps, unsupported/custom kinds, stale bindings, disabled settings, and cross-kind replacements
- promote `autoResume` only in the restore copy; persisted bindings stay unchanged

## Regression coverage

1. `3a6d1347a6` adds the failing workspace + Dock recovery tests and rejection cases
2. `ee5d7486c0` implements the restore policy

## Validation

- Swift parser on changed sources/tests
- strict test determinism
- pbxproj test wiring
- workspace package grouping
- Package.resolved policy
- pbxproj normalization
- focused macOS test and tagged cloud build dispatched at the fixed SHA

## Summary by CodeRabbit

* **New Features**
  * Improved crash recovery for Codex, Claude, and Kimi sessions.
  * Recent, trusted sessions can resume automatically across workspaces and dock splits.
  * Sessions preserve their configured agent identity during restoration.

* **Bug Fixes**
  * Stale, unsupported, incomplete, or disabled resume configurations remain manual instead of resuming unexpectedly.
  * Previously running sessions restore using their correct running state.

### DIFF
--- a/Sources/App/WorkspaceRuntimeSettings.swift
+++ b/Sources/App/WorkspaceRuntimeSettings.swift
@@ -218,12 +218,66 @@ enum AgentSessionAutoResumeSettings {
     static let autoResumeAgentSessionsKey = "terminal.autoResumeAgentSessions"
     static let defaultAutoResumeAgentSessions = true
     static let didChangeNotification = Notification.Name("cmux.agentSessionAutoResumeSettingsDidChange")
+    static let crashRecoveryBindingMaxAge: TimeInterval = 2 * 60 * 60
+    private static let crashRecoveryKinds: Set<String> = ["claude", "codex", "kimi"]
 
     static func isEnabled(defaults: UserDefaults = .standard) -> Bool {
         guard defaults.object(forKey: autoResumeAgentSessionsKey) != nil else {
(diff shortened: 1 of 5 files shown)


krandder:
Closing after the revised verification requirement. An isolated build based on unmodified main (only unrelated CLI ref-resolution changes) passed the full fresh-hook hard-crash check: Claude prompt completed, the autosave held a fresh local agent-hook binding with autoResume=true, the app was killed with SIGKILL, and relaunch eagerly restored the Claude UI and prior exchange without manual Enter. The built-in terminal.autoResumeAgentSessions=true path therefore covers this case; no additional restore policy should ship without a failing real-world repro. The branch test lane also exposed unrelated/pre-existing suite failures, so keeping this speculative behavior change open would add risk without demonstrated need.

PR #10214, opened 2026-08-15
fix(tui): restore native Windows GNU launch behavior

### LINKED ISSUE #8904: Native Windows (x86_64-pc-windows-gnu): three small platform-guarded fixes (build.rs / graphics.rs / file_url)

lawrencecchen:
## Summary

- force Zig to use the GNU ABI on a native `x86_64-pc-windows-gnu` Rust host
- suppress terminal capability queries on platforms that cannot read their replies
- produce valid drive, UNC, and verbatim Windows file URLs, and reject device namespaces
- extend the hosted Windows check with launch-query, Unicode ConPTY input, and resize proof

## Credit

This is a current-main recut and extension of https://github.com/manaflow-ai/cmux/pull/9980 by Jung-hoon Lee (`blackdurumi`). The original diagnosis and fixes for the GNU Zig target, unread startup queries, and drive file URLs came from that PR.

This PR is intended to supersede https://github.com/manaflow-ai/cmux/pull/9980 after hosted verification. It preserves that work on current main and adds UNC, verbatim-path, device-path, localization, and real ConPTY coverage.

## Testing

- `git diff --check`
- Python syntax check for `cmux-tui/scripts/smoke-windows-conpty-resize.py`
- red hosted run: https://github.com/manaflow-ai/cmux/actions/runs/31863645558
  - Linux failed in the shared Zig download with exit 28, before the selected tests
  - macOS red test is still queued
- focused green hosted verification pending
- full green hosted verification, including Windows ConPTY, pending

Local Cargo, Rust, and Zig execution was not run because `cmux-tui/AGENTS.md` requires hosted verification.

## Issues

- Closes https://github.com/manaflow-ai/cmux/issues/8904
- Supersedes https://github.com/manaflow-ai/cmux/pull/9980 after hosted verification

No Windows npm or PyPI package is added or published.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10214"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789360079&installation_model_id=21920&pr_number=10214&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10214&signature=e023c53f16fe1dfa476870d6379db0648a65dc00d86d0b489a9287274a22d164"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Restores native Windows GNU launch behavior and filters documented ConPTY init sequences so Windows attaches start clean. Previously MinGW hosts could select the MSVC ABI and startup queries leaked into attach output; now Zig uses GNU on native MinGW, Windows-only probes are skipped, and the OpenConsole startup prefix is ignored before validating output.

- Build (`ghostty-vt-sys`): move Zig target mapping to `build_support::zig_target_arg`; force GNU ABI on native MinGW; tests cover native and cross targets.
- Terminal probes (`cmux-tui`): gate `write_terminal_probe_queries` on reply readability; Unix sequences unchanged; Windows emits no unread queries; tests added.
- ConPTY attach (`cmux-tui`): detect and ignore the OpenConsole startup prefix; smoke test verifies launch, Unicode input, and resize.
- File URLs (`cmux-tui`): `sidebar_files::file_url` now returns `Result`; normalize drive/UNC/verbatim paths; reject device and non-absolute paths; `FileCommand::OpenBrowser` surfaces `sidebar.unsupported_browser_path`.
- CI: add Windows launch tests; ConPTY job verifies launch, input, and resize; crossterm parser tests run with `NO_COLOR=""` to preserve ANSI fixtures.

- Migration: callers of `sidebar_files::file_url` must handle `Result<String, FileUrlError>` and surface an error for unsupported Windows paths.

<sup>Written for commit e37ae735ecd85e07c9bd21f0232ebae01d5c8060. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10214?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved opening local files in browsers on Windows, including drive, network, and extended paths.
  * Added localized messaging when a Windows path cannot be opened in a browser.
  * Prevented unsupported terminal capability queries during Windows startup.
  * Improved Windows terminal startup, input, and resize reliability.

* **Build Improvements**
  * Improved support for Windows and cross-platform builds across supported target environments.
  * Added targeted Windows launch and ConPTY verification during builds.


lawrencecchen:
Closing as superseded by draft replacement https://github.com/manaflow-ai/cmux/pull/10266 (author Lawrence Chen). Git comparison shows the five commits are already present in that replacement. Keep #10266 for a fresh current-main rebase and review; this old head conflicts.

PR #10238, opened 2026-08-17
Fail closed on unavailable Windows startup evidence

lawrencecchen:
Stacked on https://github.com/manaflow-ai/cmux/pull/10131.

The trusted preflight currently reports Windows job, process, privilege, and handle observations that it does not collect. The red test requires unavailable observations to stay null. The follow-up commit will relay only the optional child job observation and leave unsupported signals unavailable, so existing validation rejects an unproven Windows claim.

Regression proof uses two commits: test first, implementation second. Hosted CI only; no local Rust/Cargo/Zig/Xcode tests.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10238"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789536799&installation_model_id=21920&pr_number=10238&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10238&signature=547cdd295eba90cca73e92b7cbbd13e9885d87cc90b70ee15dbb72161e235a5c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Fails closed when Windows startup preflight evidence is unavailable and publishes an explicit skipped Windows claim. Previously, missing Windows signals implied success; now Windows-only unavailability exits 78 with a Windows-specific reason, sets `claim_status=unverified` and `claim_reason`, attests the account-process AppContainer probe and staging failures, and gates packaging, paired runs, and profile capture.

**Review notes**
[...]

PR #10263, opened 2026-08-17
Fix startup benchmark tool path construction

lawrencecchen:
Depends on https://github.com/manaflow-ai/cmux/pull/10238.

This branch replays the red test and green Path-based fix from PR10237 onto the current PR10238 head. The workflow expression `trusted_release /= "release" / "examples"` evaluates string division before pathlib division and raises TypeError. The helper now constructs release/example paths as Path operands and has a focused behavior test.

Checks: focused path test, cmux-tui SDK/spec checks, startup contract/claim checks. Rust/Cargo/Zig/Xcode builds were not run locally.


lawrencecchen:
Closing this stale benchmark-path stack. The referenced startup benchmark files are absent from current main. Recreate the benchmark against the current runner contract when the measurement target is defined.

PR #10280, opened 2026-08-17
Add `cmux themes here` for per-surface theming

sergiokas:
## Problem

`cmux themes set` is the only way to change terminal colors, and it is global by
construction: it writes a managed `theme` override into the cmux-owned Ghostty config
and posts a reload, so it repaints every window, every split, and every future session.

That leaves no way to make one window look different from the rest — a common wish when
you keep prod and dev shells side by side, or want an agent pane visually distinct from
the one you type in. Today the only recourse is hand-writing OSC escape sequences.

## Change

Adds `cmux themes here <theme|path>`, which recolors **only the surface it runs in**.

It resolves a theme the same way `themes set` does — by bundled name, through the
existing `themeDirectoryURLs()` search path — and writes OSC dynamic-color sequences to
stdout:

| Key in theme file | Sequence |
| --- | --- |
| `background` | OSC 11 |
| `foreground` | OSC 10 |
| `cursor-color` | OSC 12 |
| `palette = N=#hex` | OSC 4;N |

libghostty scopes those to the surface they are written into, which gives the property
this command is for: no config file is touched, no other surface changes, and the effect
ends with the pane. `cmux themes here --reset` writes the matching resets
(OSC 110/111/112/104) to fall back to whatever the config specifies.

```bash
cmux themes here "Catppuccin Mocha"   # this pane only
cmux themes here ~/my-themes/custom   # a theme file by path
cmux themes here --reset              # back to configured colors
```

The argument accepts a bundled theme name or a path to a theme file. `--json` is
rejected, since the command's entire output is raw escape bytes.

## Notes

- **Nothing existing changes.** `set`, `list`, `clear`, and the interactive picker are
  untouched; `here` is a new subcommand alongside them.
- **Colors do not survive session restore.** A restored pane comes back with the config
  theme. That is inherent to OSC-scoped color, and the help text says so.
- **No broadcast.** `here` themes one surface. Fanning out across a window would need a
  socket-side primitive to write bytes to a surface without going through typed input;
  that felt like a separate change, so this PR stays deliberately small.
- Help text and `docs/cli-contract.md` are updated.

## Testing

`cmuxTests/CLIThemesHereTests.swift` (wired into `project.pbxproj`; `lint-pbxproj-test-wiring.sh`
passes) covers, through the built CLI against a pinned `GHOSTTY_RESOURCES_DIR`:

- a named theme produces exactly the expected sequences, with no trailing newline — a
  stray byte here would scroll the user's surface
- `--reset` emits the reset sequence with no theme argument
- an unknown theme exits non-zero and writes nothing to stdout, so a failed lookup cannot
  leave a half-applied theme
- parsing skips comments, malformed colors, and out-of-range palette slots, and expands
  short hex

**I could not build or run these locally** — this machine has only Command Line Tools (no
full Xcode) and no `zig`, so `GhosttyKit.xcframework` cannot be built. The changed files
pass `swiftc -parse`, and the logic mirrors the existing Ghostty theme-file parser in
`CLI/cmux_open.swift`, but CI is the first real compile and test run. Happy to fix
whatever it turns up.

---
## Summary by cubic
Adds `cmux themes here` to recolor only the calling terminal surface. Previously, `themes set` changed colors globally; now, `here` writes OSC dynamic-color sequences to stdout for per-pane theming without touching config or other surfaces.

- Resolves a theme by bundled name (via the existing search path) or a file path, then emits OSC 10/11/12/4 for `foreground`, `background`, `cursor-color`, and `palette N` respectively; `--reset` sends OSC 110/111/112/104.
- Parser trims quotes/comments, expands short hex, enforces palette index 0–255, and ignores unknown keys. Unknown flags error; `--json` is rejected because output is raw escape bytes. Unknown theme exits non-zero and writes nothing; no trailing newline is emitted.
- No behavior changes to `set`, `list`, `clear`, or the picker. Per-surface colors do not survive session restore.
- Tests added in `cmuxTests/CLIThemesHereTests.swift` (exact byte stream, reset, unknown theme, parser); wired in `cmux.xcodeproj`; CLI help and `docs/cli-contract.md` updated.

<sup>Written for commit 29081cb52d2ba17dd36d2fb053cd582628e44cf1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10280?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CLI/CMUXCLI+Themes.swift
+++ b/CLI/CMUXCLI+Themes.swift
@@ -279,6 +279,11 @@ extension CMUXCLI {
                 targetBundleIdentifier: targetBundleIdentifier,
                 explicitPassword: explicitPassword
             )
+        case "here":
+            try runThemesHere(
+                args: Array(commandArgs.dropFirst()),
+                jsonOutput: jsonOutput
+            )
         case "clear":
             if commandArgs.count > 1 {
                 throw CLIError(message: "themes clear does not take any positional arguments")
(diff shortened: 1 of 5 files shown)


sergiokas:
Closing this for now — I opened it before it had been built or tested, and the scoping design is being reworked (per-surface theming will hang off `themes set` with mutually exclusive scope flags rather than a new `here` subcommand). Sorry for the noise; I will open a fresh PR once it is actually verified.

PR #10287, opened 2026-08-17
fix(ios): make artifact failure and scope switches exhaustive

azooz2003-bit:
https://github.com/manaflow-ai/cmux/commit/04ff18eea6ceb793252583e5c8f9df74b130a5e9 added `ChatArtifactError.unknown(code:)` and the `.panel` loader scope but missed two switches, so the cmux-ios scheme stopped compiling on Aug 14 at 9:18 PM, three minutes after the last successful TestFlight upload (build 20260814211222). Every iOS TestFlight run since has failed with `switch must be exhaustive` in `ChatArtifactFailurePresentation.swift:26` and `ChatArtifactInlineViewer.swift:163`, which is why the scroll-drain watchdog crash fix from https://github.com/manaflow-ai/cmux/pull/10186 never reached any TestFlight device.

This adds the `.unknown` failure presentation (localized en/ja; copy names the unrecognized code and never blames connectivity, matching the case's doc contract) and routes the `.panel` loader scope to the `.panel` viewer scope. Test coverage: `.unknown` joins the every-case presentation table and a new test asserts the code appears in the message without connectivity blame.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10287"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789594473&installation_model_id=21920&pr_number=10287&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10287&signature=4e939500dcb117a65e9fb820074dd7282e34fa91541ba9ff79a18c219b17ab21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
[...]

PR #10289, opened 2026-08-17
test(ios): wire CmuxAgentChatUITests into the cmux-ios test plan

azooz2003-bit:
Follow-up to https://github.com/manaflow-ai/cmux/pull/10287. The `CmuxAgentChatUI` package was never referenced by `ios/cmux.xcworkspace`, so its test target was unroutable in test-ios.yml and none of its suites have ever run in CI; a `test_filter` naming them fails with "isn't a member of the specified test plan or scheme" before building. Adds the workspace FileRef and test-plan entry, mirroring the existing CmuxMobileShellUI wiring. `check-workspace-package-groups.py` passes.

Wiring the target immediately surfaced two things the silence had been hiding, both fixed here:
- `ChatArtifactViewerErrorStateTests` never compiled: it exercises `ChatArtifactViewerModel.state(for:stat:)`, which is `private` and invisible to `@testable`. Opened to internal.
- Every Debug simulator build of main is still broken: the DEBUG-only `MacSurfaceGalleryPreviewView` picker fixture predates `TerminalPickerMenuActions.selectSimulatorStream` (from the https://github.com/manaflow-ai/cmux/commit/04ff18eea6ceb793252583e5c8f9df74b130a5e9 fallout; https://github.com/manaflow-ai/cmux/pull/10295 restored the store state but missed this call). Release archives compile the file out, which is why TestFlight went green while test-ios stayed red.

Validation: test-ios dispatch with `test_filter=CmuxAgentChatUITests` — https://github.com/manaflow-ai/cmux/actions/runs/32081148075. The `package-conventions-lint` and `mobile-core-package` reds are pre-existing main debt unrelated to this PR.

## Summary by CodeRabbit

* **Tests**
  * Added iOS coverage for the agent chat interface and chat artifact error states.
  * Improved startup and duplicate-request coverage for injected attachments.
  * Expanded scrollback interaction coverage and simplified test configuration for more consistent execution.
* **Improvements**
  * Updated the Mac surface gallery preview to handle simulator-stream selection consistently during demonstrations.

### DIFF
--- a/Packages/iOS/CmuxAgentChatUI/Sources/CmuxAgentChatUI/Artifacts/ChatArtifactViewerModel.swift
+++ b/Packages/iOS/CmuxAgentChatUI/Sources/CmuxAgentChatUI/Artifacts/ChatArtifactViewerModel.swift
@@ -339,7 +339,9 @@ final class ChatArtifactViewerModel {
         await temporaryFileStore.remove(temporaryFileURL)
     }
 
-    private static func state(
+    // Internal (not private) so the error-state tests can exercise the
+    // mapping directly via @testable import.
+    static func state(
         for error: any Error,
         stat: ChatArtifactStat?
     ) -> ChatArtifactViewerState {
--- a/Packages/iOS/CmuxAgentChatUI/Sources/CmuxAgentChatUI/Resources/Localizable.xcstrings
+++ b/Packages/iOS/CmuxAgentChatUI/Sources/CmuxAgentChatUI/Resources/Localizable.xcstrings
@@ -980,7 +980,7 @@
         "en": {
           "stringUnit": {
             "state": "translated",
-            "value": "Couldn’t load file"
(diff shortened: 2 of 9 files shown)


azooz2003-bit:
Validated: [run 32087301533](https://github.com/manaflow-ai/cmux/actions/runs/32087301533) — `ios-simulator (iphone)` and `(ipad)` green with `test_filter=CmuxAgentChatUITests/ChatArtifactFailurePresentationTests`. The unfiltered [round-8 run](https://github.com/manaflow-ai/cmux/actions/runs/32085623100) executed the whole plan: 134 tests in 32 suites, all passing (its red conclusion was the verdict-script format bug fixed in 4cb5f5e976). Remaining red is `package-conventions-lint` only, pre-existing main debt (namespace enums + free functions from the Aug-14/17 merges), untouched by this PR.

PR #10293, opened 2026-08-17
Fix Claude wrapper settings re-entry loop

### LINKED ISSUE #10230: Wrapper re-entry through a custom Claude Binary Path duplicates injected --settings without bound, pinning a CPU core and breaking session restore
**Environment:** cmux 0.64.22 (102) [ddd4a01bc], macOS 26.5.1 (arm64), Claude Code 2.1.233. Settings → Automation → Claude Binary Path set to a user script; Terminal → Resume Agent Sessions on Reopen enabled.

## Summary

`cmux-claude-wrapper` injects its hook configuration as an inline `--settings <json>` argument and then execs the configured Claude binary. If that binary is a user script that ultimately resolves `claude` through `PATH`, the lookup finds cmux's own shim (`$TMPDIR/cmux-cli-shims/<surface-uuid>/claude`, which cmux places on `PATH`), and control returns to the wrapper. The wrapper treats the `--settings` it produced on the previous pass as user-supplied and merges its hook configuration into it again. Because the merge concatenates arrays, every pass appends a full copy of the hook block. The whole cycle happens inside a single PID via `exec`, so it is invisible in the process tree.

## Observed failure

After a cmux update, eight `cmux-claude-wrapper` processes ran at ~95% CPU for eight minutes without ever launching Claude Code; system load average reached 113 and `cmux restore` never returned. Each process had a 243 KB `--settings` argument containing 492 copies of the `cmux hooks feed --source claude` block.

Two distinct failures follow from the oversized argument:

1. **CPU exhaustion.** The wrapper's own argument loop matches every argument with shell glob patterns (`case "$arg" in -*) … *=* …`). On a 243 KB string, bash's pattern matcher backtracks recursively; `sample` on the stuck PID shows an unbroken recursion inside bash. The process never reaches the exec and never terminates. This is a denial of service triggered by the wrapper's own output, and it would equally affect any user who legitimately passes a large `--settings`.

2. **Merge failure.** `argv` plus the two base64 copies the wrapper places in the environment (`CMUX_USER_SETTINGS_B64`, `CMUX_AGENT_LAUNCH_ARGV_B64`) exceed macOS `ARG_MAX` (1 MB). `node` cannot be executed, producing `node: Argument list too long` followed by `cmux: warning: --settings merge failed; your --settings was ignored`.

## Reproduction

1. Create `/tmp/claude-shim` containing `#!/bin/sh` and `exec some-launcher claude "$@"`, where `some-launcher` locates `claude` via `PATH` (Python's `shutil.which`, `command -v`, `execvp`, etc.).
2. Set Settings → Automation → Claude Binary Path to `/tmp/claude-shim`.
3. Start Claude Code in a cmux pane, or restore a pane with a saved Claude session.
4. Watch the argument length of the pane's process:
   ```
   while :; do ps -o command= -p <pid> | wc -c; sleep 0.4; done
   ```
   It grows by roughly 1.7 KB per cycle, and `ps -o comm= -p <pid>` alternates between the wrapper and the launcher.

A quick check for whether a given process is affected:

```
ps -o command= -p <pid> | grep -c 'hooks feed --source claude'
```

3 is the normal count (one per hook event). Tens or hundreds mean the loop is active.

## Notes on existing guards

`find_real_claude` skips candidates identified as the shim via `CMUX_CLAUDE_WRAPPER_SHIM` / `CMUX_CLAUDE_WRAPPER_SHIM_ROOT`, but that only covers the wrapper's own resolution — it does not cover a `PATH` lookup performed by a downstream launcher. The re-exec counter `cmux_claude_wrapper_reexec_guard` (limit 16) also failed to bound the observed run, which produced 492 copies rather than at most 16; the counter appears to be reset somewhere along the chain.

## Suggested fixes, in order of how much they remove

1. **Pass the hook configuration as a file path rather than inline JSON.** The wrapper's merge helper already accepts a path (`load()` reads a file when the value does not begin with `{` or `[`). Writing the JSON to a temp file and passing `--settings /path/to/file.json` eliminates the `ARG_MAX` failure and the bash pattern-matching failure at once, and caps growth structurally.
2. **Make the merge idempotent.** Mark cmux's injected block with a sentinel key and replace it on re-entry instead of concatenating, or de-duplicate identical hook entries after merging.
3. **Detect self-injection.** Before merging, drop any incoming `--settings` whose content already contains the cmux hook block.
4. **Keep the shim directory out of `PATH`,** or place it last, so a downstream `PATH` lookup reaches the real binary first.
5. **Guard the argument loop.** Bail out with a clear error when a single argument exceeds a sane size (say 256 KB) instead of pattern-matching it, and prefer fixed-prefix tests over glob patterns on unbounded input.

## Workaround for affected users

Strip cmux shim directories from `PATH` inside the custom Claude Binary Path script before handing off, mirroring what the cmux shim itself does in its fallback branch:

```bash
clean=""
IFS=:
for e in ${PATH:-}; do
    case "$e" in */cmux-cli-shims|*/cmux-cli-shims/*) continue ;; esac
    [ -n "$e" ] || continue
    clean="${clean:+$clean:}$e"
done
unset IFS
export PATH="$clean"
exec some-launcher claude "$@"
```

Recovery from a live incident: `pkill -f cmux-claude-wrapper`.

austinywang:
Closes #10230

## Root cause

`cmux-claude-wrapper` serialized its generated hook settings inline in `argv`. A configured Claude launcher that resolved `claude` through PATH could hit cmux’s per-surface shim again; the next wrapper pass interpreted the previous generated JSON as user settings, concatenated hook arrays, and reset the re-entry guard when the launcher was misclassified as a real binary. The resulting argv and base64 environment captures grew until bash spent unbounded time matching the huge argument or Node hit `ARG_MAX`.

## Fix

- Hook settings are now written to a mode-600 temporary file and passed as a path-valued `--settings`; user settings inputs are streamed through a temporary NUL-delimited file and the launch capture is made after normalization.
- The generated document carries a cmux sentinel and hook fingerprints. Re-entry removes/replaces only cmux-owned groups, so repeated passes converge to one block while genuine user settings and hooks remain.
- Managed settings are recognized before merge, including older generated payloads, and repeated managed re-entry can fall back to the real PATH binary without re-running the custom launcher.
- Custom launcher execution receives a PATH with cmux shim roots removed; delegated launcher forms are classified as re-entry candidates so the hop/history guard remains bounded.
- Argument inspection is capped at 256 KiB and uses fixed-prefix parsing; oversized inline inputs fail clearly instead of entering expensive shell glob matching.

## Validation

Focused Python behavior harnesses pass: `test_claude_wrapper_hooks.py`, `test_claude_wrapper_mutual_shim_loop.py`, `test_claude_wrapper_user_binary_resolution.py`, `test_claude_wrapper_shim_root_survives_tmpdir_change.py`, and shell dispatch/Claude Teams wrapper tests where a CLI binary was available. No app build or Xcode/UI test was run, per the issue task.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10293"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789598021&installation_model_id=21920&pr_number=10293&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10293&signature=753f5c4af6c3e0c0238bfafb7adad4a376e1c1539f4de99924890dbd202d5209"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes a re-entry loop in the Claude wrapper that duplicated hooks and blew up argv/env. Settings are now file-backed and launch metadata is bounded; managed re-entry resolves to the real `claude` so launches converge and custom launchers no longer loop into the shim. Addresses #10230.

- `--settings` now points to a mode-0600 temp file containing a combined document with `{"__cmux":{"managed":"claude-hooks","version":1,"hookFingerprints":[...]}}`; re-entry replaces only cmux-owned hook groups and preserves genuine user hooks.
- Managed re-entry is detected from inline or file-based args; repeated managed re-entry execs the real `claude` from `PATH`, skips any `CMUX_CUSTOM_CLAUDE_PATH` launcher, and removes shim roots from `PATH` before exec to prevent loopback.
- Indirect launcher forms (`exec`/spawn/`which`/`command -v`) are treated as re-entry candidates; a direct absolute `/.../claude` is the real boundary and resets the hop guard. On non-shim boundaries the guard clears and the target inherits a shim-free `PATH`.
- Argument inspection is capped at 120 KiB (byte-accurate, multibyte-safe); oversized inline `--settings` fail fast. Launch argv capture is now post-merge and compact: previous captures are cleared on re-entry, and only filtered args (with a path-valued `--settings`) are encoded.
- Deep-merge uses `node` and writes the combined document to a temp file; if `node` is missing or merge fails, user settings are preserved unchanged and cmux hooks are disabled for that launch with a clear warning.
- Tests cover file handoff and permissions, genuine user hook preservation, large-input limits and multibyte handling, large-file merging without argv growth, `PATH` cleanup for custom launchers, and convergence of re-entry.

<sup>Written for commit fc4f5800657e320564b8d14f90d81f7c2dcb61a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of custom launchers and indirect re-entry scenarios.
  * Prevented repeated hook injection and launcher path leakage.
  * Added safeguards for oversized arguments and settings values.
  * Preserved user-provided hooks during settings merges.
  * Added cleanup and fallback behavior when settings processing fails.

* **Tests**
  * Expanded coverage for inline and file-based settings, large inputs, custom launchers, timeouts, and hook preservation.


austinywang:
Review follow-up (head 1c0724bdde):\n\n- Merge failures now preserve the original user --settings arguments, omit cmux's settings flag, disable hooks for that launch, and emit a generic actionable warning. Timeout diagnostics are decoded before returning.\n- The oversized-argv regression uses a 122,880-byte production cap and a 125,000-byte payload (below Linux's per-argument exec limit); the separate 200 KB file-based test covers legitimate large settings.\n- The re-entry assertion was extracted into a helper; TMPDIR is fixture-local and generated files are mode 0600.\n- The __cmux marker is intentional and required for idempotent self-injection detection. I verified the generated file with Claude Code 2.1.233 and 2.1.234 using --settings ... --version and doctor; both accepted it without an unknown-key error. The published Claude settings schema also permits additional properties.\n- I did not prune arbitrary cmux-claude-settings.* files: mktemp names are private per launch, and deleting matching files could race another active surface. Intermediate files are explicitly removed; final settings files remain in the OS temp area for the Claude process.

PR #10302, opened 2026-08-18
spec(tui): aggregate several machine providers in one machines column

lawrencecchen:
## Summary

The machines column is already the leftmost native rail, but its catalog supports exactly one source. `main.rs` rejects more than one of `--machine-provider`, `--machine-provider-command`, and `--cloud`, and it refuses a static `machines` array beside a socket or command provider. The only composition that exists stacks the static v0 catalog over one v1 cloud catalog with a hardcoded `2^63` key split. So SSH remotes and sandbox suppliers cannot appear together.

The wire contract does not change. Every provider still speaks `machine-provider-v1`, negotiates its own version, and owns its own bearer, so a provider written against the existing contract works unchanged. Aggregation is a client concern.

## What has landed

**Specification** — `cmux-tui/spec/machine-provider-aggregation.md` describes the registry, the key namespace, per-provider isolation and capability scoping, the port of the static SSH catalog onto the same abstraction, the `machine-provider` plugin manifest kind, and where sandbox state lives.

**Slice 1a, the key namespace** — `machine_key.rs` splits `MachineKey` into a 16-bit provider slot and a 48-bit per-provider ordinal, replacing the `2^63` split. Slot 0 stays the current session. A per-provider allocator keeps one key for one provider-stable id across snapshot refreshes, so a refresh cannot move a row or invalidate a pending action. It fails closed on ordinal exhaustion and on a registry larger than the slot field, rather than alias another provider's rows.

**Slice 1b, the registry** — `machine_registry.rs` resolves an ordered `machine_providers` array with `builtin-ssh`, `unix`, `command`, and `ssh` kinds. `env_passthrough` lists variable names only, so an API key stays in the provider process and never enters `cmux-tui.json`. A legacy configuration keeps working with no edit: the built-in SSH entry always exists, and an enabled `machine_provider.cloud` desugars into an `ssh` entry. Explicit entries come first and can replace a desugared one by reusing its id. Truncation of an oversized registry happens before the desugared entries are appended, so it cannot push the static catalog out.

## Verification

`./scripts/verify-cmux-tui-hosted.sh --filter machine_` passes on hosted Linux and macOS. 13 `machine_key` tests and 14 `machine_registry` tests run there.

## Do not merge yet

Nothing consumes the registry yet, so `machine_providers` parses without an effect. The aggregator lands in the next slice and removes the `allow(dead_code)` on `machine_key`. `docs/configuration.md` deliberately does not document the key until it does something. This stays a draft until then.

## Remaining slices

2. Per-provider generations, isolation, and capability scoping in `machine_provider_runtime.rs`. This is the real refactor and the point where the registry starts to matter.
3. Port the static SSH catalog onto the provider trait.
4. `kind = "machine-provider"` in `cmux-plugin.toml`, reusing the existing installer.
5. The first supplier provider.

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/config.rs
+++ b/cmux-tui/crates/cmux-tui/src/config.rs
@@ -51,6 +51,16 @@
 //!     "max_width": 0,
 //!     "create_sources": []
 //!   },
+//!   "machine_providers": [
+//!     {"id": "ssh", "kind": "builtin-ssh", "name": "SSH"},
+//!     {
+//!       "id": "e2b",
+//!       "kind": "command",
+//!       "name": "E2B",
+//!       "command": ["cmux-provider-e2b"],
+//!       "env_passthrough": ["E2B_API_KEY"]
+//!     }
+//!   ],
 //!   "machine_provider": {
 //!     "cloud": {
 //!       "enabled": false,
@@ -176,6 +186,8 @@ struct RawConfig {
     machine_sidebar: RawMachineSidebar,
     #[serde(default)]
     machine_provider: RawMachineProvider,
+    #[serde(default)]
+    machine_providers: Vec<crate::machine_registry::RawMachineProviderEntry>,
     #[serde(default)]
     machines: Vec<RawMachine>,
     #[serde(default)]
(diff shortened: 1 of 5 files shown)


lawrencecchen:
Closing this stale provider aggregation spec. Current main already has provider abstractions and consumers; a future change needs a concrete consumer and current protocol contract.

Issue #10348, opened 2026-08-18
Idle app holds 48-56% of one core: History CommandMenu rebuilds the full focus history on every App body evaluation


STRML:
Idle cmux holds a steady 48-56% of one core, and none of it is terminal work. The App body re-evaluates continuously; each evaluation rebuilds the History `CommandMenu`, which walks the entire focus history in both directions to render 10 rows. `AppDelegate.makeMainMenu` then calls `ViewGraphRootValueUpdater.invalidateProperties` inside the same update, so the graph is dirty again before the pass finishes.

0.64.22 (102) stable, macOS 26.6.1 (25G76), arm64. One window, 7 workspaces, 20 live terminal surfaces running Claude Code, app up 8 days.

## Measurements

Process launched 2026-08-10, 18 hours of CPU consumed at capture time:

```
$ top -l 2 -pid 11270 -stats cpu,th
48.3 110
48.1 110
55.9 109/1
```

`/usr/bin/sample 11270 5`. Main thread, 3808 samples, 2592 idle in `mach_msg2_trap`. The busy remainder:

```
509  NSHostingView.beginTransaction()
505   GraphHost.flushTransactions()
497    GraphHost.runTransaction(_:do:id:)
491     AG::Subgraph::update(unsigned int)
483      AG::Graph::UpdateStack::update()
215       DynamicBody.updateValue()
204        ViewBodyAccessor.updateBody(of:changed:)

182  GraphDelegate.beginTransaction()
182   AppGraph.updateGraph<A>(body:)
179    GraphHost.flushTransactions()
104     AppGraph.graphDidChange()
102      AppBodyAccessor.updateBody(of:changed:)
 95       Scene.commands<A>(content:)
 58        CommandMenu.init(_:content:)
 55         FocusHistoryModel.focusHistoryMenuSnapshot(direction:maxItemCount:)
 75     AppGraph.graphDidChange() + 932
 63      AppDelegate.scenesDidChange(phaseChanged:)
 59       AppDelegate.makeMainMenu(updateImmediately:)
 58        AppKitMainMenuItem.updateMainMenu(...)
 58         AppKitMainMenuItem.updateMenuHost(...)
 58          ViewGraphRootValueUpdater.invalidateProperties(_:mayDeferUpdate:)
```

`AppBodyAccessor.updateBody` at 102 samples in a 5-second idle window is the thing to look at. Nothing on screen was changing.

[...]

PR #10349, opened 2026-08-18
fix(focus-history): resolve each history entry once per menu snapshot

STRML:
`focusHistoryMenuSnapshot` resolves every history entry three times. The `compactMap` calls `resolvedFocusHistoryEntry(for:)`, then `focusHistoryEntryIsNavigable` resolves the same entry again, and that calls `focusHistoryEntryResolvesToCurrent`, which resolves it a third time. Each resolution runs a `workspaceExists` check plus `resolvedFocusHistoryPanelId`, which reaches `PaneTreeModel.panelId(forSurfaceId:)` through the host.

Against the 50-entry cap across both directions, one History menu build costs up to 300 host resolutions to render 10 rows. The History `CommandMenu` rebuilds on every App body evaluation rather than on open (`Sources/cmuxApp+HistoryMenu.swift:9-11`), so this is paid continuously, not per-open.

Found while profiling an idle app holding 48-56% of one core: #10348.

## Change

Add an overload of `focusHistoryEntryIsNavigable` that takes the already-resolved entry plus the hoisted `scope`, and call it from the menu `compactMap`. The existing single-argument version delegates to it, which also drops one redundant resolution for its five other callers.

`focusHistoryEntryResolvesToCurrent` had no other caller, so its two lines are folded into the overload.

The snapshot output is unchanged. Only the number of host lookups moves.

Equivalence, since the overload skips two guards the old path ran:

- The `resolvedFocusHistoryEntry(for:) != nil` guard is subsumed by the caller already holding a non-nil `resolvedEntry`.
- `focusHistoryEntryResolvesToCurrent` reduced to `resolvedEntry == currentEntry` when `currentEntry` is non-nil, which is what the overload does inline.
- `navigationScope()` is a closure property read once per snapshot; it cannot change mid-`compactMap`, so hoisting it is safe.

## Verification

Two commits, red then green, per the regression policy in CLAUDE.md.

The test counts `workspaceExists` calls on the fake host across one snapshot of 9 navigable entries:

```
before:  Expectation failed: (host.workspaceExistsCalls → 27) == 9
after:   ✔ Test menuSnapshotResolvesEachEntryOnce() passed
```

`swift test` in `Packages/macOS/CmuxWorkspaces`: 209 tests in 30 suites pass.

I verified at package level only. I did not run a tagged Xcode app build; the change is confined to two private methods inside `CmuxWorkspaces` and adds no new user-facing strings, so there is no localization surface.

## What this does not fix

This trims one hot path, not the loop in #10348. The App body still re-evaluates continuously and still rebuilds the whole History menu each time, and `AppDelegate.makeMainMenu` still calls `ViewGraphRootValueUpdater.invalidateProperties` inside the same update. Those need someone who knows which publisher is driving the App body; I could not name it from a stripped release build.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10349"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789657090&installation_model_id=21920&pr_number=10349&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10349&signature=aaa9c3a521799dde03ebab4e52b2439ff12a12d24691199ebe172373c8dd79d8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Resolve each focus history entry once per History menu snapshot to remove redundant host lookups while keeping the snapshot identical. Previously the menu resolved each entry up to three times; now it resolves once, reducing continuous CPU cost because the History `CommandMenu` rebuilds on every App body evaluation.

- Add an overload of `focusHistoryEntryIsNavigable` that takes `resolvedEntry` and hoisted `scope`, and call it from the menu `compactMap`; the existing method delegates to it.
- Inline and remove the unused `focusHistoryEntryResolvesToCurrent`; compare against `currentEntry` directly.
- Hoist `navigationScope()` to one read per snapshot.
- Add a regression test that counts host `workspaceExists` calls: for 9 navigable entries, 27 → 9.
- No public API or menu output changes; changes are confined to private code in `CmuxWorkspaces`.

<sup>Written for commit 52617ff4317becf573e853791ff4d865a822b221. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10349?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved back-navigation menu generation to reduce redundant workspace lookups and keep entries consistent.
  * Ensured history menu building uses pre-resolved context to prevent repeated resolution during snapshot creation.

* **Tests**
  * Added a regression test verifying that a back-navigation menu with ten history entries displays nine entries.
  * Confirmed each workspace is checked exactly once during menu snapshot generation.

### DIFF
--- a/Packages/macOS/CmuxWorkspaces/Sources/CmuxWorkspaces/Navigation/FocusHistoryModel.swift
+++ b/Packages/macOS/CmuxWorkspaces/Sources/CmuxWorkspaces/Navigation/FocusHistoryModel.swift
@@ -262,20 +262,32 @@ public final class FocusHistoryModel: FocusHistoryNavigating {
         )
     }
 
-    private func focusHistoryEntryResolvesToCurrent(_ entry: FocusHistoryEntry, currentEntry: FocusHistoryEntry?) -> Bool {
-        guard let currentEntry,
-              let resolvedEntry = resolvedFocusHistoryEntry(for: entry) else { return false }
-        return resolvedEntry == currentEntry
-    }
-
-    private func focusHistoryEntryIsNavigable(_ entry: FocusHistoryEntry, currentEntry: FocusHistoryEntry?) -> Bool {
-        guard resolvedFocusHistoryEntry(for: entry) != nil else { return false }
-        if navigationScope() == .workspacesOnly,
+    /// Navigability given an already-resolved entry. Callers that have just
+    /// resolved `entry` pass the result here instead of resolving it again;
+    /// `resolvedFocusHistoryEntry` reaches the host for every panel lookup, and
+    /// the menu path runs this once per history entry per rebuild.
+    private func focusHistoryEntryIsNavigable(
+        _ entry: FocusHistoryEntry,
+        resolvedEntry: FocusHistoryEntry,
+        scope: FocusHistoryNavigationScope,
+        currentEntry: FocusHistoryEntry?
+    ) -> Bool {
+        if scope == .workspacesOnly,
            entry.workspaceId == currentEntry?.workspaceId {
             return false
         }
-        if focusHistoryEntryResolvesToCurrent(entry, currentEntry: currentEntry) { return false }
+        if let currentEntry, resolvedEntry == currentEntry { return false }
         return true
+    }
+
+    private func focusHistoryEntryIsNavigable(_ entry: FocusHistoryEntry, currentEntry: FocusHistoryEntry?) -> Bool {
+        guard let resolvedEntry = resolvedFocusHistoryEntry(for: entry) else { return false }
+        return focusHistoryEntryIsNavigable(
+            entry,
+            resolvedEntry: resolvedEntry,
+            scope: navigationScope(),
+            currentEntry: currentEntry
+        )
     }
 
     private func navigationEntry(for entry: FocusHistoryEntry) -> FocusHistoryEntry {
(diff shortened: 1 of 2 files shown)


STRML:
Superseded by #10661 (8185825617), closing.

That PR added the same overload this one proposes. focusHistoryMenuItem now resolves each entry once and passes resolvedEntry into focusHistoryEntryIsNavigable(entry:resolvedEntry:scope:currentEntry:), so the menu path no longer resolves twice per entry. resolvedFocusHistoryEntry(for:) is down to two call sites on main, one of them the convenience overload.

Found this merging main into my integration branch: the conflict was upstream having independently written the fix.

PR #9094, opened 2026-07-28
Collapse spinner-frame titles before ingress dedup (~31% main-thread reduction under title churn)

Solarrixs:
Ports TerminalTitleChurnFilter from #6907 (credit @austinywang, co-authored-by on the commit) into GhosttyTitleUpdateIngress.submit. Spinner frames (`⠋ Working` -> `⠙ Working`) always differ, so the existing `update != lastSubmittedUpdate` guard never fires - every frame pays yield -> task enqueue -> workqueue syscall on the main thread. That's 31.2% of main-thread samples under heavy churn, and 3-5x worse click latency vs constant titles at matched rates.

What changed vs #6907:

1. Filter loses @MainActor - submit runs in ghostty's serialized per-view callback, same argument the ingress already makes for its own state (lines 13-15)
2. Per-surface LRU dropped - the ingress is per-view, so it collapses to a single lastStableTitle
3. FIFO drain gone - superseded by the AsyncStream + dispatcher that landed on main

What survives is the ~30 lines that matter: the braille U+2800-28FF leading-run collapse + its tests.

To be clear, this doesn't fix saturation-level unresponsiveness (https://github.com/manaflow-ai/cmux/issues/9092). Closes nothing, should help #6599.

---
[...]

Issue #6599, opened 2026-06-22
perf: coalesce high-frequency terminal title updates before workspace/sidebar refresh


Eridanus117:
## Problem

After the TabManager ownership fanout fix in #6570/#6551, a single active terminal can still drive high CPU when it emits continuously changing title updates. The remaining path is no longer duplicate cross-manager enqueueing. It is the per-owned-workspace title path itself: terminal title changes mutate the workspace/panel title, then wake workspace panel refresh, sidebar invalidation, and mobile workspace observation at the source title-change cadence.

This shows up with a small workspace count, so it is not primarily a scale-with-many-tabs issue.

## Evidence

Local controlled title-spam repro on current main-derived builds, same workload and machine:

```text
default title path:
  duration: 56.32s
  avg CPU: 36.99%
  workspace.title.updatePanel: 261

full TabManager title suppression experiment:
  avg CPU: 2.9%

1000ms title-update coalescing experiment:
  duration: 48.71s
  avg CPU: 3.18%
  workspace.title.updatePanel: 18
```

Interpretation: title churn is the highest-ROI source in this repro. A 1000ms coalescing window reduced downstream title/panel refresh work from hundreds of updates to tens, with CPU near the full-suppression lower bound.

This is separate from #6546. Splitting summary/detail sidebar snapshots can reduce the cost of each refresh, but it does not address the source rate when a terminal title changes continuously.

## Proposed direction

Add an opt-in terminal title update coalescing setting:

```json
{
  "terminal": {
    "titleUpdates": {
      "coalescing": {
        "enabled": true,
        "milliseconds": 1000
      }
    }
  }
}
```

Constraints:

- Default behavior should remain unchanged.
- When disabled, keep the current ~30 Hz coalescing behavior.
- When enabled, clamp the coalescing interval to a bounded range.
- Flush pending title updates before workspace ownership transfer/detach so a delayed title is not lost.
- Cover config parsing and the coalescer behavior with tests.

## Related

- #6551 / #6570: removes cross-manager fanout before enqueue.
- #6546: reduces cost of sidebar snapshot refreshes after a title/immediate update.
- #6564: earlier draft direction; this issue tracks the narrower source-rate control backed by the controlled title-spam data.

Eridanus117:
Opened draft PR #6600 with the opt-in implementation: https://github.com/manaflow-ai/cmux/pull/6600

[...]

PR #10351, opened 2026-08-18
fix: keep spinner titles animating without waking the app on every frame

STRML:
**Status:** updated onto `main` (`78f4c82504`) as merge commit `1e34d3640a`. Two conflicts, both resolved keeping the upstream side too:

- `Sources/Workspace+TitleOwnership.swift`: `main` now projects a cloud title (`cloudProjectedResource(forPanel:).cloudProcessDisplayTitle`) into the trim step. That projection is preserved, and the stable-title mechanism sits on top of it. The tab label still takes the animating frame through `refreshTabLabel`; `panelTitles` takes the stable title.
- `Sources/TabManager.swift`: `main` added a `Bool`-returning `updatePanelTitle(tabId:panelId:title:)` that the cloud rename paths call (`TabManager+CloudAgentTitle`, `TerminalController+AgentTitle`). Both entry points now exist: that one unchanged, plus the spinner-aware `stableTitle` variant the coalescer calls.

**Reviewer's guide**
- Read `TerminalTitleChurnFilter` (frame classification) first, then `GhosttyTitleUpdateIngress` and `Workspace+TitleOwnership.updatePanelTitle`.
- Tests: `TerminalTitleChurnFilterTests.swift`, `cmuxTests/GhosttyTitleUpdateIngressTests.swift`.
- The point of the change: a frame that only advances a spinner refreshes the tab label and stops. Everything else keys off the stable title, so it never writes `@Published` state and never reaches the sidebar or titlebar.

---

A spinner frame is cheap to draw and expensive to announce. cmux currently announces it: every tick of `✳ Working` or `π ⠋ safety-rating` is a new title string, so it flows through the workspace title, the sidebar's observation stream, the titlebar, and the toolbar. On a window running several agents that is the app re-laying itself out to move one glyph.

This keeps the animation and cuts the announcement.

## Why the earlier approach was wrong

The first version of this PR removed the frame from the title, which is what `TerminalTitleChurnFilter` already did for Braille spinners on main. It works, and dogfooding showed why it should not ship: the tab stops animating. Widening the frame set to the glyphs agents actually use just made the existing behavior visible. Losing the animation to save CPU is paying with the feature.

## Change

`GhosttyTitleUpdate` and `GhosttyTitleChange` carry two strings:

- `title` stays the frame the terminal emitted, spinner included.
- `stableTitle` is that frame with a standalone spinner glyph removed, so consecutive ticks share one.
- `isSpinnerFrameOnly` is `title != stableTitle`.

The ingress forwards every frame again instead of collapsing them, so the label still has something to animate with. The saving comes from what each consumer does with it:

| consumer | on a spinner-only frame |
|---|---|
| `Workspace.updatePanelTitle` | refreshes the AppKit tab label, then returns before touching `@Published panelTitles` or `applyFocusedPanelTitle` |
| `TabManager.updatePanelTitle` | skips the window title and the display-title comparison |
| `ContentView.onReceive` | skips `scheduleTitlebarTextRefresh` |
| `WindowToolbarController` | skips `scheduleFocusedCommandTextUpdate` |

`panelTitles` doubles as the memo, since it already holds the last stable label, so the change adds no new state.

The tab label path is `bonsplitController.updateTab`. `BonsplitController` is a plain `final class` with no `@Published` and no `ObservableObject` conformance, so that call is an AppKit label redraw rather than a SwiftUI invalidation. That is what makes it safe to run on every frame.

Also extends the frame set past Braille to `U+2731...U+273D` (asterisks, which is what Claude Code animates) and `U+25D0...U+25D3` / `U+25F4...U+25F7` (circles), and drops the leading-position requirement so a branded title like `π ⠋ label` is recognized. Sampling a live 20-surface window, every animated title used a glyph or a position the old filter missed, so it collapsed nothing there.

## Deleted test, deliberately

`spinnerFramesCollapseBeforeAsyncStreamEnqueue` asserted `submitted == (index == 0)`, i.e. that frames are dropped at the ingress. That assertion is the animation bug, so it is replaced by:

- `spinnerFramesAreForwardedSoTheTabLabelKeepsAnimating`
- `repeatedIdenticalFrameIsStillRejected`, so the plain duplicate guard is still covered
- three tests over the `stableTitle` / `isSpinnerFrameOnly` contract, including the legacy-payload fallback

## Verification

`xcodebuild -scheme cmux -configuration Debug`: **BUILD SUCCEEDED**.

I cannot run `cmuxTests` locally, so CI is the gate there. The churn filter itself I exercised through a standalone harness compiling the real source file alongside every pre-existing assertion plus the new ones: 23 pass, and against unmodified `main` the new cases fail.

Both new fields default to `nil` and fall back to the raw title, so the other `GhosttyTitleChange(...)` and `updatePanelTitle(...)` call sites compile unchanged and treat their input as a real change, which is the safe direction.

Running it: the tab animates as before, and the sidebar's own `✦ Running` indicator is unaffected since it comes from agent status rather than the title.

## Not in this PR

The per-frame `NotificationCenter` post stays. Removing it needs a display-only sink from the ingress to the tab bar, which is a bigger change; this cuts the cascade behind the post, which is the expensive half.

Related: #10348, #6599. Supersedes #9094.

## Summary by CodeRabbit

* **Bug Fixes**
  * Terminal titles now remain stable while spinner animations update, reducing unnecessary tab, sidebar, and workspace refreshes.
  * Spinner-only titles are ignored, while meaningful title changes continue to appear immediately.
  * Title normalization now supports additional spinner styles without altering ordinary symbols or branded titles.

* **Tests**
  * Expanded coverage for spinner frames, embedded titles, stable-title handling, deduplication, and legacy notification behavior.

### DIFF
--- a/Packages/macOS/CmuxTerminalCore/Sources/CmuxTerminalCore/TitleChurn/TerminalTitleChurnFilter.swift
+++ b/Packages/macOS/CmuxTerminalCore/Sources/CmuxTerminalCore/TitleChurn/TerminalTitleChurnFilter.swift
@@ -1,8 +1,15 @@
 /// Normalizes animated terminal titles before title-update deduplication.
 ///
-/// Command-line spinners commonly prefix a stable label with one of ten
-/// Braille Pattern frames. Collapsing that standalone animation token makes
-/// successive frames equal while preserving ordinary titles byte-for-byte.
+/// Command-line spinners animate one glyph next to a stable label. Removing
+/// that standalone animation token makes successive frames equal while
+/// preserving ordinary titles byte-for-byte.
+///
+/// The token is removed wherever it sits, not only in leading position, because
+/// agents brand their titles: OMP writes `π ⠋ label`, so a leading-only rule
+/// leaves its frames churning. A glyph is only removed when whitespace or a
+/// string edge bounds it on both sides, and a title containing a run of two or
+/// more spinner glyphs is treated as text and returned untouched, which is what
+/// keeps Braille words such as `⠋ ⠑⠇⠇⠕` intact.
 public struct TerminalTitleChurnFilter: Sendable {
     /// Creates a stateless terminal-title normalizer.
     public init() {}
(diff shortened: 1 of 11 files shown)


STRML:
Pushed 9ef27258b7, which extends this past the leading-position rule to cover branded titles.

The gap was OMP. It writes its brand glyph before the spinner, so the frame is never first:

```
π ⠋ safety-rating
π ⠦ Increase retry baseDelayMs for OpenRouter fallback
π > Continue after corruption          <- idle frame
```

`⠋` U+280B and `⠦` U+2826 were already in the Braille set. They sit at index 2, and `stableTitle` only tested the first non-whitespace character, so `isKnownSpinnerFrame('π')` came back false and the raw title went through. Three surfaces in my sample churned that way for want of two characters of offset.

The rule is now positional-free. Remove any known frame that whitespace or a string edge bounds on both sides, then collapse the whitespace it leaves. A title holding a run of two or more spinner-range glyphs is treated as text and returned untouched. A title with nothing removed is returned byte-for-byte.

```
π ⠋ safety-rating   -> π safety-rating     (all 10 frames, one title)
✳ CI checks         -> CI checks
Build ⠋ step        -> Build step          <- changed, see below
⠋ ⠑⠇⠇⠕              -> ⠋ ⠑⠇⠇⠕              (run guard)
✳ ✳✳✳               -> ✳ ✳✳✳               (run guard)
⠋-project           -> ⠋-project           (not standalone)
⣿ Building project  -> ⣿ Building project  (not a frame)
```

This drops `"Build ⠋ step"` from `preservesOrdinaryTitlesExactly`. That assertion is the one thing standing between the filter and branded titles, and I don't think it describes real input: a lone dots-frame with whitespace on both sides is an animation wherever it sits. Braille text is a run, and the run guard is what actually protects `⠋ ⠑⠇⠇⠕`. If you disagree, the alternative is a narrower rule keyed on the prefix (single token, non-alphanumeric, shorter than N) and I'll write that instead.

Every other pre-existing assertion is unchanged. All three agents in my window are now covered: Claude Code's asterisk, OMP's branded Braille, and the half-circles.

STRML:
Do not merge as pushed. Dogfooding this caught a regression I missed: stripping the frame from the title means the tab stops animating. That was already main's behavior for Braille spinners, and widening it to the glyph families agents actually use made it visible.

Rewriting to keep the animation and cut only the expensive half: carry the stable label alongside the raw frame, give the raw frame to the AppKit tab bar, and let the workspace title, titlebar, and toolbar bail when only a spinner frame changed. Will force-push here.

STRML:
Rebased onto `main` (78f4c82504). Two conflicts: `Workspace+TitleOwnership.swift` (main's cloud-title projection is preserved, the stable-title mechanism sits on top of it) and `TabManager.swift` (both entry points now exist: main's Bool-returning cloud path unchanged, plus the spinner-aware variant). Reviewer's guide at the top of the description. Only red check is CLA. @lawrencecchen relates to open issue #6599.

STRML:
recheck

STRML:
Rebased onto `main` (`0be8ab34ad`) as a linear 6-commit stack, no merge commits. Force-pushed `fix/title-churn-non-braille-spinners` @ `2f8d326a20b`. Upstream's `AutomaticTerminalTitle` bounding (#13009) now runs before this PR's position-independent spinner removal; all 12 `TerminalTitleChurnFilterTests` pass locally via `swift test --filter TerminalTitleChurnFilterTests`.

STRML:
Rebased onto `main` (623e5bcbb80) as a linear 6-commit stack. Force-pushed fix/title-churn-non-braille-spinners @ 1dd098fa588 (supersedes the earlier 0be8ab34ad-based stack). All 12 TerminalTitleChurnFilterTests pass.

STRML:

STRML:

PR #10398, opened 2026-08-19
macOS runtime backend-environment switcher (Production/Staging)

azooz2003-bit:
Stacked on https://github.com/manaflow-ai/cmux/pull/10393 (base branch `feat-backend-env-core`); part of the staging-first program with https://github.com/manaflow-ai/cmux/pull/10389.

Release/nightly Macs hard-resolve to cmux.com, api.cmux.sh, and the prod Stack project. This adds a persisted runtime override (`CMUXBackendEnvironmentOverride`, UserDefaults) that replaces only the build-default layer of `AuthEnvironment` resolution and applies on relaunch:

- Staging flips web origin / API base / VM API / iroh broker fallback to `https://cmux-staging.vercel.app`, the Stack channel to the development project, and the presence default to the dev worker (which verifies dev-Stack tokens). `pushAPIBaseURL` and the sign-in origins follow existing fallbacks; verified, not special-cased.
- Precedence everywhere: explicit `CMUX_*` env vars (including the `LSEnvironment` origins reload.sh bakes into tagged dev builds) > DEBUG-only `~/.cmux-dev.env` > runtime override > build default. Dev-lane isolation is untouched, and the Settings card shows a "pinned by launch environment" note in baked builds instead of silently ignoring the picker.
- The credential-bearing app-session handoff origin allowlist grows by exactly the compiled-in staging origin, only while the override is staging; env values can never widen it.
- Settings > Account: Production/Staging picker visible when `CMUXBackendEnvironmentSwitchGate.allows(currentUser)` (verified @manaflow.ai) OR DEBUG OR the state is already non-production (switch-back always possible), consequence footer (separate accounts/data, signs you out, Mac + iPhone must match to pair), relaunch notice with a "Relaunch cmux" button composing the updater's `persistSessionForUpdateRelaunch` with the existing `restartApp` seam, orange Staging badge on the identity card, settings-search anchor.
- Session wipe on flip rides the existing `detectAuthProjectSwitch` (resolved Stack project id changes).
- Feature flags (`Sources/FeatureFlags.swift`, PostHog via cmux.com) deliberately stay on prod.

Tests: 5 new precedence tests in `cmuxTests/AuthEnvironmentTests.swift` (run in CI; local test runs are forbidden on dev Macs), CmuxSettingsUI package suite green locally (154 tests, includes 4 new). Localization audited: 8 new keys in `Resources/Localizable.xcstrings`, en + ja.

[...]

PR #10389, opened 2026-08-19
Flip web deploy polarity: staging auto-deploys main, prod promotes manually

azooz2003-bit:
Backend changes currently reach production the moment they merge: `web/vercel.json` had `git.deploymentEnabled.main: true` on the git-linked `cmux` project, while `cmux-staging` deployed manually. That leaves no production-like environment to validate a change before real users (including cmux BETA) get it.

This PR reverses the polarity:

- **`web-deploy-staging.yml`** (new): every push to main touching `web/**` or `CHANGELOG.md` (the changelog page reads it at build time) runs the staging Aurora migration, deploys the `cmux-staging` Vercel project, then smokes it. Queued via a `concurrency` group so a mid-flight migration is never cancelled.
- **`web-promote-prod.yml`** (new): manual dispatch with a `ref` input. Resolves the sha, refuses anything not on `origin/main` history, runs the staging-then-production migration chain at that sha under the `cloud-vm-production` environment, deploys the `cmux` project, then smokes. Rollback = re-dispatch an older main sha (migrations are forward-only) or Vercel Instant Rollback.
- **`web/vercel.json`**: `main: false`, so the git integration stops deploying prod. The merge of this PR is itself the last auto-deploy.
- **`cloud-vm-migrate.yml` / `cloud-vm-smoke.yml`**: gain `workflow_call` triggers so the new workflows reuse them; `workflow_call` keeps the `environment:` bindings that carry the env-scoped secrets (a composite action cannot). Migrate accepts an optional main-history `ref` for promotions; the existing `migrate-production needs: migrate-staging` chain still enforces staging-first migrations. Dispatch behavior is unchanged.
- **`web/scripts/cloud-vm/deploy-vercel.mjs`** (+ `cloud-vm:deploy` script): links the target project from `projects.mjs` exactly like `docs-deploy-reusable.yml` (writes `.vercel/project.json` at repo root, `vercel deploy --prod --yes`), refuses to clobber an existing `.vercel` link. `bun run cloud-vm:deploy -- staging|production` is the break-glass CLI path.

No new secrets or environments: `VERCEL_TOKEN` and the `cloud-vm-staging`/`cloud-vm-production` environments already exist.

First-run verification after merge:
1. Watch the push-triggered staging run end to end (migrate, deploy, smoke).
2. Dispatch `web-promote-prod.yml` once and confirm cmux.com serves the promoted sha.
3. Confirm no prod auto-deploy appears on the next web-touching merge.
[...]

PR #10399, opened 2026-08-19
iOS runtime backend-environment switcher (Production/Staging)

azooz2003-bit:
Top of the stack: https://github.com/manaflow-ai/cmux/pull/10393 (shared model) → https://github.com/manaflow-ai/cmux/pull/10398 (macOS) → this. Part of the staging-first program with https://github.com/manaflow-ai/cmux/pull/10389.

`MobileAuthComposition` merges the persisted `CMUXBackendEnvironmentOverride` into the auth override table per key with precedence `LocalConfig.plist` > baked Info.plist > runtime override > build default. Staging contributes `AuthEnvironment=development` plus the staging origin for `ApiBaseURL` and `WebOriginURL` (magic-link callback moves with the API base); production contributes nothing, so today's resolution is byte-identical. TestFlight/App Store builds bake nothing, so the override applies exactly there; tagged dev builds stay pinned and Settings says so instead of showing a dead picker.

Verified follow-through (no changes needed): the iroh broker falls through to the resolved `apiBaseURL` for Release bundles (`MobileIrohRuntimeComposition.resolvedBrokerBaseURL`), presence picks the dev worker via `isDevelopmentAuthChannel`, and `detectAuthProjectSwitch` wipes the session on flip in both directions (covered by a new test).

Settings: Backend Environment section after Account with the Production/Staging picker (visible for verified @manaflow.ai users, DEBUG builds, or any already-non-production state so switch-back is always possible), consequence footer (separate accounts/data, signs you out, Mac + iPhone must match to pair), persistent "close and reopen cmux to apply" notice when pending ≠ active (no programmatic exit), and a staging badge in the section header and About. A new shared `CMUXBackendEnvironmentSwitchState` (CMUXAuthCore) threads active/pending/pinned from the composition into the SwiftUI environment; nil in previews hides the section.

Tests: 11 new cases in `MobileAuthEnvironmentOverrideTests` covering the full precedence matrix, pinned detection, pending-write isolation, session-clear on flip, and active-origin tracking. The iOS package graph cannot `swift test` on macOS (pre-existing: iOS-only platform declarations), so they are compile-verified via `swift build --build-tests --triple arm64-apple-ios18.0-simulator` and execute in CI's iOS lane. Localization audited: 8 new `mobile.settings.backend*` keys, en + ja.

Merge gate (per program plan): the five-method auth matrix (password, magic link, Apple, Google, GitHub) must pass on staging on both platforms first; the outstanding config item is adding `https://cmux-staging.vercel.app` to the dev Stack project's trusted domains (Stack dashboard).

Dictionary:
- **override table**: the string key/value map merged from LocalConfig.plist and Info.plist that steers iOS auth resolution.
- **bake**: a value written into Info.plist at build time by the reload scripts to pin a dev build's backend.
- **triple**: the compiler target descriptor (arm64-apple-ios18.0-simulator) that lets SPM compile iOS code without xcodebuild.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10399"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789712107&installation_model_id=21920&pr_number=10399&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10399&signature=02830cf0b4c4e6e5d2a45fc89a2d9bb863fb4362e15d79e51d832b6032f7c06f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Live iOS Production/Staging backend switch with session parking and per‑environment sign‑out. Previously the picker persisted an override and required a relaunch; now a confirmed switch parks the current session, shuts down the old graph, commits the choice, rebuilds the root, and restores the target session without exiting.

- Behavior and UX
  - Resolution precedence: explicit persisted choice (wholesale) > `LocalConfig.plist` > baked `Info.plist` > build defaults. Magic links, the iroh broker, and presence follow the resolved selection.
  - Picker: production‑lane builds keep Production/Staging; non‑production lanes add “Build lane (…)”. Full picker is gated to verified `@manaflow.ai`, DEBUG, or any already‑non‑production state; others on staging see a recovery‑only “Switch Back to Production”. Pinned builds remain read‑only. Badge and About reflect staging.
  - Sign‑out routing: required `MobileAccountSignOutActions` ensures all paths use the root teardown. Interactive sign‑out on explicit staging warns and then chains a return to the build lane; the delete‑account path skips the warning but still chains. Production and staging‑LAN rigs keep plain sign‑out.
  - Switch UI: confirmation, disabled state during the transaction, and a progress overlay that survives the root swap; tiers cover parking/retargeting/reverting, an inline sign‑in banner with “Back to Production,” and outcome notes. One‑shot marker suppresses the launch clear for switcher rebuilds only; organic build‑time flips still clear.

- Review and rollout
  - Composition: `AppCompositionHolder` replaces the SwiftUI root by generation. Adds shutdown paths across auth, iroh (auth observation, connectivity subscriber, network‑path observation), reachability, lifecycle diagnostics, and observers; process‑global one‑timers move to a single bootstrap. Exposes `CMUXBackendEnvironmentSwitchState` from `CMUXAuthCore`, injects it as `backendEnvironmentSwitch`, and adds a `BackendEnvironmentSwitchAction` seam in `CmuxMobileShellUI`.
  - Presence: `PresenceClient` now heads resolution with an explicit choice before env/defaults/bake tiers, matching auth.
  - Tests and CI: extends override/rebuild/clear cases; adds auth and iroh shutdown suites plus picker visibility, confirmation, and sign‑out interception tests. iOS CI runs; simulator triple compile‑verifies. Localized strings (en, ja).
  - Dev rig and config: add https://cmux-staging.vercel.app to the dev Stack project's trusted domains. `reloads.sh` drops restricted entitlements for local unpinned Release dogfood builds by default (set `CMUX_RELOADS_KEEP_ENTITLEMENTS=1` to keep them). `reload.sh` and `mobile-dev-launch.sh` clear any persisted backend choice and the one‑shot marker to keep reloads deterministic.

<sup>Written for commit bd292f7cf904f20ab0d92f70b6d5399bb8d1cfc4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


azooz2003-bit:
Update: the "close and reopen cmux" UX is gone. The switch now applies live: the static composition root became a replaceable `AppCompositionHolder` whose generation the root scene keys on via `.id()`, with the progress overlay mounted above that boundary. New `shutdown()` APIs on `AppCompositionRoot`/`MobileAuthComposition`/`MobileIrohRuntimeComposition` cancel everything the old graph kept alive (push reachability loop, feature-flag scheduler, lifecycle observers, diagnostic tap, debug-log mirror, protected-data observation, iroh auth observation, connectivity subscriber, network-path observation), and process-global one-timers (Sentry, debug-log arming, transport reporter) moved to a once-only bootstrap. Sign-out runs under the old defaults so push unregister and iroh binding revoke hit the old backend; rebuilding re-runs `detectAuthProjectSwitch` for the clear semantics. Picker confirms via dialog, disables while switching; pinned dev builds can never start the transaction. Six new localized keys en+ja, relaunch notice removed. New tests: rebuild-over-same-defaults, release-shaped clear, auth/iroh shutdown suites, confirmation state machine.

azooz2003-bit:
Part E landed on the iOS side: switch steps park via `detachSessionLeavingTokens` (skipping the iroh sign-out preparation that would revoke the parked pairing), restore through the rebuilt root, and prompt through a held `waitForNextSignedInUser` task that the establishing banner's "Back to Production" button cancels — SignInView itself is the prompt, fully interactive under a non-blocking banner. Sign-out on staging warns at the root choke point and returns to Production restoring the parked prod session; the delete-account bypass is closed with a required interactive/direct actions value threaded through every chrome entry point. Settings gains the strict visibility tiers (full picker for gate/DEBUG, staging recovery card otherwise, hidden on production). Keychain store project-slotted; rebuild marker suppresses the launch clear for switcher rebuilds only, pinned by organic-flip-still-clears tests. Shared suites re-ran green; iOS package tests compile-verified for the simulator triple and run in CI. Strings en+ja. One dogfood watch item: the staging sign-out alert presents at the root right as the Settings sheet dismisses; if SwiftUI drops it, the fix is deferring through the root presentation coordinator.

azooz2003-bit:
Part F on iOS: the runtime override table becomes a fresh three-key table under an explicit choice (beating bakes and LocalConfig, Stack-key overrides documented as stripped), and the resolvers that read baked keys outside that table gained explicit-choice heads — the iroh broker's baked `CMUXIrohBrokerBaseURL`, the presence service's env/defaults/bake tiers, and a third presence consumer found during migration (the connectivity-invalidation URL). Device-lane builds (baked to staging) now classify as a staging LANE distinct from an explicit staging choice, unblocking that pick; Settings shows Build lane / Production / Staging on non-production lanes; sign-out interception fires only for explicit staging with the chained return targeting the lane. Rig determinism: the sim leg deletes both keys at install and `mobile-dev-launch` injects `CMUX_DEV_CLEAR_BACKEND_ENV_CHOICE` for both legs, consumed before the choice is read. Strings en+ja; shared suites green; the iOS graph compile-verifies with only the documented pre-broken files failing, identically to base.

PR #10401, opened 2026-08-19
Custom sidebars: reactive JS runtime, right-sidebar panels, agent rosters with child subagents

lawrencecchen:
A reactive JS scene runtime for custom sidebars, plus the right-sidebar extension lane built on it.

## JS sidebar runtime (`.js` lane)
`~/.config/cmux/sidebars/<name>.js` programs run ONCE in JavaScriptCore, build a retained scene graph, and bind to live data through signals; data changes re-run only the bindings that read them (per-prop scene ops onto `@Observable` nodes). No per-tick re-parse, no full-tree swaps. Hot-reloads on save, validates via `cmux sidebar validate`, falls back to last-good render on error.

Interaction quality is the point: Arc-style drag reorder (leading-edge swap rule, hysteresis, X/width nesting preview, Escape cancel, group block drag, cmd/shift multi-select drag), optimistic everything (select/close/rename/collapse/order paint the same frame; commands echo behind and self-reconcile), derived double-click with zero single-click latency, inline rename, hover-revealed accessories, marquee for overflowing titles (flicker-proof hover arm, animated return), `.rotation`/`.fade` chainables, glass surface, live `TextField` editing (`onEdit`, `autofocus:false`).

Select-burst coalescing in the action dispatch: a workspace switch costs 100-250ms of main-actor view swapping, so clicking through rows fast now runs only the newest queued select (each click still paints instantly).

## Right-sidebar custom panels
Revives the dormant `RightSidebarMode.customSidebar` seam: `cmux right-sidebar set custom <name>` renders any sidebar file as a right panel (own render worker, same data keys and `cmux(...)` actions, hot reload). Localized en/ja. Palette entry deferred until the mode gets a real shortcut action.

## Agents data + panels
`workspaces[i].agents[]` projects the agent-chat session registry (hook-driven) per workspace: kind, status with timestamps, title, surface binding. Child subagents are tracked from the parent's hook events (Claude `Task`/`Agent` tool brackets, Codex `SubagentStart/Stop`) into `agents[i].children[]` with labels and start/end times — the first time cmux records nested agents at all.

## Examples (all validated in the package test suite)
`workspaces` (the Dia-style groups sidebar), `panel-info`, `panel-sessions` (searchable, workspace/all scope), `panel-todo`, `panel-subagents` (t3code-style roster with an indented child tree and jump-to-terminal), five agent-roster variations (`agents-board|cards|timeline|focus|dense`), and `kitchen-sink` (every runtime feature on one panel, doubles as a visual regression page).

## Tests
38 CmuxSwiftRenderUI package tests (runtime, reorder math, validation of all 19 examples against a rich fixture), 53 CmuxSidebar tests (data-context projection incl. agents/children), 8 app-target `AgentChatChildRunTests`, plus updated right-sidebar persistence/CLI tests.

## Follow-ups (specced, not in this PR)
- Drag agents from the panel into terminals to view them: `docs/subagents-panel-plan.md` (UTType, `.draggable` chainable, terminal drop route, one shared open-session action).
- Palette/shortcut for the right-sidebar Custom mode.


lawrencecchen:
Round 2 (dogfood-driven): the example sidebar is now Dia-style (one-line rows, no header/clock/leading dot, hover wash, squircle selection). Runtime additions: `hoverBackground` (host-side hover state, no JS round trip), continuous corner curvature for all `cornerRadius`, `paddingHorizontal`/`paddingVertical`, and a `spacing` option on `Reorderable` that feeds both layout and slot geometry. Also two fixes found by dogfood: leading-edge swap rule with a 6pt hysteresis dead band (real-mouse jitter flip-flopped the target on the shared boundary), and a scene-prop decode bug where NSNumber(0/1) bridged to Bool so `lineLimit(1)`/`opacity(1)` were silently dropped (regression test added).

lawrencecchen:
Round 3: context menus, groups, cross-group drag. Right-click on any row/header via `.contextMenu([...])` (Button/Menu/Divider scene nodes, live-binding labels, `.destructive()`), wired to `workspace.action`, `workspace.close`, `workspace.group.action`, `workspace.group.collapse/expand` (all smoke-tested over the socket). Groups ship in the data context (`data.groups()` + `workspaces[i].group`, both app data paths). The example is now ONE flat drag surface: headers are `.fixed()` rows (not grabbable, still open gaps), so a workspace drags between groups / into a group / out to ungrouped in one gesture; the drop resolves container + anchor from the flat slot and dispatches `workspace.group.add/remove` + `workspace.reorder` (verified live). Headers are chevron-only; chevron toggles collapse (optimistic local flip synced to the app), row selects the anchor. Perf with 54 seeded workspaces + 3 groups: 1.3-2.4% idle CPU. Known gaps: dragging a group as a block (use its Move menu), no auto-scroll when dragging near the scroll edge, and Move to New Window needs window-id piping.

lawrencecchen:
Review disposition before merge: the three actionable majors are fixed in 4998adcd07 (disposeScope now detaches from its parent so churning keyed lists cannot retain unmounted row scopes; a runtime error no longer marks skipped data keys as delivered; Reorderable item keys are collected during the reconcile loop instead of a second scan). The remote-renderer glass P1 is documented as a known limitation of the opt-in out-of-process lane in docs/custom-sidebars.md rather than plumbing the surface preference through the worker protocol. The watchdog-injection and remaining per-item-serialization suggestions are deliberate: the serialized item IS the change detector for keyed rows, and the watchdog resolver is process-wide by design; both are noted for the follow-up hardening PR. Local gpt-5.6-sol review of the exact head: zero P0/P1, 67 P2 policy lints (style tier). RightSidebarChromeHeightUITests fails on the e2e runners for main and this branch identically ('Failed to activate application... Running Background', no logged-in GUI user) - runner environment, tracked separately.

lawrencecchen:
recheck

PR #10404, opened 2026-08-19
cmux-tui: give a bare session its shell at attach

### LINKED ISSUE #10391: cmux-tui: attaching to a session whose workspaces are all empty shows screenless workspaces instead of offering a shell
After the 0.10.1 startup repair (https://github.com/manaflow-ai/cmux/pull/10299, https://github.com/manaflow-ai/cmux/pull/10304) prunes every dead terminal, a session can legitimately hold only empty workspaces. Attaching then lands the user in a workspace with no screens, and the other workspaces are equally bare. Pressing the new-tab binding populates a workspace correctly (`create_terminal_in_workspace` gives an empty workspace its first screen and pane), but nothing tells the user that, and a fresh session by contrast starts with a shell.

Seen in real dogfood on the session repaired from the wedge in https://github.com/manaflow-ai/cmux/pull/10299: four workspaces, zero screens each, felt broken despite being healthy state.

Suggestion: when a client attaches and the active workspace has no screens, auto-create the default shell there (parity with fresh-session startup), or render an explicit empty-workspace placeholder naming the key that creates a tab. Empty workspaces stay first-class (GUI frontends create them deliberately via create_empty_workspace); this is only about the attach experience.

lawrencecchen:
Fixes https://github.com/manaflow-ai/cmux/issues/10391.

The startup repair from https://github.com/manaflow-ai/cmux/pull/10299 and https://github.com/manaflow-ai/cmux/pull/10304 legitimately leaves a session holding only screenless workspaces, and a bare `uvx cmux` launch then renders pure emptiness with no hint that the new-tab binding would populate it. A fresh session by contrast starts with a shell.

`Session::ensure_initial` now classifies the refreshed tree once. No workspaces creates the first workspace, unchanged. Every workspace bare creates the default shell in the active workspace (`create-terminal` by stable key, numeric id fallback), then verifies the daemon exposed the screen it created. One surviving screen anywhere leaves the layout untouched, which keeps deliberately empty workspaces created by GUI frontends intact. The provider-owned and session-unavailable gates in `ensure_initial_for_machine_ui` sit in front of this and still suppress creation entirely.

Regression test first, fix second, so CI shows red then green. The test's fake writer refuses `new-workspace`, so it also pins that attach must not bolt an extra workspace onto a session that already has four. Verified on a Blacksmith Testbox at 5a4780614c: the two `ensure_initial` tests pass, all 1360 `cmux-tui` bin tests pass, clippy clean.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10404"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789714268&installation_model_id=21920&pr_number=10404&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10404&signature=b613893c9eb286bf7391e8f19d43f987c4a336fbf8464ae0b3402365590e16f1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Attach now auto-creates the default shell in the active workspace when a session’s workspaces are all bare; previously attach rendered an empty screen until users created a tab. When revision metadata is missing or bareness cannot be proven, attach skips creation to avoid duplicates and still succeeds, leaving the session bare.

- Centralizes bootstrap logic in `initial_bootstrap`, used by `Session::ensure_initial` in `cmux-tui`.
- Local: one tree snapshot drives both the decision and the target selection to avoid races.
- Remote: re-reads `list-workspaces`; only an explicit empty `screens: []` per workspace proves bareness. If all are bare and metadata exists, send guarded `create-terminal` to the active workspace (stable key, numeric id fallback) with `expected_generation`, `expected_revision`, a unique `mutation_id`, and `origin: attach-bare-session-bootstrap`; then unconditionally refresh the tree to update the cache whether the create was executed, rejected, or skipped, and accept whichever attach created the shell.
- Still creates the first workspace via `new-workspace` when none exist; never adds a workspace otherwise; preserves layouts with any surviving screen and deliberately empty workspaces. Tests cover bare-session attach, losing the guarded-create race without failing startup, skipping unguarded creates on old daemons, and cache refresh when another attach wins between snapshots.

<sup>Written for commit 5cd30251a1d0cc1bf0c958409c21c7788bcac7de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10404?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup handling for empty and partially initialized sessions.
  * Sessions with existing workspaces now create a terminal in the active or fallback workspace instead of creating an unnecessary workspace.
  * Preserved sessions that already contain an active screen.
  * Improved verification of remotely created terminals during startup.
  * Prevented startup actions from relying on inconsistent session state.
  * Safely handled simultaneous startup attempts without creating duplicate terminals.
  * Added clearer error reporting when startup terminal creation or verification fails.

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/session/mod.rs
+++ b/cmux-tui/crates/cmux-tui/src/session/mod.rs
@@ -606,10 +606,13 @@ impl Session {
                         let workspaces = snapshot["workspaces"].as_array();
                         let bare = workspaces.is_some_and(|workspaces| {
                             !workspaces.is_empty()
+                                // An explicit empty screen array is the only
+                                // proof of bareness; a missing or malformed
+                                // field fails closed and skips the create.
                                 && workspaces.iter().all(|workspace| {
                                     workspace["screens"]
                                         .as_array()
-                                        .is_none_or(|screens| screens.is_empty())
+                                        .is_some_and(|screens| screens.is_empty())
                                 })
                         });
                         // Fail closed: without the revision metadata the
--- a/cmux-tui/crates/cmux-tui/src/session/mod.rs
+++ b/cmux-tui/crates/cmux-tui/src/session/mod.rs
@@ -295,6 +295,30 @@ fn default_true() -> bool {
     true
 }
 
+/// How attach must bootstrap a session so a bare `cmux` launch never lands on
+/// pure emptiness. A brand-new session gets its first workspace. A session
+/// whose every workspace has lost its screens (the legitimate outcome of the
+/// startup repair that prunes dead terminals) gets the default shell in the
+/// active workspace, because empty is indistinguishable from broken at
+/// attach. One surviving screen anywhere means the user's layout is intact,
+/// and includes the deliberately-empty-workspace case, so startup must not
+/// mutate anything.
+enum InitialBootstrap {
+    FirstWorkspace,
+    ShellInActiveWorkspace,
+    LayoutIntact,
+}
+
+fn initial_bootstrap(tree: &TreeView) -> InitialBootstrap {
+    if tree.workspaces.is_empty() {
+        return InitialBootstrap::FirstWorkspace;
+    }
+    if tree.workspaces.iter().all(|workspace| workspace.screens.is_empty()) {
+        return InitialBootstrap::ShellInActiveWorkspace;
+    }
+    InitialBootstrap::LayoutIntact
+}
+
 /// Attach optional cols/rows fields to a remote command.
 fn with_size(mut cmd: serde_json::Value, size: Option<(u16, u16)>) -> serde_json::Value {
     if let Some((cols, rows)) = size {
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Local review gate at 5cd30251 reports one remaining P2: the fail-closed branch returns success while leaving the session bare when list-workspaces omits generation/terminal_revision. Refuted with evidence: every released daemon since the first published version emits both fields in list-workspaces (cmux-tui-v0.9.7 server.rs sets workspaces["generation"] and workspaces["terminal_revision"] unconditionally), so no supported daemon reaches that branch; it hardens against malformed responses only. The finding's proposed alternatives regress real behavior: a compatibility error would fail an attach that works today, and an atomic server operation cannot exist on the hypothetical old daemon either. Earlier findings from the same gate (two-snapshot race, unguarded create, missing-screens bareness, stale cache on the skip path) were all fixed in commits 73232488c2..5cd30251a1.

PR #10408, opened 2026-08-19
Spike: cmux-tui daemon terminal backend behind a dev flag (tier A, deletable bridge)

lawrencecchen:
Tier-A spike from the cmux-tui GUI-frontend migration plan: prove that a cmux DEV terminal can survive quitting the app by backing it with a cmux-tui daemon terminal. **This is a deletable bridge, not a feature.** The agreed shipped data path is the native CMTH renderer inside GhosttyKit; this PR exists only to validate daemon supervision, reattach, and exit semantics ahead of that work. Deletion criterion is recorded in the migration working doc (tier-A spike section).

## What it does

Behind `terminal.beta.tuiBackend.enabled` (off by default, Settings > Beta Features > "cmux-tui Terminal Backend"):

- New plain local main-grid terminal tabs run `workspace create` against a cmux-tui daemon session named `cmux-<tag>` (spawning `server start --session cmux-<tag> --headless` if the socket under `$TMPDIR/cmux-tui-<uid>/` is absent), then the Ghostty surface command becomes `<cmux-tui-bin> attach --session cmux-<tag> --terminal <id>` instead of the shell.
- The daemon `terminal_id` persists in `SessionTerminalPanelSnapshot.tuiTerminalID`. On session restore, if the daemon socket is alive and `terminal list` still has that id, the surface reattaches; any missing link falls back to today's fresh spawn (which, with the flag on, provisions a new daemon terminal).
- Flag off: zero behavior change; every hook is gated by `TuiTerminalAttachPolicy.shouldProvisionNewTerminal` / `restoreDecision` which return the legacy path immediately.

Key files:
- `Sources/TuiTerminalAttachPolicy.swift` - pure decision logic (unit-tested)
- `Sources/TuiTerminalAttachBridge.swift` - daemon ensure + CLI calls (spike-only I/O)
- `Sources/Workspace.swift` - creation hook in `newTerminalSurfaceLocal`, snapshot capture, restore reattach
- `Sources/SessionPersistence.swift` - new optional `tuiTerminalID` snapshot field (legacy snapshots decode unchanged)

## Spike-only shortcuts (deliberate, documented)

- **Binary path is a dev setting** (`terminal.beta.tuiBackend.binaryPath`) defaulting to `/Users/lawrence/.local/bin/cmux-tui-npm`. There is no bundled cmux-tui artifact yet; embedding one is build item 1 in the migration plan.
- **Bounded socket poll** (50ms/5s cap) after spawning the daemon, and short blocking CLI calls on the main actor. Violates the no-sleep-sync code rule knowingly; the principled replacement is launchd supervision (build item 2), which this spike exists to motivate.
- **Attach-client exit is indistinguishable from shell exit** (accepted quirk from the plan).
- **CMUX_* identity env is injected into the attach client, not the daemon-side shell**, so shell integration / agent hooks inside daemon-backed terminals are incomplete (accepted quirk; env contract on `terminal create` is build item 4).
- **Scope: new main-grid terminal TABS and restored tab panels only.** The initial terminal of a brand-new workspace (workspace scaffold path), split panes (`splitPaneWithNewTerminal` and the split creation function), dock terminals, SSH/remote workspaces, and agent-command terminals keep today's local spawn path. Closing a daemon-backed tab leaves the daemon terminal running (orphan). Working directory is not propagated to the daemon shell (`workspace create` has no cwd flag in the pinned binary). All acceptable for a dev-flag spike.
- The GUI parents the daemon on first spawn; the daemon survives GUI exit (verified below) but not logout/reboot.

## Preflight evidence (tag `tuispk`)

All driven headlessly through the tagged debug socket (`/tmp/cmux-debug-tuispk.sock`), never the user's cmux; daemon session `cmux-tuispk`, untouched pre-existing sessions (`hq`, `cloud-demo`, `demo-14`).

1. Launched `cmux DEV tuispk` with `terminal.beta.tuiBackend.enabled = true` (temporary `defaults write` on the tagged bundle id, which is the flag's real settings store).
2. `tab-action --action new-terminal-right` created a new tab; debug log shows `tuiAttachSpike.daemon.start session=cmux-tuispk` then `provision.ok terminal=term_ee51ed...` (daemon cold start + provision took ~240ms). Daemon-side `terminal list` showed the terminal resized to the surface's 108x45, proving the attach client owns the PTY.
3. Sent `echo spike-alive && sleep 999` + Enter through the socket. GUI `read-screen` of the surface showed the prompt line and `spike-alive`; daemon `screen read` showed the same. Recorded `sleep` pid 61184.
4. Killed the app (`pkill` the tagged binary; the 8s autosave had already persisted `tuiTerminalID` into `session-com.cmuxterm.app.debug.tuispk.json`, verified by parsing the file). Daemon pid and `sleep` pid survived the app exit.
5. Relaunched the same app. Debug log: `tuiAttachSpike.restore.decision terminal=term_ee51ed... alive=true decision=reattach`. `read-screen` of the restored surface (same tab position) showed the original scrollback (`echo spike-alive && sleep 999` / `spike-alive`) and `ps` confirmed the same `sleep` pid 61184 still running, started before the quit.
6. The two restored panels that predated the flag had no persisted terminal id and correctly fell back to fresh spawn (which, flag-on, provisions fresh daemon terminals).

Round 2 on the pushed HEAD (includes the scrollback-replay fix): new tab -> `echo spike-round2 && sleep 888` (pid 90288) -> autosave persisted `term_19e663...` -> app killed -> pid 90288 confirmed alive -> relaunch -> `restore.decision ... decision=reattach` for all four persisted terminals -> `read-screen` showed `spike-round2` scrollback in the same tab position with no duplicated scrollback, and both `sleep 888` (pid 90288) and round-1's `sleep 999` (pid 61184) still running.

Every app instance I launched was killed afterward. The `cmux-tuispk` daemon session is deliberately left running so the deeplinked build reattaches to the demo terminals on first launch; `cmux-tui-npm server stop --session cmux-tuispk` removes it.

## Tests

`cmuxTests/TuiTerminalAttachSpikeTests.swift`: reattach-vs-fresh-spawn decision matrix, provisioning gate, session naming, attach-command quoting, CLI JSON parsing, and `SessionTerminalPanelSnapshot` round trip including legacy decode. Pure unit tests; they run in CI (not run locally per the local-test ban).

Localization audit: the new Settings strings (`settings.betaFeatures.tuiTerminalBackend*`) are in `Resources/Localizable.xcstrings` with en and ja translations; the curated settings-search entry is registered. No other user-facing strings added.

## Quit dialog (round 3)

With the flag on and the daemon owning at least one live terminal, every quit path (`applicationShouldTerminate`, plus the Cmd+Q shortcut warning path, which defers to it) asks keep-vs-stop: "Keep Sessions and Quit" (default), "Stop Sessions and Quit", "Cancel". The dialog replaces the generic quit warning, never stacks on it. Stop closes every daemon terminal first and then runs `server stop`, because `server stop` alone leaves PTY hosts adoptable by a later daemon; the close-then-stop sequence was verified empirically on a throwaway session (`terminal list` empty, no `__terminal-host` processes left). The stop runs as owned async work behind `terminateLater`. Presenting before terminate on the shortcut path also fixes a pre-existing wedge: socket-driven `simulate_shortcut cmd+q` called `NSApp.terminate` inside a main-queue drain and deadlocked `_shouldTerminate`, whose terminate reply is itself a main-queue task.

**Known bridge gap (found today, not fixed this round):** attach clients do not reconnect when the daemon restarts underneath a running app; the surfaces show dead attach processes until a full app relaunch, which then reattaches normally. Tracked for the bridge-deletion milestone rather than patched in the spike.

## Close semantics (round 4)

Two user-visible cuts fixed on daemon-backed tabs:

1. **Idle close no longer prompts.** Closing a daemon-backed tab always showed the "process is running" confirmation because the surface child is the always-running attach client. The shared `panelNeedsConfirmClose` path (tab close, pane close, workspace close, ghostty runtime close requests) now consults the DAEMON terminal's real process state via `terminal <id> process show --json`: prompt when the shell has a child process or when the root process is not a shell at all, skip the prompt for an idle shell. If the daemon cannot be queried (dead socket, missing binary, CLI error, malformed payload) the decision is "unknown" and the existing prompting behavior stands, so the daemon path can never silently skip a confirmation it cannot justify. Decisions are cached 2s per terminal so one close gesture costs one bounded CLI call.
2. **Close now closes the daemon terminal.** Tab/pane/workspace close runs `terminal <id> close` (fire-and-forget off the main actor) from the shared panel-discard path, fixing the known "close orphans the daemon terminal" cut. Two deliberate exceptions, decided by a pure policy function: detach transfers (moving the tab to another workspace/window) carry `tuiTerminalID` inside `DetachedSurfaceTransfer` so the destination keeps daemon-aware close behavior and closes the terminal when the tab finally closes; and app termination never closes terminals from panel teardown, because quit owns keep-vs-stop through its round-3 dialog. Detach-without-close therefore stays available only through quit (Keep Sessions).

Decision logic lives in `TuiTerminalAttachPolicy` as pure functions (`closeConfirmationDecision(fromProcessShowJSON:)`, `processShowArguments`, `terminalCloseArguments`, `shouldCloseDaemonTerminalOnPanelDiscard`) with unit tests covering the idle/child/non-shell/login-shell/malformed-payload matrix and the discard truth table. No new user-facing strings; the existing localized `dialog.closeTab.*` confirmation is reused, so no xcstrings changes were needed.

Remaining cuts, stated not hidden: closing the daemon terminal leaves its empty daemon-side workspace behind (`workspace close` does not end processes; cleanup needs a terminal-to-workspace round trip; daemon-side junk only, invisible to the GUI). A daemon-backed tab moved into the Dock loses the mapping (docks have no `tuiTerminalIDsByPanelId`), reverting that one tab to pre-fix behavior. `respawnTerminalSurface` on a daemon-backed tab replaces it with a plain local shell and orphans the daemon terminal (pre-existing).

This round also commits an env(1) fix found uncommitted in the spike worktree: ghostty wraps surface commands in `bash -c "exec -l <cmd>"`, where a leading `CMUX_TUI_CONFIG=...` prefix parses as the program name and reattach fails to launch; the attach command now uses `env CMUX_TUI_CONFIG=... <bin> attach ...`, with the quoting test updated to the new shape.

### Preflight evidence (round 4, tag `tuispk`, PR head build)

Driven through the tagged debug socket only; the user's two live workspaces and all 10 daemon terminals were left untouched (verified identical terminal-id set before and after).

- Idle close, GUI path: created workspace `tui-close-preflight`, `new-surface --type terminal` provisioned `term_7f3bf397...` (`provision.ok` breadcrumb), daemon `process show` reported `children: []`. `debug.shortcut.simulate cmd+w` on the focused tab closed it with NO dialog (screenshot captured), breadcrumbs `closeConfirm.decision ... required=0` then `surfaceClose.done ... ok=1`, and the terminal disappeared from daemon `terminal list`.
- Running close, GUI path: second daemon-backed tab (`term_c058ca34...`), ran `sleep 500` (child pid visible in `process show`), cmd+w showed the "Close tab?" dialog (screenshot captured; breadcrumb `required=1`); Escape canceled and the tab survived.
- Running close, socket path: `close-surface` is non-interactive BY DESIGN (`force: true`, "Socket API must be non-interactive"), so it bypasses the GUI confirmation; the policy decision for this tab was verified through the cmd+w dialog and the `required=1` breadcrumb instead. The socket close still went through the shared discard path: `surfaceClose.done ok=1`, the daemon terminal left `terminal list`, and the `sleep 500` process ended.
- Cleanup: test workspace closed, focus restored to the user's workspace, daemon terminal set byte-identical to the pre-test baseline, app left running on the PR-head build with the user's tabs reattached (`restore.decision ... decision=reattach` for all three persisted terminals).

CI on this head (run [32346908229](https://github.com/manaflow-ai/cmux/actions/runs/32346908229)): `TuiTerminalAttachSpikeTests` passed (suite log: "passed after 44.316 seconds", shard 4). The dispatch run itself is red with the same broad pre-existing unit-failure set as the previous head's run ([32328403367](https://github.com/manaflow-ai/cmux/actions/runs/32328403367)): ~140 failing tests across AppDelegateIssue2907Routing/Browser lifecycle/CLINotify integration suites, identical by name modulo a handful of known-flaky SSH/hook-install tests, none touching this round's files. That set predates this commit and is branch-level drift, not introduced here.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10408"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789717102&installation_model_id=21920&pr_number=10408&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10408&signature=a2cb7dcfad2cd7e7fc7f3a80119865f67fb9b19338fe78131c051e5fd1e05515"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds a beta `cmux-tui` daemon terminal backend so new plain local main-grid terminals survive app quit and reattach on relaunch. Previously these terminals exited with the app; with the flag on, a daemon terminal backs each new tab, preserving shell, scrollback, input, mouse, and cursor state across quit and reattach.

**Behavior**
- Quit prompts Keep Sessions (default) vs Stop Sessions when the daemon owns live terminals; Stop closes each daemon terminal before stopping the daemon.
- Closing a daemon-backed tab consults the daemon's real process state for the close confirmation, then closes the daemon terminal; transfer paths carry the terminal id so the destination keeps daemon-aware close behavior.
- Reattach persists the daemon `terminalID`, skips scrollback replay, reasserts host mouse capture on focus-in and resize, and preserves the last-set extended mouse encoding (SGR, with an SGR-preferring fallback for old daemons).
- Every bridge subprocess (surface command, daemon spawn, CLI) runs with an app-managed `CMUX_TUI_CONFIG`, never the user's config.
- Flag off is byte-for-byte the legacy path; splits, dock, SSH/remote, agent, and the initial workspace terminal are unchanged.

**Rollout and tests**
- Enable via Settings > Beta Features: `terminal.beta.tuiBackend.enabled` (default off) and point `terminal.beta.tuiBackend.binaryPath` at a `cmux-tui` binary.
- This is a throwaway bridge; the shipped data path is the native CMTH renderer.
- Unit tests cover attach policy, restore decisions, close logic, snapshot round-trip, quit flow, and env isolation.
- `TuiBridgeReattachMouseUITests` drives the full Keep-quit→relaunch→click loop in CI; CI conditionally builds `cmux-tui` and exports `TEST_RUNNER_CMUX_UI_TEST_TUI_BINARY`.

<sup>Written for commit 71830ef5bf981ae332698ed11c860243d2e9d85b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an optional beta terminal backend that can preserve shell sessions when the app is closed.
  * Added settings to enable the backend and configure its executable path.
  * Added session restoration for terminals managed by the beta backend.
  * Added quit-time options to keep or stop active backend sessions.
  * Added English and Japanese localization for the new beta setting and quit dialogs.

* **Bug Fixes**
  * Added fallback behavior to continue using standard terminals when beta backend setup is unavailable.
  * Improved handling of custom cancellation choices in quit dialogs.

* **Tests**
  * Added coverage for terminal provisioning, restoration, command generation, quit handling, and legacy session compatibility.

---

## Round 6: TuiBridgeReattachMouseUITests (CI closure of the quit→relaunch→click loop)

`103d77ab7f` adds an XCUITest that drives the whole round-6 dogfood loop in hosted `test-e2e.yml`: launch the tagged app with the bridge flag pointing at a CI-built cmux-tui, open a daemon-backed tab, run a fixture that enables mouse tracking in btop's order (`1002h,1015h,1006h`, SGR last) and execs `cat -v`, click with a real window-server event and assert the forwarded bytes echo as SGR (`^[[<0;`), quit through the **Keep Sessions and Quit** dialog (the app is frontmost under XCUITest, so the dialog holds), relaunch, click the reattached terminal, and assert the click still arrives as SGR and never urxvt (`^[[32;`). Cleanup quits through **Stop Sessions and Quit**, which also stops the per-run daemon session. `test-e2e.yml` gains a conditional step that builds cmux-tui when the filter targets `TuiBridge*`; the test resolves the binary via `GITHUB_WORKSPACE` (or `CMUX_UI_TEST_TUI_BINARY`) and skips when absent.

**Runs**: RED — the test against this branch's pre-fix cmux-tui (`103d77ab7f`) fails at the reattach assertion ("Timed out waiting for reattach click forwarded as SGR, not urxvt", with the fresh-attach SGR click visible on the captured screen), i.e. the user's bug reproduced end-to-end in CI: https://github.com/manaflow-ai/cmux/actions/runs/32374888567. GREEN — with https://github.com/manaflow-ai/cmux/pull/10428's cmux-tui merged into this branch (`20e8436fa6`, plus `bdfa13d327` passing the built binary to the UI-test runner via `TEST_RUNNER_` env, and `624b82d85f` absorbing the headless-CI launch-activation flake that failed run 32374891250 before the test body started): https://github.com/manaflow-ai/cmux/actions/runs/32377571125.

**Ordering**: land 10428 first; its commits then drop out of this PR's diff. The merge exists only so CI could prove green here.


lawrencecchen:
Dogfood round 2 root cause and fix: the dead clicks and wrong Ghostty cursor style after reopen are host-passthrough bugs in `cmux-tui attach --terminal`, fixed in https://github.com/manaflow-ai/cmux/pull/10428.

The attach client reused the full-TUI startup: unconditional mouse capture (`DECSET 1000/1002/1003/1015/1006` + `XTSHIFTESCAPE`) and a session-default cursor override (`OSC 12 #c0c1b5` + `DECSCUSR 5`) for a plain zsh that requested nothing. First attach and reattach streams are byte-identical; it only surfaced after reopen because only restored panels are daemon-backed in this spike. The fix makes scoped attach mirror the inner terminal: host capture follows the inner mouse-tracking state per frame (vim/btop still capture, including across reattach), and cursor escapes are emitted only for application-authored DECSCUSR. Full byte-level before/after and the sibling-bug audit are in the fix PR body.

New binary requirement for this spike: the bridge needs a cmux-tui build containing PR 10428. On this Mac the dev setting now points at it (temporary, dogfood-only): `terminal.beta.tuiBackend.binaryPath = ~/.local/bin/cmux-tui-fix` in `com.cmuxterm.app.debug.tuispk`, and daemon `cmux-tuispk` is running that binary (restored terminals survived the swap). The npm default (`cmux-tui-npm`, 0.10.1) still has both bugs; bump the published binary or land 10428 before widening the dogfood.

Known bridge trade-off introduced by correct passthrough: wheel scrollback in plain-shell daemon tabs no longer works (host alt screen has no scrollback; keyboard scrollback works; the native CMTH renderer is the durable fix).

lawrencecchen:
Round 4 on the bridge: the reattached-btop dead-clicks repro is addressed on https://github.com/manaflow-ai/cmux/pull/10428 (commits 159a447223 red, 517b1fd510 green). Root-cause class: the scoped client derived host mouse capture from the rendered-frame projection (`rendered_terminal_pointer_semantics`), which `draw_content` clears at the start of every draw and restores only after a successful render, so a frame with a failed or skipped pane render wrote capture-off to the Ghostty pty while the inner app still owned the mouse. Capture now follows the mirror terminal's canonical mode state (replay-restored, live-updated) with keep-last-applied when the state is momentarily unknowable.

Two byte-level findings relevant to the spike itself, no code change here:

- The daemon side was clean the whole time: terminal-host replays carry all non-default modes (DECSET 1002/1006/1015 observed in the live btop terminal's state read), and replay application restores them to the client mirror.
- pty taps on the real GUI relaunch confirmed reattach asserts capture for mouse-mode tabs and asserts nothing (no mouse modes, no OSC 12, no DECSCUSR) for plain-shell tabs, so the round-2 passthrough contract holds at the byte level in the app, not just in harnesses.

lawrencecchen:
The repeated "could not checkpoint terminal ... reconnect: session changed during checkpoint capture" toasts on daemon-backed terminals are addressed in two parts. https://github.com/manaflow-ai/cmux/pull/10484 (merged) stopped the destructive half: a failed reconnect checkpoint no longer disconnects the healthy host and re-runs the reconnect, whose journal writes had been re-poisoning every other terminal's capture window. https://github.com/manaflow-ai/cmux/pull/10501 fixes the capture layer itself: the consistency fence in journal_checkpoint.rs compared a journal head and a session snapshot read under separate lock holds, so any concurrent journal write between them spuriously aborted the checkpoint; capture now reads both under one projection lock as a single cut, and any residual skip reports one status event total (further skips go to the daemon log) instead of one toast per terminal per reconnect.

lawrencecchen:
Closing this obsolete Tier-A spike. Its own description marks it as a deletable bridge, and the branch is conflicting. The surviving implementation work is tracked in the descendant cloud/manual-IO stack [#11062](https://github.com/manaflow-ai/cmux/pull/11062), while the local daemon pieces need a fresh, selective rebase. No code from this conflicting spike is being merged.

lawrencecchen:
Superseded by the current selective relay stack; see the preceding evidence comment.

PR #10428, opened 2026-08-19
cmux-tui: scoped attach becomes a transparent host passthrough (mouse capture, cursor style)

lawrencecchen:
Dogfood of the tier-A bridge spike (#10408) found two host-state bugs in `cmux-tui attach --session <s> --terminal <id>`, the command the GUI runs inside each Ghostty surface: clicks in the terminal did nothing, and the user's Ghostty cursor style was replaced.

## Root cause

`attach --terminal` reused the full-TUI client startup: `app.rs` executed `EnableMouseCapture` unconditionally, and `ui/pane.rs` mirrored the resolved cursor visual, which bakes session defaults in. Outer Ghostty forwarded every click to the attach client instead of doing native selection, and the host cursor got the daemon session's default style. First-attach and reattach byte streams are identical, so both bugs existed from the first attach; the reporter only noticed after reopening because only restored panels are daemon-backed in the spike.

## Byte-level evidence (pty harness, plain zsh inner terminal that requested nothing)

Before (cmux@0.10.1), startup: `DECSET 1049, 1000, 1002, 1003, 1015, 1006, 1004, 2004`, `XTSHIFTESCAPE >1s`, kitty push, `OSC 12 #c0c1b5` + `DECSCUSR 5`. Mid-session, inner `DECSET 1002h/l` produced no host transitions (capture already always on), and inner `DECSCUSR 0` re-asserted the session default (`OSC 12` + `DECSCUSR 5`) instead of restoring the host default.

After (this branch, commit 6a7a843c63), startup and reattach: `DECSET 1049, 1004, 2004`, kitty push, cursor-visibility toggles only. No mouse tracking, no `XTSHIFTESCAPE`, no `OSC 12`, no `DECSCUSR`. Mid-session mirroring:

```
inner \e[?1002h  ->  host DECSET 1000,1002,1003,1015,1006 + XTSHIFTESC >1s
inner \e[?1002l  ->  host XTSHIFTESC >0s + DECRST 1006,1015,1003,1002,1000
inner \e[6 q     ->  host OSC 12 + DECSCUSR 6
inner \e[0 q     ->  host OSC 112 + DECSCUSR 0   (restores the host default)
```

Reattach with inner mouse still enabled re-enables host capture right after the daemon replay lands, so vim/btop keep mouse across detach/reattach.

## Fix (two commits, red then green)

- Scoped startup asserts no mouse tracking and no shift-bypass report. After each frame the client mirrors the inner terminal's mouse-tracking state to the host, so plain shells give the host native clicks and selection while mouse apps still capture.
- Host cursor escapes are emitted only when the inner application authored a cursor style, tracked by a new streaming `CursorStyleProvenance` detector fed exclusively by raw inner-PTY output bytes; daemon-built replays reset it because they carry resolved state with defaults baked in. A scoped client starts from an applied `Reset` and writes no cursor escapes until an application authors one.

Deliberately kept host-enabled in scoped mode, with rationale in code: alternate screen and raw mode (needed to render), kitty keyboard negotiation (keys are re-encoded per the inner terminal's actual protocol), focus reporting, and bracketed paste (the client consumes these and re-encodes paste per the inner terminal's `2004` state). Full-TUI behavior is unchanged.

## Live dogfood preflight (tag tuispk)

Daemon `cmux-tuispk` swapped to this branch's binary; app quit and relaunched twice. Restored panels reattached by terminal id with inner state alive (a btop with 2 days uptime, a background `sleep` marker). Daemon-side vt state after reattach: the btop terminal carries `DECSET 1002/1006/1015` (capture mirrored on for that tab), the plain-shell terminal carries no mouse modes and no DECSCUSR (host owns clicks; host cursor untouched). No orphaned attach processes after either quit.

## Audit of sibling bugs in the same class

| Item | Status | Evidence |
| --- | --- | --- |
| Unconditional host mouse capture (dead clicks) | fixed | before/after startup captures above |
| Session-default DECSCUSR + OSC 12 override | fixed | before/after cursor phases above |
| Reattach replay restores inner-requested mouse | fixed, tested | `daemon_replay_restores_inner_mouse_tracking_to_the_mirror`; reattach capture re-enables capture post-replay |
| Bracketed paste | correct by design | host 2004 stays on; client rewraps per inner mode 2004 before writing to the PTY |
| Kitty keyboard protocol | correct by design | host push retained; keys re-encoded per inner terminal's negotiated state (`KeyEncoder::sync_from_terminal`) |
| Focus reporting (1004) | pre-existing gap, deferred | client consumes FocusGained/Lost and never forwards them to inner apps that requested 1004 (both modes) |
| OSC 52 from inner app | pre-existing gap, deferred | only the client's own selection copy emits OSC 52; inner-app OSC 52 is not propagated to the host clipboard |
| Title propagation | not needed for bridge | host title writes only via the daemon SetWindowTitle command; the GUI renders its own tab titles |
| Resize mismatch on reattach | correct | vt-state replay applies daemon cols/rows, then the client asserts its viewport through the reservation/barrier resize path |
| Double scrollback | n/a | client runs on the alt screen; host scrollback never accumulates |
| Orphaned attach processes | none observed | 0 attach clients after quit, fresh set after relaunch, two cycles |

Known trade-off: with capture off, wheel input in a plain-shell bridge tab reaches a host alt screen with no scrollback, so wheel scrollback browsing is lost for non-mouse inner apps (keyboard scrollback still works). The shipped data path is the native CMTH renderer, which restores host-native scrollback.

Hosted verification green on this head: tests plus artifact build (https://github.com/manaflow-ai/cmux/actions/runs/32251772365).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal mouse-capture behavior for attached sessions, including smoother transitions when applications change mouse-tracking modes.
  - Preserved focus reporting and bracketed paste support while preventing unnecessary host mouse capture at startup.
  - Improved cursor handling so application-defined cursor styles are preserved and displayed only when appropriate.
  - Added more reliable restoration of mouse and cursor state across terminal sessions, resizing, and replayed output.

## Round 4: reattached btop tab loses clicks (capture derived from the rendered-frame projection)

User repro: btop in a daemon-backed tab has working clicks and drag-to-scroll; quit with Keep Sessions and Quit; reopen; clicks and drag are dead in the reattached tab.

Byte-level investigation first, against a throwaway daemon session and against the live tuispk app with a pty tap wrapped around the attach binary:

- The daemon replay is not the culprit. The terminal host's real attach replay (`vt_replay_bounded_theme_portable_with_aliases`) emits every non-default mode: the live btop terminal's `state read` carried `DECSET 1002/1006/1015`, and a mirror applying that replay reports `mouse_tracking = true`.
- The round-2 client re-asserted capture in every harness combination (replay-only modes, live DECSET after a replay, mid-session resize), and tap captures from a real GUI relaunch show both the btop tab and a fresh btop tab asserting `DECSET 1000/1002/1003/1015/1006 + XTSHIFTESCAPE >1s` on reattach while plain-shell tabs assert nothing and write no cursor escapes.

Why the round-2 reattach test lied: coverage was split. One test proved the daemon replay restores mouse modes to the mirror; another proved capture follows `rendered_terminal_pointer_semantics`. The link between them exists only through a successful pane render. `draw_content` (ui/pane.rs) removes the surface's entry from that map at the start of every draw and re-inserts it only after a successful nonzero-rect `render_frame`; any frame that skips or fails the scoped pane's render (renderer error while the network thread applies daemon output, zero-sized pane during restore layout churn, surface lookup miss during attach handoff) ends with the map empty, and the end-of-frame sync in `draw_terminal` writes capture-off to the host while the inner application still holds the mouse. A reattach with btop running lands replay application, resize, and full-screen redraws in the same frames, exactly that storm.

Fix (two commits, red then green): `desired_host_mouse_capture` now reads the inner terminal's canonical mode state straight from the session surface's pointer-semantics probe, the same state the daemon replay restores and live output updates, so render success is no longer load-bearing for mouse ownership. When the state is momentarily unknowable (terminal lock contended, surface gone during teardown) the client keeps the capture it last applied instead of toggling the host. The round-2 cursor rule is untouched: host cursor escapes still require application-authored DECSCUSR recovered from raw output bytes, and daemon replays still reset that provenance (0 OSC 12 and 0 DECSCUSR writes across every GUI tap, btop and plain-shell alike).

New tests pin the previously untested level: `scoped_host_mouse_capture_follows_canonical_state_without_a_rendered_frame` and `scoped_host_mouse_capture_keeps_last_applied_when_state_is_unknowable` (both red before the fix), plus `daemon_theme_portable_replay_restores_mouse_tracking_to_the_attach_probe`, which upgrades the replay-path test from a hand-written DECSET to the daemon's real replay builder output observed through the probe the App reads.

Honest caveat: the reported dead-click state could not be reproduced after the fact; the failing session's host-side terminal state died with the app process, and taps installed afterwards show correct assertions. This fix removes the only mechanism found in the client that can write capture-off while the inner app holds the mouse, and recovery no longer depends on a later successful render.

Client-only change: the daemon and terminal hosts run unmodified code paths, so the live daemon was not restarted for this round.

Hosted verification green on 517b1fd510: https://github.com/manaflow-ai/cmux/actions/runs/32336286736 (first attempt failed only the known-flaky Linux smoke step, the OSC 11 probe race in scripts/smoke-tui.py whose own comment notes it raced before; every Rust suite on Linux and macOS was green in the first attempt; the rerun of that job passed)

---

## Round 6: reattach clicks were re-encoded as urxvt, not SGR (byte-proven, fixed)

Rounds 2-5 fixed real capture-assertion bugs, but capture was never this failure. Tap evidence from the user's live app: Ghostty delivered an SGR press (`Down(Left) (35,20)`, admission=true) and the client enqueued `\x1b[32;36;21M` — urxvt — to the inner PTY. btop parses only SGR, so the click vanished.

Root cause: the extended-coordinate DEC modes (1005/1006/1015/1016) are one **last-set-wins selector**, not independent flags. btop enables `1002h, 1015h, 1006h` (SGR last). The daemon replay serializes the mode flags in numeric order (`1002, 1006, 1015` — proven at offsets 5534/5542/5550 of `terminal state read`), so a mirror rebuilt from replay ends with urxvt active, and the client re-encodes every forwarded mouse event with it. Live-attached clients watch the DECSETs in true order and stay SGR; every reattach flips to urxvt.

Fix (red `a480c119b1`, green `1a2e2da5ef`), three layers keyed on Ghostty's own parsed state:

- ghostty-vt tracks the ACTIVE wire format per terminal (`MouseWireFormat`) by probing Ghostty's mouse encoder whenever the mouse-mode revision refreshes, and exposes it on `TerminalPointerSemanticSnapshot`. Never guessed from the boolean flags.
- Replay serialization (all daemon attach/resize/checkpoint paths) appends a correction suffix whenever the numeric dump would activate the wrong format: reset the non-active format flags, re-assert the active selector **last** — replay reproduces the semantic, not the flag dump.
- The attach client's mirror normalizes replays from daemons that predate the suffix: both 1006 and 1015 flagged after a daemon replay proves the dump ordering ran, so re-assert SGR (every known app that sets both prefers SGR). Fixed daemons never leave both flagged, so a deliberate urxvt-last choice survives reattach (pinned by test).

Press, release, wheel, and motion all encode through encoders synced from the same mirror state, so every forwarded-event path inherits the fix.

Red tests fail with byte-for-byte the tap's failure (`\x1b[32;37;21M` vs `\x1b[<0;37;21M`): a ghostty-vt replay round-trip (SGR-last must stay SGR; urxvt-last must stay urxvt), the remote surface against the daemon's real theme-portable replay, and against a legacy numeric flag-dump replay (old-daemon compatibility).

Headless end-to-end preflight (Python pty harness, real daemon + real attach client, SGR click written to the client's host pty): fixed client vs an **old fix2 daemon** → inner `cat -v` shows `^[[<0;5;5M^[[<0;5;5m` (SGR; was urxvt); fixed client vs a **fixed daemon** → SGR as well. So the fixed client works without a daemon restart.

`d4a7a4cd9e` fixes the `-D warnings` clippy gate (round-5 tap tool `if_same_then_else`, one test `redundant_clone`); no behavior change.

User-requested CI closure of the human loop (quit → relaunch → click) lives on the spike branch: `TuiBridgeReattachMouseUITests` on https://github.com/manaflow-ai/cmux/pull/10408 — green requires this PR's cmux-tui, land this first.

---

**Round 8 (2026-08-21): current `main` merged in** (merge `3d16fbe8bc`, fmt `ad32cdd5ce`). The branch base was missing fixes the user hits daily; this merge brings in the checkpoint-fence family (#10484, #10501 incl. `5e798b82eb` one-cut fence + once-per-condition skip reporting), `COLORTERM=truecolor` (#10429), `TERM=xterm-ghostty` pass-through (#10465), client-focus-v1 (#10331), and machine resilience (#10512). One textual conflict (the `app.rs` test-module import list), resolved as the union of both sides. Verified on the merged head: ghostty-vt 74/74, cmux-tui-core 1100 green, cmux-tui 1407 green (two known timing-flaky pointer tests failed only under full-suite load on a busy Mac, 20/20 green isolated, same family as the round-7 gate forensics); hosted full gate dispatched on `ad32cdd5ce`.


lawrencecchen:
Closing: main absorbed this branch's full content via https://github.com/manaflow-ai/cmux/pull/10603 (chatmux-relay Rust rewrite). The diff between this branch and main restricted to every file this PR touches is empty, so there is nothing left to merge.

PR #10488, opened 2026-08-20
iOS: preserve workspace-list scroll position across workspace enter/exit

### LINKED ISSUE #10481: iOS: exiting a workspace scrolls the main menu back to the top instead of preserving scroll position
## Bug

On the iOS app, when you exit a workspace and return to the main menu (the workspace list), the list scrolls all the way back to the top. It should preserve where you were scrolled so you can keep browsing downward from the same spot.

## Repro

1. On iOS, scroll partway down the main menu / workspace list (enough that the top rows are offscreen).
2. Open a workspace.
3. Exit the workspace back to the main menu.

**Actual:** the list is reset to the very top.
**Expected:** the list is exactly where it was before entering the workspace, so you can continue scrolling down from that point.

## Notes

- Likely the list view's scroll state is being lost on navigation (view rebuilt on return, data source refresh resetting the offset, or an explicit scroll-to-top somewhere in the enter/exit path) rather than the position being saved and restored.
- The fix should preserve scroll position across the workspace enter/exit round trip, including when list content refreshed while the workspace was open.

austinywang:
Closes #10481

## Bug

On iPhone, exiting a workspace back to the main menu (the workspace list) always snapped the list back to the very top instead of staying where the user had scrolled.

## Root cause

`WorkspaceListView.workspaceListWithToolbar` branched the **entire list subtree** on `showsNavigationToolbar`:

```swift
if showsNavigationToolbar {
    content.toolbar { ...items... }
} else {
    content
}
```

On the compact stack, `showsNavigationToolbar` is `compactNavigationPath.isEmpty`, so it flips **off** when a workspace is pushed and **on** again when the user exits. A ViewBuilder `if/else` is `_ConditionalContent`; switching branches changes the structural identity of everything inside, so SwiftUI dismantled the `WorkspaceListTable` `UIViewControllerRepresentable` and built a fresh `UITableView` — at content offset zero — on every workspace enter/exit round trip.

## Fix

Keep one stable view tree: always apply `.toolbar` and move the visibility condition **inside** the `ToolbarContentBuilder` (the same pattern `WorkspaceShellView` already uses for its root toolbar). The toolbar items still only render while the list is topmost, but the list's identity — and therefore the table and its scroll position — survives the round trip. Content refreshes that arrive while a workspace is open flow through the existing `WorkspaceListTableCoordinator` diff path, which preserves the offset (anchoring behavior is unchanged and already covered by the apply-route tests).

## Regression test (red → green in CI)

Commit 1 adds `WorkspaceListScrollPreservationTests` only; commit 2 adds the fix. The test hosts the real `WorkspaceListView` at the root of a `NavigationStack` and flips `showsNavigationToolbar` exactly as the shell does on enter/exit:

- `workspaceTableSurvivesNavigationToolbarRoundTrip` — same `WorkspaceListUITableView` instance and content offset after the flip round trip.
- `workspaceTableKeepsOffsetWhenContentRefreshesWhileCovered` — rows appended while the workspace is "open" still land through the coordinator without losing the offset.

Dispatched `test-ios.yml` runs on both SHAs to demonstrate red on the test-only commit and green on the fix (links in comments once complete).

## Localization audit

No user-facing strings added or changed: the toolbar items are byte-for-byte the same views, only their position in the builder moved; the new file is test-only.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10488"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789795522&installation_model_id=21920&pr_number=10488&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10488&signature=175216330f24e92e824c70b0417e94eefe3ea3e7a86b2810c64e6f566fe12e88"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Preserves the workspace list scroll position on iPhone when returning from a workspace. Previously the list reset to the top because the entire list subtree was branched on `showsNavigationToolbar`, changing view identity; now `.toolbar` is always applied and items are gated inside the builder, so the table instance and offset survive the round trip.

- Moves the `showsNavigationToolbar` check inside `.toolbar {}` in `WorkspaceListView+Toolbar.swift`; toolbar items still render only when the list is topmost, but the list’s structural identity stays stable.
- Adds `WorkspaceListScrollPreservationTests` that host the real list in a `NavigationStack` and assert the same `WorkspaceListUITableView` instance and content offset after a toolbar-visibility round trip, including during a content refresh while covered.

<sup>Written for commit a52ec57fa4d273cc4d20b594e4b817e4e406ac03. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Packages/iOS/CmuxMobileShellUI/Sources/CmuxMobileShellUI/WorkspaceListView+Toolbar.swift
+++ b/Packages/iOS/CmuxMobileShellUI/Sources/CmuxMobileShellUI/WorkspaceListView+Toolbar.swift
@@ -58,9 +58,15 @@ extension WorkspaceListView {
         filterMachines: [WorkspaceFilterMachine]
     ) -> some View {
         #if os(iOS)
-            if showsNavigationToolbar {
-                content
-                    .toolbar {
+            // The toolbar-visibility flip (off while a workspace is pushed on
+            // the compact stack, back on at exit) must stay inside the toolbar
+            // content builder. Branching the whole subtree on it changes the
+            // list's structural identity on every workspace enter/exit, which
+            // dismantles the represented workspace table and resets its scroll
+            // position to the top (issue #10481).
+            content
+                .toolbar {
+                    if showsNavigationToolbar {
                         if !usesExternalSharedToolbar {
                             ToolbarItem(id: "workspace-list-settings", placement: .topBarLeading) {
                                 settingsMenu
@@ -89,9 +95,7 @@ extension WorkspaceListView {
                             }
                         }
                     }
-            } else {
-                content
-            }
+                }
         #else
             content
                 .toolbar {
--- a/Packages/iOS/CmuxMobileShellUI/Tests/CmuxMobileShellUITests/WorkspaceListScrollPreservationTests.swift
+++ b/Packages/iOS/CmuxMobileShellUI/Tests/CmuxMobileShellUITests/WorkspaceListScrollPreservationTests.swift
@@ -1,0 +1,150 @@
+#if os(iOS)
+import CmuxMobileShellModel
+import SwiftUI
+import Testing
+import UIKit
+@testable import CmuxMobileShellUI
+
+/// Entering a workspace on the compact stack flips the list's
+/// `showsNavigationToolbar` off, and exiting flips it back on. The list's
+/// structural identity must not depend on that flag: an identity change
+/// dismantles the represented workspace table, so the freshly created table
+/// comes back scrolled to the top (issue #10481).
+@MainActor
+@Suite struct WorkspaceListScrollPreservationTests {
+    @Test func workspaceTableSurvivesNavigationToolbarRoundTrip() async throws {
+        let fixture = Fixture(workspaceCount: 40)
+        defer { fixture.tearDown() }
+        await fixture.render(showsNavigationToolbar: true)
+
+        let table = try #require(fixture.currentTable())
+        let offset = CGPoint(x: 0, y: 240)
+        table.setContentOffset(offset, animated: false)
+
+        // Enter a workspace (list covered, toolbar hidden), then exit back.
+        await fixture.render(showsNavigationToolbar: false)
+        await fixture.render(showsNavigationToolbar: true)
+
+        let tableAfter = try #require(fixture.currentTable())
+        #expect(
+            tableAfter === table,
+            "Toolbar visibility must not change the list's structural identity; a rebuilt table loses the scroll position."
+        )
+        #expect(tableAfter.contentOffset == offset)
+    }
+
+    @Test func workspaceTableKeepsOffsetWhenContentRefreshesWhileCovered() async throws {
+        let fixture = Fixture(workspaceCount: 40)
+        defer { fixture.tearDown() }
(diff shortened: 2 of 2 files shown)


austinywang:
## Red/green verification (iOS Simulator)

The shared `test-ios.yml` lane is currently red on `main` itself (pre-existing: `cmuxFeatureTests` fails to build — `MobileIrohCooldownBroker`/`MobileIrohRevocationBroker` fakes no longer conform to `CmxIrohClientBrokerServing`/`CmxIrohBindingRevoking` — plus a pre-existing `lint-ios-package-conventions` failure; every `main` dispatch of that workflow has failed for days, e.g. runs 32302940492 / 32296619121). That masks any per-PR signal there, so the regression suite was run directly against the `CmuxMobileShellUI` package (whose graph does not include the broken target) on a fleet Mac (`cmux7s-mac-mini`, Xcode 26.3), on a dedicated freshly created simulator (`cmux-verify-10481`, explicit UDID `89106B2D-9D2F-4F73-99FA-8CF377E0A7B8`), via `xcodebuild test -scheme CmuxMobileShellUI -only-testing:CmuxMobileShellUITests/WorkspaceListScrollPreservationTests`.

**Red — commit 159c93e3d2 (test only, no fix):** `** TEST FAILED **` (exit 65), deterministically across two runs. The test process actually **segfaults** (`EXC_BAD_ACCESS` / `KERN_INVALID_ADDRESS`, SIGSEGV) while the `showsNavigationToolbar` flip dismantles and rebuilds the represented table — the identity churn is not just a scroll reset, it's a use-after-free-class hazard in the dismantle path.

**Green — commit a52ec57fa4 (with fix):**
```
✔ Test workspaceTableSurvivesNavigationToolbarRoundTrip() passed after 0.510 seconds.
✔ Test workspaceTableKeepsOffsetWhenContentRefreshesWhileCovered() passed after 0.531 seconds.
✔ Test run with 2 tests in 1 suite passed after 0.533 seconds.
```

The simulator was deleted and the fleet lease released after the runs.

PR #10407, opened 2026-08-19
cmux-tui: live loading states for machine opens (connection-progress-v1)

lawrencecchen:
Stacked on https://github.com/manaflow-ai/cmux/pull/10386. The connect interstitial showed one static "connecting" for every machine open; a resumed (paused) machine looked identical to a cold connect, and providers had no way to say what was happening.

**Generalized loading states.** The interstitial's status line now has three sources, in precedence order: a hard failure ("unavailable"), the provider's live progress message, and a status-aware default (a sleeping or stopped machine shows "waking" instead of "connecting"; localized en+ja).

**`connection-progress-v1`.** New capability-gated push event `connection_progress {machine_id, message}` in the machine-provider protocol. Gated because v1 event decoders reject unknown event names; the client offers it in `negotiate_client_capabilities` next to `provider-action-targets-v1`. The update pump forwards it as a lightweight `MachineUpdate` without a snapshot reload - the provider's serialized control loop is busy inside `open_machine` exactly while it pushes these, so a snapshot request would stall the very message it is meant to show. Progress is presentation-only per machine: newest wins, carried across snapshot rebuilds mid-connect, and cleared when the switch settles, fails, or is re-aimed.

The Freestyle provider (cmuxterm-hq `workspaces/cmux-cloud-beta`) now negotiates the capability and pushes "resuming the machine", "waiting for the machine agent" (pty transport), and "waiting for sshd" (ssh transport).

Tests: wire-document round trip for the new event, app-level precedence test (`machine_transition_shows_progress_then_status_aware_default`), negotiation test updated. Verified on a Blacksmith testbox: fmt, clippy `-D warnings`, full suite green.

---
[...]

PR #10386, opened 2026-08-19
cmux-tui: instant warm machine switching and a bounded client log

lawrencecchen:
Two client-side quality-of-life features for the machine rail, stacked on https://github.com/manaflow-ai/cmux/pull/10331.

**Instant warm machine switching.** The connection hub already kept every machine connection alive and unbounded. It now keeps the last N most recently used connections (default 5, `CMUX_TUI_WARM_MACHINES`, floor 2 so the currently presented connection can never be evicted mid-switch) and evicts older ones back to Disconnected with their connector intact. `connect_tracked` reports whether a connection was reused; a reused session skips the managed-workspace guard and default-colors round-trips it already ran when first opened, and a switch to a hub-Ready machine keeps painting the current machine instead of blanking the content area with a connecting interstitial or flashing the rail badge. Warm switches settle in one tree round-trip. New test: `connection_hub_evicts_least_recently_used_beyond_the_warm_limit`; the existing lease-retention test is unchanged.

**Bounded rolling client log.** Red bottom-bar status messages, durable provider notices, and toasts now persist to `client.log` at the cmux-tui state root (`CMUX_TUI_LOG_FILE` override; rolls to `client.log.1` at 2 MiB, two files kept, so at most ~4 MiB on disk). Client `eprintln!` diagnostics mirror into it via a `stderr_log!` macro, and while the TUI owns the terminal fd 2 is routed into the log, so panics and stray library writes stop corrupting the raw-mode screen and are kept for diagnosis. The original stderr is saved and restored when the terminal is handed back. Docs: `configuration.md` (client log), `machines.md` (warm pool).

Residual risks: a reused session keeps the default colors from its first open (a mid-run theme change applies on the next cold connect), and input during the ~one-round-trip warm switch window still goes to the previous machine because its content stays painted.

Verified on a Blacksmith testbox: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, full `cargo test` suite green.

---
[...]

PR #10512, opened 2026-08-20
cmux-tui: machine resilience - bounded warm pool, live open progress, pause-aware wake

lawrencecchen:
The remainder of the machine-rail stack (#10386/#10407), rebuilt on main after its generic fixes landed separately (#10483-#10487, #10504).

**Warm pool.** The connection hub kept every machine connection alive, unbounded, and handed dead connections back to switches. It now keeps the last N most recently used (default 5, `CMUX_TUI_WARM_MACHINES`, floor 2 so the presented connection is never evicted mid-switch), validates liveness on reuse (a dead Ready slot reconnects fresh; its lease cleans up), reports dead slots honestly in phases/readiness, and a reused session skips the preparation round-trips it already ran. Warm switches settle in one round-trip with no interstitial flash - measured 0.25-0.31s click-to-prompt vs 0.7-1.1s cold in the field.

**Live open progress.** New capability-gated `connection_progress` push event (`connection-progress-v1`): providers narrate opens ('resuming the machine', 'waiting for sshd'). The update pump forwards it as a lightweight update without a snapshot reload, because the provider's serialized control loop is busy inside `open_machine` exactly while it pushes. The interstitial layers failure > sleeping hint > live progress > status-aware default ('waking' for sleeping/stopped machines; localized en+ja).

**Pause-aware wake.** Stream death for a sleeping/stopped machine no longer auto-reconnects (that force-resumed deliberate pauses); it presents as asleep with a localized 'sleeping — press any key to wake' hint, and the first keystroke wakes it through the normal switch. A running machine that lost its stream still reconnects automatically.

Tests: hub LRU eviction, wire-document round trip for the event, interstitial precedence, sleep-no-auto-resume + wake-on-key. All field-verified on the Freestyle dogfood (pause -> sleeping hint -> keystroke -> resume with live progress). Supersedes #10386 and #10407.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10512"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789861918&installation_model_id=21920&pr_number=10512&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10512&signature=957aed978cee09213f0657257a943093c4a57009d66390ae1c2eb75f2417d55a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Improves machine switching resilience and UX: keeps a bounded MRU warm pool, renders provider‑pushed open progress, and treats paused machines as deliberate until a key wakes them. Before: every connection stayed warm, transitions hid context, and paused machines auto‑resumed; now only recent connections stay warm, interstitials render when needed, sleeping machines wake on input, keys never reach the old machine during warm switches, and stale progress clears.

- Warm pool: last N MRU connections stay warm (default 5 via CMUX_TUI_WARM_MACHINES, min 2); never evict the presented connection (including the initial open); validate liveness on reuse and reconnect dead slots; readiness/phases report dead warm slots honestly; reused sessions skip only default-colors and always re‑run the managed‑workspace guard; evicted sessions shut down promptly.
- Live open progress: negotiate `connection-progress-v1`; providers may push `connection_progress` during `open_machine`; client forwards without snapshot reload, coalesces latest per machine via a shared cell (registry hard‑capped at 128 ids, pruned to the snapshot, drops new ids at the cap), and rate‑limits to ≤1 update/machine/50ms; strips control chars and caps provider text at 512 bytes; applies only while Connecting and flips a stale Ready target to Connecting when progress arrives; clears on failure/settle/sleep; pump stops on forward failure.
- Pause‑aware wake and transitions: sleeping/stopped machines don’t auto‑reconnect on stream end; UI shows “sleeping — press any key to wake,” and the first key queues a normal switch; reselecting a sleeping presented machine renders Connecting end‑to‑end; warm‑target interstitial suppression now also requires a live presented session; keystrokes during any in‑flight switch are gated and never delivered to the old machine.
- Protocol/spec/CI: older providers require no changes; compare `OpaqueId` by string; register `connection_progress` in the spec inventory; run vendored crossterm tests single‑threaded; extend smoke test timeouts for slower CI runners.

<sup>Written for commit d635b05d24a71c223935ba823a4b29fd734b85b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10512?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added live connection-progress messages while machines reconnect or wake.
  * Recently used machine connections remain warm for faster switching.
  * Added sleeping and waking states, with paused machines resuming on keyboard input.
  * Added configurable limits for retained warm connections, with five retained by default.

* **Bug Fixes**
  * Dead retained connections now reconnect automatically.
  * Progress messages clear after failures or completed transitions.
  * Paused machines no longer resume unexpectedly when disconnected.

* **Documentation**
  * Documented warm connection behavior and configuration options.


lawrencecchen:
Review-triage record for the local autoreview loop (14 rounds):

Fixed across rounds: spec registration and prose for `connection_progress`, Ready-phase masking on reselect of an asleep machine, interstitial suppression and input re-aim during switches away from a dead machine, provider event-loop exit on send failure, presented-connection eviction protection (including the initial open and the read-under-lock race), per-machine latest-value coalescing for progress with a strict cap on distinct ids, settle-gated progress application, a 512-byte cap on provider progress text, and stale-progress clearing when a machine goes to sleep.

Refuted with evidence and not actioned: the repeated "reused connection skips preparation after a failed prepare" finding — every controller's abort_replacement removes the failed candidate from the hub (StaticMachineController, the local overlay, and ProviderMachineRuntime all do), and the worker calls abort_replacement on every prepare error, so a poisoned warm entry cannot survive; the only skip is candidate == active, which was prepared when it first became active.

Known accepted limitation: a pooled connection that dies silently (no stream-death event reaches the app) can leave a stale Ready phase, so a switch to it reconnects without an interstitial for that one open. The hub revalidates liveness and reconnects correctly; the cost is cosmetic and bounded, and fixing it requires plumbing hub liveness into the render layer, out of scope here.

lawrencecchen:
Two further autoreview findings accepted as bounded cosmetic limitations rather than actioned, after 19 local review rounds:

Stale-Ready warm switch (raised rounds 14/16/19): a pooled session that dies silently right before a switch shows no interstitial for that one reconnect. It requires missing BOTH the stream-death event path and the 15s provider watchdog, the hub still reconnects correctly, input is consumed safely, and any provider open narration now flips the phase to Connecting and surfaces the reconnect. Fixing the remainder means plumbing hub liveness into the render layer.

Progress rate floor (raised round 17, counter-raised round 19): per-machine progress delivers at most every 50ms; two stages inside one 50ms window show the earlier one until the next event or the settle. Normal providers emit stages seconds apart; the floor exists because round 17 flagged unbounded redraw work from event-rate-controlled providers. Latest-wins and rate-bounding are in direct tension; the floor errs toward bounded work for pathological providers at a cosmetic, settle-cleared cost.

PR #10513, opened 2026-08-21
cmux-tui: detect, journal, and announce unexpected terminal host death

lawrencecchen:
Round 7 on the daemon-backed bridge tabs, stacked on #10428 (base is `feat-tui-attach-passthrough` so 10428 stays merge-ready).

Incident (2026-08-20, session cmux-tuispk, terminal `term_a07b3cc0`, btop): the `__terminal-host` family went silent at 09:44:52 and the GUI bridge tab kept the last frame for six hours, accepting mouse input and a 311x72 resize that the daemon journaled as state at 15:30:19. The daemon never marked the terminal exited, logged nothing (its stderr log holds 3 startup lines), and the 15:39:29 tab close tombstoned it, so at relaunch the terminal was silently gone. Forensic detail: 31 bytes of terminal output were journaled at the same second as the close, so the stream was wedged rather than cleanly dead all along; the daemon-side reader thread observed nothing for six hours. Evidence archived in `cmuxterm-hq` `artifacts/tui-frozen-round7/` (client tap, ptytap, journal extract, daemon stderr).

Five red->green pairs (regression-test commit ordering, red first):

1. Daemon, reconnect give-ups: every permanent give-up in the hosted reconnect loop funnels through one choke point that logs a timestamped reason to daemon stderr and either commits the durable exit latch (surface-exited + terminal-registry-changed broadcast, journaled `terminal-exited` event with reason) or, when the host was demonstrably reachable, marks the connection failed and emits a status event. While the host's liveness flock is still held the daemon now waits indefinitely at the backoff cap (warn once) instead of silently abandoning after 16 attempts, so an unreachable-but-alive host is never falsely killed and its later death still commits.
2. Daemon, wedged reader: a `terminal-liveness-sweep` thread (30s, `CMUX_TUI_HOST_LIVENESS_SWEEP_MS`, started only for served daemons) probes each hosted terminal's process-lifetime flock proof while its surface still claims a healthy connection, and commits a positively-proven death through the same first-writer-wins latch on a dedicated thread. This is the direct detector for the incident shape; the red test wedges the reader with a debug-only env hook and SIGKILLs the host.
3. Client, visible endings: the scoped attach records why it stopped and prints it on the host terminal after leaving the alternate screen: "the terminal's process ended" (surface-exited), "the terminal was closed" (live-transport Empty), or "lost the connection to the cmux-tui session daemon (reason)" with nonzero exit (transport loss). Localized EN/JA. Full-TUI quits stay silent.
4. Client, reader-loop audit (the silent `break` at remote.rs:1860): every way the remote reader loop can end now returns a reason (oversized frame, EOF, read error), recorded first-writer-wins and appended to the connection-lost line; daemon error replies that match no pending request (fire-and-forget pty input rejected with "unknown surface") surface one status event per distinct error instead of being dropped.
5. Client, liveness: `connect_for_terminal_attach` spawns a 30s heartbeat (identify request under the existing 10s request timeout); a daemon that stops answering while the socket stays open converges to the visible connection-lost exit instead of an indefinite freeze.

Reproduction work: plain and adopted-host `kill -9` were already handled on this branch (exit committed, client ended) and did NOT reproduce the incident; the zombie states requiring these fixes were reproduced with an ambiguous liveness proof (chmod), a held flock past backoff exhaustion, and a wedged-reader stall hook, each captured as a red test. The exact 09:44:52 trigger remains unproven; the checkpoint-race family fixed on main by #10484/#10501 (absent from this branch's base) is the leading suspect, and with this PR any recurrence leaves stderr, journal, and client-notice evidence.

Dictionary:
- liveness flock: an advisory file lock the host process holds for exactly its lifetime; acquiring it proves the host is dead.
- exit sidecar: a small fsynced file the host writes with its child's exit status before announcing exit.
- first-writer-wins latch: the durable exit commit is idempotent, so the reader thread and the sweep can race safely.
- scoped attach: `cmux-tui attach --terminal <id>`, the PTY-in-PTY bridge client a GUI tab runs.

---

**Round 8 (2026-08-21): rebased base (merge `db62d258a0`).** The updated #10428 base now carries main's checkpoint work (#10484, #10501), which replaces the destructive checkpoint-failure reconnect branch this PR had routed through its give-up choke point. Conflict resolution drops that one give-up site (the failure mode no longer exists: capture retries in place and a persistent skip is reported once, host stays connected); every other give-up path of this PR is unchanged. `RemoteSession` keeps both `is_shut_down` (main's caller) and this PR's `transport_lost`/`transport_disconnect_reason`.

Second daily wedge analyzed with `sample` stacks (session cmux-tuispk, terminal `term_25f13201`, artifacts in cmuxterm-hq `artifacts/tui-frozen-round7/wedge-*.txt`): daemon and host were both healthy-idle; the frozen btop itself was deadlocked in its own stall handler (main thread in `pthread_join` inside `Runner::run`, runner thread parked in `atomic_wait`, which is not a cancellation point). btop enters that path only after its output write blocks >5s, i.e. after a >=5s terminal-output admission stall in the daemon path. The journal shows a 7s output gap at 18:18:09 with the main 2.6GB session DB file written exactly then (WAL checkpoint), then permanent silence. So the checkpoint-race family is one producer of such stalls (and is fixed by this stack's new base), but journal-writer stalls on a multi-GB session DB remain a producer this stack does not fix; follow-up: journal growth management/GC.

Merged-head verification on a throwaway daemon (fix6 = `db62d258a0`): host kill -> "the terminal's process ended" + exit commit + both daemon stderr lines in 0.6s; daemon binary swap re-adopted 5 running terminals; 12 busy terminals through a daemon restart produced 19 non-destructive reconnect-checkpoint skips with zero observed output stall (old binary: same load, no reproducible wedge, so prevention of the exact incident is NOT claimed).


lawrencecchen:
Closing: main absorbed this branch's full content (terminal host death detection, journaling, recovery tests) via https://github.com/manaflow-ai/cmux/pull/10603 and follow-ups. The diff between this branch and main restricted to every file this PR touches is empty, so there is nothing left to merge.

PR #10515, opened 2026-08-21
terminal: strip trailing line references before path probing

### LINKED ISSUE #9830: cmd+click / cmux open on "path:line" opens Safari — :line suffix never parsed; file:// URLs mangled
## Summary

Cmd-clicking a `path:line` reference in a cmux terminal (the format every AI coding agent and compiler prints — `src/main.py:1292`) opens **Safari** instead of the file. The underlying dispatcher doesn't parse the `:line` suffix at all, and it also mishandles `file://` URLs.

## Environment

- cmux 0.64.22 (102) [ddd4a01bc]
- macOS 26.6 (25G72), arm64
- Reproduced with the bundled `cmux` CLI — no third-party terminal involved

## Repro 1 — `cmux open` treats `path:line` as a literal path

```bash
mkdir -p /tmp/cmux-test/src && touch /tmp/cmux-test/src/main.py && cd /tmp/cmux-test
cmux open src/main.py          # ✅ OK files=1 ... (opens file preview)
cmux open "src/main.py:1292"   # ❌ Error: Path does not exist: /tmp/cmux-test/src/main.py:1292
```

The whole `path:line` string is checked against the filesystem verbatim. The `:1292` is never stripped, even though the binary contains a line/colon parser (`unexpectedBetweenLineColonAndLineNumber`, `unexpectedAfterLineNumber`, `nonDecimalLineNumber` symbols) — it just isn't wired into the open/click path.

Expected: `cmux open "src/main.py:1292"` opens the file at line 1292 (or at minimum the file itself, ignoring the suffix).

## Repro 2 — `file://` URLs are mangled

```bash
cmux open "file:///tmp/cmux-test/src/main.py"
# ❌ Error: Path does not exist: /tmp/cmux-test/file:/tmp/cmux-test/src/main.py
```

The URL is treated as a relative path: scheme is collapsed (`file:///` → `file:`) and the cwd is prepended. Same failure with a `:line` suffix appended.

Expected: `file://` URLs are recognized and opened as local files (this is the OSC 8 hyperlink target format terminals receive from tools like pi, gcc, clang, ripgrep).

## Repro 3 — cmd+click on `path:line` terminal text → Safari

In a cmux terminal:

```bash
echo "src/main.py:1292"
# then Cmd+click the printed text
```

**Actual:** Safari opens (the text falls through to the URL/browser handler).
**Expected:** file opens in the cmux preview / configured editor.

Note the contrast: cmd+click on a bare `src/main.py` (no suffix) correctly opens the file preview. So the click handler *can* resolve relative paths — it just refuses/fails when a `:line` suffix is present.

## Impact

`file:line` and `file:line:col` are the standard location format printed by compilers, test runners, linters, and AI coding agents (Claude Code, pi, Codex…). Cmd+click-to-source is the primary workflow for these tools, and it's currently broken in cmux whenever a line number is present: every such click opens a browser.

Suggested fix: in the open/click dispatcher, first parse `path[:line[:col]]` (the parser seems to exist already), verify `fileExistsAtPath` on the path part, and route to the file preview / `app.preferredEditor` with the line — and only treat the token as a URL if it has a real URL scheme.

Happy to test a fix.

Reyansh14:
Fixes #9830.

## Summary

Cmd-clicking a `path:line` reference in a cmux terminal — the format every compiler, linter, and AI coding agent prints — opened `https://path:line` in the system browser instead of the file. `cmux open "path:line"` failed the same way.

## Root cause

`pathResolutionCandidates()` in `String+TerminalPathTokens.swift` generates the spellings that `TerminalPathResolver` probes against the filesystem. It trims trailing sentence punctuation, quotes, and bracket pairs, but never a trailing `:<digits>` run — so the existence check ran against the literal `src/main.py:1292`, found nothing, and `resolveOpenURLFilePath` returned nil.

Control then reached `BrowserURLResolver.navigableURL(from:)`:

```swift
if trimmed.contains(":") || trimmed.contains("/") || trimmed.contains(".") {
    return URL(string: "https://\(trimmed)")
}
```

which produced the `https://src/main.py:1292` that ended up in the browser.

## Change

Adds `String.strippingTrailingLineSuffix()` and probes it as an additional candidate. It handles `:line`, `:line:column`, and `:startLine-endLine`, matches ASCII digits only (so a non-numeric final segment like `notes.md:draft` is left alone), and removes at most two segments so genuine path components aren't eaten.

It's applied to both the raw token and the punctuation-trimmed one, so `notes.md:144).` resolves too.

**Non-regressive by construction:** `resolveQuicklookPath` returns the first *existing* candidate, and the new candidates are appended last. Any input that resolved before resolves to the same path — the change can only turn a `nil` into a hit. A file genuinely named `backup:2024` still wins over `backup`, mirroring the existing `prefersLiteralPathThatReallyEndsWithDot` / `...Paren` invariants.

## Scope

Deliberately limited to the `:line` half of #9830. Two things left out:

- **`file://` mangling** (Repro 2 in the issue) — separate defect in a different code path.
- **Jumping to the line** once the file opens. The suffix is parsed and discarded here; threading the line number through to the preview would touch `TerminalLinkOpenCoordinator` and the file-opening protocol. Happy to follow up if wanted.

## Testing

26 tests added to `TerminalPathResolverTests.swift` (24 → 50), in the existing Swift Testing style and reusing the `existsIn` helper: single-segment stripping, ranges and columns, negative cases (non-numeric segment, bare colon, digits without colon, leading-colon token), resolver-level resolution for relative and absolute paths, punctuation-after-suffix, candidate-depth ordering, bare-filename open-URL resolution, and the literal-path-wins invariants.

**Disclosure on how this was verified.** The `CmuxTerminalCore` library target builds cleanly in-tree under the package's real settings (Swift 6 language mode, `ExistentialAny`, `InternalImportsByDefault`), and the test file passes `swiftc -parse`. I was **not** able to run `swift test`: this machine has Command Line Tools but not full Xcode, and `_Testing_Foundation` — the cross-import overlay `import Testing` needs — ships only with Xcode. To verify behavior I compiled `String+TerminalPathTokens.swift` and `TerminalPathResolver.swift` standalone with the same flags and ran every assertion, including four ported from the existing suite as regression checks: 30/30 pass. So the logic is verified but the suite has not been executed as Swift Testing — worth a reviewer running it.

## Summary by CodeRabbit

* **New Features**
  * Improved file path recognition for references with line numbers, line ranges, and line-and-column locations.
  * Supports paths with trailing punctuation while preserving valid matching files.
  * Resolves annotated paths for Quick Look and open-in-editor actions, including bare filenames.

* **Bug Fixes**
  * Prevents invalid or incomplete line suffixes from being interpreted as file paths.
  * Preserves literal files whose names legitimately end with numeric segments.
  * Continues rejecting genuine web and file URLs as local paths.

### DIFF
--- a/Packages/macOS/CmuxTerminalCore/Sources/CmuxTerminalCore/PathResolution/String+TerminalPathTokens.swift
+++ b/Packages/macOS/CmuxTerminalCore/Sources/CmuxTerminalCore/PathResolution/String+TerminalPathTokens.swift
@@ -61,61 +61,49 @@ extension String {
         return String(characters[..<end])
     }
 
-    /// Strips a trailing editor-style line reference from the receiver.
+    /// Strips one trailing editor-style line reference from the receiver.
     ///
     /// Compilers, linters, ripgrep and AI coding agents all print locations as
     /// `path:line`, `path:line:column` or `path:startLine-endLine`. The
     /// reference is not part of the file name, so path probing needs the bare
     /// path as a candidate too.
     ///
-    /// Only ASCII digit runs qualify, so a path whose final colon-separated
-    /// segment is non-numeric is left alone. At most two segments are removed,
-    /// which covers `:line:column` without eating genuine path components.
+    /// Exactly one `:<line>` or `:<startLine>-<endLine>` run is removed per
+    /// call so callers can probe every stripping depth in order, letting a file
+    /// whose name genuinely ends in `:<digits>` still win. Only ASCII digit runs
+    /// qualify, so a non-numeric final segment is left alone.
     ///
-    /// - Returns: The receiver without its trailing line reference, or `nil`
-    ///   when there is no such reference or nothing would remain.
+    /// - Returns: The receiver without its final line reference, or `nil` when
+    ///   there is no such reference or nothing would remain.
     public func strippingTrailingLineSuffix() -> String? {
         func isASCIIDigit(_ character: Character) -> Bool {
             character.isASCII && character.isNumber
         }
 
-        /// Removes one `:<digits>` or `:<digits>-<digits>` run from the end.
-        func strippingOneSegment(_ characters: [Character]) -> [Character]? {
-            var end = characters.count
+        let characters = Array(self)
+        var end = characters.count
 
-            var trailingDigits = 0
+        var trailingDigits = 0
+        while end > 0, isASCIIDigit(characters[end - 1]) {
+            end -= 1
+            trailingDigits += 1
+        }
+        guard trailingDigits > 0 else { return nil }
+
+        // Optional `-<digits>` prefix, so `:144-198` is one reference.
+        if end > 0, characters[end - 1] == "-" {
+            end -= 1
+            var leadingDigits = 0
             while end > 0, isASCIIDigit(characters[end - 1]) {
                 end -= 1
-                trailingDigits += 1
+                leadingDigits += 1
             }
-            guard trailingDigits > 0 else { return nil }
-
-            // Optional `-<digits>` prefix, so `:144-198` is one reference.
-            if end > 0, characters[end - 1] == "-" {
-                end -= 1
-                var leadingDigits = 0
-                while end > 0, isASCIIDigit(characters[end - 1]) {
-                    end -= 1
-                    leadingDigits += 1
-                }
-                guard leadingDigits > 0 else { return nil }
-            }
-
-            guard end > 0, characters[end - 1] == ":" else { return nil }
-            end -= 1
-            return Array(characters[..<end])
+            guard leadingDigits > 0 else { return nil }
         }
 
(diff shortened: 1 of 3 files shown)


Reyansh14:
Closing as a duplicate of #9967, which predates this by ten days and is the more complete fix — it also handles `file://` URLs and forwards the parsed location to the preferred editor, neither of which this PR does. I should have searched open PRs before opening this one.

The one case #9967 does not cover is line **ranges**: `parseLocationSuffix` validates with `positiveInteger`, so `docs/synthetic-monitoring.md:144-198` never parses as a location. I have opened that as a small follow-up against #9967's branch (blackdurumi/cmux#1) so it lands as part of that PR rather than competing with it.

No action needed here — #9967 is the one to review.

PR #10577, opened 2026-08-22
iOS: trim toolbar title reserves 12pt with a collapse-recovery ratchet

azooz2003-bit:
The workspace-detail title pill truncated beside a ~35pt dead gap before the trailing button cluster, because the estimate reserve from https://github.com/manaflow-ai/cmux/pull/10376 must over-count chrome to stay safe. Two approaches died on the way here, both in-branch: shrinking the constants by 30pt collapsed the trailing items into the system More menu on an iPhone 17 Pro Max, and sizing the title from measured window-space positions proved unreliable on iOS 26 (a grouped ToolbarItem's content can render composited without its own window attachment, so cross-item positions are unobtainable; DEBUG probe logs in the history document this).

What lands: the reserves give back 12pt of the ~22pt of slack measured on a 402pt iPhone 17 (per-item chrome 24→20, margins 84→80), well under the trim that broke the Pro Max, and a collapse-recovery ratchet backstops the rest. The always-structural trailing cluster carries a UIKit probe view; its content leaving the window while the SwiftUI view is alive is the observable signature of the system folding the item into More. When that fires, a 28pt recovery reserve ratchets on for the view's lifetime, which is strictly roomier than the original constants, so the bar un-collapses within a frame. Worst case on a device with unusually wide chrome is a one-frame flicker, never a persistent More menu.

Conditional items (Changes chip, alt-screen notice) do not wire the detector because their structural removal also detaches the probe and is indistinguishable from a collapse; the cluster is the widest item and collapses first in practice.

No user-facing strings changed (localization audit: none needed).

### DIFF
--- a/Packages/iOS/CmuxMobileShellUI/Sources/CmuxMobileShellUI/AgentChatDemoScreen.swift
+++ b/Packages/iOS/CmuxMobileShellUI/Sources/CmuxMobileShellUI/AgentChatDemoScreen.swift
@@ -108,7 +108,7 @@ struct AgentChatDemoScreen: View {
                 measuredTrailingItemsWidth: 0,
                 measuredTrailingItemCount: 0,
                 trailingItemCount: 0,
-                measuredTitleToTrailingSpan: nil,
+                hadTrailingCollapse: false,
                 isEnabled: true,
                 workspaceName: inlineWorkspaceTitle ?? "",
                 hasUnread: false,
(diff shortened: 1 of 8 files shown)


azooz2003-bit:
Dogfood on the iPhone 17 Pro Max immediately reproduced the trailing-items-into-More collapse with these tighter reserves (sim-calibrated values undershoot the device's real chrome). Reverted; not worth the regression. Any future tightening should measure the rendered capsule frames on-device instead of tuning constants.

azooz2003-bit:
Reopening per Aziz: the goal stands (tighter title-to-trailing gap), only the constant-shrinking approach is dead. Next revision grows the title cap only into the measured on-screen gap, with the #10376 reserves as the floor.

azooz2003-bit:
Superseded by the consolidated https://github.com/manaflow-ai/cmux/pull/10620 (per Aziz: one PR with all toolbar changes).

PR #10590, opened 2026-08-22
fix(ssh): self-heal daemon bootstrap uploads on remote hosts

### LINKED ISSUE #10587: Nightly update strands tinybox ssh/cmux-ssh sessions: bootstrap parks on 'remote daemon integrity verification failed' and never self-heals (post-#10555)
## Summary

After updating to the current nightly (which includes https://github.com/manaflow-ai/cmux/pull/10555, merged 2026-08-22, the fix for https://github.com/manaflow-ai/cmux/issues/10541) and reopening the app, cmux cannot reconnect to existing `ssh` / `cmux ssh` sessions on `tinybox` at all. The workspace shows:

> Remote daemon bootstrap failed: remote daemon integrity verification failed. Automatic reconnect paused because bootstrap cannot proceed; repair the reported remote failure and use Reconnect to try again.

Both terminal panes show `[cmux] remote PTY bridge closed; reattaching (attempt 1/20).` and then sit dead; notifications show "SSH reconnect paused (tinybox)". Plain `ssh tinybox` works (screenshot 2026-08-22 ~16:19 local; login banners at 16:18:53 and 16:19:27 show the host itself is reachable and fast to auth).

## Question / task

Is this expected fail-closed behavior from #10555, or a bug? #10555 intentionally makes bootstrap verify the uploaded daemon's byte count + SHA-256 remotely and park reconnection ("blocked failures park immediately") instead of looping on a corrupt install. Detection working is good. But the net user experience is: update nightly → daemon version bump triggers re-bootstrap → integrity verification fails → reconnect parks permanently with an error that tells the user to "repair the reported remote failure" by hand, with no indication of what to repair. If the verification failure is caused by cmux's own transfer path producing truncated/empty bytes on this link (the exact upload regression class from #10541: a backgrounded `cat` reading no stdin yields a 0-byte payload; tinybox is a slow link where this reproduced before), then bootstrap fails closed against its own broken upload forever, and "repair and press Reconnect" can never succeed. That is not expected behavior; the self-heal that #10555 promises ("repairs corrupt/nonresponsive installs") never converges.

Needs investigation:

1. Read the sanitized failure diagnostics captured at park time (versioned path, remote byte size, log-tail indicator per #10555) from the local app diagnostics and determine WHICH check failed (byte count vs SHA-256 vs unverifiable) and what the remote byte size was (0? truncated?).
2. Inspect the tinybox side (read-only) — the versioned daemon install path and any temp payloads/PID markers — to see what the transfer actually produced.
3. Decide: transfer bug (fix the upload path so verification passes), repair-path bug (probe detects corruption but the re-upload/repair loop doesn't actually re-upload or reuses the same broken transfer), or genuinely expected manual-intervention case (then the error surface must say exactly what to repair and ideally offer the repair itself).
4. A version bump on nightly update is the most common bootstrap trigger; updating the app must not strand existing remote hosts.

## Environment

- Host: tinybox (slow uplink; prior 0-byte upload repro from #10541 was on this host)
- Local: macOS, nightly updated 2026-08-22; previous session's ssh workspaces present at restore
- Related: https://github.com/manaflow-ai/cmux/issues/10541 (closed by #10555), https://github.com/manaflow-ai/cmux/pull/10555

austinywang:
## Summary

Closes #10587.

This is a transfer bug, not expected fail-closed behavior. PR https://github.com/manaflow-ai/cmux/pull/10555 correctly computes the local byte count and SHA-256 and refuses to promote an unverified payload, but its backgrounded remote `cat` loses SSH stdin under a non-interactive POSIX shell and deterministically writes 0 bytes.

Read-only evidence from tinybox and the local nightly snapshot:

- `/home/tiny/.cmux/bin/cmuxd-remote/0.64.22-nightly.3254304245301/linux-amd64/cmuxd-remote` was mode 755 and **0 bytes** at `2026-08-22T23:19:16Z`.
- The matching local Linux artifact was **6,295,704 bytes**; the older `0.64.22-nightly.3252776701701` remote binary was 6,295,704 bytes and hash-matched its cache. The current `0.64.22-nightly.3260048785601` final path was absent after the rejected attempt.
- The persisted session at park time showed the integrity-class suspension after repeated hello failures. Upload verification occurs before the existing diagnostic probe, so the parked user-facing record did not include path/size; this PR does not weaken the sanitized logging or fail-closed behavior.
- The exact old and new stdin forms were exercised over SSH to tinybox without writing remote state: the old form returned 0 bytes, and the fd-preserving form returned all 22 bytes.

## Fix and coverage

The upload script now uses the portable `exec 3<&0` / `cat <&3` pattern and closes fd 3 after the reader exits. Byte-count/SHA-256 verification, temporary cleanup, atomic promotion, and bounded retry/fingerprint accounting are unchanged. Commit `2cc7fcb5ff` adds a behavioral regression test first; commit `bea3179c4a` fixes it. The test executes the generated command with real file-backed stdin and verifies the complete temporary payload and writer-marker cleanup.

No remote state was modified. No user-facing strings changed, so localization catalogs were unaffected.

### DIFF
--- a/Packages/macOS/CmuxRemoteSession/Sources/CmuxRemoteSession/Session/RemoteSessionCoordinator+DaemonUpload.swift
+++ b/Packages/macOS/CmuxRemoteSession/Sources/CmuxRemoteSession/Session/RemoteSessionCoordinator+DaemonUpload.swift
@@ -96,7 +96,12 @@ extension RemoteSessionCoordinator {
         pid_path=\(quotedRemoteTempPIDPath)
         printf '%s\\n' "$$" > "$pid_path"
         trap 'if [ -n "$cat_pid" ]; then kill "$cat_pid" 2>/dev/null || true; fi; if [ -n "$watchdog_pid" ]; then kill "$watchdog_pid" 2>/dev/null || true; fi; rm -f -- "$temp_path" "$pid_path"; exit 1' HUP INT TERM
-        cat > "$temp_path" &
+        # POSIX shells give an asynchronous command /dev/null for stdin unless
+        # the parent explicitly preserves the descriptor first. Without this
+        # dup, cat exits 0 after writing an empty payload even though ssh had
+        # a file-backed stdin stream to forward.
+        exec 3<&0
+        cat <&3 > "$temp_path" &
         cat_pid=$!
         printf '%s\\n' "$cat_pid" > "$pid_path"
         (
@@ -126,6 +131,7 @@ extension RemoteSessionCoordinator {
         wait "$cat_pid"
         cat_status=$?
         cat_pid=
+        exec 3<&-
         if [ -n "$watchdog_pid" ]; then kill "$watchdog_pid" 2>/dev/null || true; wait "$watchdog_pid" 2>/dev/null || true; fi
         watchdog_pid=
         rm -f -- "$pid_path"
--- a/Packages/macOS/CmuxRemoteSession/Tests/CmuxRemoteSessionTests/RemoteDaemonUploadRecoveryTests.swift
+++ b/Packages/macOS/CmuxRemoteSession/Tests/CmuxRemoteSessionTests/RemoteDaemonUploadRecoveryTests.swift
@@ -5,6 +5,78 @@ import Testing
 @testable import CmuxRemoteSession
 
 extension RemoteDaemonUploadTests {
+    @Test("Background upload reader consumes the SSH stdin stream")
+    func backgroundUploadReaderConsumesSSHStdin() throws {
+        let fileManager = FileManager.default
+        let root = fileManager.temporaryDirectory.appendingPathComponent(
+            "cmux-remote-daemon-upload-stdin-\(UUID().uuidString)",
+            isDirectory: true
+        )
+        let remoteDirectory = root.appendingPathComponent("remote", isDirectory: true)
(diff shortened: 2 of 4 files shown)


austinywang:
CI note: the three failed-job retries all stop before app-host test execution on shared `CMUXAuthCore` / `CmuxAuthRuntime` undefined-symbol linker errors (the same failure is present on unrelated PR branches). The Swift package matrix also repeatedly times out in the unrelated `CmuxIrohTransport` `CmxConnectivityPeerSessionTests` suite after its built-in retry. The focused remote-daemon Go lane, runtime/warning lane, Linux preflight, and all non-app-host checks are green; no actionable failure points to this diff. Attempt 3 is still running for the final package result.

austinywang:
Final CI retry result (attempt 3): the focused `CmuxRemoteSession` package suite passed **164 tests in 27 suites**, including `Background upload reader consumes the SSH stdin stream`; the runtime/warning lane and all Linux/non-app-host checks also passed. The aggregate workflow remains red only because (1) all four app-host shards fail before test execution on missing `CMUXAuthCore`/`CmuxAuthRuntime` linker symbols, reproduced across three fresh-runner attempts and unrelated branches, and (2) the package matrix later times out twice in the unrelated `CmuxIrohTransport` `CmxConnectivityPeerSessionTests` suite. No actionable failure involves this PR. PR remains mergeable/clean.

Issue #10547, opened 2026-08-22
8s+ hangs materializing SF Symbols in NSWindow constraint layout (NSImageSymbolRepProvider, 6k users)


austinywang:
## Summary

App hangs of **8+ seconds** with the main thread inside `-[NSImageSymbolRepProvider init]` triggered from window constraint layout (`NSWindow layoutIfNeeded` → NSISEngine → `_changeWindowFrameFromConstraintsIfNecessary`). SF Symbol image materialization is happening synchronously on the main thread during layout.

- Sentry: [CMUXTERM-MACOS-PZ](https://manaflow.sentry.io/issues/7291999772/)
- **62,142 events / 6,120 users** all-time; still in the top hang cluster for the last 14 days (11,450 events)
- Mechanism: AppHang (≥8000 ms); latest event on `com.cmuxterm.app@0.63.2+79`

## Stack (latest event)

```
at -[NSImageSymbolRepProvider init]
at _nsis_frameInEngine
at -[NSWindow(NSConstraintBasedLayoutInternal) _changeWindowFrameFromConstraintsIfNecessary]
at -[NSISEngine withBehaviors:performModifications:]
at -[NSWindow(NSConstraintBasedLayoutInternal) layoutIfNeeded]
```

## Relationship to existing issues

The `NSImageSymbolRepProvider` *crash* on macOS 27 is tracked in https://github.com/manaflow-ai/cmux/issues/7254 and https://github.com/manaflow-ai/cmux/issues/8162, and a launch-time FileExplorer/NSImage hang was fixed in https://github.com/manaflow-ai/cmux/issues/4853 (closed). This is a distinct, still-live signature: a *hang* (not an exception) during window frame layout, hitting 6k users.

## Suggested direction

- Find which view resolves SF Symbol images inside layout (window frame changes) and pre-resolve/cache those `NSImage`s outside the layout pass.
- Symbol caching added for the macOS 27 crash workaround may be reusable here.

Sourced from a Sentry triage sweep on 2026-08-21.

PR #10591, opened 2026-08-22
Fix degraded SF Symbol rasterization after #10549

### LINKED ISSUE #10586: Titlebar SF Symbol icons render degraded after #10549 bitmap materialization (context CTM captured before rep size set)
## Summary

After https://github.com/manaflow-ai/cmux/pull/10549 ("Fix SF Symbol materialization during window layout", merged 2026-08-22) landed, the titlebar icons render visibly degraded: the sidebar-toggle, bell, and plus glyphs draw thin/soft/washed-out instead of crisp SF Symbol strokes (user screenshot 2026-08-22 16:21 local). https://github.com/manaflow-ai/cmux/pull/10584 reverts #10549 to restore correct icons, which re-opens the 8s+ layout hang https://github.com/manaflow-ai/cmux/issues/10547 that #10549 fixed. We need the hang fix without the rendering regression.

## Root cause (analysis)

`RenderableSystemSymbol.materializedImage(_:size:)` (added in #10549, Sources/RenderableSystemSymbol.swift) rasterizes the configured symbol into an owned `NSBitmapImageRep` with `pixelsWide/High = size * 2`. The bug is order-of-operations:

```swift
let graphicsContext = NSGraphicsContext(bitmapImageRep: bitmap)  // CTM captured here
...
bitmap.size = size                                               // set AFTER context creation
...
source.draw(in: NSRect(origin: .zero, size: size), ...)
```

`NSGraphicsContext(bitmapImageRep:)` derives its transform from the rep's state at creation time. At that moment the rep's `size` still equals its pixel dimensions (72dpi default), so the context maps 1 user-space unit → 1 pixel. Setting `bitmap.size = size` afterwards changes only how the rep is scaled at display time, not the already-created context's CTM. Drawing the symbol into a `size`-point rect therefore rasterizes it at the wrong scale/position within the 2x-pixel bitmap, and the final image (rep scaled to `size` points) draws the glyph distorted/soft — the degraded icons in the screenshot.

The fix direction: rasterize with correct point→pixel mapping (set the rep's `size` before creating the context, or scale the CTM explicitly by the pixel scale before `source.draw`), keep `isTemplate = true`, and rasterize at the actual backing scale rather than hardcoding 2x if feasible. Add a regression test that inspects the rasterized bitmap's pixel coverage (the current test in cmuxTests/RenderableSystemSymbolTests.swift only asserts a bitmap rep exists, which passes even when the glyph is drawn at the wrong scale).

## Expected

Titlebar (and all `CmuxSystemSymbolImage` call sites) render pixel-identical to pre-#10549 `Image(systemName:)`/lazy-NSImage output at 1x and 2x backing scales, while symbol images handed to AppKit layout remain eagerly materialized bitmaps so https://github.com/manaflow-ai/cmux/issues/10547 stays fixed.

austinywang:
## Summary

NSGraphicsContext(bitmapImageRep:) captures the bitmap representation's point-to-pixel transform when it is created. The materializer in https://github.com/manaflow-ai/cmux/pull/10549 created that context while the bitmap still reported its pixel dimensions, then assigned the logical point size afterward. The symbol was therefore rasterized at 1x into the upper-left portion of a 2x bitmap and displayed soft, thin, and displaced.

This change reapplies the eager materialization and titlebar prewarming from https://github.com/manaflow-ai/cmux/pull/10549 on top of the merged revert, but assigns each bitmap's point-space size before creating its context. The context CTM now maps the configured symbol rect across the full backing bitmap. Each returned image carries owned 1x and 2x bitmap representations, allowing AppKit to choose a native scale without resampling. The template flag remains true, preserving AppKit tinting and enabled/disabled appearance behavior.

## Regression coverage

- The first commit adds a pixel-level regression test that draws the returned image and a fresh configured AppKit symbol into independently initialized bitmaps.
- The test runs at both 1x and 2x and compares alpha-weighted centroid, bounding box, and coverage, so the pre-fix scale/placement regression fails.
- The eager-materialization test verifies both representations are bitmap-backed and that no lazy symbol provider is returned.
- The test-only regression commit precedes the fix commit intentionally.

This supersedes the stopgap revert https://github.com/manaflow-ai/cmux/pull/10584: it keeps https://github.com/manaflow-ai/cmux/issues/10547 fixed while restoring crisp titlebar icon rendering.

## Validation

- xcrun swiftc -parse passed for all changed Swift files.
- ./scripts/lint-pbxproj-test-wiring.sh passed.
- ./scripts/check-pbxproj.sh passed.
- python3 scripts/check-package-resolved-policy.py passed.
- No local Xcode build or test was run, per task instructions; hosted CI is the test environment.

Closes #10586

### DIFF
--- a/Sources/RenderableSystemSymbol.swift
+++ b/Sources/RenderableSystemSymbol.swift
@@ -138,6 +138,25 @@ enum RenderableSystemSymbol {
         renderabilityCache.isRenderable(symbol)
     }
 
+    /// Resolves a bounded set of symbols before a view enters an AppKit
+    /// window's constraint/layout cycle.
+    @MainActor
+    static func prewarmAppKitImages(
+        systemNames: [String],
+        pointSizes: [CGFloat],
+        weight: Font.Weight? = nil
+    ) {
+        for systemName in systemNames {
+            for pointSize in pointSizes {
+                _ = configuredAppKitImage(
+                    systemName: systemName,
+                    pointSize: pointSize,
+                    weight: weight
+                )
+            }
+        }
+    }
+
     static func clampedRasterPointSize(_ pointSize: CGFloat) -> CGFloat {
         guard pointSize.isFinite else {
             return minimumRasterPointSize
@@ -186,9 +205,14 @@ enum RenderableSystemSymbol {
             weight: fontWeight
         )
         let configuredImage = baseImage.withSymbolConfiguration(configuration) ?? baseImage
-        let image = (configuredImage.copy() as? NSImage) ?? configuredImage
+        let imageSize = symbolImageSize(configuredImage.size, fallbackDimension: rasterSize)
+        guard let image = materializedImage(configuredImage, size: imageSize) else {
+            return nil
+        }
+        // Keep the template contract used by the SwiftUI and AppKit callers,
+        // while replacing AppKit's lazy symbol representation with a bitmap
+        // that cannot be materialized again from an NSWindow layout pass.
         image.isTemplate = true
-        image.size = symbolImageSize(configuredImage.size, fallbackDimension: rasterSize)
         appKitImageCache[cacheKey] = image
         appKitImageCacheInsertionOrder.append(cacheKey)
         while appKitImageCacheInsertionOrder.count > appKitImageCacheLimit {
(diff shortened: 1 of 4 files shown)


austinywang:
Addressed the review feedback in 97bf365186:

- The raster-geometry test now runs the same centroid, bounds, and alpha-coverage assertions at both 1x and 2x.
- The materialized image now contains native 1x and 2x bitmap reps, so neither scale is produced by resampling the other.
- The branch has been rebased onto the merged revert on origin/main; gh pr view reports mergeable CLEAN.

austinywang:
Hosted validation note: the focused cmuxTests/RenderableSystemSymbolTests lane was attempted on both Warp (run https://github.com/manaflow-ai/cmux/actions/runs/32606603970) and Blacksmith macOS 15 (run https://github.com/manaflow-ai/cmux/actions/runs/32607449481). The changed source compiled, but the shared cmux-unit test bundle failed before executing tests with the existing undefined CMUXAuthCore/CmuxAuthRuntime linker symbols (the same failure appears in prior main test-depot runs, e.g. https://github.com/manaflow-ai/cmux/actions/runs/32546133025). No symbol-raster test failure or new warning was reported; the PR’s required checks remain green and mergeability is CLEAN.

PR #10592, opened 2026-08-23
ios: hold scroll position through replays mid-gesture + frame-rate audit probes

azooz2003-bit:
Two changes driven by on-device probe data from the pixel-scroll dogfood (follow-up to https://github.com/manaflow-ai/cmux/pull/10523):

**Fix: replays no longer fight an active scroll.** The device log showed 117 verified replays in one session under agent output; each reset the mirror to the bottom, the anchor restore stood down for user interaction (`viewport_restore.skipped reason=user_interaction` every few hundred ms), and the next scroll frame rebased from the reset viewport, so the app visibly fought the gesture. While a gesture or its deceleration is active, the pixel pump is now the position authority: batches rebase from the last position the pump itself applied rather than the live viewport, and a replay completing mid-gesture triggers one zero-delta batch so a stationary finger is restored too. Idle behavior is unchanged (live viewport stays authoritative; dock/typing snaps and row-space changes clear the held position). Also rounds the scroll offset to whole device pixels so glyph antialiasing doesn't resample every frame.

**DEBUG audit probes:** per-second `perf.refresh scroll_fps/link_fps/missed/max_gap_ms` while scrolling, one-shot `perf.display max_fps`, rate-limited `perf.render_now`. Together with the merged `perf.pixel_scroll`/`perf.process_output` probes they cover the whole scroll path. First device data: 120Hz granted and 113fps achieved when quiet; replay-storm windows downshift Core Animation to 80Hz with 23-53ms gaps, and all pipeline stages measure fast, which makes storm reduction (separate follow-up) the remaining smoothness work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scrolling smoothness by aligning pixel movement to device pixels.
  * Preserved active scroll positions during viewport restoration, reducing unexpected jumps during gestures.
  * Reduced visual inconsistencies caused by fractional offsets, replay resets, and switching between scrolling modes.
  * Improved momentum handling for line-based scrolling.

* **Diagnostics**
  * Added debug metrics for scroll frame rates, delayed frames, display timing, and rendering duration.
  * Added rate-limited reporting for slow rendering operations to support performance troubleshooting.

### DIFF
--- a/Packages/iOS/CmuxMobileTerminal/Sources/CmuxMobileTerminal/Debug/GhosttySurfaceView+ScrollScriptDebug.swift
+++ b/Packages/iOS/CmuxMobileTerminal/Sources/CmuxMobileTerminal/Debug/GhosttySurfaceView+ScrollScriptDebug.swift
@@ -50,6 +50,55 @@ extension GhosttySurfaceView {
         enqueueScrollToBottom()
         debugScrollScriptDone = true
     }
+
+    /// Scroll-smoothness audit: logs the display's max rate once, then one
+    /// per-second summary of achieved display-link cadence while a scroll
+    /// gesture or its deceleration is active. `scroll_fps` is what the app
+    /// achieved, `link_fps` is the rate Core Animation currently grants, and
+    /// `missed` counts ticks that arrived over 1.6x the granted frame time.
+    func debugRecordScrollFrameRateTick(now: CFTimeInterval) {
+        if !debugScrollFrameRateStats.loggedDisplayInfo,
+           let screen = window?.windowScene?.screen {
+            debugScrollFrameRateStats.loggedDisplayInfo = true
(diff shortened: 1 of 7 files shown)


azooz2003-bit:
Merge gate: local pinned autoreview clean for 3f0a53d8 ("patch is correct") after three review rounds fixed real findings: the fraction carry moved out of the flush guard (idle interaction-generation churn), epoch-gated batch commits (in-flight batch resurrecting cleared authority), exact position carry (sub-pixel stalls), and a revision-guarded held anchor (eviction/reflow renumbering). Remaining P2 policy advisories dispositioned earlier: the debug helper lives in the dedicated Debug/ folder per the policy's fallback, and the frame-rate stats struct matches the neighboring debug members.

PR #10615, opened 2026-08-23
fix(ios): stop the diff pager taking over mid-swipe gestures

azooz2003-bit:
Swiping between files in the iOS diff viewer stuttered: halfway through a page switch the pager would cancel the finger-driven transition and animate on its own (reported from phone dogfood). Two earlier passes on this branch fixed child-level artifacts (placeholder crossfades, cold pages) but not the takeover.

Root cause: the page-style TabView's content was a function of an LRU presentation cache held in sheet-level `@State`, three levels up. Every async cache write rebuilt the paging container: prefetch landings while reading, the LRU `touch` fired on every page settle, and evictions (cap 7 against a radius-2 prefetch window). SwiftUI's page TabView is backed by a UIKit paging scroll view; when its content dependencies change mid-gesture it cancels the interactive transition and animates to the resolved page. Two owners were fighting over one interaction.

The redesign gives UIKit exclusive ownership of the gesture and page lifecycle, and makes the cache invisible to rendering:

- `DiffPagerContainerView` wraps `UIPageViewController`. Its update pass is a guarded no-op unless the changed-file snapshot itself changed, and never while a transition is in flight, so no SwiftUI re-render can rebuild the container mid-swipe.
- `FileDiffPresentationStore` is a render-inert reference type (deliberately not `@Observable`) holding parsed diffs and per-page scroll positions. Pages pull from it at mount; prefetch and load completions write into it without invalidating any view.
- A small `@Observable` `DiffPagerPageEnvironment` carries live font size and the selected index, so pinch and selection gating re-render only mounted page subtrees.
- Deleted: `DiffPagerMountPolicy`, the `Color.clear` placeholder swap, and `.transition(.identity)` — `UIPageViewController` instantiates only the current page and its neighbor natively.
- Binary-preview toolbar actions forward through the composition layer because preference keys do not cross `UIHostingController` boundaries.

`FileDiffPageView` is unchanged. The prefetch policy stays (warming now lands silently in the store). 73 package tests green, including new store coverage; the mount-policy test was replaced by the store test.

HIG: https://developer.apple.com/design/human-interface-guidelines/gestures — respond to gestures consistently with people's expectations and keep direct manipulation uninterrupted with predictable feedback; a pager that takes over an in-progress drag violates both.

No new user-facing strings (localization audit: none needed).

## Summary by CodeRabbit

* **Performance**
  * Nearby file diffs now prefetch as you browse, helping pages open faster.
  * Prefetching prioritizes the next page and skips already cached files.
  * Recently viewed diff pages and scroll positions are preserved for smoother navigation.

* **Bug Fixes**
  * Diff pages no longer crossfade unexpectedly while loading or when the visible page range changes.
  * Loading, success, and error states now transition more smoothly without unwanted animations.
  * Paging remains stable when file lists change or selections fall outside the available range.


azooz2003-bit:
Verification evidence (isolated simulator `cmux-dev-dtree`, `CMUX_UITEST_CHANGES_PREVIEW=diff` harness with its 150ms artificial per-file load delay, identical scripted idb swipe sequence on both builds; recordings and 20fps frame grids on the dev Mac under /tmp/dtree-pager-{before,after}.mp4 and /tmp/dtree-{before,after}-grid.png):

Before: landing pages rendered as full-page redacted placeholders that popped into content well after settle; two files' diff text visibly crossfaded/superimposed during transitions; the 400-line RenderPipeline page never painted during its pass.

After: every transition is a clean side-by-side page slide with both pages rendered, including the 400-line page; no superimposed text in any extracted frame; the only remaining placeholder is a single frame on a page being flung past during a hard multi-page fling, replaced by content on the next frame (its prefetch cache hit had not yet committed). Placeholder-heavy frames (>10% placeholder-gray, 20fps sampling): before 3 mid-gesture, after 0 mid-gesture (one at app launch, which is the intended initial-load placeholder).

PR #10686, opened 2026-08-25
fix(ghostty): recover incomplete renderer state after #10037

### LINKED ISSUE #10037: [ghostty fork] Renderer thread SIGSEGV in rebuildCells: cursor indexes a never-built row (empty cell buffer) after an incremental RenderState update
## Summary

_Filed here because `manaflow-ai/ghostty` has issues disabled; this is a bug in that fork's renderer (the engine cmux and kooky both embed)._

kooky (a macOS terminal embedding libghostty via the GhosttyKit xcframework built at `88c3325d`) received two user crash reports with byte-identical stacks: `EXC_BAD_ACCESS (KERN_INVALID_ADDRESS at 0x18)` on the thread named `renderer`, triggered while switching windows on long-lived surfaces (one process was ~6.4h old). Report with full .ips files: https://github.com/iAmCorey/kooky/issues/53

Release binaries are stripped, so we symbolicated by matching instruction byte sequences from the crash PC back into the symbol-rich `libghostty-internal.a`:

| frame | function |
|---|---|
| 0 (crash) | `renderer.generic.Renderer(Metal).rebuildCells + 0xf0c` |
| 1 | `renderer.generic.Renderer(Metal).updateFrame` |
| 2 | `watcher.async.AsyncMachPort(...).wait.callback` |
| 3 | `backend.kqueue.Loop.tick` |
| 4 | renderer thread entry |

## Crash site

[`src/renderer/generic.zig` (rebuildCells, cursor block)](https://github.com/manaflow-ai/ghostty/blob/88c3325dc9698d887da7e07ee0f9b79c53020be2/src/renderer/generic.zig#L2878-L2889):

```zig
const cells = state.row_data.items(.cells);
const cell = cells[cursor_vp.y].get(cursor_vp.x);   // <- crash
```

Disassembly matches exactly: the per-row `MultiArrayList(Cell)` for the cursor's row has `capacity == 0`, so `items()` yields the empty-slice sentinel pointer `0x8`, and `.get(x)` with `cursor_vp.x == 2` reads `0x8 + 2*8 = 0x18` unchecked in ReleaseFast. Registers in the .ips confirm it: `x10 = 0x8` (sentinel), `x28 = 2` (cursor x), `far = 0x18`.

So at rebuild time the render state held a row whose cell buffer was **never built**, while `cursor.viewport` pointed at it.

## How the invariant breaks

In [`RenderState.beginUpdate`](https://github.com/manaflow-ai/ghostty/blob/88c3325dc9698d887da7e07ee0f9b79c53020be2/src/terminal/render.zig#L356), cursor detection and row building are independent: cursor detection runs unconditionally per chunk, while row builds are gated on dirty flags. The implicit invariant is that any row with an empty cell buffer is covered by a `redraw` pass before an incremental frame runs. Two paths can break it:

1. **Mid-update allocation failure.** Dirty flags are consumed (`p.dirty = false`, `page_row.dirty = false`) before/while calling `try builder.row(...)`, and `self.rows` is updated early. If a `try` fails partway through a redraw that populated fresh rows (a resize, or a state reset), the next update sees `redraw == false` and skips the never-built rows forever — until the cursor lands on one.

2. **Short viewport page iteration.** The row loop is `while (y < self.rows) { const chunk = page_it.next() orelse break; ... }` with only a Debug-mode `assert(y == self.rows)` after it ([here](https://github.com/manaflow-ai/ghostty/blob/88c3325dc9698d887da7e07ee0f9b79c53020be2/src/terminal/render.zig#L631)). In a release build a short iteration silently leaves rows `[y..rows)` unbuilt, and the selection pass right below dereferences their undefined pins.

### Amplifier: the periodic reset never resets its own counter

[`updateFrame`](https://github.com/manaflow-ai/ghostty/blob/88c3325dc9698d887da7e07ee0f9b79c53020be2/src/renderer/generic.zig#L1417-L1422):

```zig
if (self.terminal_state_frame_count >= max_terminal_state_frame_count) {
    self.terminal_state.deinit(self.alloc);
    self.terminal_state = .empty;
}
self.terminal_state_frame_count += 1;
```

Once a surface passes 100k frames, **every** subsequent frame tears down and fully rebuilds the terminal state — defeating the incremental optimization and re-opening the "fresh empty rows being populated" window on every single frame. This matches the crashes being accumulation-dependent (hours-old surfaces).

## Reproduction

We verified the mechanism end-to-end with a fault-injection build: an env-gated hook that, every 120 frames, destroys the cursor row's cell buffer right after a fully successful `beginUpdate` (recreating the corrupt state byte-for-byte). Results:

- **Unpatched `88c3325`**: crashes within seconds. All five stack frames land at the user's reported imageOffsets shifted by exactly +184 bytes (the size of the injected hook code) — i.e. the identical call chain, same faulting instruction, same `0x8` sentinel, and `far == 0x8 + cursor_x*8`.
- **Patched build** (fixes below): survives the identical injection indefinitely; the bounds check skips the cursor for one frame.

## Fixes we ship downstream (happy to PR any/all)

Full patch (against `88c3325d`, incl. two regression tests): https://github.com/iAmCorey/kooky/blob/main/scripts/kooky-libghostty-issue53.patch

1. **Bounds-check the cursor cell access** in `rebuildCells` (and the `row_dirty[cursor_vp.y]` preedit read): skip the cursor for one frame instead of faulting.
2. **Reset `terminal_state_frame_count = 0`** inside the periodic reset.
3. **`RenderState.incomplete` flag + explicit `UpdateError`**: set at the start of `beginUpdate`, cleared only when the row loop accounted for every viewport row. A short iteration now returns `error.IncompleteUpdate` so the caller discards the frame (the renderer thread's existing `catch { log.warn; return }` handles it), and the next update forces a full redraw — covering both the error path and the silent short-iteration path. Two implementation notes that bit us: the recovery branch must NOT sit under `assert(y == self.rows)` (in ReleaseFast the assert lowers to `unreachable` and the optimizer deletes the recovery branch entirely), and the allocation-failure regression test must start each fail-point injection from a brand-new `RenderState` (retained capacities otherwise absorb allocations and skip failure points).

One semantic note we chose not to change: the C API folds `IncompleteUpdate` into its out-of-memory result. Behaviorally correct (caller discards the update) but you may want a distinct error code at the ABI layer.

## Upstream note

The render-state rewrite this stems from (`446f80f4`, "render state update optimizations") also landed in `ghostty-org/ghostty`, so the unchecked cursor access and the short-iteration path likely affect upstream as well — you may want to relay it there.

austinywang:
Closes #10037

## What changed

The renderer could read a cursor row whose cell buffer was never built after an allocation failure or a short viewport page iteration. This change:

- marks `RenderState` updates incomplete until every viewport row is accounted for and forces a full rebuild after an interrupted update;
- returns `error.IncompleteUpdate` for short iterations in ReleaseFast;
- bounds-checks cursor/preedit row and cell reads before GPU-cell rebuilding;
- resets the periodic terminal-state teardown counter;
- adds allocation-failure recovery coverage and a regression for an unavailable row.

The Ghostty fork implementation and review branch are [manaflow-ai/ghostty#208](https://github.com/manaflow-ai/ghostty/pull/208). The final pinned universal ReleaseFast artifact is documented in `docs/ghostty-fork.md`; both test and fix SHAs are pinned in `scripts/ghosttykit-checksums.txt` and verified by checksum.

## Test-first history

1. `7e02cf05e2` — test-only parent pointer to Ghostty `0ce0157de`, with its matching test-SHA artifact/checksum (reproduces the incomplete-row behavior).
2. `83fbfea680` — Ghostty pointer `c1d6d8769`, docs, and the final artifact checksum.

## Verification

- Focused Zig tests passed:
  - `interrupted render state rebuilds empty row on next update`
  - `allocation failure mid-update recovers on the next update`
  - `renderer rebuild row preedit catch-up tolerates empty tail after covered glyph`
  - `incremental updates match full rebuild`
- ReleaseFast allocation-failure test passed.
- `zig fmt --check` and `git diff --check` passed.
- `scripts/download-prebuilt-ghosttykit.sh --verify-only` passed against the published 123 MB fix artifact; the test-SHA artifact also passed archive validation and checksum verification.
- `test_ci_ghosttykit_checksum_present.sh`, `test_ci_ghosttykit_release_check.sh`, `test_ci_ghosttykit_checksum_verification.sh`, and package-resolved policy checks passed.

Per the issue instructions, no local Xcode build, XCUITest, tagged reload, or app launch was run. A broader `begin and end update` filter also exercised unrelated search tests and hit an existing libxev/kqueue timeout; the targeted renderer filters above are green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal rendering recovery after interrupted updates.
  * Hardened cursor and preedit display when terminal rows are incomplete.
  * Improved scrolling and rendering behavior across embedded terminal views.

* **Documentation**
  * Updated pinned renderer information and documented recovery behavior.
  * Added verification details for the updated embedded terminal build.


austinywang:
Review audit for refreshed head `da73f0490dc00540da44d04e7b02fb4e612ff27f` (base `2ead47750ab2f47c13972d0709d99cdcbaa8ad73`): the only inline Cubic thread on `docs/ghostty-fork.md` is resolved and outdated; the paragraph separation is present in the current file. CodeRabbit’s `ios/scripts/upload-testflight.sh:942` suggestion was verified as outside this PR: `git diff origin/main...HEAD -- ios/scripts/upload-testflight.sh` is empty and the file matches base, so no unrelated iOS change was made. The effective PR diff remains only `docs/ghostty-fork.md`, the `ghostty` gitlink, and `scripts/ghosttykit-checksums.txt`; no unresolved actionable review thread remains.

PR #10688, opened 2026-08-25
Keep minimized windows minimized when the global hotkey activates cmux

### LINKED ISSUE #9991: Global hotkey un-minimizes minimized windows when cmux is in the background
**cmux version:** 0.64.22
**macOS version:** macOS 27.0 (build 26A5388g)
**Mac chip:** Apple Silicon
**Installation method:** Direct download (DMG)
**Reproduces on NIGHTLY:** not tested, but the code path below is unchanged on `main` as of 22c82ea176.

### Bug description

When the system-wide hotkey (Show/Hide All Windows) fires while cmux is in the background but not hidden, every minimized cmux window is un-minimized and brought to the front, not just the windows that were on screen.

iTerm2 does not behave this way — its equivalent activation brings windows forward but leaves minimized windows in the Dock. macOS convention generally treats minimizing as an explicit "put this away" that only the user or the Dock undoes.

### Expected behavior

The hotkey shows the windows that were on screen and activates the app. Windows the user deliberately minimized stay minimized.

### Steps to reproduce

1. Enable the system-wide hotkey in Settings → Global Hotkey.
2. Open two cmux windows. Minimize one.
3. Click into another app (Finder, a browser — just switch focus; do **not** use Hide cmux, and do not use the hotkey to hide it).
4. Press the system-wide hotkey.

Both windows come to the front — the minimized one is restored along with the visible one.

### Where this happens

`MainWindowVisibilityController.showApplicationWindows` (`Sources/App/MainWindowVisibilityController.swift:270`) picks reveal targets in three branches:

1. **App was hidden by the hotkey** (`:283`) — restores the targets captured at hide time, filtered to visible-and-not-miniaturized in `captureHiddenWindowRestoreTargets` (`:222`). Minimized windows are correctly left alone here, and there's a test for it: `testHotkeyRestoreUsesCapturedVisibleTargetsWithoutDeminiaturizingMiniaturizedWindows` (`cmuxTests/MainWindowVisibilityControllerTests.swift:84`).
2. **App hidden, nothing captured** (`:297`) — `revealTargets = allWindows.filter { isMiniaturized }`. Deliberate, and tested at `:146`. This one seems right as a last resort: without it the hotkey would be a no-op when every window is minimized.
3. **App merely in the background** (`:299`) — `revealTargets = visibleOrMiniaturizedTargets`, which includes minimized windows.

`reveal()` then keeps miniaturized windows whenever `makeKey` is true (`:375-377`) and deminiaturizes each one (`:383`).

Branch 3 is the one that produces the report above, and it's also the most common way to reach the hotkey — switching to another app doesn't hide cmux. So the "leave minimized windows alone" rule that branch 1 carefully implements is bypassed in the ordinary case.

### Question for maintainers

Is branch 3 intended, or did the minimized-window handling in branch 1 simply not get extended to it?

Two ways to resolve it, and I'd rather hear a preference than guess:

- **Just fix branch 3** — filter to visible windows, keep branch 2's fallback so the hotkey is never a no-op. Matches iTerm2 and macOS convention, no new settings surface, but it is a behavior change for anyone used to the current restore-everything.
- **Add a setting** — e.g. `app.globalHotkeyRestoresMinimizedWindows` alongside `systemWideHotkeyEnabled` in `AppCatalogSection`, with a toggle in `GlobalHotkeySection`, threaded into the controller through `Dependencies` so it stays testable. Governs branches 1 and 3; branch 2 stays unconditional.

Happy to send a PR for whichever you prefer, with regression coverage for the background-not-hidden path.

austinywang:
## Summary

- Keep already-minimized main windows in the Dock when the system-wide Show/Hide hotkey activates cmux while it is inactive but not hidden.
- Reveal only currently visible, non-miniaturized main windows in that path; preserve the existing hidden-application fallback that restores minimized windows when no captured or dismissed targets exist.
- Add behavior-level coverage for the reported visible-plus-minimized window state and correct the hidden fallback fixture.

The inactive path previously grouped visible and miniaturized windows together, then sent every target through `reveal()`, which deminiaturizes its window.

Fixes #9991

## Testing

- Red proof: [test-only commit `13dcefdd`](https://github.com/manaflow-ai/cmux/commit/13dcefddf9d37627be0e452b46c1ecdecdbaf35f) in [hosted run 32807694065](https://github.com/manaflow-ai/cmux/actions/runs/32807694065) executed the exact regression test once and failed at the minimized-window assertion.
- Fixed head `792917af`: [lifecycle suite](https://github.com/manaflow-ai/cmux/actions/runs/32809635183) passed 4 tests in 1 suite, including `backgroundHotkeyLeavesMiniaturizedWindowsInTheDock()`.
- Fixed head `792917af`: [controller suite](https://github.com/manaflow-ai/cmux/actions/runs/32808808046) passed 11 tests with 0 failures.
- Passed `git diff --check`, Swift frontend parsing for all three changed Swift files, test wiring lint (703 files), pbxproj normalization check, Package.resolved policy, workspace package grouping, and strict test-determinism check.
- No local xcodebuild, XCUITest, tagged reload, or app launch was run, per the task constraint.
- The requested `.github/swift-file-length-budget.tsv` and `scripts/swift_file_length_budget.py` do not exist on current `main`. The production file remains 507 lines, the pre-existing 520-line XCTest file does not grow, and the lifecycle test file is 207 lines. No warning-budget file was changed.
- Localization audit: no user-facing strings or message catalogs changed.

## Demo Video

- Not recorded: the task explicitly prohibited a tagged reload or launched build. Runtime dogfood remains for handoff.

## Checklist

- [ ] I tested the change locally
- [x] I updated docs/changelog if needed (not needed for this targeted bug fix)
- [ ] I requested bot reviews after my latest commit

---
## Summary by cubic
Fixes #9991 by keeping minimized main windows in the Dock when the global Show/Hide hotkey activates `cmux` while the app is inactive but visible. Previously, the hotkey revealed both visible and minimized windows; now it reveals only visible windows and restores minimized windows only when no visible or dismissed targets exist.

- Adds lifecycle coverage for mixed visible/minimized and all-minimized states, and corrects the hidden-app fallback fixture.
- Routes the browser skill contract through configured `LINUX_RUNNER`, defaulting to the Blacksmith runner.

<sup>Written for commit 68435c7530898e7af8892997f6609345f3377ddf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10688?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved window visibility toggling so minimized windows remain in the Dock when a visible window is available.
  - Restores a minimized window when all application windows are minimized, ensuring the app can still be brought forward correctly.
  - Corrected restoration behavior when switching between hidden and visible application states.

- **Tests**
  - Added coverage for visibility toggles involving minimized and visible windows.
  - Updated restoration scenarios for hidden applications.
  - Added lifecycle coverage for global-hotkey visibility toggles.


jacob-carlborg-apoex:
Reporter of #9991 here. Thanks for picking this up — the direction is right and the filter change fixes the case I reported. I built this branch's behaviour locally and confirmed the visible-plus-minimized case now works.

But it doesn't fully close #9991: it trades the original bug for a state where the hotkey does nothing at all.

### The missing case

When **every** main window is minimized and the app is not hidden:

- `visibleTargets` is empty, so the `else if` at [L299](https://github.com/manaflow-ai/cmux/blob/792917af03d2fdae531c3ee9e84d2e316a844140/Sources/App/MainWindowVisibilityController.swift#L299) doesn't take.
- The `else` at [L301-L307](https://github.com/manaflow-ai/cmux/blob/792917af03d2fdae531c3ee9e84d2e316a844140/Sources/App/MainWindowVisibilityController.swift#L301-L307) only considers `dismissedWindowRestoreTargets`, which is populated solely by titlebar dismiss and is normally empty.
- So `revealTargets == []`, and `reveal()` hits its `guard !windows.isEmpty` at [L378](https://github.com/manaflow-ai/cmux/blob/792917af03d2fdae531c3ee9e84d2e316a844140/Sources/App/MainWindowVisibilityController.swift#L378) and returns **before** `activate()` at [L393](https://github.com/manaflow-ai/cmux/blob/792917af03d2fdae531c3ee9e84d2e316a844140/Sources/App/MainWindowVisibilityController.swift#L393).

Repro: minimize every cmux window, click into another app (don't Hide cmux), press the hotkey. Nothing happens — the app doesn't even come forward. Same dead end if you minimize your only window while cmux is still frontmost. Before this PR that state deminiaturized and activated.

The hotkey path is the only caller with no fallback: `applicationShouldHandleReopen` and `showMainWindowFromMenuBar` both fall through to `ensureInitialMainWindowIfNeeded`, which focuses a window and deminiaturizes it. `toggleApplicationVisibility` just does `_ = showApplicationWindows(...)`.

### Suggested fix

The hidden branch already has exactly the right last-resort at [L297](https://github.com/manaflow-ai/cmux/blob/792917af03d2fdae531c3ee9e84d2e316a844140/Sources/App/MainWindowVisibilityController.swift#L297); it just needs to apply to the not-hidden branch too:

```diff
             dismissedWindowRestoreTargets.removeAll()
-            revealTargets = dismissedTargets
+            revealTargets = dismissedTargets.isEmpty
+                ? allWindows.filter { dependencies.windowOperations.isMiniaturized($0) }
+                : dismissedTargets
         }
```

Preference order then reads the same in both branches — on-screen windows first, then dismissed, then minimized only when there is nothing else to show. Minimized windows stay in the Dock whenever anything else can be revealed, and the hotkey is never inert. (Worth pulling that filter into a small helper so the two branches don't duplicate the rule.)

### The tell

`testShowApplicationWindowsStillRestoresMiniaturizedWindowsWhenNoHiddenTargetsWereCaptured` was exactly this scenario — one minimized window, `isApplicationHidden: { false }`, reason `.globalHotkey`, asserting deminiaturize + makeKey. This PR flips its fixture to `isApplicationHidden: { true }` and renames it. That retarget is the regression showing through at the test level: the old assertions describe the behaviour that is now lost.

With the fallback above, that test passes **unmodified** — which seems like the strongest signal that the fallback is the missing piece rather than the test being wrong.

Happy to send this as a PR against your branch if that's easier than patching it yourself.

PR #10743, opened 2026-08-25
fix(tui): filter tree tabs against attached surfaces

lawrencecchen:
Filter remote tree snapshots against the attached surface catalog before publishing them. This closes the detach/overflow race where a stale pane tab could create a TabView for a removed surface. Clamp active tab indexes after filtering.

Validation: cargo fmt --all -- --check; git diff --check. Rust tests/builds were not run locally per instructions.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10743"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790259720&installation_model_id=21920&pr_number=10743&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10743&signature=77b1e9cdfd17a81ced02bfd5cded77a21e09e16fa8e567b7513b453a75604530"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Filters remote tree snapshots to only include tabs backed by currently attached surfaces, closing the detach/overflow race that let a stale pane tab render a removed surface. After filtering, clamps each pane’s active tab index to the remaining tabs to avoid out-of-bounds; the active tab may shift when a surface detaches.

<sup>Written for commit 470252914f76bd3124d38a5e19c61c9716cd1fb3. Summary will update on new commits.</sup>

[...]

PR #10747, opened 2026-08-25
fix(tui): close filtered tree publication race

lawrencecchen:
Follow-up to PR #10743.

Preserve the active SurfaceId while removing detached tabs, use a deterministic clamped fallback when it disappears, and keep empty panes valid. Hold the surface catalog lock through filtering and tree-cache publication so detach events cannot publish a mismatched tree/catalog pair.

Focused tests cover active-tab identity and all-tabs-removed panes.

---
## Summary by cubic
Fixes a race in TUI tree publication that could produce a tree/catalog mismatch and scramble the active tab when surfaces detached mid-update. The renderer now sees a coherent snapshot, and panes remain valid even when all tabs are removed.

- Bug Fixes
- Hold the surface catalog lock through tree filtering and cache publication to prevent mismatched revisions.
- Preserve the previously active surface when earlier tabs are removed; clamp to the last tab only if the active surface no longer exists.
- Keep empty panes valid with active_tab=0 and no active surface.
- Capture surfaces under the lock and drop it before updating browser sources to avoid lock-held iteration.
- Add focused tests for active-tab preservation and all-tabs-removed panes.

<sup>Written for commit 35ef21fa3b41f528709b2c932468737aa6475369. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10747?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/session/remote.rs
+++ b/cmux-tui/crates/cmux-tui/src/session/remote.rs
@@ -3377,8 +3377,12 @@ impl RemoteSession {
         // A tree snapshot can race a detach or overflow event. The surface
         // catalog is authoritative for what can be rendered, so filter the
         // snapshot while holding one catalog snapshot before publishing it.
-        let attached_surface_ids = self.surfaces.lock().unwrap().keys().copied().collect();
         let mut tree = tree;
+        // Keep the catalog lock through filtering and cache publication. This
+        // makes the tree and attached-surface set one coherent revision from
+        // the renderer's point of view, so a detach cannot land between them.
+        let surfaces_guard = self.surfaces.lock().unwrap();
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Closing this stale branch. Its publication-fencing intent is already present on current main in commits 81bd256cf4, f0e5120d8c, c4d7ee7520, e00dc4a686, f56a596547, and 55228a023b, with the authoritative tree path now published atomically. No current-main recut is justified.

PR #10749, opened 2026-08-25
fix(tui): preserve authoritative tabs across lazy attach

lawrencecchen:
Retain server-authoritative tree tabs during lazy startup and reconnect. Filter only explicit retired surface evidence, avoiding separate-mutex catalog publication races. Adds startup catalog-empty coverage.

---
## Summary by cubic
Preserves server-authoritative tabs across lazy startup and reconnect. Previously we filtered against the local surface catalog (which could be empty or stale), causing tabs to disappear and active_tab to point to nothing; now we drop only tabs with explicit retire evidence and clamp active_tab, preventing tab loss and flicker.

**Bug Fixes**
- Stop using the local catalog as a negative filter and retain tabs unless the server snapshot shows them retired.
- Re-prune with the latest retired set immediately before publishing the tree to fence races.
- Keep retire evidence based on raw server snapshot IDs, not the filtered view, to avoid prematurely clearing it.
- Add a test ensuring tabs persist when the local catalog is empty at startup.

<sup>Written for commit eb701b708a4dd402eb71f5b38b9dee5001edcb3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10749?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/session/remote.rs
+++ b/cmux-tui/crates/cmux-tui/src/session/remote.rs
@@ -3374,12 +3374,13 @@ impl RemoteSession {
             },
         );
         drop(capabilities);
-        // A tree snapshot can race a detach or overflow event. The surface
-        // catalog is authoritative for what can be rendered, so filter the
-        // snapshot while holding one catalog snapshot before publishing it.
-        let attached_surface_ids = self.surfaces.lock().unwrap().keys().copied().collect();
+        // The server tree is authoritative. The local surface catalog is a
+        // lazy mirror and may be empty during startup or reconnect, so it
+        // cannot be used as a negative filter. Remove only explicit retire
+        // evidence captured at the detach boundary.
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Closing this stale branch. Its authoritative-tab and lazy-attach intent is already present on current main in commits 81bd256cf4 and f0e5120d8c, with follow-up cleanup commits c4d7ee7520, e00dc4a686, f56a596547, and 55228a023b. No current-main recut is justified.

PR #10775, opened 2026-08-26
iOS: reclaim dismantled Ghostty surfaces

### LINKED ISSUE #7199: iOS: dismantled GhosttySurfaceView leaks (view<->bridge retain cycle kept for libghostty uiview pointer safety)
Found during review of https://github.com/manaflow-ai/cmux/pull/7150 (see the Codex inline finding on GhosttySurfaceBridge).

`GhosttySurfaceBridge` holds its `GhosttySurfaceView` strongly while the view owns the bridge, so a dismantled terminal's view + bridge + libghostty surface are never reclaimed by dismantle: `deinit` is the only caller of `disposeSurface()` and cannot run while the cycle exists (the render-pipeline recovery rebuild is currently the only path that frees a surface).

Making the back-reference weak is NOT safe on its own: `makeSurface` hands libghostty the raw view pointer via `ghostty_platform_ios_s(uiview: Unmanaged.passUnretained(self))`, and queued surface operations (render_now etc.) can dereference it after the view deallocates — the leak fix becomes a use-after-free. The bridge userdata is already protected across the async free (`Unmanaged.passRetained(bridge)` in `disposeSurface`), but the uiview pointer is not.

Proper fix needs lifetime choreography, e.g.:
- call `disposeSurface()` from `prepareForDismantle()` (not just deinit), and
- keep the view retained until `ghostty_surface_free` has drained the output queue (retained uiview userdata released in the free completion, mirroring the retainedBridge pattern), then break the bridge cycle.

Until then the strong back-reference is deliberate; see the comment in GhosttySurfaceBridge.swift.

austinywang:
Fixes #7199

GhosttySurfaceView now retires its libghostty surface from `prepareForDismantle`. The strong bridge back-reference remains intact for raw UIKit-pointer safety, while a main-actor retention token keeps the view alive through the FIFO surface queue and synchronous `ghostty_surface_free`. The token is released exactly once after free completion; recovery frees use the same path.

Regression coverage is split across two commits: a real surface dismantle test and deterministic repeated fake retention/free cycles.

Validation: Swift 6 parsing and git diff checks locally; full iOS package/simulator validation is delegated to the iOS workflow because this host has CommandLineTools but no Xcode.app. No merge requested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iOS terminal surface cleanup when views are dismantled.
  * Ensured surfaces remain valid during teardown and are released safely exactly once.
  * Made repeated cleanup operations safe and prevented lingering callbacks or retained views.

* **Tests**
  * Added coverage for dismantling terminal views, callback detachment, repeated cleanup, and surface retention behavior.

### DIFF
--- a/Packages/iOS/CmuxMobileTerminal/Sources/CmuxMobileTerminal/GhosttySurfaceBridge.swift
+++ b/Packages/iOS/CmuxMobileTerminal/Sources/CmuxMobileTerminal/GhosttySurfaceBridge.swift
@@ -6,12 +6,12 @@ import UIKit
 
 /// Bridges libghostty C callbacks (which run on the IO read thread or
 /// other Ghostty-internal threads) onto the main actor where the
-/// `GhosttySurfaceView` lives. The single mutable property is the
-/// `weak var surfaceView`; we serialise reads/writes through the main
-/// actor, which lets us conform to `Sendable` for the `Task { @MainActor }`
-/// hops below.
+/// `GhosttySurfaceView` lives. The deliberate strong back-reference keeps the
+/// raw UIKit pointer valid until dismantle detaches it; we serialise
+/// reads/writes through a lock because callbacks arrive off-main, which lets
+/// us safely perform the `Task { @MainActor }` hops below.
 final class GhosttySurfaceBridge: @unchecked Sendable {
-    // lint:allow lock — sanctioned carve-out: serial low-level primitive hidden behind the type, guarding a single weak ref on the libghostty-callback / typing-latency path; actor rewrite tracked as the GhosttySurfaceView split follow-up.
+    // lint:allow lock — sanctioned carve-out: serial low-level primitive hidden behind the type, guarding a single strong ref on the libghostty-callback / typing-latency path; actor rewrite tracked as the GhosttySurfaceView split follow-up.
     private let lock = NSLock()
     // Deliberately STRONG: libghostty holds the raw view pointer
     // (`ghostty_platform_ios_s.uiview`, passUnretained in `makeSurface`), so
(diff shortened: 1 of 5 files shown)


smundhra-git:
I hit the same dismantle leak (#7199) while auditing iOS memory. This PR covers it, so I haven't opened a second one. There are two cases where the queued `ghostty_surface_free` can still be lost. Both apply here and in render recovery:

1. **The queue is full.** `GhosttySurfaceWorkQueue.enqueue` refuses work once 256 operations are pending, and `enqueueSurfaceFree` ignores the `Bool` that `async` returns. When the free is refused, the C surface and the retained bridge are never released. The view-retention token is never released either, so the view leaks too. `pendingSurfaceFreeCount` also stays up, and recovery stops for good after 8 lost frees. A capacity-exempt enqueue used only for frees fixes this.
2. **The queue is released first.** `scheduleNext()` holds the work queue only weakly (`queue.async { [weak self] in guard let self else { return } ... }`), and the pending closures live in the instance's arrays. Recovery replaces `outputQueue` and keeps the old queue only in a local, so it can be freed before its worker block runs. When that happens the free is skipped. Having the free closure capture its queue strongly (for example `withExtendedLifetime(queue) {}` after the free) keeps the queue alive until the free has run.

For reference, the version I had locally:

```swift
func asyncBypassingCapacity(_ work: @escaping @Sendable () -> Void) {
    enqueue(work, priority: false, bypassingCapacity: true)
}
```

```swift
queue.asyncBypassingCapacity { [weak self] in
    let userdata = ghostty_surface_userdata(surface)
    ghostty_surface_free(surface)
    GhosttySurfaceBridge.releaseRetainedOpaque(userdata)
    withExtendedLifetime(queue) {}
    Task { @MainActor in self?.surfaceFreeDrainWatchdog.cancel(generation: generation); completion?() }
}
```

A related fix for the browser-stream pane leaks is in #14274.

PR #10798, opened 2026-08-26
Cloud VM observability: Axiom span export + server-side PostHog product analytics

lawrencecchen:
Two observability legs for the Cloud VM control plane (`web/`).

## Leg 1: Axiom trace export

`web/instrumentation.ts` now wires an `OTLPHttpJsonTraceExporter` (from `@vercel/otel`, already a dependency) into `registerOTel`, targeting `https://<AXIOM_DOMAIN|api.axiom.co>/v1/traces` with `Authorization: Bearer $AXIOM_TOKEN` and `X-Axiom-Dataset: $AXIOM_DATASET`. Gated on `AXIOM_TOKEN` + `AXIOM_DATASET`: with either absent the exporter is not constructed and `registerOTel` behaves exactly as on main. The VM routes' spans (`withAuthedVmApiRoute` / `withVmSpan`) already carry per-stage create timings as `cmux.vm.timing.<stage>_ms` attributes via `VmTimingRecorder`, so create latency is analyzable per stage once spans land in Axiom.

Live-verified against the preview dataset: the exact exporter class exported a span with `ExportResult {"code":0}` and a raw OTLP POST returned 200 with empty `partialSuccess`. Query-back was blocked because the ingest token has no query scope.

## Leg 2: server-side PostHog product analytics

All captures live in ONE module, `web/services/vms/productAnalytics.ts`, invoked from two places:

- The usage-event chokepoint `VmRepository.recordUsageEvent`/`recordUsageEvents` mirrors every persisted lifecycle event under its `eventType` name: `vm.created`, `vm.create.requested`, `vm.create.failed`, `vm.resumed`, `vm.destroyed`, `vm.attach` (with a derived `reattach` flag), `vm.snapshot.created`, `vm.forked`, `vm.create.credit.reserved`/`granted`/`refunded`/`grant_failed`, `vm.exec`, `vm.open_port`, `vm.ssh_endpoint`, `base.*`. `workflows.ts` gains only additive resume-timing hooks (see the follow-up section).
- Route handlers add what the chokepoint cannot know: `vm.create.completed` (outcome, http_status, duration_ms, per-stage `timing_*_ms`, provider, image, plan, memory), `vm.attach.completed` (reconnect vs fresh, duration, transport), `vm.limit_hit` (the free-plan paywall moment, with `upgrade_shown`), `vm.desktop.opened` (noVNC wrapper opens).

[...]

PR #10805, opened 2026-08-26
vm: Exit-based Effect boundary, crash-safe billing, status guards, Sentry capture

lawrencecchen:
Backend-only refactor of the Cloud VM Effect boundary plus crash-safety and Sentry error handling. Nine commits, each reviewable alone; wire shapes are byte-compatible with what VMClient.swift parses.

1. **Exit-based Effect boundary** (`417706e`): `runVmWorkflowExit` runs workflows with `Effect.runPromiseExit` + `Cause.failureOption`, returning `{ok, value} | {ok, error: VmWorkflowError}`; defects rethrow raw. `runVmWorkflow` now rejects with the tagged error itself, so the `effect/Runtime/FiberFailure/Cause` symbol-walking (`vmWorkflowErrorCause`) is deleted. Also fixes a latent bug: `VmSnapshotNotFoundError` was missing from the tag set, so a failed restore lookup surfaced as a 500 instead of its 404 mapping.
2. **Central create-error map** (`5902c1f`): the 409/500/402/402 create-like mapping that was copy-pasted in `vm/route.ts`, `restore`, `fork`, and `base/routeShared.ts` now lives in `vmCreateWorkflowErrorResponse` next to `vmWorkflowErrorResponse`; routes pass only verb-specific copy.
3. **bestEffort reporting** (`3bd2f4d`): ~27 `Effect.catchAll(() => Effect.void)` swallows (refunds, usage events, rollbacks) now go through `bestEffort(operation, context)`: scrubbed log, `cmux.vm.best_effort_failure` OTel counter, active-span event, Sentry capture via `reportError` (SENTRY_DSN-gated, ids only). A failed refund additionally writes a `vm.create.credit.refund_failed` usage event so lost refunds are ledger-visible.
4. **compensated() dedupe** (`91ff27d`): the create→fail→refund→markFailed→usage-event compensation copy-pasted across `createVm`/`finishBaseCreate`/`forkVm` is one helper; a failed compensation step no longer skips the remaining steps.
5. **Route-boundary Sentry** (`7352f14`): `withAuthedVmApiRoute` captures every non-tagged failure before shaping the 5xx, with route/method/operation/path/user id/team id context, never tokens or lease material.
6. **zod request schemas** (`fa36e77`): `services/vms/requestSchemas.ts` replaces the hand-rolled typeof validation for create, Base, exec, attach/sessions identifiers, PATCH rename, restore/fork fields. Accepted/rejected inputs, field precedence, byte bounds, and error envelopes are unchanged (verified against route tests plus an edge-case probe).
7. **Status guards** (`c264b36`): `markCreateRunning`/`markBaseCreateRunning` require `status = 'provisioning'`, so a slow create can no longer resurrect a row the reconcile cron marked destroyed (a phantom active VM that eats a plan slot); the guard miss flows into the existing compensation, which destroys the orphan provider VM and refunds. `markCreateFailed` skips destroyed rows. `recordLease`'s token-hash conflict path no longer sets `revokedAt: null`, so sign-out revocation can never be undone by a colliding lease write.
8. **Crash-safe create credits + stuck-row sweeper** (`c264b36`, `d4b2f1c`): new `cloud_vm_credit_reservations` table (migration `20260825010000`) records the intent to debit BEFORE the Stack debit RPC and advances pending → debited → committed/refunding/refunded/refund_failed with atomic `UPDATE ... WHERE status IN (...)` transitions, so the request and the cron can never both refund one debit. `reconcileCreditReservations` (runs in the existing `vm-reconcile` cron) commits debits whose VM runs, refunds debited/refund_failed rows whose create failed, skips in-flight creates, and reports pending rows whose debit outcome is unknowable (Stack debits are not idempotent or queryable per request) instead of guessing. `sweepStuckProvisioningVms` runs first in the same cron and fails provisioning rows without a `providerVmId` older than 30 min (crash between the insert and the finalize write); they previously counted against the 3-VM free limit forever while being invisible to provider-status reconciliation. The `provisioning_stuck` failure code is retryable so idempotent retries are not blocked.
9. **Modeled-5xx Sentry capture** (`3e13ec4`): mapped workflow failures that still return 5xx (provider outage 502, database outage 503, create-failed 500) now also reach Sentry, tagged with route, method, operation, and the error `_tag`; `reportError` promotes a fixed key allowlist (subsystem, route, operation, provider, errorTag, vmId) to searchable Sentry tags.

Tests: full `bash scripts/run-tests.sh` — 1448 pass / 0 fail. Full `CMUX_PORT=4342 bun db:test` (Docker Postgres, migrations applied twice, `CMUX_DB_TEST=1`) — all gated files pass, including the new `tests/vm-repository-guards.test.ts` (guards, sweeper, reservation transitions/races against real Postgres) and the 76-test `vm-workflows` suite. New unit files: `tests/vm-workflow-exit.test.ts`, `tests/vm-best-effort.test.ts`, `tests/vm-credit-reservations.test.ts` (reservation ordering, refund claim races, reconcile decision table, sweeper). `tsgo --noEmit` clean; eslint errors unchanged from baseline (16 pre-existing, none in touched files).

Known limits: a crash in the milliseconds between the Stack debit RPC and the debited-mark write leaves a pending row that is reported for manual review, not auto-refunded — Stack items expose no per-request debit ledger, so refunding blindly could mint free credits. The sweeper marks only the VM row failed; a Base generation swept this way needs a Base reset (existing recovery path). This PR does not change provider drivers, `providerGateway.ts`, or telemetry export.

Note for mergers: a sibling branch (#10798) touches `repository.ts` `recordUsageEvent` and `vmActiveLimitExceededResponse`; conflicts are expected to be resolved at merge time, this branch deliberately did not rebase onto it.

Dictionary:
- **Exit**: Effect's result value for a finished fiber, either success or a Cause describing the failure.
- **Cause**: Effect's failure tree distinguishing typed failures from defects (unexpected thrown bugs).
- **defect**: an error the workflow did not model; here it becomes a 500 plus a Sentry capture.
- **best-effort**: a side effect whose failure must not fail the parent workflow.
- **compensation**: the cleanup sequence that undoes a partially completed create (refund credit, mark row failed, record the failure event, destroy the orphaned provider VM).
- **reservation**: the durable Postgres row that tracks one create-credit debit from intent to settlement.
- **reconcile cron**: the scheduled `/api/cron/vm-reconcile` job that repairs state a crashed request left behind.

### DIFF
--- a/web/app/api/account/route.ts
+++ b/web/app/api/account/route.ts
@@ -65,7 +65,6 @@ import {
   VmAccountDeletionIdentityRevocationError,
   isVmAccountDeletionIdentityRevocationError,
   isVmProviderOperationError,
-  vmWorkflowErrorCause,
 } from "../../../services/vms/errors";
 import type { ProviderId } from "../../../services/vms/drivers";
 import { jsonResponse } from "../../../services/vms/routeHelpers";
@@ -905,8 +904,8 @@ async function destroyPersonalCloudVms(
 }
 
 function didVmDestroyReachProvider(error: unknown): boolean {
-  const workflowError = vmWorkflowErrorCause(error) ?? error;
-  return isVmProviderOperationError(workflowError) && workflowError.operation === "destroy";
+  // runVmWorkflow rejects with the tagged error itself; no unwrapping needed.
+  return isVmProviderOperationError(error) && error.operation === "destroy";
 }
 
 async function listAccountDeletionCloudVms(
--- a/web/app/api/cron/vm-reconcile/route.ts
+++ b/web/app/api/cron/vm-reconcile/route.ts
@@ -1,6 +1,8 @@
 import {
+  reconcileCreditReservations,
   reconcileVmProviderStatuses,
   runVmWorkflow,
+  sweepStuckProvisioningVms,
 } from "../../../../services/vms/workflows";
 
 
@@ -11,8 +13,17 @@ export async function GET(request: Request): Promise<Response> {
   }
 
   try {
-    const result = await runVmWorkflow(reconcileVmProviderStatuses());
-    return Response.json({ ok: true, ...result });
+    const providerStatuses = await runVmWorkflow(reconcileVmProviderStatuses());
+    // Order matters: sweep first so a crashed create's row is durably failed
+    // before the reservation reconciler decides whether to refund its credit.
+    const stuckProvisioning = await runVmWorkflow(sweepStuckProvisioningVms());
+    const creditReservations = await runVmWorkflow(reconcileCreditReservations());
+    return Response.json({
+      ok: true,
+      ...providerStatuses,
+      stuckProvisioning,
+      creditReservations,
+    });
   } catch (err) {
     console.error("[VM] cron status reconcile failed", err);
     return Response.json({ error: "vm_reconcile_failed" }, { status: 500 });
--- a/web/app/api/vm/[id]/attach-endpoint/route.ts
+++ b/web/app/api/vm/[id]/attach-endpoint/route.ts
@@ -6,6 +6,7 @@ import {
 } from "../../../../../services/vms/routeHelpers";
 import { setSpanAttributes } from "../../../../../services/telemetry";
 import { isVmNotFoundError } from "../../../../../services/vms/errors";
+import { optionalVmClientIdentifier } from "../../../../../services/vms/requestSchemas";
 import { openAttachEndpoint, runVmWorkflow } from "../../../../../services/vms/workflows";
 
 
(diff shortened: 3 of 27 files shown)


lawrencecchen:
Fleet instruction update for head `fbec544e008e441db8193710bf5c08b77d76e922`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #10811, opened 2026-08-26
Cloud VMs: single driver contract, shared cmuxd attach module, schema-validated provider JSON

lawrencecchen:
Generalizes the Cloud VM provider layer (web/services/vms/drivers/) so Blaxel can be swapped out easily: adding a provider is now one new driver file implementing `VmProviderDriver` plus a registry entry.

**Driver contract.** `types.ts` names the interface `VmProviderDriver` (old `VMProvider` kept as a deprecated alias) and [`drivers/CONTRACT.md`](../blob/feat-vm-provider-abstraction/web/services/vms/drivers/CONTRACT.md) documents the surface workflows actually use (create, fork, pause/resume, destroy, exec, attach/lease install, SSH, desktop/port ingress, status/stats, revocation), with per-method idempotency rules and the error taxonomy. `workflows.ts` already consumes only this interface through `VmProviderGateway`, so it needed no changes.

**Declared capabilities.** Every driver declares a `VmProviderCapabilities` matrix (`ssh`, `snapshot`, `fork`, `pause`, `getStatus`, `getStats`, `openPort`, `revokeEndpointLeases`), and the gateway gates optional operations on the declared flag before any provider round-trip instead of calling a method and catching `NotImplementedError` (how 3 of 4 drivers grew throw-only stubs). Each gate raises the same error the old stub produced, so external behavior is unchanged; `snapshot`/`restore` are now optional interface methods, Blaxel's throwing stubs are gone, and the gateway exposes `capabilities(provider)` so workflows/routes can branch on the matrix directly. `tests/vm-provider-capabilities.test.ts` pins the per-provider matrix, the flag-implies-implementation invariant, and each gate's error.

**One attach ritual.** `cmuxdAttach.ts` + `cmuxdConstants.ts` extract the previously copy-pasted cmuxd attach dance, parameterized by a `ProviderTransport` (exec into the guest): lease-path discovery from the daemon's flags, reusable-RPC-lease read, PTY+RPC lease minting, and the lease install, which is now atomic (temp file + `mv -f`, previously a direct `>` redirect the daemon could observe half-written). The `/healthz` retry loop, which existed in three hand-rolled idioms, is one Effect `Schedule` policy that rethrows the final probe error unchanged so Freestyle's SSH-fallback message classifier keeps working.

**Validated JSON.** `schemas.ts` puts zod at the driver boundary. `blaxelFetch` takes a schema per operation instead of `as T` casts, and Freestyle exec responses must carry a numeric `statusCode`. That fixes a real bug: a malformed exec response with no `statusCode` was treated as exit-0 success, so failed lease installs and repairs could look clean. Commits are split red→green: the first commit adds only the behavior-level regression test (stubbed provider HTTP response, fails on the old driver), the second turns it green.

**One scoped hardening from the prod audit.** Blaxel desktop/open-port preview tokens now live 24h instead of 7 days. The token rides in the `/vm/desktop/<id>?cmux_token=…` wrapper URL (browser history, edge logs) and keeps working until expiry, so the TTL is a real exposure window; the wrapper already shows an honest reopen screen on expiry, and reopening from cmux mints a fresh token. Sign-out and destroy already revoke desktop previews (`revokeEndpointLeases` deletes every preview including `port-*`; sandbox deletion kills the rest).

Audit findings verified but intentionally NOT fixed here (out of scope for this refactor):
- `/vm/desktop/[id]` is an unauthenticated capability URL by design; binding it to a Stack session is a product decision.
- `POST /api/vm/leases/revoke` invalidates the native-auth verification cache only on the serving instance; other warm instances can honor a revoked session for up to the 30s cache TTL (`web/services/vms/auth.ts`).
- `PREVIEW_ENDPOINT_LEASE_TTL_MS` (12h lease rows in `workflows.ts`) is now shorter than the 24h token; reconciling the two belongs with the workflows/billing changes on `feat-vm-effect-boundary`.

Duplication removed: `git diff --stat` shows 631 insertions / 720 deletions in the four drivers themselves; the four copies of the attach ritual and three healthz loops (~400 lines of drift-prone duplication) collapse into the two shared modules.

Coordination: this PR owns `drivers/` and the attach/health modules only. `errors.ts`, `repository.ts`, and the billing/compensation logic in `workflows.ts` are untouched, so it composes with the sibling `feat-vm-effect-boundary` branch.

Tests: `bun test tests/vm-cmuxd-attach.test.ts tests/vm-provider-schemas.test.ts tests/vm-freestyle-exec-statuscode.test.ts` (new, 54 tests) and `tests/vm-provider-capabilities.test.ts` (12 tests) plus the existing vm-* driver/workflow suites and `freestyle-cloud-shell-repair` all pass; `bun run typecheck` clean. The 14 failures seen in a whole-directory `bun test tests` run reproduce identically on `main` (cross-file interference in unrelated subrouter/coderouter/billing tests) and are not from this change.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10811"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790320820&installation_model_id=21920&pr_number=10811&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10811&signature=6467d79e5285dc4d977ec98cfe0b940583fd80acde8c6b29a034142912307f15"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added validation for virtual machine provider responses.
  - Improved handling of execution results, including timeouts and missing output fields.
  - Added a documented provider integration contract covering lifecycle, networking, snapshots, and error handling.

- **Bug Fixes**
  - Invalid or incomplete execution responses now produce clear provider errors instead of being treated as successful results.

- **Tests**
  - Added coverage for response validation, timeout handling, exit codes, and output normalization.

### DIFF
--- a/web/services/vms/drivers/CONTRACT.md
+++ b/web/services/vms/drivers/CONTRACT.md
@@ -1,0 +1,73 @@
+# VmProviderDriver contract
+
+`types.ts` defines `VmProviderDriver`, the single interface every VM provider implements
+(`blaxel.ts`, `freestyle.ts`, `daytona.ts`, `e2b.ts`). Workflows never import a concrete driver:
+they reach providers through `VmProviderGateway` (`../providerGateway.ts`), which resolves the
+driver by `ProviderId` and lifts each call into Effect. Adding a provider means one new driver
+file plus a registry entry in `index.ts` (and, to make it selectable, a kill-switch env key in
+`../config.ts` and an image entry in `../images/manifest.json`).
+
+## Shared modules a driver must use
+
+- `cmuxdConstants.ts` — cmuxd port, lease paths, TTLs, timeout clamps. Never re-declare these.
+- `cmuxdAttach.ts` — the attach ritual: lease-path discovery, reusable-RPC-lease read,
+  atomic lease install, endpoint assembly inputs, and the `/healthz` poll (`waitForCmuxdHealthy`,
+  the ONE retry idiom). A driver supplies only a `ProviderTransport` (`exec` into the guest).
+- `schemas.ts` — zod schemas for provider JSON. Every raw JSON payload from a provider API is
+  validated at the driver boundary; `as T` casts of provider responses are forbidden. Schema
+  mismatch raises `ProviderError` naming the provider and operation.
+- `wsLease.ts` — lease minting primitives (tokens, sha256 lease bodies, shell quoting).
+
+## Identity and state
+
+- `providerVmId` is the provider's own machine identifier and the key workflows store in
+  Postgres. Blaxel uses the human-friendly sandbox name; the others use SDK ids.
+- `providerMetadata` is an opaque, driver-owned bag persisted on the VM row and handed back on
+  later calls (`AttachOptions.providerMetadata`). Only the server may populate it; workflows
+  overwrite anything a client sent. Blaxel stores `sandboxUrl`, `previewUrl`, `homeVolume`,
+  `image`, `memoryMb` (resurrection needs them); Freestyle stores its daemon admin token.
+- `VMStatus` is the four-state cmux view (`creating | running | paused | destroyed`); each
+  driver maps its provider's richer state machine onto it and must map unknown states to
+  `running`, never throw.
+
+## Method semantics
+
+| Method | Semantics | Idempotency |
+| --- | --- | --- |
+| `create` | Provision a machine from a resolved image and return a running handle. Must roll back the provider resource if bootstrap fails partway (Blaxel deletes the sandbox but keeps the durable home volume). | Not idempotent; caller (workflow) owns dedup via the repository create-lock. |
+| `destroy` | Release the machine's compute and ingress. | Idempotent: an already-gone machine (provider 404) is success, not an error. Cleanup paths retry it. |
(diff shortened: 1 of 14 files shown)


lawrencecchen:
Fleet instruction update for head `83456907bdc973376d88a5e631b9ae6aa23bceb7`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #10800, opened 2026-08-26
Cloud cmux-tui daemon: transport spike + migration design

lawrencecchen:
Spike toward moving Cloud VMs off the Go `cmuxd-remote` daemon onto the cmux-tui remote daemon (manual-IO rendering on macOS, snapshot-based reconnects, drag-from-right-pane attach). Design: `docs/cloud-cmux-tui-daemon.md`. Spike automation: `scripts/spike-cmux-tui-blaxel.sh`.

## Proven end to end against a live Blaxel sandbox

1. Static `x86_64-unknown-linux-musl` `cmux-tui` (built from this branch's base `main` commit on a Blacksmith testbox, 1m47s warm) runs unmodified inside a `blaxel/base-image` microVM, injected through the same gzip+base64 filesystem channel `web/services/vms/drivers/blaxel.ts` uses today (chunked: the ~30 MB encoded payload exceeds the sandbox API body cap).
2. `cmux-tui server start --session cloud --remote-ws 0.0.0.0:1337 --remote-ws-insecure-bind` serves `/v1/link` behind the sandbox's **private** preview: the preview token rides as `?bl_preview_token=...` and the Rust ws dialer passes the URL through verbatim, so the single exposed HTTPS port needs no header support and no adapter. No token: 401.
3. Enrollment over that URL (invitation created in-VM, `remote connect --invite-file` from the Mac, approval in-VM), then the reconnect evidence:

```
{"type": "process-write-accepted", "process": "e979c629-...", "write_id": 2}
through_sequence: 10
'(none):~# echo SPIKE-MARKER-42 $(uname -m) $(date -u +%H:%M:%S)'
'SPIKE-MARKER-42 x86_64 05:56:32'          <- written before the client was SIGKILLed
'(none):~# echo RECONNECTED-AFTER-KILL $(date -u +%H:%M:%S)'
'RECONNECTED-AFTER-KILL 05:57:19'          <- fresh connection, same PTY (pid 54)
'(none):~#'
```

The state came back through `snapshot-process-terminal` (structured ghostty-vt rows + `through_sequence`), not raw byte replay. The interactive TUI (`remote connect`) was also driven over the same preview URL from a scripted PTY. An Aug-20 client binary interoperated with the `main`-tip daemon (both protocol 5).

## Local repro without Blaxel credentials

[...]

PR #10812, opened 2026-08-26
cmux Cloud: sessions run on the cmux-tui remote daemon (no cmuxd-remote on new machines)

austinywang:
Moves cmux Cloud sessions off the Go `cmuxd-remote` daemon onto the **cmux-tui remote daemon**, implementing the next open item of the design in #10800 (`docs/cloud-cmux-tui-daemon.md`): *wire the attach endpoint to inject and start the new daemon* — and, per the owner's call, making cmux-tui **the** session daemon rather than a side-by-side option. On by default (`CMUX_VM_CMUX_TUI_ENABLED=0` is the kill switch): new machines get cmux-tui only (no `cmuxd-remote` install or process), and `cmux vm shell` / the Machines panel open cmux-tui sessions, falling back to the websocket attach only when the control plane reports cmux-tui disabled (machines from before the migration). #10742 (manual-IO pump) remains the rendering half; until it lands the pane runs the real cmux-tui client TUI.

## Backend (Blaxel driver, workflows, routes)

- **Daemon on every machine**: the driver resolves the daemon build from the artifacts manifest the `cmux-tui-artifacts` workflow publishes — rolling `https://files.cmux.com/cmux-tui/latest/manifest.json` by default, `CMUX_VM_CMUX_TUI_MANIFEST_URL` to pin one commit's manifest for a deployment — and installs the static-musl `cmux-tui` it names onto the persistent home volume (`/root/.cmux/bin/cmux-tui`; the machine `curl`s it in-region and verifies the manifest's sha256 itself). No sha in env, nothing hand-rolled. It runs `cmux-tui server start --session cloud --remote-ws 0.0.0.0:1337 --remote-ws-insecure-bind` under the sandbox supervisor; daemon identity and enrolled devices live under `/root` so they survive resurrection; an installed daemon that no longer matches the manifest is reinstalled on attach. The smart-sleep watcher counts connections on 1337 and cmux-tui's terminal children, so idle machines still sleep.
- **`POST /api/vm/[id]/attach-endpoint` with `{"transport":"cmux-remote"}`** returns `{transport, route, token, expiresAtUnix, session, invitation?}`: `route` is `wss://<hash>.preview.bl.run/v1/link?bl_preview_token=…` (12 h token, hashed into the lease ledger like previews, so sign-out revokes it); `invitation` is a single-use `cmux://enroll/…` minted with `remote enroll create --ttl 300` only when the caller's `deviceFingerprint` is not among the daemon's enrolled devices.
- **`POST /api/vm/[id]/cmux-remote/approve {invitationId}`** approves the pending claim (`remote enroll pending` → `approve`), returning `state: pending|approved` plus the device fingerprint — the control plane is the daemon owner and minted the invitation for this authenticated user, so this is the honest encoding of "already authenticated". Pre-approved invitations in `cmux-remote` (Rust; `approval_required` is hardcoded) stay the follow-up named in #10800.
- **Measured gateway behaviour**: the branded `vm.cmux.sh` CloudFront distribution refuses WebSocket upgrades without a `User-Agent` (403), which the Rust dialer never sent; the raw `*.preview.bl.run` host does not enforce it. Fixed at the source (see Rust commit); the daemon preview is unbranded only until that build is in `latest`.
- Provider/gateway types: `CmuxRemoteEndpoint`, `openCmuxRemote?`, `approveCmuxRemoteEnrollment?`; unmigrated providers simply don't implement them. Revocation also stops the cmux-tui daemon process, matching cmuxd's semantics.

## Mac app + CLI

- **The app bundles the client.** `scripts/install-cmux-tui-client.sh` puts a universal, sha256-verified `cmux-tui` into `Contents/Resources/bin` next to the bundled `cmux` CLI and Ghostty helper — from the same manifest the backend uses — and `reload.sh`, `ci.yml`, `release.yml` and `nightly.yml` run it. The CLI looks there first (then `CMUX_TUI_CLIENT`, `~/.cmux/bin/cmux`, `cmux-tui` on PATH; every candidate must answer `remote-probe --json` as cmux-tui), and compares its `remote_protocol` with the machine daemon's (`daemonBuild` on the endpoint) before opening a pane, naming the stale side instead of hanging.

- Socket methods `vm.cmux_remote_info` / `vm.cmux_remote_approve` (listed in `capabilities`), `VMClient.openCmuxRemote` / `approveCmuxRemoteEnrollment`.
- **`cmux vm tui <id>`** opens a workspace whose pane runs the hidden `vm-tui-connect` helper: it hands the terminal to the local cmux-tui client (`remote connect <route> --invite-file … --state-dir ~/.cmuxterm/cmux-tui-client`, own process group in the foreground like the other interactive-child paths) and, while the client claims the invitation, approves the enrollment through the socket; the device fingerprint is remembered per machine in `~/.cmuxterm/vm-tui-devices.json` so later attaches skip the invitation. The client binary is found via `CMUX_TUI_CLIENT` → `~/.cmux/bin/cmux` → `cmux-tui` on PATH, and every candidate must answer `remote-probe --json` as cmux-tui (the SSH-remote bootstrap leaves a shell wrapper at `~/.cmux/bin/cmux` that is executable but not a client).
- Localized en+ja strings for the new CLI messages; `vm --help`, the usage contract and `docs/cli-contract.md` list `tui`.
- **`cmux-tui-artifacts.yml` publishes on main pushes again** (paths-filtered), so `latest/` tracks main; `latest` was republished by dispatch and now carries the musl target (it had been stale since July 15 with a glibc-named binary).
- **`cmux-remote` sends a `User-Agent` on direct WebSocket dials** (Rust, unit-tested): the CloudFront distribution on the branded `vm.cmux.sh` domain refuses upgrades without one (measured: 403 without, 101 with; `Origin` also 403s). Once this is in `latest`, the daemon preview goes back to the branded host (`<machine>-tui.vm.cmux.sh`); until then the driver keeps the raw preview host.

## Verified live on Blaxel (`smooth-wren`)

Through the driver against the real machine: install + `server status` ready → route on the raw host → `remote connect --invite-file --headless` from the matching macOS client → control plane approves in 2 s → `known-daemons` lists the cloud daemon → a second attach with the device fingerprint mints no invitation. Over the enrolled link, workspace RPC spawned a `bash` PTY, wrote a marker, and `snapshot-process-terminal` returned the styled grid: `PHASE1-TERMINAL-OK smooth-wren root`. Alpine/musl compatibility of the pinned binary was checked first (the older glibc `latest/` artifact fails there; `gcompat` doesn't rescue it).

## Tests

- `web/tests/vm-blaxel-cmux-tui.test.ts`: on by default with the `=0` kill switch; manifest URL defaults to `latest` and must be https; the manifest parser takes the musl build + sha256 and fails closed without a commit or musl target; install command verifies the pin before and after download, installs `curl` or falls back to `wget`, and installs onto `/root`; daemon command shape; invitation URI parsing and rejection of foreign/malformed URIs.
- `web/tests/vm-route-auth.test.ts`: `transport: "cmux-remote"` reaches `openVmCmuxRemote` with the account scope and device fingerprint and never the websocket workflow; unknown transports are rejected before any workflow; the approve route validates the invitation id (shell-unsafe input rejected) and passes the scope through.
- Existing Blaxel provider, workflow and route suites unchanged and green; `tsc` clean; CLI contract-help probes pass.

## Rollout

Opt-in: set the two env vars on a deployment to start shipping the daemon; no client changes for existing attaches. Phase 2 (default new attaches to `cmux-remote`, keep websocket as fallback for one release) and Phase 3 (delete the Go daemon path per provider, then `daemon/remote`) follow the sequencing in #10800; e2b/freestyle/daytona take the same musl binary in `build-cloud-vm-images.ts`.

Refs #10800, #10742.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2250,6 +2250,7 @@ jobs:
           ./scripts/install-prebuilt-ghostty-cli-helper.sh \
             ghostty-cli-helper/ghostty \
             build-universal/Build/Products/Release/cmux.app
+          ./scripts/install-cmux-tui-client.sh build-universal/Build/Products/Release/cmux.app
 
       - name: Validate Release artifact slices
         run: |
@@ -2260,6 +2261,9 @@ jobs:
           DIFF_SIDECAR="build-universal/Build/Products/Release/cmux.app/Contents/Resources/bin/cmux-diff-sidecar"
           test -x "$APP_BINARY"
           test -x "$CLI_BINARY"
+          TUI_CLIENT="build-universal/Build/Products/Release/cmux.app/Contents/Resources/bin/cmux-tui"
+          test -x "$TUI_CLIENT"
+          lipo "$TUI_CLIENT" -verify_arch arm64 x86_64
           test -x "$HELPER_BINARY"
           test -x "$DIFF_SIDECAR"
           file "$APP_BINARY" "$CLI_BINARY" "$HELPER_BINARY" "$DIFF_SIDECAR"
--- a/.github/workflows/cmux-tui-artifacts.yml
+++ b/.github/workflows/cmux-tui-artifacts.yml
@@ -11,7 +11,7 @@ name: cmux-tui artifacts
 # Layout in R2:
 #   cmux-tui/<commit-sha>/cmux-tui-<rust-target>   immutable, commit-addressed
 #   cmux-tui/<commit-sha>/manifest.json
-#   cmux-tui/latest/cmux-tui-<rust-target>         rolling, manual publishes only
+#   cmux-tui/latest/cmux-tui-<rust-target>         rolling, every main push
 #   cmux-tui/latest/manifest.json
 #   cmux-relay/<commit-sha>/cmux-relay-<rust-target> immutable, commit-addressed
 #   cmux-relay/<commit-sha>/manifest.json
@@ -26,7 +26,18 @@ name: cmux-tui artifacts
 #   chatmux-relay/latest/manifest.json
 
 on:
-  # Temporarily manual-only beginning 2026-07-13 to pause automatic CI/CD.
+  # Every main push that can change the binaries republishes the immutable
+  # commit-addressed artifacts and the rolling latest/ manifest: cmux Cloud
+  # machines install the daemon from latest/manifest.json and the macOS app
+  # bundles the matching client from it (scripts/install-cmux-tui-client.sh),
+  # so latest must track main. Manual dispatch stays for re-publishing.
+  push:
+    branches: [main]
+    paths:
+      - "cmux-tui/**"
+      - "ghostty"
+      - ".github/workflows/cmux-tui-artifacts.yml"
+      - ".github/workflows/cmux-tui-build-package.yml"
   workflow_dispatch: {}
 
 concurrency:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -483,6 +483,11 @@ jobs:
           install -m 755 nightly-inputs/ghostty/ghostty "$DEST"
           echo "Injected Ghostty CLI helper architectures: $(lipo -archs "$DEST")"
 
+      - name: Bundle the cmux-tui client
+        run: |
+          set -euo pipefail
+          ./scripts/install-cmux-tui-client.sh build-universal/Build/Products/Release/cmux.app
+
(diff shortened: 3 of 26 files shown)


lawrencecchen:
Superseded by merged [PR #10887](https://github.com/manaflow-ai/cmux/pull/10887), which contains the current Cloud resource/projection and remote-daemon baseline. Residual security and provisioning concerns remain tracked as follow-up debt. Closing this older stacked implementation.

PR #10832, opened 2026-08-26
Fix fish login-shell initial-command dispatch (#9822)

### LINKED ISSUE #9822: fish login shell: opencode attach viewer fails with "command not found" (dispatcher has no fish) branch)
# cmux bug: login-shell dispatcher has no `fish)` branch → `opencode attach` viewer (and other injected initial-commands) fail with `command not found` under a fish login shell

## Summary

When the user's login shell is **fish**, cmux's internal login-shell dispatcher falls through to a bare `/bin/sh` fallback branch. Any *initial command* injected through that path — most visibly the **`opencode attach` session viewer** — is then evaluated by a non-login, non-interactive `/bin/sh` that has **not** run `path_helper`, so it gets `PATH=/usr/bin:/bin` only. On Homebrew (Apple Silicon) or Nix installs, `opencode` lives at `/opt/homebrew/bin/opencode` (or a `/nix/store/...` path), neither of which is on that bare PATH, so the launch dies:

```
/bin/sh: opencode: command not found
Ghostty failed to launch the requested command
```

The user-visible symptom is: **the opencode viewer pane fails to open unless the pane is manually focused first** (focusing routes the command into an already-running login shell that has the full PATH, masking the bug).

## Environment

- macOS (Apple Silicon), nix-darwin + Determinate Nix, Homebrew present.
- Login shell: **fish** (`users.users.<u>.shell = pkgs.fish`), resolved via Nix to `/run/current-system/sw/bin/fish` → `/nix/store/.../fish-<ver>/bin/fish`.
- `opencode` installed via Homebrew: `/opt/homebrew/bin/opencode` (symlink into Cellar).
- cmux app bundle: `/Applications/cmux.app`; launch templates compiled into `Contents/MacOS/cmux` (48 MB Mach-O), **not** in editable JS resources.

## Root cause

The dispatcher (strings extracted from `Contents/MacOS/cmux`) branches on the basename of `CMUX_LOGIN_SHELL`:

```sh
CMUX_LOGIN_SHELL="${SHELL:-/bin/zsh}"
case "${CMUX_LOGIN_SHELL##*/}" in
  sh|dash|ksh|mksh|ash|yash|posh)  exec ... "$CMUX_LOGIN_SHELL" -i -c 'eval "$1"; exec "$0" -i' ... ;;
  csh|tcsh)                        exec ... "$CMUX_LOGIN_SHELL" -i -c 'eval "$argv[2]"; exec "$argv[1]" -i' ... ;;
  nu|nushell)                      exec ... "$CMUX_LOGIN_SHELL" --execute "$cmux_initial_command" ;;
  pwsh|powershell)                 exec ... "$CMUX_LOGIN_SHELL" -NoExit -Command "$cmux_initial_command" ;;
  *)                               exec ... /bin/sh /bin/sh -c 'eval "$1"; exec "$2" -i' cmux-initial-command "$cmux_initial_command" "$CMUX_LOGIN_SHELL" ;;
esac
```

There is **no `fish)` branch.** fish therefore falls into `*)`, which runs `eval "$cmux_initial_command"` in a bare `/bin/sh` and only then `exec`s the login shell. For an interactive pane the trailing `exec fish -il` eventually loads the user's PATH (which is why agent panes work). But for a **one-shot injected command** like `opencode attach ...`, `opencode` is invoked in the `/bin/sh` step *before* fish is ever exec'd, with `PATH=/usr/bin:/bin`.

### Why stock zsh/bash users don't see it
A default macOS login shell (zsh) runs `/etc/zprofile` → `path_helper`, which folds `/etc/paths` + `/etc/paths.d/homebrew` (i.e. `/opt/homebrew/bin`) into PATH. So even the cold shell resolves `opencode`. fish + the `*)` fallback denies that entirely; Nix amplifies it because the toolchain lives under `/nix/store` and `/opt/homebrew`, none of which the bare fallback PATH reaches.

## Reproduction

On a fish login shell (Homebrew opencode), with no pane pre-focused, trigger the opencode-attach viewer → `command not found`.

Minimal repro of the failing environment (bare `/bin/sh`, no path_helper):

```sh
# FAILS — mimics the *) fallback's /bin/sh with the OS default env:
env -i /bin/sh -c 'command -v opencode || echo NOTFOUND'   # -> NOTFOUND (PATH=/usr/bin:/bin)

# WORKS — the zsh branch a stock user would take:
env -i /bin/zsh -lc 'command -v opencode'                  # -> /opt/homebrew/bin/opencode (path_helper ran)
```

## Latent sibling bug (same gap)

Because fish falls through to `*)`, `eval "$cmux_initial_command"` is executed by **`/bin/sh`, not fish**. Any initial command using fish-specific syntax is silently POSIX-sh-evaluated and will misparse/fail. So the missing branch is not only a PATH issue — it's a shell-semantics mismatch for every injected initial command under fish.

## Suggested fix

Add a `fish)` branch mirroring the interactive path that already works (the agent-pane launcher uses `fish -il --init-command 'source "$CMUX_FISH_INTEGRATION_FILE"'`). Something equivalent to:

```sh
  fish)  exec ... "$CMUX_LOGIN_SHELL" -il -c "$cmux_initial_command" ;;
```

so the injected command runs in a **login-interactive fish** with the user's real PATH and fish semantics — instead of a bare `/bin/sh`. This closes both the `command not found` bug and the POSIX-vs-fish eval mismatch, and lets fish users drop the `launchctl config user path` workaround.

## User-side workaround (until fixed)

Set the launchd user-domain PATH so GUI apps (and thus cmux's cold `/bin/sh` children) inherit Homebrew:

```sh
/bin/launchctl config user path "/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
```

Applies at next boot; the fix is only inherited by cmux instances launched *after* it takes effect (relaunch cmux after the reboot). On nix-darwin, place this in `system.activationScripts.postActivation.text` so it's reapplied on every `darwin-rebuild switch`.

---

🤖 Generated with [OpenCode](https://opencode.ai) (Claude Opus 4.8)

austinywang:
## Summary

- Run remote initial commands through a login-interactive fish shell when the fallback dispatcher sees fish.
- Add a runtime regression covering fish-only syntax, login-provided PATH, and one-shot command claiming.
- Install fish only for the focused remote-initial-command test workflow so the gated test executes on hosted runners.

Fixes #9822.

## Verification

- Hosted suite run passed: https://github.com/manaflow-ai/cmux/actions/runs/32957812745 (4 tests, including fishFallbackRunsInitialCommandAfterLoginStartup).
- swiftc -parse Sources/RemoteInitialCommandBootstrap.swift
- swiftc -parse cmuxTests/RemoteInitialCommandBootstrapFailureTests.swift
- ruby YAML parse of .github/workflows/test-e2e.yml
- git diff --check
- The requested Swift file-length budget script/table are absent from this checkout; no policy files were recreated.

Commits remain split into test-only then fix commits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote session startup when using the Fish shell, ensuring initial commands run correctly after login initialization.
  * Prevented initial commands from running more than once during repeated startup attempts.

* **Tests**
  * Added coverage for Fish shell startup, command execution, and one-time command consumption.
  * Updated end-to-end test setup to verify Fish shell availability.


austinywang:
Reviewed the embedded pre-merge findings in the earlier [CodeRabbit report](https://github.com/manaflow-ai/cmux/pull/10832#issuecomment-5423488732). They were generated for the superseded review range `1f98632eb7d30f0d371c72e78e8e8ca484486872` → `1188fecafd926d3ae4791cf3dc80c4fc614920a7`, while the live PR is now head `21c59f4b610cc7927850f6e99fbcd6594bdb4f1e` on base `b40cadee00cabed65d24ad17116ca4d9a912f586`; live review state is 0 threads / 0 inline comments and the required policy checks pass. The package-extraction and docstring suggestions are advisory and would expand this focused fish-dispatch fix, so no code change is needed. The remaining non-code blockers are the two pending Vercel deployments.

PR #10833, opened 2026-08-26
Filter modifier-hold shortcut hints by held modifier

### LINKED ISSUE #9842: Modifier-hold shortcut hints should be filtered by the held modifier
With `shortcuts.showModifierHoldHints` enabled, holding **either** Cmd or Control shows hint chips for everything at once — so holding Cmd overlays the surface tab strip with `⌃1…⌃9` chips. The information is correct (surfaces are on Control-digits, workspaces on Cmd-digits by default), but the moment is misleading: while holding Cmd, a chip that reads `⌃2` suggests the wrong continuation of the chord the user has already started, and pressing the shown digit does something else (selects workspace 2, not surface 2).

**Expected:** chips shown while a modifier is held are filtered to bindings that begin with that modifier — hold Cmd, see only `⌘…` completions (workspace rows); hold Control, see only `⌃…` completions (surface tabs). Rebound shortcuts would follow their configured modifiers automatically.

**Current:** both modifiers trigger the full map regardless of which one is held.

Happy to take a swing at a PR if the direction sounds right — same author as #9801.

austinywang:
## Summary

- Filter Bonsplit surface-tab modifier-hold hints by the held Command/Control family.
- Preserve the full configured modifier label for custom numbered-surface bindings.
- Fixes #9842. The dependency change is in [Bonsplit #225](https://github.com/manaflow-ai/bonsplit/pull/225).

## Testing

- Regression policy is split into two commits: `cba4103e87` (test) then `93ea8888f6` (fix).
- Before the fix, the focused policy run was red with 5 failures.
- After the fix, the focused policy run passed 4/4 cases.
- `swift test --disable-sandbox` in `vendor/bonsplit`: 222 XCTest cases and 27 Swift Testing cases passed.
- Parent checks passed: package-resolved policy, workspace package grouping, pbxproj normalization, pbxproj test wiring, Swift warning-budget guard, and CI change-area filter.
- No local Xcode/XCUITest or tagged launch was run, per the recovery dispatch restrictions.

The repository no longer contains `.github/swift-file-length-budget.tsv` or `scripts/swift_file_length_budget.py` (the budget was removed on main in #8125), so that requested audit command is unavailable; no parent Swift files were changed, and the two touched Bonsplit Swift files have no applicable parent budget row.

## Demo Video

Not recorded: local UI launch is explicitly prohibited for this recovery session.

## Review Trigger

Not requested; background review agents are disabled by the task instructions.

## Checklist

- [x] Tested the change locally
- [x] Added or updated tests for the behavior change
- [x] Updated docs/changelog if needed (no user-facing strings changed)
- [ ] Dogfood/UI video (blocked by explicit no-launch restriction)
- [ ] Merge (requires user approval; do not merge)

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled Bonsplit component to a newer version.
  * No direct user-facing feature or behavior changes.


austinywang:
recheck

PR #10856, opened 2026-08-26
fix(portal): ask bonsplit which pane owns a tab instead of scanning

STRML:
**Status:** updated onto `main` (`78f4c82504`) as merge commit `75b1f9a63f`. No conflicts.

**Reviewer's guide**
- Three call sites, two shapes: `Workspace.paneId(forPanelId:)`, `Sources/Panels/TerminalPanelView.swift`, `Sources/WorkspaceContentView.swift`.
- Both replacement APIs (`paneId(containing:)`, `selectedTabId(inPane:)`) are already in the pinned bonsplit, and `DockSplitStore` and `PaneDropContainer` already use them.
- Behaviour is unchanged; the diff is about not reading `TabItem` fields inside a view update.

---

Portal ownership checks were reading tab titles. They run inside a SwiftUI update, so an animating spinner title invalidated every terminal surface in the window. This stops that.

Three call sites, two shapes.

`Workspace.paneId(forPanelId:)` walked every pane:

```swift
bonsplitController.allPaneIds.first { paneId in
    bonsplitController.tabs(inPane: paneId).contains(where: { $0.id == tabId })
}
```

`tabs(inPane:)` maps each `TabItem` through `Tab.init(from:)`, which copies all fourteen properties, `title` among them. Reading a title inside a view update subscribes that update to the title, so a scan looking for one id subscribed the caller to every tab in the window.

Bonsplit already keeps a pane-ownership index. `paneId(containing:)` is a dictionary lookup and touches no tab fields. `DockSplitStore.paneId(forPanelId:)` was already calling it.

The other two sites asked `selectedTab(inPane:)?.id == tabId` when the id was all they wanted. `selectedTabId(inPane:)` answers that without building a `Tab`.

## Measurements

Instruments SwiftUI template, 20 seconds, attached to a Release build with eight agent tabs writing spinner frames into their titles.

| | before | after |
|---|---|---|
| `currentPortalPaneOwner` | 75 ms | 0.1 ms |
| `Workspace.paneId(forPanelId:)` | all of the above | 0.1 ms |

The after capture had the workload to do: 560 `TabItem.title` invalidations in 20.7 seconds, about 27 a second. The call site just stopped answering them.

What this doesn't claim. Process CPU didn't move measurably, because 75 ms out of a 5000 ms capture is under 2%. And the 68 ms of `tabs(inPane:)` still in the after capture is real, but every sample of it comes from `Workspace.sidebarOrderedPanelIds()`, which has the same shape and is its own change.

Behavior is unchanged. `paneId(containing:)` and the scan return the same pane, and `selectedTabId` returns the id `selectedTab?.id` would have.

### DIFF
--- a/Sources/Panels/TerminalPanelView.swift
+++ b/Sources/Panels/TerminalPanelView.swift
@@ -207,7 +207,9 @@ struct TerminalPanelView: View {
               let tabId = workspace.surfaceIdFromPanelId(panel.id) else {
             return false
         }
-        return workspace.bonsplitController.selectedTab(inPane: currentPane)?.id == tabId
+        // See the resolver in WorkspaceContentView: selectedTab would
+        // subscribe this update to every tab title in the pane.
+        return workspace.bonsplitController.selectedTabId(inPane: currentPane) == tabId
     }
 
     private var effectiveTerminalAgentContext: String {
--- a/Sources/Workspace+WorkspaceSurfaceTreeReading.swift
+++ b/Sources/Workspace+WorkspaceSurfaceTreeReading.swift
@@ -18,11 +18,17 @@ extension Workspace: WorkspaceSurfaceTreeReading {
         paneTree.surfaceId(forPanelId: panelId)
     }
 
+    /// The pane that currently holds this panel's surface.
+    ///
+    /// Bonsplit keeps a pane-ownership index, so ask it. The scan this replaced
+    /// built a `Tab` for every tab in every pane, and `Tab.init(from:)` copies
+    /// all fourteen `TabItem` properties. Run inside a SwiftUI update, as the
+    /// portal-ownership resolver does, that subscribed the update to every tab
+    /// title in the window, so an animating title invalidated the terminal
+    /// surfaces. `DockSplitStore.paneId(forPanelId:)` already did it this way.
     func paneId(forPanelId panelId: UUID) -> PaneID? {
         guard let tabId = surfaceIdFromPanelId(panelId) else { return nil }
-        return bonsplitController.allPaneIds.first { paneId in
-            bonsplitController.tabs(inPane: paneId).contains(where: { $0.id == tabId })
-        }
+        return bonsplitController.paneId(containing: tabId)
     }
 
     func indexInPane(forPanelId panelId: UUID) -> Int? {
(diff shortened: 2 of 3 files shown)


STRML:
One follow-up this doesn't cover, in case someone wants it before I get there.

`tabs(inPane:)` still costs 68 ms in the after capture, and 676 of its 677 samples come from a single caller:

```
closure #1 in Workspace.sidebarOrderedPanelIds()
```

Same shape as `paneId(forPanelId:)`. It wants ids and builds `Tab` values to get them. Separate change, and I'd rather land this one first.

The other thing worth recording while the trace is open, because it moves where the next win is: the sidebar is now the largest single thing in the profile at 21.4% (1059 ms of 4944 ms). It was 0.1-0.6% in every capture I took before the tab bar work.

```
3480  VerticalTabsSidebar.appKitWorkspaceScrollArea(renderContext:)
2711  VerticalTabsSidebar.workspaceTableRowConfiguration(_:input:listSnapshot:renderContext:)
1315  closure #3 in VerticalTabsSidebar.appKitWorkspaceTableRows(renderContext:)
 797  VerticalTabsSidebar.workspaceTableActions(renderContext:)
```

Top leaf frame in the whole trace is `initializeWithCopy for VerticalTabsSidebar` at 215 ms, so that view struct is large enough that copying it costs real time.

STRML:
Rebased onto `main` (78f4c82504) and re-verified: merges without conflicts now, and the description opens with a reviewer's guide (files to read first, which tests cover it). The only red check is CLA, which I'll sign. @lawrencecchen would you take a look when you get a moment? Same idle-work family as #12009, #12023, #10351.

STRML:
recheck

STRML:
Rebased onto `main` (`0be8ab34ad`, 2 commits, linear). Force-pushed `fix/portal-owner-avoids-tab-title-observation` @ `d78d3d04b25`. Siblings: #12009 #11483 #5848.

STRML:
Rebased onto `main` (623e5bcbb80), 2 commits, linear. Force-pushed fix/portal-owner-avoids-tab-title-observation @ 854648d1b37 (supersedes the 0be8ab34ad stack).

STRML:

STRML:

Issue #9178, opened 2026-07-29
iOS: session whose iroh paths all closed is never marked closed (no sessionClosed, stale pool entry)


azooz2003-bit:
In the 2026-07-29 13:03 outage export (cmuxdiag, iOS client, build 1.0.4/20260729110215), the foreground control session established at 13:01:10 lost both of its iroh paths at 13:03:14 (`transportPathEvent closed` for relay and privateNetwork, then `selectedPathChanged unknown`), but no `sessionClosed`, `transportCloseAttribution`, or session-lifecycle removal event was ever emitted for it during the remaining 73 s of the log. The QUIC session object stayed in the local pool as a corpse while recovery dialed fresh sessions around it.

Expected: when a pooled session's last path closes and no path reopens within a bounded grace window, the owner should treat the session as dead: emit its close attribution and `sessionClosed`, evict it (`closedSessionEvicted` or `explicitlyInvalidated`), and let recovery proceed against a clean pool.

Note this is the phone-side sibling of the Mac-side runtime wedge fixed in https://github.com/manaflow-ai/cmux/pull/8930; that PR does not touch the client session pool. Same incident evidence as https://github.com/manaflow-ai/cmux/issues/9166 and https://github.com/manaflow-ai/cmux/issues/9167.

PR #10858, opened 2026-08-26
iroh host: one peer stalled mid-handshake must not wedge the accept pipeline

lawrencecchen:
## Problem

During the 2026-08-26 itest simulator timing batch (31 launches, relayOnly through the GCP test relay, branch feat-iroh-integration-test @ 84050588d0), the Mac host repeatedly became unreachable through the relay after a client session ended abnormally (sim killed mid-session, host logged `Transport idle timed out`), and never recovered until the host app was restarted. It cost 15/26 warm launches their Mac session.

The relay's own logs rule out relay attachment: across the cycle-5 wedge onset the host's relay TCP connection was attached at 11:48:42Z and stayed up unchanged until 11:58:52Z, yet launches at 11:53:27Z and 11:56:11Z never got a session while 11:49-11:52 launches connected in ~35s. The client logs show route discovery succeeding (host still broker-registered, 21 bindings) and one last QUIC leg connecting in 205 ms at 11:10:47Z before every subsequent handshake timed out. The host was discoverable, relay-attached, and not accepting.

Root cause: `CmxIrohLibEndpoint.accept()` completed the entire server-side handshake (`Incoming.accept`, ALPN read, handshake completion) inline in `CmxIrohEndpointServer`'s single serial accept loop. One peer that initiates a connection and then stops making handshake progress (killed app, dead relay path) suspends the loop for the driver's full handshake timeout, while the retrying client refills the accept queue with attempts it has already abandoned by the time the host reaches them. The loop never catches up; occasional 80-240s successes were dials that happened to hit a momentarily empty queue. This is the host-side sibling of the reconnect-wedge family (https://github.com/manaflow-ai/cmux/issues/9178 is the phone-side session-corpse variant; https://github.com/manaflow-ai/cmux/pull/8930 fixed broker-rejection recovery but not accept-path ownership).

## Mechanism

`CmxIrohEndpoint.accept()` now returns an un-handshaken `CmxIrohIncomingConnection`. Its `establish()` runs inside the same per-connection admission task that owns the rest of that connection's lifetime, bounded by the existing 15s admission deadline and the driver's own handshake timeout. The accept loop's only job is draining the accept queue. Identity-scoped capacity checks move to post-handshake registration, where the identity is TLS-verified; the global pending-admission cap (10) still bounds pre-handshake work, and all existing close-reason strings are preserved. An accept queue that ends while its generation is still current now routes through endpoint recovery with the existing bounded backoff instead of the loop dying silently under a still-published binding.

Principled or hacky: principled; it fixes the ownership defect (a connection's handshake belongs to that connection's task, not to the shared accept loop) rather than adding a watchdog. Residual risk: Swift task cancellation may not unwind an in-flight FFI handshake future, so a timed-out establish can linger until the driver's own timeout resolves it; it holds only its own slot and is closed on resolution.

## Tests

Two commits so CI proves the test catches the bug: commit 1 adds `CmxIrohEndpointServerStalledHandshakeTests` (a peer stalled mid-handshake, healthy peer queued behind it must still be admitted) which fails on the old pipeline; commit 2 adds the fix. Full package suite: 618 tests / 69 suites pass.

## Base caveat

Branched from `origin/feat-iroh-integration-test` (ideal-shape series), so this PR shows that branch's commits until it merges; the wedge fix itself is the top two commits.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10858"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790341052&installation_model_id=21920&pr_number=10858&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10858&signature=73467097d8bf499ae96a1899ae5d7581747ed81d5d8b6b24ead0bd49f178ac4b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added signed relay-policy retrieval without client-held relay credentials.
  * Added relay attach/detach reporting and relay status diagnostics.
  * Added configurable connection dial timeouts and improved stalled-handshake handling.
  * Added limited reuse of recently expired, verified cached policies.
  * Added debug relay override support for development environments.
  * App termination is now scheduled safely through the main run loop.

* **Changes**
  * Simplified release verification to focus on standard connectivity checks.
  * Removed legacy relay-token and offline-pairing flows.
  * Incoming connections now complete handshakes through an explicit admission step.


lawrencecchen:
Triage note from the review-triage pass that produced https://github.com/manaflow-ai/cmux/pull/10865, recorded here so future reviewers stop re-flagging it: the "host serves under an unconfirmed cached policy / unconfirmed-policy admission" finding is a FALSE POSITIVE.

Disproof: `validateCachedPolicy` re-verifies the signed endpoint attestation for the exact account/identity/endpoint/settings tuple and rejects after the signed expiry (`CmxIrohHostRuntime+PolicyRefresh.swift:355-386`), so the cache-first path never serves under a policy the broker did not sign for this binding. Cache-first `start()` schedules the authenticated broker resolve immediately, and that refresh fails closed: a non-transient broker rejection tears the runtime down to `.failed` (`CmxIrohHostRuntime+PolicyRefresh.swift:652-668`). The remedy the finding proposes is the pre-existing `cachedPolicy(after:)` fallback, so offline exposure is unchanged by cache-first start. Net delta is a one-broker-round window while the broker is reachable, and a revocation inside that window stops the host.

PR #10659, opened 2026-08-24
settings: make the Pairing Port row describe both listeners (TCP pairing + Iroh UDP)

azooz2003-bit:
The Pairing Port preference in Settings > Mobile is bound by two listeners: the legacy TCP pairing listener and the Iroh endpoint's UDP socket, which is the port iOS Direct addresses dial. The Settings copy claimed TCP only, the Apply note said connected devices reconnect on the new port (true for the TCP listener, while the Iroh endpoint only adopts the new port on its next activation, in practice app relaunch), and the Reachable-at diagnostics dropped the Iroh route entirely because it is a peer endpoint rather than host:port.

This PR makes the section reflect the true state:

- Port row subtitle names both bindings: pairing listener (TCP) and the Iroh endpoint that Direct addresses dial (UDP).
- The Apply note states the split: the pairing listener rebinds now, the Iroh endpoint adopts the new port the next time cmux starts.
- The live status strings are scoped to the pairing listener so they cannot be read as covering the UDP bind.
- The Reachable-at list now renders the Iroh route's usable direct-address path hints as Iroh rows (deduplicated, expired hints dropped), so the UDP port actually in effect is visible, including when the preferred port was taken and iroh silently fell back to an ephemeral one.
- The stale routes-empty hint no longer claims Tailscale is required for network pairing.
- Doc comments on the catalog key, `configuredPort`, and the settings snapshot models updated to match; en and ja localizations updated for every changed string.

Known limitation: the Iroh rows show the broker-registered public-disclosure hints, so LAN-scoped addresses do not appear there; surfacing the local bind directly would need new plumbing from the Iroh runtime into the status cache and is left as a follow-up.

Localization audit: changed keys `settings.mobile.port.subtitle`, `settings.mobile.port.note`, `settings.mobile.port.apply.inUse`, `settings.mobile.port.apply.saved`, `settings.mobile.port.status.ok`, `settings.mobile.port.status.fallback`, `settings.mobile.routes.empty` in both en and ja; no new keys.

Tests: `MobileHostServiceSettingsTests` gains coverage for the snapshot mapping (host:port rows preserved, Iroh direct-address rows added, dedupe, expired-hint filtering) and the socket-literal parser.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #10378, opened 2026-08-19
Replace the mobile transport: CmuxPeerTransport on stock upstream iroh, zero forks

azooz2003-bit:
Tears down the forked transport stack and replaces it end to end.

**Deleted:** `Packages/Shared/CmuxIrohTransport` (29k src + 31k test lines), the entire fork dependency chain (`manaflow-ai/iroh-ffi v1.0.2-cmux.7` → `manaflow-ai/iroh` @4152d81 + `manaflow-ai/noq`), the two-phase admission barrier and candidate-deferral machinery those forks existed for, and the iroh release-gate scripts/workflow. Net: −62k lines against +30k.

**Replacement:** `Packages/Shared/CmuxPeerTransport` on byte-identical upstream `n0-computer/iroh-ffi v1.1.0` binaries (checksum-pinned release zip, rewrapped framework-style by `scripts/ensure-iroh-xcframework.sh` so its module map cannot collide with GhosttyKit's; generated Swift bindings vendored unmodified from the same tag — no fork repos, no fork builds, no binary hosting). Protocol is `cmux/mobile/2`: single-phase admission — client presents its Ed25519 pair grant on the first control stream, host verifies + broker-revalidates (same ≤30s cadence, sticky online denials, offline continuation only on broker connectivity failure), one ack, then the control stream carries the unchanged length-prefixed JSON RPC. Same lane model (control / serverEvents+cursor / terminal+cursor / artifact+offset).

**Why:** field logs (Aug 13–18) showed median launch→ready 12.7s (p90 105s), 286 dials racing endpoint activation, 35/57 reconnect failures "superseded by a newer attempt", and a 2s retry loop that kept dialing while connected. Each pathology now has a structural owner, pinned by behavior tests: a single-owner `PeerConnectionSupervisor` (automatic triggers join in-flight attempts, explicit intent replaces, success cancels the ladder by construction, denials sticky, inactive-scene triggers park), an awaitable endpoint-activation barrier, account-scoped broker cooldowns that outlive teardown, decorrelated-jitter backoff, generation fencing after every await, and single-flight discovery with failure-classified staleness. 184 package tests in 23 suites, including a live two-endpoint QUIC loopback that drives dial → admission (accept and deny) → RPC bytes over the `CmxByteTransport` seam → terminal and server-event lanes → close attribution.

[...]

PR #10859, opened 2026-08-26
Remove the iroh transport end to end: Tailscale/TCP is the only connectivity path

azooz2003-bit:
Erases all iroh code from the iOS app, the macOS host, the web backend, the presence worker, and CI, per the decision to rebuild the transport from scratch without bias from the current implementation. 893 files, -146k lines. Only the TCP path remains: the phone dials the Mac's TCP pairing listener directly over Tailscale (or loopback for simulators), authorized by the existing Stack same-account gate or a user-entered pairing code.

What changed on the wire: `CmxAttachTransportKind` has no `iroh` kind and `CmxAttachEndpoint` has no `peer` case. Compatibility with already-deployed peers comes from tolerant decoding (`CmxTolerantAttachRouteList`): tickets, registry payloads, presence pushes, and persisted stores drop unknown route kinds per element instead of failing. Old Macs' iroh routes are ignored; a retired v3 (iroh) pairing QR decodes as an unrecognized version and shows update-the-app copy. "Tailscale Only" previously rode an iroh QUIC lane pinned to the tailnet address; `.tailscale` now always means the raw authorized TCP dial. Persisted per-Computer methods ("iroh", "automatic", "direct") read as Tailscale so no stored pairing is lost.

Load-bearing pieces were relocated, not deleted: the connection-exit bookkeeping is now `MobileHostConnectionExit`, the transport diagnostics ring is `MobileHostDiagnostics.hostDiagnosticLog`, and `CmxConnectivityInvalidationSubscriber` moved into CMUXMobileCore because the phone inline-reply inbox rides the presence worker's `connectivity.invalidate` nudge; that worker lane and `replies.ts` are untouched. The iroh-only QUIC lanes (independent server events, terminal, artifact, simulator-stream v2) are gone on both sides; their existing RPC/control-lane fallbacks are now the only path. Same-device evidence keeps reading the legacy iroh-named keychain service through the neutral `KeychainIdentityEvidenceProbe` so durable device identity survives the upgrade (the one intentional iroh string left).

Backend: `/api/devices/iroh/*`, `/api/relay/*` (including the #10730 access-control hook), `/api/connectivity/{v2,v3}/sync`, the trust-broker/relay/connectivity services, and the Rust `iroh-relay-minter` are deleted; a forward migration drops the seven iroh tables; Vercel builds no longer gate on the relay catalog. The device-registry API and presence worker drop `kind:"iroh"` routes and pass all other kinds through opaque. Released clients lose these endpoints on deploy and degrade to the TCP path, which is the intended end state. Attach tooling gates flipped from iroh-route-required to tailscale-route-required with the same fail-closed semantics.

Deliberately out of scope: `cmux-tui`'s own iroh transport (`cmux-remote`'s default-on `iroh-transport` feature, `--iroh` daemon flag). It is a separate surface needing its own pass. Historical DB migrations, CHANGELOG, and retired design docs (now header-noted) keep their iroh references. This PR conflicts by design with the open iroh PRs (https://github.com/manaflow-ai/cmux/pull/10629, https://github.com/manaflow-ai/cmux/pull/10378, https://github.com/manaflow-ai/cmux/pull/10737, https://github.com/manaflow-ai/cmux/pull/10739).

Verification: fleet macOS Debug build green (tag noiroh); web typecheck green and web suite 1204 pass / 0 fail; presence worker 191 pass / 0 fail including inline-reply and connectivity-invalidation coverage; Swift package suites pass locally with every remaining failure reproduced at the base commit; iOS fleet build in progress. Branch includes a merge of current main (900 commits) with the iroh halves of #10659 and #10730 stripped in follow-up commits.

Dictionary: relay (server that forwards traffic between peers that cannot connect directly), trust broker (backend service that registered device endpoints and issued connection grants), tolerant decoding (dropping unrecognized entries during decode instead of failing the whole payload), pairing listener (the Mac's TCP socket the phone connects to), forward migration (a new database migration that drops tables without rewriting history).

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10859"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790341396&installation_model_id=21920&pr_number=10859&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10859&signature=f84728187959c004b104c459f2683e4cef13b8176e3b6575684137ad6c5d5b0f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


azooz2003-bit:
Preflight findings from the tag `noiroh` sim run, all now on the branch:

- Restored `mac_client_namespace` in authenticated host status via a transport-neutral `MobileHostBundleNamespace` (067aa305f8). Development phones fail closed without it, so its iroh-named producer could not simply be deleted; pairing failed with `build_incompatible` until this fix.
- Fixed a main-inherited launch stall unrelated to the teardown (b2ed8846fa + 8b09023123, red/green): `PushRegistrationService.retryPendingUnregisterIfPossible` acquired an authenticated session snapshot even with an empty cleanup queue. Inside the post-sign-in hook that snapshot self-deadlocks on `awaitBootstrapped()` for the full 15s timeout, and the cancelled continuation then poisons `RPCStackTokenGate` for 30s, turning the first pairing attempt into an instant `requestTimedOut`. With the fix, cold-launch pairing goes from ~18s-with-one-failure to 2.7s, and the launcher readiness gate passes consistently. This mechanism is a plausible cause of the recurring sim/phone pairing-gate false negatives.
- Follow-ups deliberately not in this PR: (1) with push enabled and a cached APNs token the upload path can still hit the same self-deadlock; the `awaitBootstrapped()` re-entrancy should be broken structurally. (2) `RPCStackTokenGate` 30s self-poisoning on a cancelled waiter is aggressive. (3) `ios/scripts/reload.sh` bakes `CMUX_API_BASE_URL` from ambient `CMUX_PORT`, which inside a cmux terminal is the user's own app port; dev sim builds then point at a dead web origin.
- CI notes: the four new Swift warning buckets and the `SidebarDropPlannerTests` failure reproduce from files this branch does not touch (main drift under Xcode 26.3); the warning buckets are budgeted in 304ce800ba with rationale.

azooz2003-bit:
CI triage: the remaining failures reproduce main's standing red set, verified against a main-only baseline rather than assumed. Main run https://github.com/manaflow-ai/cmux/actions/runs/32212616081 (Aug 19, the last main push that exercised the mac lanes; later pushes path-skipped them) fails the same jobs — app-host unit tests (1/4, 3/4, 4/4) and swift-package-tests — including the identical test names now failing here (e.g. `AppDelegateIssue2907RoutingTests.testIssue2907BonsplitTabLookupUsesRecoveredRoute`, `testIssue2907TabManagerDependentSocketCommandsRecoverLiveSurfaceContext`). The failures also survived a full failed-jobs rerun on this PR, so they are deterministic main-red, not flakes introduced here.

Branch-caused failures were found and fixed earlier: the `mac_client_namespace` unit test (067aa305f8) and the Swift warning budget (304ce800ba). No Vercel preview deployment was produced for this PR by the integration; web changes are covered by the green web-typecheck lane and the local web suite (1204 pass / 0 fail).

PR #10905, opened 2026-08-27
Clear the three Swift warning buckets over the CI budget on main

austinywang:
`tests-build-and-lag` on main fails its warning-budget gate with three buckets over budget:

- `Sources/Mobile/MobileHostIrxRuntime.swift` ×2 — `enabledDefaultsKey`/`forceRelayDefaultsKey` are main-actor-isolated statics read from the nonisolated activation gate (inherited from #10782). Marked `nonisolated` (they are string constants).
- `Sources/Surfaces/SurfaceSocketCommands.swift` ×2 — worker-lane handlers read `SurfaceCatalog.shared` synchronously; now awaited like the surrounding calls.

No behavior change. Verified with `swiftc -parse`; the gate is `python3 scripts/swift_warning_budget.py` on the CI build log.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10905"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790383878&installation_model_id=21920&pr_number=10905&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10905&signature=6c80a71c39cce6044a3519e01e3dc7d3bec4763fc8f57e624d28184d0877f1f7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Clears the three Swift warnings pushing the `tests-build-and-lag` CI job over its warning-budget gate on main. No behavior change.

[...]

PR #10662, opened 2026-08-24
Keep pinned workspace groups when their last workspace closes

austinywang:
Closes #10560

## Root cause
Workspace groups were lifecycle-owned by a real member workspace. The close-path anchor fixup removed the group when that member was the last one, and session persistence/sidebar rendering had no representation for a memberless group.

## Fix
- Give group anchors an explicit live/empty state with a stable header identity.
- Keep pinned groups as header-only groups when their last workspace closes; unpinned groups retain their existing dissolve behavior.
- Render and reorder empty pinned headers, promote the first workspace added or dropped into one, and keep explicit Delete Group as the destructive removal path (including confirmation in the sidebar).
- Persist and restore pinned empty groups, including crash-pruning and autosave fingerprint coverage.
- Add an empty-group drop resolver path.

## Verification
- Regression test commit first: `0bcc843762` (fails on the parent).
- Fix commit: `fdab2a91ad`.
- `swift test --package-path Packages/macOS/CmuxWorkspaces` (210 tests passed).
- Focused `swift test --package-path Packages/macOS/CmuxFoundation --filter droppingWorkspaceOnEmptyGroupHeaderProducesExplicitGroupPlan` passed.
- `python3 scripts/check-package-resolved-policy.py`, `./scripts/lint-pbxproj-test-wiring.sh`, and `./scripts/check-pbxproj.sh` passed.

---
[...]

PR #10907, opened 2026-08-27
Fix the two red CmuxFoundation package tests on main

austinywang:
Six `main` failures in the `swift-package-tests` CI job, fixed as six commits (plus a merge of `origin/main`). The combined `CmuxFoundation` filter suite passes 34/34 locally (twice), the full `SSHForegroundAuthenticationRetryPolicy` suite 23/23 on both macOS 26 and a dedicated macOS 15.7.4 host, and the `CmuxSettingsUI` anchor suite 5/5.

## 1. `SidebarDropPlannerPackageTests.rootDropBeforeEmptyHeaderProducesPlan`
**Root cause:** #10662 special-cased an *empty* group header in `SidebarWorkspaceReorderDropResolver.groupScopeCandidate` — its entire header (top **and** bottom half) became a group adopt zone — while the same PR added this test expecting the header's **upper** half to plan a *root* reorder *before* the group (`usesTopLevelRows`, `explicitGroupId == nil`, `edge == .top`), exactly like every non-empty header. The special case contradicted the test.

**Fix:** removed the empty-header branch so headers behave uniformly (upper half → before, lower half → adopt). The two sibling tests that hit the lower half still pass, as do the four other empty-group tests.

## 2. `SSHForegroundAuthenticationRetryPolicyTests.processTreeTerminationUsesOneOverallDeadline`
**Root cause (macOS-15 CI specific):** the test spawns a 25-process, TERM-ignoring `sh` chain and expects the whole tree gone within one shared two-second deadline. The ordered walk only force-kills processes it *reaches* before the deadline, so a subtree the deadline cut short was left frozen and alive. It passes on a dedicated macOS 15.7.4 fleet host (15/15 in a loop) and on macOS 26, but is **intermittently red on the shared Blacksmith macOS 15 runner** — that runner is *fork-starved*, and the terminator spawned ~100 short-lived `ps`/`pgrep`/`kill`/`date` processes. Any one failing with `EAGAIN` silently skips a signal, leaving survivors and a non-zero exit. The trace confirmed it: the walk ran to completion, yet a fork in the harness/walk failed. A local `ulimit -u` sweep on real macOS 15 reproduces `alive_after=24` exactly at the fork ceiling.

**Fix — make liveness independent of the fork-heavy walk and of the clock:**
- Snapshot the whole tree with a **single** `ps` after freezing the root, compute the descendant closure in-shell (no per-node fork), and STOP each member.
- Use the shell builtin `kill` throughout the terminator instead of `/bin/kill` (~20 forks removed; the post-deadline path now forks zero times).
- Take the snapshot **before** reading the clock; if `date` can't fork, treat the deadline as already lapsed instead of bailing out, so cleanup still kills.
- Backstop the ordered walk by force-killing every snapshotted member together — a builtin `kill` of the frozen PID list, needing no fork at all.

A full cleanup now needs only **two** successful `ps` forks (root validation + the snapshot); everything after is fork-free, and the frozen snapshot means a reparent during the walk cannot hide a survivor. The graceful per-process walk is unchanged for the small trees the replacement / terminal-mode tests exercise. Timing assertion loosened `< 3` → `< 5` (honest walk + 2 s grace ≈ 3.5–4 s on a loaded runner); the **liveness** assertion (no survivor) stays exact. On failure the harness now runs under `sh -x` and attaches the trace, harness facts, cleanup RC, and a `ps` state dump.

## 3. `SettingsRowAnchorResolutionTests.everyCuratedSettingEntryIsReachable`
**Root cause:** the Cloud Machines beta row (`.json("cloud.beta.machines.enabled")`, anchor `setting:betaFeatures:cloudMachines`) shipped without its path in the test's `rowConfigPaths`, which mirrors the `.json(...)` annotations across the section views, so the search result had no registered row to scroll to.

**Fix:** added the path next to the other beta toggles; the row and curated entry already agree.

## 4. `CmuxTerminal` no longer compiles on the CI toolchain (Xcode 16.4 / Swift 6.1)
**Root cause:** `TerminalSurfaceRuntimeTeardownFenceTests.swift:58` spawned a `Task` whose closure captured the non-Sendable `ghostty_surface_t` raw pointer just to compute `UInt(bitPattern:)` inside it — `error: passing closure as a 'sending' parameter risks causing data races`. Newer toolchains accept it; the runner's does not, and the compile error aborts the whole `swift-package-tests` job before any later package runs. Inherited from main (86743c203f).

**Fix:** compute the address in the `freeSurface` hook and let the `Task` capture the `UInt`. (Same change already present on the recently green `feat-computer-use` branch.) Fence suite 2/2 on macOS 15.7.4.

## 5. `SSHPTYAttachRetryScriptBuilderTests.signalInterruptsReconnectBackoffPromptly` (intermittent on the CI runner)
**Root cause:** same fork-starvation class as #2, but in shipped code. The reconnect backoff's signal handler ran `/bin/kill -TERM "$backoff_pid" || true; wait "$backoff_pid"` inside a trap. When that external `kill` can't fork, the failure is swallowed and `wait` blocks for the full backoff `sleep` (30 s in the test, up to the max reconnect delay in production), so the wrapper doesn't exit on HUP/INT/TERM promptly — the contract the test asserts.

**Fix:** use the shell builtin `kill` in every signal/cleanup handler of the SSH retry scripts (backoff builder's shared handler branch, the foreground-auth capture wrapper's signal forwarding and cleanup, and the test-authored handlers that mirror them). Nothing on the exit path spawns anymore. Full CmuxFoundation package 273/273 on macOS 15.7.4.

## 6. `TerminalSurfaceStartupRestorePolicyTests` "Explicit input cancels a deferred agent resume before admission" (deterministic on main)
**Root cause:** #9266 made `cancelStartupRestoreAdmissionForExplicitInput()` start the runtime with the `.inputDemand` source (comment: "do not make the first keystroke wait behind the unrelated paced-restore queue"), so the spawn goes straight to `createSurface` and never touches the restore spawn scheduler — but the test the same PR added still asserted the pre-#9266 paced behavior (`scheduledSurfaceIds == [surface.id]`). CI never caught it because `CmuxTerminal` could not compile on the runner (fix #4) and `CmuxFoundation` was red before it. Not a flake: sources and every test double are byte-identical to the green `feat-computer-use` run — that branch simply carries the corrected assertion.

**Fix:** assert the intended contract — scheduler stays empty and exactly one direct runtime-create attempt is made (same text as on `feat-computer-use`, #7151).

`origin/main` (a1f0cf9a72) is merged in so `tests-build-and-lag` picks up the #10905/#10922 warning-budget buckets.


austinywang:
CI triage for `main` after #10887 (dispatched run 33027651263):

- **swift-package-tests** — fixed by this PR (both suites reproduce on pristine `main`).
- **tests-build-and-lag** — warning budget, fixed by #10905.
- **app-host unit tests** — the lane is unstable on `main` regardless of the merge: a baseline dispatch on the pre-merge commit (`baseline/pre-cloud-surfaces` = `4202645fe5` + only the determinism-gate fix, run 33029450243) fails shard 2/4 with a dozen suites of its own (Dock terminal selection focus, File search workspace scope, Vault pane-transfer lifecycle, BrowserPanelViewIdentityTests, GhosttyPhysicalInputFocusReassertionTests, …), while post-merge `main` fails a *different* set (CLI hook tests timing out with status 15, main-window routing, TextBox IME layout). The new surfaces suites (`SurfaceCatalogTests`, `LocalSurfaceProviderTests`, `CmuxTuiSurfaceProviderTests`) pass, and the failing logs contain no runtime reference to the catalog — this is the pre-existing app-host storm (post-test 45 s timeout → batch status 125), not a regression from #10887.
- All four push-triggered workflows on `main` (cmux-tui artifacts, nightly macOS build, cmux-tui SDKs, docs) are green.

PR #10914, opened 2026-08-27
iroh: consume AddressLookupService behind a debug flag (registry resolve + record publish)

lawrencecchen:
Consumes the foreign-implementable `AddressLookupService` from [iroh-ffi PR #11](https://github.com/manaflow-ai/iroh-ffi/pull/11) in `CmuxIrohTransport`, behind a Debug-only flag (default OFF), per [plans/iroh-custom-discovery-ffi.md](https://github.com/manaflow-ai/cmuxterm-hq/blob/feat-iroh-custom-discovery-plan/plans/iroh-custom-discovery-ffi.md) migration step 2.

**Base caveat:** part of the ideal-shape series, based on `feat-iroh-integration-test`; until that lands on `main` this diff shows the whole stack. This PR owns the last two commits (red contract tests, then the resolve/publish implementation).

**Fork-pin caveat: not mergeable to `main` yet.** The pin is `manaflow-ai/iroh-ffi` exact `1.0.2-cmux.9-dev.1`, a published prerelease built from PR #11's head (c8a3029) plus regenerated Swift bindings, through the fork's own release pipeline (`release/v1.0.2-cmux.9-dev.1` branch, CI-built 5-slice xcframework, checksum baked by the release bot, attested, tag-promoted). Merge only after fork PR #11 lands and a real `v1.0.2-cmux.9` tag exists, then bump the pin (Package.swift + 4 Package.resolved) in one commit. Merging into `feat-iroh-integration-test` is fine now: the prerelease asset is a normal GitHub release the fleet resolves like any other.

## Consumption mechanics (teammates)

Nothing special: the dep is a normal SPM version pin; `swift build`/`swift test` in `Packages/Shared/CmuxIrohTransport`, tagged reloads, and fleet builds all fetch the prerelease xcframework by URL+checksum. To cut another dev prerelease from a fork branch: clone manaflow-ai/iroh-ffi, branch `release/v<version>` from the feature head, regenerate bindings (`cargo build --locked --lib`, then the uniffi-bindgen + sed + patch steps from `make_swift.sh`, commit `IrohLib/Sources/IrohLib/IrohLib.swift`), run `python3 scripts/release/bump_version.py --swift-release <version> --swift-repository manaflow-ai/iroh-ffi`, push, wait for `release_swift.yml` to bake the checksum, verify per RELEASING.md, tag that exact commit, and let `release.yml` promote the draft. Do not merge the release branch.

## Swift side

`CmxIrohRegistryAddressLookup: AddressLookupService`:

- **resolve** answers in order: in-memory record cache (zero network) → injected persisted-record source → at most ONE broker fetch riding the discovery snapshot (and its backpressure gate). The fetch is single-flight (concurrent resolves join it) and cools down on failure by the #10865 taxonomy: `preservesVerifiedStateDuringRefresh` failures back off on `.foregroundClient`, trust failures fail closed on a 5-minute-floor schedule. All well-formed fetched records are cached per endpoint id.
- **Record policy** (`CmxIrohEndpointRecordPolicy`): parse+signature verification via the fork's `parseEndpointRecord` (Rust re-verifies again in magicsock), freshness owned by Swift (1h window, 5m skew — matches `maximumPrivateHintTTL`), and the relay allowlist rule: a record naming ANY relay outside managed catalog / custom profile / debug override is dropped whole (the ed25519 signature covers the full packet, so partial stripping is impossible). Mirrors the `untrusted_relay` rule in `web/services/relay/report.ts` and the `unmanagedRelayURL` hint filter.
- **publish** verifies the endpoint's own record, primes the resolve cache, and POSTs the blob to the new endpoint-record route.

Install is flag-gated: `CMUX_IROH_ADDRESS_LOOKUP` (env or defaults, Debug builds only). ON = `EndpointOptions.addressLookup` gets the lookup while hint dials stay untouched, so magicsock merges `Source::AddressLookup` next to `Source::App` (safe A/B). OFF = nil lookup, byte-identical bind options (pinned by test). Wired on the Mac host runtime (with an `iroh-diag` section: installed-since, last resolve outcome/source, last publish outcome, counters, via the #10814 lock-mirror pattern) and on the iOS client composition.

## Web side

Publish = **new route** `POST /api/devices/iroh/endpoint-record` rather than a field on the registration payload: iroh fires publish on address changes independent of the registration cadence, and forcing a full challenge-signed re-register per address change would couple the lookup to registration state. Flagged as the deliberately smaller change. The route requires the binding-request proof (no legacy account-credential fallback), checks the record's leading 32-byte public key equals the caller binding's endpoint id, caps size (105–1200 bytes decoded, canonical base64), and stores it as an opaque blob (`iroh_endpoint_bindings.endpoint_record` + CHECK constraints). Deliberately no server-side ed25519 verification and no reader trust: readers verify the signature in Rust, so any cache/replica may serve records. Discovery (`publicBinding`) now returns `endpoint_record`; revocation wipes it with the same posture as path hints. No account-revision bump on record writes (resolve pulls on demand; hints already bump revision on registration refresh).

## Honest limitations (A/B phase, flag OFF by default)

- A publish that fails before registration completes retries on the next address change, not on registration completion.
- iOS allowlist for resolved records is frozen per endpoint generation (like hint dials); the Mac host reads the live policy mirror.
- `CmxIrohBackpressuredClientBroker/HostBroker.publishEndpointRecord` uses an `as?` runtime check on the wrapped broker so existing fakes stay untouched; a non-record broker fails typed.
- Stored records are not retention-swept beyond revocation wipe (single bounded column per binding; client applies its own freshness window).

## Tests

- Red commit: 14 contract tests against a stubbed resolver (9 failing); green commit implements resolve/publish, full `CmuxIrohTransport` suite 659/659 green locally against the prerelease pin.
- Coverage: cache-hit resolve with zero network, persisted-cache tier, single fetch then cache, transient vs trust failure classes + cooldown + re-arm, relay-allowlist drop + trailing-slash tolerance, stale-record rejection, wrong-endpoint rejection, publish upload/prime/reject/upload-failure, sign/parse round trip + tamper detection, flag default-off/opt-in/env-precedence, byte-identical options with flag off.
- Web: `bun run typecheck` clean; iroh unit suites green (trust broker publish/mismatch/proof/malformed/revocation-wipe, route handler wiring); DB behavior lane on local postgres:16 covers store/guard/wipe plus the CHECK constraint (which also caught that Postgres caps regex repetition at 255 — the length bound lives outside the regex).

Dictionary:
- **xcframework**: Apple's multi-platform binary bundle; the fork ships the Rust core as one, checksum-pinned in Package.swift.
- **prerelease tag**: a semver tag with a suffix (`-dev.1`) marking a temporary dev artifact.
- **binding-request proof**: three signed headers proving the caller holds the endpoint key of a registered binding.
- **backpressure gate**: the per-operation serializer in front of trust-broker calls.
- **DB behavior lane**: the CI job running repository tests against a real postgres:16 with migrations applied.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10914"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790390734&installation_model_id=21920&pr_number=10914&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10914&signature=f60faabc93cb79221c36a2b810d64ee8bcc9263887f5117ce07c1e26eb10e9e2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Consumes the foreign-implementable `AddressLookupService` from `manaflow-ai/iroh-ffi` PR #11 in `CmuxIrohTransport`, behind a Debug-only flag (default off), and removes the client-held relay token machinery — managed relays are now tokenless with admission decided by the server-side relay allow hook.

**Address lookup**
- `CmxIrohRegistryAddressLookup.resolve` answers from the in-memory record cache (zero network), then the persisted-record source, then at most one single-flight broker fetch that cools down per the transient/trust failure taxonomy.
- Records naming a relay outside the managed catalog or custom profile are dropped whole; freshness is a 1h window with 5m skew, and resolved hex IDs use a nibble table instead of per-byte formatting.
- `publish` verifies the endpoint's own record, primes the resolve cache, and posts the blob to the new `POST /api/devices/iroh/endpoint-record` route; discovery now returns `endpoint_record` and revocation wipes it.
- Flag OFF binds byte-identical endpoint options, and release builds compile the flag and lookup away.

**Token removal and fork pin**
- Deletes the relay credential coordinator, token types, `issueRelayToken` broker calls, the `/api/relay/token` route, and the dormant n0-hosted relay minter.
- Pins `manaflow-ai/iroh-ffi` to prerelease `1.0.2-cmux.9-dev.1`; not mergeable to `main` until the fork PR lands and a real `v1.0.2-cmux.9` tag exists, then bump the pin in `Package.swift` plus four `Package.resolved` files in one commit.
- Bundles prior iroh fixes: bounded dial phases, per-connection server handshakes, admission capacity release on connection liveness, register-when-ready publication, paired-peer allowlist admission, and client cache-first activation.

<sup>Written for commit 9df8825ae1e8e8f33b737cb70f7f9587d29cd88c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added server-managed relay admission without client-held credentials.
  * Added signed endpoint record publishing, address discovery, and relay attachment reporting.
  * Added credential-free admission for previously paired devices.
  * Added debug diagnostics and optional relay overrides.

* **Bug Fixes**
  * Added bounded connection admission timeouts and improved capacity recovery.
  * Prevented quit actions from deadlocking through debug controls.

* **Improvements**
  * Enabled cache-first startup and limited reuse of recently expired, verified relay policies.
  * Added single-round registration proofs and an additional relay policy trust key.


lawrencecchen:
Pinned local review (gpt-5.6-sol, high, isolated): two rounds.

Round 1 verdict was "incorrect" on two findings; dispositions:

- **P0 `String(format:)` in `canonicalEndpointID`** — fixed in 9b46aca3eb: pure-Swift nibble table (resolve runs the conversion per dial and per fetched record, so the per-byte Foundation format path was a real allocation hazard).
- **P1 "relay readiness bypassed at `CmxIrohClientRuntime.swift:526`"** — false positive and out of scope. This PR touches zero lines of that file (it is part of the `feat-iroh-integration-test` base this PR is stacked on, so the review bundle shows the whole stack), and the flagged `hasConfiguredRelay()` guard is immediately followed by `try await connectivityEngine.waitForUsableHomeRelay()` at line 530, so the claimed bypass does not exist.
- **P2 policy findings (57)** — most target stack files outside this PR. On this PR's own files: missing DocC on the new public outcome fields and record-broker extension methods is fixed in 9b46aca3eb; the `OSAllocatedUnfairLock` diag mirrors are intentional per the in-tree `MobileHostIrohRuntime+RelayDiag` / `AgentChatThemeSync` precedent (synchronous write visibility from non-async funnels, readable while the main thread is wedged), with the reasoning inline; `CmxIrohDebugAddressLookupFlag` mirrors the `CmxIrohDebugRelayOverride` namespace-enum shape deliberately; the test fixture/fake types beside their suite follow the existing suite convention.

Round 2 (post-fix head 9b46aca3eb): the P0 is gone. One finding remains, again against stack code this PR does not touch: [P1] per-identity admission limits checked only after handshake completion (`CmxIrohEndpointServer.swift:273`, from the integration branch's accept-loop redesign commits e1ed22e230 / 4661baec04 / 616fc69829). Out of scope here; flagged to the `feat-iroh-integration-test` owners as a possible pre-authentication pending-slot monopolization concern. Zero findings against this PR's own commits after round 1's fix.

PR #10867, opened 2026-08-26
Withhold managed relays from a fresh Mac host until registration is acknowledged

lawrencecchen:
Part of the iroh ideal-shape series. **Base caveat: opened against `main` but the diff is on top of `feat-iroh-admission-race` @ 484a6c4ba6 (https://github.com/manaflow-ai/cmux/pull/10857), so the shared withhold mechanism composes; it should merge after that branch lands (the "Files changed" tab shows the whole stack until then).**

## Problem

Host-side twin of the client cold-launch admission race fixed in https://github.com/manaflow-ai/cmux/pull/10857, flagged independently by two pinned reviews at `CmxIrohHostRuntime.swift:189` ("Withhold managed relays until host registration completes"). A freshly installed Mac host binds its iroh endpoint with the managed relay map installed (`start()` builds `CmxIrohEndpointConfiguration` with the resolved managed profile before any broker round), native iroh dials the home relay immediately, and the relay's allow hook asks the broker about an endpoint whose registration is still in flight: on a fresh install `cachedHostPolicy` is nil, so the blocking `resolveInitialPolicy` performs `broker.register` strictly after bind (the registration payload needs the post-bind endpoint address, so registration structurally cannot precede the dial). The broker answers deny and cmux-relay negatively caches it (manaflow-ai/cmux-relay#9: first deny 5s, escalating), which stalls the very `waitForUsableHomeRelay` gate the host's register-when-ready publication (https://github.com/manaflow-ai/cmux/pull/10737) then waits on. #10737 orders publication after relay readiness, not the dial after registration, so it does not prevent this.

## Fix

`CmxIrohHostRuntime.start()` mirrors the client mechanism with the shared pieces: when no cached policy cryptographically verifies for this endpoint, the endpoint binds with `CmxIrohEndpointRelayProfile.unavailableManagedSelection`, and the real managed profile is installed through the existing `CmxConnectivityEngine.replaceRelayProfile(_:expectedIdentity:)` immediately after the initial policy resolve returns, which a fresh host can only reach with an acknowledged `broker.register` (the `cachedPolicy(after:)` fallback requires the same verified cached policy whose absence made the bind withhold). The relay-required debug gate (`hasConfiguredRelay` + `waitForUsableHomeRelay`) then waits on a dial the relay can admit.

[...]

PR #10857, opened 2026-08-26
iroh: order a fresh endpoint's first managed-relay dial after acknowledged registration

lawrencecchen:
Part of the iroh ideal-shape series. **Base caveat: opened against `main` but the diff is on top of `feat-iroh-integration-test` @ 84050588d0; it should merge after that branch lands (the "Files changed" tab shows the whole integration-test stack until then).**

## Problem

The 31-launch iOS simulator timing batch (feat-iroh-integration-test @ 84050588d0, GCP test relay, relayOnly) found a cold-launch admission race on 3/5 cold launches: a freshly installed app binds its iroh endpoint with the managed relay map installed, native iroh dials the home relay immediately, and the relay's allow hook asks the broker about an endpoint whose `POST api/devices/iroh/register` is still in flight. The broker answers deny, cmux-relay negatively caches it (NEGATIVE_TTL 60s), the client's 15s usable-home-relay gate times out, and the retry lands 40-60s after launch.

Log evidence (cycle 2, cold): `appLaunched` 10:54:47.383Z, `Iroh endpoint starting` 10:55:04.704Z, `Relay policy refreshed` 10:55:08.878Z, `Iroh endpoint failed (Iroh endpoint unavailable)` 10:55:29.161Z (gate timeout), retry `starting` 10:55:31.806Z, `active` 10:55:41.523Z = 54.1s launch-to-active. Cycles 4 and 5 cold launches show the same shape (63.4s and 60.5s).

## Fix

`CmxIrohClientRuntime` orders the first managed-relay dial after broker admission with a deterministic signal (no sleeps): when no cached binding proves the broker has already acknowledged this endpoint, the endpoint binds with `CmxIrohEndpointRelayProfile.unavailableManagedSelection`, and `start()` installs the real managed profile through the existing `replaceRelayProfile` machinery immediately after `install(policy:)`, which a fresh endpoint can only reach with an acknowledged registration (cooldown and offline fallbacks both require prior broker proof). The relayOnly gate then waits on a dial the relay can admit. Cached-binding (warm) endpoints keep the old behavior: relays installed at bind, dial overlaps registration refresh. Custom relays are user-operated, not admission-gated by the cmux broker, and stay installed at bind so a broker outage cannot disable them.

[...]

PR #10888, opened 2026-08-26
iroh: close orphaned admissions, hold consumed-handshake slots, re-arm the publication ready gate

lawrencecchen:
Fixes the last three P1 findings from the local pinned reviews of the ideal-shape stack (#10880 dispositions flagged the first two for the capacity-release stack owner; the third was flagged in the #10858 and #10867 reviews, todo F9).

**Stacked-base caveat:** branched from `feat-iroh-integration-test` (8700ea6821), so the diff against `main` carries the whole ideal-shape stack. This PR's own change is the top two commits (red tests f30713e9c5, fix 616fc69829). It merges into `feat-iroh-integration-test`; this PR exists for review anchoring.

1. **Connection-orphan race** ([CmxIrohEndpointServer.swift:443](https://github.com/manaflow-ai/cmux/blob/616fc69829/Packages/Shared/CmuxIrohTransport/Sources/CmuxIrohTransport/CmxIrohEndpointServer.swift#L443)) — CONFIRMED. Capacity can fill between `registerEstablished` and the admission marker; `markAdmitted` removed the pending entry, returned `false`, and no table owned the established QUIC connection, so `finishHandler` never closed it. Fix: close it (`connection_capacity`) before returning, mirroring `rejectEstablished`.

2. **Non-abortable consumed handshake** ([CmxIrohLibIncomingConnection.swift:28](https://github.com/manaflow-ai/cmux/blob/616fc69829/Packages/Shared/CmuxIrohTransport/Sources/CmuxIrohTransport/CmxIrohLibIncomingConnection.swift#L28)) — CONFIRMED. `refuse()` on a consumed `Incoming` is a no-op and the IrohLib uniffi bindings have no cancellation propagation (`uniffiRustCallAsync` has no cancellation handler and frees the rust future only on completion), so the admission timeout released the slot while the native handshake stayed live; a flood could mint more handshake work than `maximumPendingAdmissions`. Fix: a timeout on an in-flight handshake refuses the dialer but keeps the slot (`abandoned` flag) until `establish()` resolves; resolution releases it. The driver's handshake/idle timeout bounds the hold.

[...]

PR #10919, opened 2026-08-27
Client cache-first activation, warm cache-first dials, one-round registration

lawrencecchen:
Client-side leg of the platonic-ideal cold start (gap #2 in the iroh goal pipeline, F3 decomposition): a warm phone dials with ZERO blocking control-plane calls. Cached device records and the cached verified policy carry activation and the dial; every network refresh runs behind the session and keeps the existing fail-closed taxonomy. The host side already had this (#10737); this PR gives the client the same shape and also collapses the genuinely-cold path.

> **Base caveat:** stacked on `feat-iroh-integration-test` (head e841b8ab15). The diff shows the whole integration series until that branch lands; only the 4 commits ending at this head are this PR. Do not merge to `main` directly; this lands on the integration branch after review.

## Round-trip accounting

| Path | Before | After |
| --- | --- | --- |
| Warm activation, automatic mode | Stack token refresh + 1 blocking discovery sync (activation awaited the overlapped prefetch) | 0 blocking broker rounds |
| Warm activation, relayOnly mode | blocking policy GET (median 3.8 s in the F3 measurements) + the discovery sync above | 0 blocking broker rounds |
| Warm dial | 1 discovery GET + 1 pair-grant POST (in-memory grant cache is empty after relaunch) | 0 blocking broker rounds |
| Cold registration | challenge POST + register POST, serialized | 1 register POST (self-proof) |

## 1. Client cache-first activation (`CmxIrohClientRuntime.start()`)

A warm client holds two persisted proofs: the cached broker binding (broker acknowledged this exact endpoint — the same proof that keeps managed relays installed at bind, #10857) and the offline route record whose pair grants re-verify against the stored Ed25519 key set. `cachedStartPolicy` resolves the start policy entirely from those caches after the local bind; the immediate registration refresh (`requiresDiscovery: true`) is scheduled right behind activation and is the fail-closed authority: a non-transient rejection, or authoritative discovery that no longer lists the binding, tears the runtime down and wipes the cached policy (`refreshRegistration`'s existing taxonomy — exactly #10737's host semantics). Transient (connectivity-class) refresh failures preserve the activation. A broker cooldown floor no longer blocks a cache-first start; a cache miss still rethrows it, preserving the previous ordering. A fresh endpoint (no cached binding, or no offline record) keeps today's blocking ordering and #10857's managed-relay withholding untouched.

relayOnly mode: the composition (`MobileIrohRuntimeComposition`) previously blocked activation on a live policy refresh (`:1843`, the largest fixed cost in the F3 sim measurements). A warm client (verified cached binding + restored policy with usable relays) now restores the verified cached policy exactly like automatic mode and refreshes immediately after activation. A fresh install keeps the blocking refresh, and the account whose previous cache-first relay-only activation failed (e.g. every cached relay dead, so the relay readiness barrier timed out) falls back to the blocking refresh instead of looping on a dead cached catalog.

## 2. Challenge/register collapse (one broker round)

The two-step's only cryptographic content is a fresh, one-use nonce inside the signed transcript. The new self-contained proof reproduces both properties without the extra round: transcript `cmux/iroh/device-registration/v2\n{issuedAt}\n{nonce}\n{payloadSha256}`, where freshness is the same ±5-minute window the broker already grants every timestamp-signed binding request (`verifyBindingRequestSignature`), and one-use is server-side consumption of the client nonce (per-user dedupe select under the existing challenge advisory lock, global `nonce_hash` unique index as a cross-account backstop). `registerWithSelfProof` mints the challenge row already consumed with the same strict per-slot monotonic mint time as `issueChallenge` and applies the registration through the extracted `applyChallengeRegistration`, so slot adoption, reincarnation, revision advance, and the `challenge_superseded` high-water gate are shared code with the two-step flow, proven by a DB test that interleaves both flows. The dedupe row outlives the proof's entire acceptance window, so pruning cannot reopen a replay.

Backward compatibility: the two-step wire contract is byte-identical for deployed clients (optional fields are omitted from the encoded body). The new client sends the one-round proof first and falls back to the interactive challenge exactly when the two-step can repair the rejection: an old broker's parse rejection of the proof shape (`400 invalid_challenge_id` / `unknown_field`) or a skewed client clock (`403 self_proof_expired`). Every other verdict propagates without a retry.

Residual security delta, stated precisely: a server-minted nonce proves the signature was created after a server-chosen instant; the self-proof bounds it by clock skew instead. An attacker who could steal a fresh signature inside the 5-minute window could register once — but stealing it requires compromising TLS or the device, which already yields the endpoint secret key itself. This matches the trust level the broker has always granted post-registration binding-proof requests.

## 3. Dial-from-cache for device records

`CmxIrohRegistryContextProvider.context(for:)` serves a dial from the verified offline record when it covers the exact requested tuple (grants re-verified against the stored key set; a still-fresh verified snapshot, when present, confirms the record without consuming its one-shot window). One shared background discovery refresh is armed behind the dial (single-flight through the existing `sharedDiscover` coalescing and backpressure gate); when it proves the cached target vanished or was replaced, the record is pruned and the peer marked stale.

Staleness-evidence composition (#10739/#10865), verified by tests: a marked-stale peer (failed dial on an empty/unreachable plan, presence push) bypasses the cached record and every reuse window and rebuilds from a fresh broker snapshot — stale record → failed dial → refresh → redial stays intact. A cached record whose hints cannot produce a dialable plan falls through to the authoritative resolve instead of failing the dial.

Test-contract change, deliberate and called out for review: the provider tests named `*NeverConsultsOfflinePolicy` previously asserted no cache READ happened before a broker rejection. Cache-first reads the record before the broker is asked, so those tests now mark the peer stale first and pin the invariant that actually matters: **a broker's authoritative rejection never FALLS BACK to cached authority**, on the fresh-discovery path they now exercise. Similarly `authoritativeRejectionCannotFallBackToStaleOfflineAuthority` now proves the post-activation form: cache-first start succeeds, then live discovery that drops the binding tears it down.

## Verification

- `CmuxIrohTransport`: 653/653 tests in 73 suites (red commits b4adf832b7 / 6ebb06eba0 show the failing halves).
- Web: `bun run typecheck` clean; `tests/iroh-trust-broker.test.ts` 62/62; full unit lane has 14 failures + 1 error that reproduce identically on a sibling worktree without these changes (full-lane cross-file flakes in billing/coderouter/subrouter files, unrelated).
- DB behavior lane vs local `postgres:16`: 15 files, 229 tests, 0 failures (includes the new atomic self-proof registration, replay dedupe, and mixed-flow ordering tests).
- macOS cloud build tag `iccf` (no launch); iOS app compile-checked against an isolated simulator DerivedData path — that compile caught one real defect (the predicate read the transport-internal `activeRelays`; fixed by the public `hasDialableRelays` accessor, commit 2ea6757586). `ios/cmuxPackage` cannot `swift test` on macOS because `CmuxMobileTerminal` pins macOS 10.13 (pre-existing), so composition behavior is covered by the pure decision-function tests in `MobileIrohRelayOnlyCacheFirstTests` plus CI's iOS lane.

Known limits, stated rather than hidden: (1) the client does not adopt a server-side replaced binding in the background refresh the way the host does (#10867's `adoptReplacedBinding`); it fails closed, wipes the persisted binding via `handlePolicyInvalidation`, and the next activation re-registers fresh — one teardown cycle, self-healing. (2) Cache-first dials extend authorization-evidence reuse from the 30-second snapshot window to the (days-long, signature-bounded) pair-grant lifetime while the broker is reachable; the bounded exposure is one dial, because the armed refresh re-validates and the Mac host and relay allow-hook still independently verify admission. (3) relayOnly cache-first is exercised at the runtime and pure-decision layers; the full composition activation path has no in-process test because relay readiness needs a live relay (same gap as before this PR).

---
## Summary by cubic
Gives warm phones a cache-first cold start: activation and dials run with zero blocking broker rounds, and cold registration drops from two serialized broker calls to one. The stored relay-token machinery is gone — managed relays admit solely via the relay's server-side allow hook — so a fresh endpoint binds relay-less until registration is acknowledged. Warm dials to already-paired Macs carry no in-band credential and are admitted from the Mac's paired-peer allowlist, so the hot path never fetches a pair grant.

**Behavior**
- Warm activation resolves the start policy from the verified cached binding and offline route record; the registration refresh runs behind activation and fails closed, tearing the runtime down and wiping cached policy on a non-transient rejection.
- relayOnly mode restores the verified cached policy like automatic mode, while fresh installs and accounts whose last cache-first relay-only activation failed keep the blocking refresh.
- Warm dials are served from the verified offline record when it covers the exact tuple, credential-less for already-paired Macs, with one shared background discovery refresh behind the dial; staleness evidence still bypasses every cached source.
- The cache-first refresh is fenced to the lifecycle that armed it, so teardown or policy identity replacement cancels it before it can mutate newer state.
- Registration tries the one-round self-proof first (signed timestamp plus one-use client nonce) and falls back to the two-step flow only for old-broker parse rejections or a skewed client clock.
- A server-side binding replacement fails closed on refresh (one teardown cycle, self-healing on next activation); discovery dropping the binding marks the cache stale.

**Migration**
- Requires the relay allow-hook fleet rollout (`cmux-relay#9` dual-accept) to be deployed before merge.
- This PR lands on `feat-iroh-integration-test`, not `main`; the diff against main shows the whole stacked series, with only the client cache-first commits belonging to this PR.
- Test contract change: cache-first reads the cache before the broker, so provider tests now mark peers stale first and pin that a broker's authoritative rejection never falls back to cached authority.

<sup>Written for commit 6a3dfcf717569cef6ec2c431aa609bfd8f0e8418. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added signed relay-policy fetching without client relay credentials or token minting.
  * Added cache-first activation with background refresh and safe fallback behavior.
  * Added one-round self-proof registration with compatibility fallback.
  * Added paired-peer allowlist admission for warm connections.
  * Added relay attach/detach reporting and relay diagnostics.
  * Added configurable dial and relay-readiness timeouts.
  * Added deferred app termination to prevent shutdown deadlocks.
* **Bug Fixes**
  * Improved stalled connection handling, capacity recovery, and handshake cleanup.
  * Added bounded reuse for recently expired verified policies.
* **Chores**
  * Removed the retired relay-token minter and related release-gate scenarios.


lawrencecchen:
Local pinned review (`gpt-5.6-sol`, high reasoning) ran twice; dispositions for every finding. Both rounds reviewed the branch against `main`, so the frozen diff includes the whole stacked integration series, not just this PR's five commits.

**Round 1, P1 — "Do not install managed relays based only on a persisted cached binding" (`CmxIrohClientRuntime.swift:170`).** Rejected as pre-existing, previously triaged design. The presence-of-cached-binding gate is #10857's client-side contract; #10867 deliberately gave only the HOST the stronger full-cached-policy validation and recorded the asymmetry. This PR does not change when relays are installed at bind, and the exposure window (bind → acknowledged registration) is the same before and after: pre-PR, a server-side-revoked binding also had relays installed at bind while the in-band discovery/registration repaired it. Mitigations already on the branch: the relay allow hook's escalating negative cache starts at a 5-second deny (relay#9), and this PR's immediate post-activation registration refresh re-authenticates the binding right away and fails closed (tears down and wipes cached policy) on a non-transient rejection.

**Round 2, P1 — "Cancel cache-first refresh tasks when the context is cleared" (`CmxIrohRegistryContextProvider.swift`).** Confirmed and fixed in 06423130b0: `cancelCacheFirstRefresh()` bumps a generation fence and cancels the task; runtime network teardown and `updatePolicy`'s identity replacement both call it, and the refresh re-checks the generation after every suspension before touching provider state. Regression test `cancelledCacheFirstRefreshCannotMutateProviderState` holds the refresh's broker call, cancels, releases, and proves the late result cannot mark the peer stale.

**Round 2, P1 — "Release stale admissions after a generation change" (`CmxIrohEndpointServer.swift:342`).** False positive, and outside this PR's delta (the file is #10888's, part of the stacked base). Disproof: every generation change goes through `handle()` → `cancelConnections(exceptGeneration:)`, which removes all stale-generation entries from `pendingAdmissions` synchronously (before its first await) and abandons/cancels them; `stop()` clears the whole table. By the time `registerEstablished` runs for a superseded handshake, its entry is already gone, so the flagged guard path closes the connection without holding any slot. The only entry that can fail the `generation == currentGeneration` check while still present would have to survive `cancelConnections`' synchronous removal, which the actor's serialization prevents.

**P2 policy findings.** The two inside this PR's delta are fixed in cb90da965c (`cmxRegistrationRetriesAsTwoStep` moved to file scope; `CacheFirstRuntimeSeed` split into its own file). The remaining Aziz lints (`CmxIrohDebugBrokerBypassHeader`, `CmxIrohDebugRelayOverride`, `CmxIrohEndpointServer`, `CmxIrohClientSessionDialBoundTests`, `CmxIrohEndpointServerCapacityReleaseTests`, `CmxIrohEndpointServerStalledHandshakeTests`, `MobileHostIrohRuntime+RelayDiag`) are on stacked-series files this PR does not touch; they belong to the PRs that introduced those files.

Post-fix verification: CmuxIrohTransport 654/654 tests in 73 suites at 06423130b0.

PR #6409, opened 2026-06-18
Add persistent Freestyle sshd cloud slot

lawrencecchen:
## Summary
- make default `cmux vm new` use one persistent Freestyle VM slot via a stable idempotency key
- open the default cloud slot through SSH into a pinned `sshd` workspace and reuse that workspace when it already exists
- resume paused VMs before minting attach or SSH credentials

## Verification
- `bunx tsc --noEmit --pretty false`
- `bun test tests/vm-workflows.test.ts` (DB-backed cases skipped without `CMUX_DB_TEST=1`)
- `xcodebuild test -quiet -project cmux.xcodeproj -scheme cmux-unit -configuration Debug -destination 'platform=macOS' -derivedDataPath /tmp/cmux-cloudsshd -only-testing:cmuxTests/CLINotifyProcessIntegrationRegressionTests/testVMNewDefaultCreatesPinnedSSHDWorkspaceOverFreestyleSSH -only-testing:cmuxTests/CLINotifyProcessIntegrationRegressionTests/testVMNewDefaultReusesPinnedSSHDWorkspaceOverFreestyleSSH -only-testing:cmuxTests/CLINotifyProcessIntegrationRegressionTests/testVMSSHOpensManagedWorkspaceThroughSharedSSHPath`

---
[...]

PR #10931, opened 2026-08-27
ci: make the app-host unit-test lane green for the right reasons

austinywang:
## Why the app-host lane was red

`app-host unit tests (n/4)` on `main` (run 33027651263: shards 1, 2, 4 red; pre-merge baseline run 33034886551: shards 2, 4 red) fails for four independent reasons. None of them is a flaky CLI hang.

### How the lane decides

`run_unit_test_batch` in `ci.yml` tolerates every ordinary assertion failure and only turns a shard red when the batch's last `Executed … (K unexpected)` line has `K > 0` — i.e. a thrown error (`XCTUnwrap failed: threw error`, `failed: caught error`, `XCTAssertEqual(try …)`) or a crash/restart. Every red shard in both runs traces to one of the classes below.

### 1. The xcodebuild wrapper killed every big batch 45 s into the Swift Testing phase

`scripts/ci/xcodebuild_noninteractive.py` armed its post-test deadline on the XCTest `Test Suite 'Selected tests' … at` line. The cmux-unit bundle runs XCTest first and Swift Testing second (`◇ Test run started.` follows two lines later and runs for minutes), so every large batch was SIGTERMed mid-Swift-Testing: `Post-test timed out after 45s; terminating xcodebuild` → `App-host unit-test batch failed with status 125`, truncated Swift Testing coverage, and orphaned CLI children/mock servers keeping the host alive. The deadline now disarms when a Swift Testing run starts after the XCTest summary and re-arms on `Test run with N tests in M suites passed|failed after`; a failed Swift Testing run keeps the batch red. Regression test in `tests/test_ci_xcodebuild_noninteractive_helper.py` (red in the first commit, green in the second).

### 2. `status: 15, timedOut: true` = libdispatch pool starvation, not a CLI hang

`ClaudeHookPIDAuthenticationTests`, `ClaudeHookLifecycleCleanupTests`, `ClaudeHookLiveDeliveryTargetTests`, `CLICodexResumeNotificationTests`, `CLIOmpSupersededCleanupTests`, `SSHRemoteCommandChainingTests` (`status: 0, timedOut: true` — the child had exited but nobody observed it) all run in the parallel Swift Testing phase. Their helpers parked `waitUntilExit` and both `readDataToEndOfFile` drains on `DispatchQueue.global`, and the mock cmux socket servers blocked `accept`/`read` loops on the same pool. Once the pool is exhausted the exit-observer block never runs and the mock never answers, so a CLI that returned in milliseconds is reported as hung and SIGTERMed.

`cmuxTests/CLITestProcessRunner.swift` owns the child lifecycle on POSIX primitives and dedicated threads (`posix_spawn` into its own process group, a `waitpid` reaper thread, a drain thread per pipe, group kill on timeout so grandchildren cannot hold the capture pipes open). The shared helpers (`CLINotifyProcessTestSupport`, `ClaudeHookSurfaceResolutionSwiftTests`, `ClaudeHookLiveDeliveryHarness`, `runCodexHookProcess`, `CMUXOpenHTMLFocusTests`, `CampfireHookNotificationTests`) delegate to it, and every `DispatchQueue.global().async` block in a Swift Testing file moved to `CLITestProcessRunner.detachBlockingThread`. Regression test `CLIProcessHelperStarvationTests` saturates the pool and asserts the helpers still complete (red first, green after).

### 2b. One-way feed telemetry raced the mock server's reader

`sendBestEffortFeedTelemetry` writes `feed.push`, closes the socket, and exits without waiting for a reply, so `ClaudeHookLiveDeliveryHarness` tests that read `context.state.snapshot()` right after the hook process exited were racing the mock server's per-client reader thread. On the runner that showed up as `subagentStopDoesNotPublishParentCompletionAttention` → `saw []` (run 33043353308, shard 4). `runHookProcess` now waits for the mock server to reach quiescence — the accept loop polls in 10 ms slices and only it may declare "backlog empty, no reader in flight" — before returning, and `cleanup()` parks the loop before closing the listener fd so a reused descriptor can never be polled by a stale loop. Regression test `oneWayFeedTelemetryIsDrainedBeforeTheCommandLogIsRead` models the late reader with an injected start delay (red first, green after).

### 3. Deterministic thrown-error tests

- `hooks grok|agy install` refuses to install without a pinned target (`CMUX_SOCKET_PATH`/`CMUX_TAG`) since 65e56c4f54; five install tests predating that contract exited 1 and then threw from `Data(contentsOf:)`. They now pin a socket path like the other pinned-install tests. Two of them also carried expectations from before the pin contract: the installed command is now `'<bundled cmux>' --socket '<pin>' hooks grok notification`, so the literal `cmux hooks grok notification` substring can never match, and the hooks-directory conflict message is "…. Remove or rename the conflicting file, then run `cmux hooks setup` again." Both assertions now match the current CLI output.
- The implicit socket-reroute notice (`cmux: default socket … is unavailable; using …`) was printed before command dispatch for every invocation, including commands that never open the socket. On the app-host runner the isolated app host is always a live fallback, so `GhosttyConfigTests.testThemesListIncludesCmuxUserThemesDirectory` (which merges stderr into `--json themes list`) threw a JSON parse error every run. The notice is now parked in `CLISocketRerouteNotice` and written by `SocketClient` immediately before its first connection, so it still appears for every command that actually connects through the rerouted socket. Regression test `CLISocketRerouteNoticeTests` (red first, green after). No new user-facing strings.

### 4. App-host crash from a Swift Testing suite

`SSHConfiguredRemoteCommandHostTests` waited through `XCTestCase.wait(for:)` on a helper `XCTestCase` with no running test; a timeout raised `NSInternalInconsistencyException` ("was recorded by test … which does not have a testCaseRun") and aborted the whole app host (baseline shard 1: `Terminating app due to uncaught exception` → `Restarting after unexpected exit`). It now uses `XCTWaiter().wait(for:)` and `#expect`s the outcome.

### 5. `BrowserPanelViewIdentityTests` measured AppKit's display cycle, not the portal

`portalPresentationRefreshDoesNotDriveSwiftUILayout` / `externalPortalGeometrySyncDoesNotDriveSwiftUILayout` dirtied the SwiftUI hosting view and then awaited two main-queue turns before asserting `layoutPassCount == 0`. XCTest drives the main run loop while an async test runs, so any tick of AppKit's display cycle in that window (`UpdateCycle` → `CATransaction` flush → `NSDisplayCycleFlush` → `NSWindow.layoutIfNeeded`) lays out every dirty window, ordered in or not — the captured backtrace has no portal frame in it. That is the per-batch flip on CI (baseline shard 2 red, main shard 4 green then red in one job) and a deterministic failure on a local Mac. Both tests now dirty the reference view, drive the portal's deferred pass synchronously through DEBUG-only hooks (`synchronizeExternalGeometryNowForTesting`, `flushPendingHostedWebViewRefreshesForTesting`), and assert before yielding; the resize-notification wiring is still exercised while the view is clean. Verified alone and inside the full 307-entry shard-2 batch recipe from job 98396168342.

### 6. Deterministic shard-1 failures (all reproduce locally, all identical on `main`)

- **`mobile.panel.artifact.*` over the socket answered `internal_error`** (`testMobilePanelArtifactMethodsRunOnSocketWorker`, `testHeartbeatMethodsSupportInProcessAndSocketDispatch`): `ControlCommandExecutionPolicy` routes the three panel-artifact reads to the socket worker, but the worker switch had no case for them, so every socket client hit the policy-drift backstop; `system.capabilities` did not advertise them either. Product fix: the worker lane now runs the same async dispatch the mobile RPC path uses, and the methods are advertised.
- **`configureRemoteConnection` seeded the initial remote surface before releasing disconnect placeholders** (`testPersistentSSHPTYRestoreRewritesStaleRemoteRelayContextIDs`): a terminal whose placeholder ownership was being reset by the new configuration was skipped by the seed and then dropped from the placeholder set — neither parked nor tracked — so its preserved persistent-PTY session id never matched again until a manual retrack. Product fix: release the set first. The relay test asserts exactly this and has been red since #6409 introduced placeholders.
- **`testOpenCodeForkSupportCachesUnsupportedVersion`**: since #8173 the convenience `supportsFork(snapshot:)` probes with a throwaway cache (the cache is owned by the caller; `SharedLiveAgentIndex` keeps one per index). The test now shares one cache and resolver across both probes, and shows a fresh cache probes again.
- **`testPersistentPTYBootstrapReinstallsOldDaemonMissingPTYCapability`** (and the sibling `testDaemonBootstrapUploadUsesAbsoluteHomePathForRemoteDestination`, red on `main` shard 2): the bootstrap probe only trusts an existing daemon when `__CMUX_REMOTE_SIZE__` is non-zero (cc9ea0aba9), so the fixture reported the daemon as missing and ran a first install — and a local Go build — before any hello; and the upload script backgrounds `cat <&3 >&4` (4fffdfc128), which the old `cat > ` / `cat <&3 > ` matchers never saw. Fixture and matcher updated; the reinstall path is what runs now (traced: probe → hello → diagnostics → rm → mkdir → upload → verify).
- **`AppDelegateIssue2907RoutingTests` (3 tests)**: since 3131545160 an agent-hook binding projects a managed restorable snapshot — a command-only binding yields typed argv, a replacement binding rebuilds argv from its own session and cwd, and a newer binding's snapshot supersedes a stale one even after the binding is cleared. The expectations predated that and asserted the compatibility-shell path.

### 7. App-host crash at the tail of a batch: `AgentChatTranscriptService.deinit` off the main actor

The local shard-2 batch recipe ended with `Restarting after unexpected exit`. The `Invalid view geometry: x is infinity` / `[FileExplorer] setRootPath` lines next to it are the *restarted* host's startup noise, not the cause. The crash report (`EXC_BREAKPOINT`, thread `com.cmuxterm.app.sessionPersistence`) is `_dispatch_assert_queue_fail ← MainActor.assumeIsolated ← AgentChatTranscriptService.deinit ← AppDelegate.__ivar_destroyer ← _Block_release ← _dispatch_lane_serial_drain`: `AppDelegate.persistSessionSnapshot`'s autosave write block captured `self` and ran on the utility persistence queue, so for an `AppDelegate()` created by a test that block became the delegate's last reference and the whole delegate — with the transcript service it owns — deinited on that queue, where the deinit's `MainActor.assumeIsolated` trapped. Any test that instantiates `AppDelegate` and triggers an autosave can end the host this way, which is exactly how one suite's cleanup turns every suite still in flight red.

Fix at the source: the deinit is nonisolated, so it now tears the prose streamer and wake driver down synchronously when already on the main thread and on the next main-queue turn otherwise (the block retains both until then); and the autosave write block captures only the `nonisolated` snapshot store instead of `self`, so the delegate is never released on the persistence queue at all. Regression test `AgentChatTranscriptServiceDeinitTests.releasingTheServiceOffTheMainActorDoesNotTrap` hands the service to a queue block as its only owner (red first: it takes the host down; green after).

## Still open (not fixed here)

- `testInitialPersistentAttachReauthenticatesAfterTransportLoss`, `testRestoredPersistentAttachReauthenticatesAfterTransportLoss`, `testDefaultFreestyleSSHAttachScopesTmuxSessionToCallerSurface`, `WorkspaceSSHFishShellTests.testMultiplexedSSHForegroundAuthDoesNotDependOnLocalCommand`: the generated foreground-auth/reconnect wrapper exits 255 on the runner before the fake `ssh`/attach ever runs (stderr `^D`), then the tests throw on the missing attempt files. Serial XCTest phase, so not the starvation above; needs a runner-side repro.
- The local reproduction of the batch-tail host restart is fixed in section 7. Two unsymbolicated app-host crashes remain from the CI logs, whose signatures differ from that one: `Bad pointer dereference at 0x5f595454534f4847` ("GHOSTTY_" as a pointer) on thread 0 right after `[FileExplorer] setRootPath` / `Invalid view geometry: x is infinity` (baseline shard 1), and a `System trap` on thread 61 during `MobileHostConnectionEventLaneTests.testIndependentEventBackpressureClosesAtBoundedQueueCapacity` (main shard 4).
- `TerminalControllerSocketSecurityTests` never finished on `main` (baseline shard 1 logs `◇ Suite … started.` and nothing after it). With the deadline fixed it runs to completion; the two panel-artifact failures are fixed above, and `testRemoteConfigureUsesLastForwardAgentOption`, `testWindowDockRemoteReadinessSurvivesLaunchWorkspaceRemoval`, and an intermittent `testRemotePTYBridgeRoutesMovedSurfaceToCurrentWorkspace` fail only on a developer Mac (local ssh agent / display state), not on the runner.
- Dozens of tolerated assertion failures (`AppDelegateIssue2907RoutingTests`, `RightSidebarModeShortcutHintTests`, `CommandPaletteShortcutCustomizationTests`, `BrowserWindowPortalLifecycleTests`, `CMUXOpenCommandTests`, …) that the gate ignores; they need their own passes.
- `IrxLiveQUICTests.swift:210` sleep-then-assert is flagged by the determinism gate on `main`; fixed separately in #10922.

## Verification

- `python3 tests/test_ci_xcodebuild_noninteractive_helper.py` → PASS (new mixed-phase cases).
- `python3 scripts/check-test-determinism.py --strict` → only the pre-existing `IrxLiveQUICTests` finding.
- `./scripts/lint-pbxproj-test-wiring.sh` → ok.
- Local `cmux-unit` run of the touched suites: see the PR comment below.
- CI dispatched with `gh workflow run ci.yml --ref fix/app-host-lane`.

## Summary by CodeRabbit

* **New Features**
  * Socket reroute notices now appear when a command establishes a connection, while socket-free commands remain quiet.
  * Added support for asynchronous mobile artifact operations, including status, retrieval, and thumbnails.
  * Remote surface restoration better preserves terminal sessions and managed command snapshots.

* **Bug Fixes**
  * Improved process handling under heavy system load.
  * Fixed cleanup when agent chat services are released off the main thread.
  * Improved remote connection setup, browser refresh behavior, and session persistence.
  * Enhanced CI handling for Swift Testing phases and failures.


austinywang:
## Local `cmux-unit` run of the touched suites (5d2cfbb086)

`xcodebuild test-without-building -scheme cmux-unit` with `TEST_RUNNER_CMUX_TEST_PROCESS=1`, focused on the suites this PR touches. No `timedOut: true`, no `status: 15`, no app-host crash or restart, no post-test kill — every remaining failure is a plain assertion that is either identical on `main` or newly surfaced because the suite now runs to completion.

**Now green (were red/thrown on `main`)**
- `CLIProcessHelperStarvationTests` (3/3), `CLISocketRerouteNoticeTests` (1/1), `GhosttyConfigTests.testThemesListIncludesCmuxUserThemesDirectory` — the new regression tests and the `--json themes list` parse.
- All five pinned Grok/Antigravity install tests: `testAntigravityHookInstallUsesNativeHooksJSONShape`, `testGrokHookInstallRoutesNotificationEventToNotificationSubcommand`, `testGrokHookInstallPreservesUserWrappedLegacyCommands`, `testGrokHookInstallPreservesLegacyFileMetadataWhenPruningOwnedHooks`, `testGrokHookInstallRejectsFileAtHooksDirectory`.
- `ClaudeHookPIDAuthenticationTests`, `ClaudeHookLifecycleCleanupTests`, `ClaudeHookSurfaceResolutionSwiftTests`, `CLIOmpSupersededCleanupTests`, `CMUXOpenHTMLFocusTests`, `Campfire hook notifications`, `Feed coordinator` — all pass, none with a starvation timeout.
- `SSHConfiguredRemoteCommandHostTests` no longer aborts the app host; `sshStartupConnectsWhenHostConfigSetsRemoteCommandAndRequestTTY` records an expectation failure locally (attach never reached the mock within 5 s on a loaded machine) instead of an `NSInternalInconsistencyException`; it passes on both baseline CI shards that ran it.

**Still failing, identical on `main` (baseline run 33034886551 / main run 33027651263)**
- `testClaudeClearSessionStartMarksWorkspaceRunning` (`{}` vs `OK`), `CLICodexResumeNotificationTests` (2 tests), `SSHRemoteCommandChainingTests.nonPersistentRestorePreservesExplicitRemoteCommandIntent`, `RightSidebarRemoteCommandTests.v1CommandsRejectCustomSidebarNames`.
- `testInitialPersistentAttachReauthenticatesAfterTransportLoss` and `WorkspaceSSHFishShellTests.testMultiplexedSSHForegroundAuthDoesNotDependOnLocalCommand` — the `255` wrapper exit listed under "Still open".

**Newly surfaced (suite never finished on `main`)**
- `TerminalControllerSocketSecurityTests`: `testRemoteConfigureUsesLastForwardAgentOption`, `testHeartbeatMethodsSupportInProcessAndSocketDispatch`, `testMobilePanelArtifactMethodsRunOnSocketWorker`, `testWindowDockRemoteReadinessSurvivesLaunchWorkspaceRemoval`, intermittent `testRemotePTYBridgeRoutesMovedSurfaceToCurrentWorkspace`. Baseline shard 1 logs `◇ Suite TerminalControllerSocketSecurityTests started.` and nothing after it.

XCTest phase: 13 tests, 8 failures (4 thrown — the two `255` tests above). Swift Testing phase: 117 tests in 11 suites, 30 issues, all assertion-only.

PR #10966, opened 2026-08-27
cmux-tui: sidebar split columns and all-scope agents view

### LINKED ISSUE #11042: cmux-tui: remote transport loss synthesizes MuxEvent::Empty, silently exiting the TUI with code 0 mid-session
Found while byte-accounting https://github.com/manaflow-ai/cmux/issues/10431 (evidence and repro on PR https://github.com/manaflow-ai/cmux/pull/11041). This is the `FileNotFoundError`-on-the-control-socket variant from that issue's comment, and the causality is not what the comment guessed: the shell does not die first.

**Behavior.** When the remote event transport ends for any reason, the `session/remote.rs` reader thread synthesizes `MuxEvent::Empty`:

```rust
// Connection lost: retain the reason before telling the app to quit.
if let Some(session) = reader_session.upgrade() {
    session.disconnect_transport_with_reason(reason);
    session.emit(MuxEvent::Empty);
}
```

The app's `MuxEvent::Empty` handler means "the session has zero workspaces" and quits cleanly (exit 0). The stored disconnect reason is never consulted; `Session::transport_disconnect_reason` is dead code (rustc warns "never used"). So any transport loss is indistinguishable from a deliberately emptied session: the TUI exits 0 with no message, unlinks its control socket, and its teardown then SIGHUPs the still-live pane shell.

**How it fires under load.** The server terminates an event stream whose per-stream outbound queue overflows (`push_regular_locked` -> `terminate_stream_locked` in `cmux-tui-core/src/server.rs`). A paste flood plus a CPU-saturated client is enough: in the issue-10431 repro (Blacksmith 32-vCPU box, ~40 spinners, 860-byte paste into an `--ephemeral` TUI) the TUI exits 0 mid-paste roughly once per 10-25 pastes.

**Instrumented timeline of one instance** (audit notes from PR #11041's `CMUX_TUI_INPUT_AUDIT`, wall-clock micros):

```
1787886054742965 quit app.rs (MuxEvent::Empty handler)
1787886054742973 event-loop-exit quit=true shutdown_requested=false daemon_shutdown=false owner_shutdown=false
1787886054756392 surface-lifecycle surface=2 local reader EOF        <- teardown closed the master
1787886054759011 surface-lifecycle surface=2 child exit: Signal { signal: 1 }  <- teardown SIGHUP
1787886054759030 surface-lifecycle mux surface_exited id=2
```

The mux-side `emit_empty_if_current` note never fired, and the workspace still held a live shell, so the `Empty` could only have come from the remote reader's synthesis. The pane shell's death is a consequence of the quit, not its cause.

**Why it matters.** In CI this looks like an unexplained clean TUI death (`read-screen` gets `FileNotFoundError` because the socket was unlinked by the clean-shutdown path). Interactively, an attached client that loses its daemon connection silently exits 0 instead of saying why.

**Suggested direction.** Distinguish "the daemon deliberately ended the session" (protocol goodbye/detach notice) from "the transport died" (EOF without notice, read error, oversized frame). The latter should surface `transport_disconnect_reason()` and exit non-zero (or reconnect for an in-process owner), not impersonate an empty session.

lawrencecchen:
The TUI sidebar gains a split primitive: a `sidebar.views` entry can now be a split group instead of a leaf view. `{"split": "vertical", "panes": [...]}` stacks panes top to bottom inside one column, `"horizontal"` places them side by side, groups nest up to three levels, and each pane takes an optional `weight`. The motivating layout ships as `cmux-tui/examples/sidebar-splits.json`: workspaces on top, all agents below.

```json
"sidebar": { "views": [
  {"id": "left", "split": "vertical", "panes": [
    {"id": "workspaces", "levels": ["workspaces"]},
    {"id": "all-agents", "levels": ["agents"], "scope": "all", "weight": 2}
  ]}
]}
```

Mechanism. Config resolution now produces the flat view list plus a layout forest (`SidebarLayoutNode::Leaf|Split`); every existing consumer of `sidebar.views` is untouched. The column solver (`sidebar_layout_for_state`) packs top-level nodes exactly as it packed views before (same collapse priorities, minimum widths, reveal hysteresis), then recursively partitions each column's rect: vertical splits carve a one-row divider between children, horizontal splits share the width, and children that no longer fit at minimum size hide lowest-collapse-priority first, mirroring column behavior. If `views` is replaced without a matching layout (tests, older paths), the solver falls back to one column per view, so the old behavior is the degenerate case.

Interaction. Dividers drag with the mouse (`Hit::SidebarSplitDivider` for vertical dividers; a horizontal split reuses the left child's border column and the press handler reroutes it), committing session-local share fractions. A width drag on any stacked rail resizes the whole column under the split group's id. Rail focus is now geometric: Left/Right pick the nearest column with the best vertical overlap, Up/Down move within a column, and pressing Up/Down at a rail's first/last row continues into the stacked neighbor (workspaces, tabs, and projection rails).

`scope: "all"` on flat `tabs`/`agents` views sweeps every workspace instead of following the selected one. All-scoped agents order by `updated_at_ms` newest first, keep `done` agents visible (it is a status board, not an active-agent list; `unknown` stays hidden), and show the workspace name as subtitle when the agent has no session title. `all`-scoped `tabs` views never map to the legacy selected-workspace tabs rail.

Sidebar layout state stays frontend-local per `spec/native-frontend.md`: fractions and width overrides are session-only, config is the durable source.

Tests: split grammar resolution (weights, scope validation, single-pane unwrap, bad-direction/ambiguous-group rejection, JSON parse), solver geometry (stacking, divider rects, fraction overrides, hidden-pane collapse), divider drag re-sharing with min clamps, column-wide width drag, vertical focus movement, and all-scope projection ordering across workspaces.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/10966"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790421777&installation_model_id=21920&pr_number=10966&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F10966&signature=4b4eee07c1f08c4d0c1a56de996dd341017c46a32ae9efc8af45aab4a46d6c2f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Sidebar columns previously held one view tied to the selected workspace; they can now contain nested split panes and views that span all workspaces. Remote transport loss previously looked like an empty session and exited quietly; it now surfaces a localized nonzero error, while genuinely empty sessions retain the quiet exit.

**Sidebar**
- `sidebar.views` entries support vertical or horizontal `panes`, nesting up to three levels with optional weights; documentation and an example config are included.
- Split dividers are draggable, keyboard focus moves between stacked panes, and low-priority panes collapse first when space runs out.
- `scope: "all"` makes flat `tabs` and `agents` views enumerate every workspace; agents hide `unknown`, retain `done`, sort by latest status change, and use the workspace name when no session title exists.
- Profile changes restore configured layouts, while bounded projection caching preserves selection by resource identity.

**Transport handling**
- Empty-session events attempt machine-session reconnects before reporting the recorded transport failure.
- Deliberate local disconnects remain quiet, and transport errors include recovery guidance.
- Fixes #11042.

<sup>Written for commit 79ffd93866754498f86abca190f6faadce9047ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/10966?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable nested sidebar split groups with vertical or horizontal panes.
  * Resize stacked panes by dragging dividers, with shared column widths and persisted layout settings.
  * Navigate focus across stacked sidebar sections using keyboard controls.
  * Added “all workspaces” views for tabs and agents, with chronological ordering and workspace context.
  * Completed agents remain visible in all-workspace views, while unknown agents are filtered out.
  * Profile switching restores the configured sidebar layout.
  * Preserve selection when sidebar resources are reordered.

* **Documentation**
  * Added sidebar split and resource-scope configuration guidance, plus an example layout.

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/app.rs
+++ b/cmux-tui/crates/cmux-tui/src/app.rs
@@ -84,6 +84,7 @@ use crate::session::{
 use crate::sidebar_files::{FileBrowser, FileCommand, file_url, shell_single_quote};
 use crate::sidebar_projection::{
     ProjectionBranch, ProjectionRailState, ProjectionRow, ProjectionTarget,
+    selected_workspace_cache_key,
 };
 use crate::ui::graphics::{
     GraphicPlacement, GraphicSourceRect, kitty_graphic_image, kitty_graphic_placement,
@@ -100,6 +101,11 @@ use crate::ui::{
 
 const DEFERRED_INPUT_CAPACITY: usize = 512;
 const MAX_PROJECTION_ROWS_CACHE_ENTRIES: usize = 8;
+
+/// Session-local divider shares, keyed by split group and stable child id.
+/// Child ids keep a drag ratio attached to its rail when visibility pruning
+/// or layout reordering changes the current child vector.
+type SidebarSplitFractions = HashMap<String, HashMap<String, f32>>;
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 struct ProjectionFocusKey {
@@ -143,14 +149,14 @@ struct ProjectionRowsCache {
     workspace_revision: u64,
     pane_revision: Option<u64>,
     invalidation_revision: u64,
-    selected_workspace: usize,
+    selected_workspace: Option<usize>,
     focus: ProjectionFocusKey,
     /// All surfaces that can contribute an agent row to this snapshot. This
     /// includes currently hidden/filtered rows so an agent state transition
     /// cannot leave an old empty snapshot cached.
-    agent_surfaces: HashSet<SurfaceId>,
+    agent_surfaces: Arc<HashSet<SurfaceId>>,
     /// Surfaces whose title can affect a visible row in this snapshot.
-    title_surfaces: HashSet<SurfaceId>,
+    title_surfaces: Arc<HashSet<SurfaceId>>,
     rows: Arc<[ProjectionRow]>,
 }
 
@@ -7074,6 +7080,11 @@ pub struct App {
     /// retired surface cannot accumulate or trigger future fanout.
     projection_agent_surfaces: HashSet<SurfaceId>,
     projection_title_surfaces: HashSet<SurfaceId>,
+    /// Dependencies are retained per active view, not only in a global union.
+    /// This lets hidden views be removed without losing dependencies for a
+    /// visible view whose bounded row snapshot was evicted.
+    projection_agent_surfaces_by_view: HashMap<String, Arc<HashSet<SurfaceId>>>,
+    projection_title_surfaces_by_view: HashMap<String, Arc<HashSet<SurfaceId>>>,
     /// Coalesce surface updates until the next successful rendered frame.
     projection_paint_pending: bool,
     pub(crate) machine_rail_follow_selection: bool,
@@ -7094,9 +7105,10 @@ pub struct App {
     tabs_sidebar_width_override: Option<u16>,
     projection_sidebar_width_overrides: HashMap<String, u16>,
     /// Session-local divider drags inside sidebar split groups for the active
-    /// profile. A profile switch clears these values so a reused group id
-    /// cannot apply another profile's geometry.
-    sidebar_split_fractions: HashMap<String, Vec<f32>>,
+    /// profile. Values are keyed by stable child id, and a profile switch
+    /// clears them so a reused group id cannot apply another profile's
+    /// geometry.
+    sidebar_split_fractions: SidebarSplitFractions,
     /// Session-local visibility overrides, keyed by profile and stable view id.
     hidden_sidebar_views: HashMap<String, HashSet<String>>,
     /// Pane region of the current frame (screen minus sidebar/status).
@@ -8087,7 +8099,7 @@ fn place_sidebar_column_node(
     node: &SidebarColumnNode,
     rect: Rect,
     layout: &mut SidebarLayout,
-    split_fractions: &HashMap<String, Vec<f32>>,
+    split_fractions: &SidebarSplitFractions,
 ) {
     use crate::config::SidebarSplitDir;
     if rect.width == 0 || rect.height == 0 {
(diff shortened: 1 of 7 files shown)


lawrencecchen:
Heads-up for whoever is iterating this branch: with https://github.com/manaflow-ai/cmux/pull/11002 an ended agent session is dropped from the roster daemon-side, so the `keep_done` special case for all-scoped agents views in `projection_rows` only ever shows stale remote-cache rows. I reverted it on the throwaway dogfood branch `feat-tui-sbsplit-dogfood` (commit "hide done agents in every agents-view scope"); please fold the same revert into this branch rather than merging keep_done as-is.

PR #11002, opened 2026-08-27
cmux-tui: journal reducer framework; derive the agent roster from the session journal

lawrencecchen:
The agents view is now a materialized view of the session journal, per the state-ownership plan's "Reducer" primitive (its first instance).

**Framework** (`journal_reducers.rs`): a reducer is a pure fold over committed journal records with a durable cursor (last folded sequence) and a state snapshot, persisted in the registry `meta` table (no schema migration). Startup restores the snapshot and folds only the journal tail; bumping the reducer version discards the snapshot and re-folds from the journal head. The live fold reads each committed record back from the journal before folding, so live folds and replays are literally the same computation over the same bytes. A test proves derivation by wiping the persisted snapshot and asserting the re-folded roster is identical.

**Agent roster reducer**: terminal -> {state, source, session, updated_at}. Hook events map session.started/turn.completed -> idle, turn.started -> working, approval/question/plan-review/error -> blocked; session.ended removes the entry (exited agents disappear from agents views; history stays in the journal and the durable projection). Deltas from the fold drive the projection commits and `agent-changed` broadcasts, so remote frontends converge on the same states.

**Single write path**: direct socket/SDK `agent report` keeps its synchronous projection commit and replay contract, and now also appends an echo journal event carrying the committed state/timestamp; the fold recognizes the echo adapter and applies it roster-only. Every agent intent is therefore in the log, and the roster has exactly one writer. Hook-beats-socket precedence is arbitrated on the durable projection row under the commit's registry lock (serializing concurrent reports) and expressed identically in the fold; a `done` projection no longer pins precedence, so a fresh agent in the same terminal starts clean. The old restored-projection roster cache is deleted.

Behavior notes: each fresh direct report publishes twice on the shared change epoch (resource commit + journal echo) - affected tests document it; terminal close retires the roster entry explicitly until terminal lifecycle flows through journaled events; a version-bump re-fold only sees retained journal history, which is correct for a live roster and the reason ledger-shaped reducers should version conservatively.

Commit narrative: red test (hook events must drive records), imperative stopgap, exit-removal, then the reducer framework superseding the stopgap's internals.

Found while dogfooding https://github.com/manaflow-ai/cmux/pull/10966 (all-agents sidebar view). Verified end to end with real claude in tmux; codex verified via `codex exec` (its hooks fire the same pipeline).

## Summary by CodeRabbit

* **New Features**
  * Agent activity statuses now update automatically from lifecycle events, including idle, working, blocked, and completed states.
  * Status updates are associated with the relevant terminal and agent when available, including child-agent sessions.
  * Hook- and socket-based updates are reflected in live terminal activity.
  * Agent status information is restored after restarting from persisted activity history.

* **Bug Fixes**
  * Prevented replayed events from applying duplicate status updates.
  * Preserved hook-owned status when direct reports are received.
  * Safely handles events for terminals that are no longer active.
  * Cleans up terminal records when agent sessions end.


lawrencecchen:
Added in 6ecbde8415: an `agent.session.ended` commit now also drops the live record after broadcasting the done state, so exiting an agent removes it from agents views (a fresh agent in the same terminal starts clean). Terminal close already purged records; this covers the agent exiting while its shell stays open.

lawrencecchen:
8da5643df4 adds the agent adapter id to the roster (reducer snapshot v2): every hook event names its agent (claude, codex, ...), agent records / the agent-changed event / AgentInfo expose it, and rows label by agent type with the detail line leading with it (claude · working · <ws>). Socket-only reports leave it absent until a hook claims the terminal. The durable projection JSON is deliberately unchanged (SDK schema and its equality contracts intact); adding the type there is a follow-up. Demoed live: labeled rows, chronological reorder on activity, exit removal, all on the combined dogfood branch.

lawrencecchen:
Closing the stale stack parent. Its reducer and agent-status work was superseded by the userland screen-detection line in https://github.com/manaflow-ai/cmux/pull/11454, which carries the active replacement architecture. This branch is hundreds of commits behind and has unresolved lifecycle findings.

Issue #11008, opened 2026-08-27
Cloud machine link: 'Connecting…' forever after machine churn; wedged attempts leak dial processes and mask the timeout error


lawrencecchen:
Repro observed live on 2026-08-27 (dev stack, tag ctw0/ctw1 dogfood, blaxel provider) while sandboxes were deleted and recreated several times in one hour.

Symptom: the cloud tree shows the machine's spinner "Connecting…" indefinitely (user screenshot at 14:57, machine created 14:14). `surface ls --json` shows `link_state: connecting`, `link_error: null` for 40+ minutes.

What was proven, in order:

1. The machine, daemon, route, tokens, enrollment, and shared client state dir are all fine. The same bundled binary (`cmux DEV ctw0.app/Contents/Resources/bin/cmux-tui remote connect <route> --state-dir ~/.cmuxterm/cmux-tui-client ...`) run from a shell, with the app's own enrolled identity and freshly minted tokens, prints its connection-snapshot line in 0.6-4 s, every attempt (5/5), including via a Python spawn with piped stdout that mirrors the app's spawn shape.
2. In the wedged app process, spawned dial children DO establish transport (3 established TCP lanes to the preview gateway), but the app's 60 s reader in `CloudMachineLink.connect` never consumes the snapshot line; each attempt dies with "cmux-tui link did not report a socket within the connect timeout".
3. The UI never shows that error: each retry immediately re-enters `connecting`, so the transient state is what the tree renders, forever. The recorded `link_error` is only visible between attempts.
4. Wedged attempts leak `remote connect` child processes. Found at once on one Mac: a 20-minute-old child of the ctw0 app, a child of the ctw1 app dialing a machine deleted an hour earlier (still holding 3 ESTABLISHED TCP connections to the gateway edge - the client has no handshake deadline), a child of cmux NIGHTLY dialing another deleted machine, and an orphan (ppid 1) from Aug 25. All four cmux instances share `~/.cmuxterm/cmux-tui-client`.
5. Restarting the app fixes it instantly: a fresh process connected and populated resources on the first try, through ongoing machine churn. The wedge is accumulated per-process state.

[...]

PR #11030, opened 2026-08-28
fix(cmux-tui): bound the remote-connect handshake with a deadline

lawrencecchen:
`cmux-tui remote connect <route>` children dialed machines that had been deleted and held 3 ESTABLISHED TCP connections to the preview gateway edge for an hour or more, never printing their connection-snapshot line and never exiting. The gateway edge still accepts TCP and completes transport for a deleted machine, so the client established its lanes and then waited forever for the Noise handshake reply.

Mechanism: the unbounded wait is in `initiate_secure_link` (`crates/cmux-remote/src/crypto.rs`). For invitation dials (the argv the macOS app spawns uses `--invite-file`), the outer deadline in `remote_runtime.rs` is the whole startup budget: invitation remaining lifetime plus the 5-minute enrollment-approval window plus 15 s. That is deliberate for the post-handshake welcome, which can wait on a human approving the enrollment, but it also covered the machine-speed prelude/Noise exchange, so a silent endpoint pinned the connection for the invitation lifetime.

Fix: `initiate_secure_link` now deadlines the pre-welcome phase (prelude send, Noise exchange, hello) with a new `ClientHandshake.handshake_timeout`, tears the link down on expiry (the close itself bounded at 2 s so a dead peer cannot block teardown), and fails with a new terminal `CryptoError::HandshakeTimeout`. The `ServerWelcome` wait stays outside the deadline, so invitation approval keeps its window. The failure carries a distinct stderr message ("secure link handshake timed out after Nms: the endpoint accepted the connection but never completed the cmux remote handshake"). Review moved its classification from terminal to a retryable carrier failure: a temporary network stall or an overloaded daemon produces the same timeout as a deleted endpoint, so the `ReconnectPolicy` (attempt limits, backoff, startup budget) decides when the client stops redialing and exits, and the torn-down link means no TCP is held between attempts. The deadline itself is what fixes the reported wedge: each dial now fails in seconds instead of pinning ESTABLISHED lanes for the invitation lifetime.

Budget: the deadline is wired from `ReconnectPolicy.attempt_timeout`, default 15 s, overridable with the existing `--reconnect-attempt-timeout-ms` flag. A healthy handshake completes in 0.6-4 s end to end, so 15 s is generous, and reusing the per-attempt timeout adds no new config surface; its doc comment already promises to bound authentication for one attempt.

Principled: the deadline separates the machine-speed handshake phase from the human-speed approval phase at the exact protocol boundary between them, rather than shrinking the invitation budget or patching one caller, and the retryable classification keeps retry policy in one place (`ReconnectPolicy`) instead of encoding it in the error type. Residual behavior to know: with the default unlimited reconnect attempts, an invitation dial to a dead endpoint cycles bounded dials (15 s hold, then teardown and backoff) until the startup budget expires, rather than exiting on the first timeout; configure `--reconnect-attempts` for a faster give-up.

Regression test in two commits so CI goes red then green: commit 1 adds `initial_connect_fails_fast_when_the_endpoint_accepts_but_never_handshakes`, a link group whose endpoint accepts and then goes silent; a 5 s outer bound converts the historical hang into a deterministic failure ("connect held a silent endpoint open past its handshake deadline: Elapsed(())", red hosted run https://github.com/manaflow-ai/cmux/actions/runs/33125881195). Commit 2 adds the fix; the test passes on hosted Linux and macOS in https://github.com/manaflow-ai/cmux/actions/runs/33135190500 (a follow-up commit only applies the rustfmt layout for the new enum variants). The review round (terminal to retryable) is verified green on hosted Linux and macOS in https://github.com/manaflow-ai/cmux/actions/runs/33136261809; the test now uses a one-attempt reconnect policy and asserts the timeout is surfaced as a retryable carrier failure.

Out of scope: the Swift `CloudMachineLink` attempt lifecycle (leaked children, cancellation-deaf stdout readers, persistent link error state) from the same issue.

Part of https://github.com/manaflow-ai/cmux/issues/11008

---
## Summary by cubic
Bounds the remote-connect handshake with a deadline so a dial against a deleted machine's preview endpoint fails fast instead of holding the connection open for the invitation lifetime. The client now tears down the link and reports a handshake-timeout error when the endpoint accepts the connection but never completes the cmux handshake.

**Bug Fixes**
- The deadline covers only the pre-welcome phase (prelude, Noise exchange, hello); the post-handshake welcome still waits for enrollment approval.
- The deadline reuses `ReconnectPolicy.attempt_timeout` (default 15s, overridable with `--reconnect-attempt-timeout-ms`), so no new config surface is added.
- `HandshakeTimeout` is a retryable carrier failure, so the `ReconnectPolicy` decides when the client stops redialing and exits.
- Adds a regression test with an endpoint that accepts the transport and then goes silent.
- Part of https://github.com/manaflow-ai/cmux/issues/11008.

<sup>Written for commit 4295e446dcdc2b0c52e0084dd5a1ec31c35d0f25. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11030?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Connection attempts now time out if the remote endpoint accepts a connection but does not complete the initial handshake.
  - Timed-out handshakes close the underlying connection and report a clear error.
  - Reconnect attempt limits now consistently apply to invitation-based connections.

- **Documentation**
  - Clarified handshake timeout behavior and documented reconnect and heartbeat options.


lawrencecchen:
Local review-gate audit note: one pass flagged a P0 claiming `matches!(error, CryptoError::DaemonKeyMismatch { .. })` moves the non-Copy error before `return Err(error)` (E0382). That is incorrect: a `{ .. }` pattern binds nothing, so the match only inspects the discriminant and does not move the scrutinee. The exact head 4295e446dc compiles and passes the cmux-remote suite on hosted Linux and macOS: https://github.com/manaflow-ai/cmux/actions/runs/33136261809. Dismissed with that evidence; a fresh review pass runs on the same head.

PR #11035, opened 2026-08-28
Reap disowned shell watchers on parent PID reuse (issue 10926)

### LINKED ISSUE #10926: Shell integration leaks two disowned zsh watchers per pane — 793 orphans / 2.1 GB / 208 concurrent `gh` after 20 days
## Summary

The zsh shell integration spawns two disowned background watchers per pane. They are never reaped when the pty closes, so they accumulate indefinitely. On my machine **793 orphaned watcher shells had piled up over 20 days, holding 2.1 GB RSS**, and collectively spawning up to **208 concurrent `gh` processes**.

This is the `Orphaned login → zsh → zsh trees accumulate after pane close` lifecycle path mentioned in passing in #2962 — this issue pins it to the specific code and quantifies it.

## Environment

- cmux 0.62.2 (build 77)
- macOS 15.6.1, Darwin 24.6.0, arm64
- zsh shell integration (`Contents/Resources/shell-integration/cmux-zsh-integration.zsh`)

## Evidence

Sampling `pgrep -x zsh` and `pgrep -x gh` every 2 seconds:

```
zsh processes:        793 orphans (PPID=1, tty=??, nice=5)
combined RSS:         2,133 MB
oldest:               20 days 18 hours
accumulation rate:    ~40/day, steady over the whole 20-day window
gh processes:         208 concurrent at peak (43 MB and 0.82 s each)
```

Age histogram was flat across the entire period — 86 from today, 66 one day old, 75 two days old, and so on out to 41 at nineteen days and 39 at twenty. Nothing ever reaps them.

Every orphan had a `sleep` child:

```
440 × sleep 45   (PR poll loop)
335 × sleep 1    (git HEAD watch)
```

Each orphan's fd table shows stdout/stderr on `/dev/null` and **no stdin at all** — the pty is long gone, but the loop keeps running.

Tracing the parents of the `gh` swarm showed every one of them was a `/bin/zsh` orphan. In my case this initially looked like a runaway daemon of my own; it was not.

## Root cause

Two loops in `cmux-zsh-integration.zsh`, both launched with `&!` (zsh disown), so no SIGHUP reaches them when the pty closes.

**1. `_cmux_start_git_head_watch` (~L335) — no parent-liveness guard at all:**

```zsh
{
    local last_signature="$watch_head_signature"
    while true; do
        sleep 1
        local signature
        signature="$(_cmux_git_head_signature "$watch_head_path" 2>/dev/null || true)"
        ...
    done
} >/dev/null 2>&1 &!
```

Nothing in this loop can ever terminate it. It polls a git HEAD file once per second, forever.

**2. `_cmux_start_pr_poll_loop` (~L299) — guard present but ineffective:**

```zsh
local watch_shell_pid="$$"
...
{
    while true; do
        kill -0 "$watch_shell_pid" >/dev/null 2>&1 || break
        _cmux_run_pr_probe_with_timeout "$watch_pwd" || true
        sleep "$interval"     # 45
    done
} >/dev/null 2>&1 &!
```

**Hypothesis (not directly verified):** the `kill -0` guard is defeated by PID reuse. macOS recycles PIDs within a day or two on a busy machine, so once the recorded `watch_shell_pid` is reassigned to any live process the guard returns true forever. Supporting evidence: the guarded loop left **more** survivors (440) than the unguarded one (335), which is hard to explain if the guard were working.

## Impact

- Unbounded process and memory growth proportional to uptime × panes opened.
- A `gh pr view` storm: ~800 watchers on a 45 s cycle is roughly 18 `gh` spawns per second, which is also a plausible contributor to the rate-limit exhaustion reported in #9843.
- Once-per-second `git` HEAD stat traffic multiplied by the number of dead panes.
- Every watcher survives cmux restarts and app updates, since they are reparented to launchd.

## Reproduction

1. Open a cmux pane in a git repo with a GitHub remote.
2. Close the pane (or the whole workspace).
3. `ps -Ao pid=,ppid=,nice=,command= | awk 'NF==4 && $2==1 && $3>0 && $4=="-/bin/zsh"'`

The watcher shells from the closed pane are still there, with `sleep 45` / `sleep 1` children. Repeat over days and they accumulate linearly.

## Workaround

Orphans are cleanly separable from live watchers — the live ones still have a real parent shell:

```bash
ps -Ao pid=,ppid=,nice=,command= \
  | awk 'NF==4 && $2==1 && $3>0 && $4=="-/bin/zsh" {print $1}' \
  | xargs /bin/kill -TERM
```

On my machine this matched 793 orphans and left all 24 live-pane watchers untouched. Result: zsh 823 → 30 processes, 2,133 MB → 69 MB, peak concurrent `gh` 208 → 1.

## Suggested fixes

1. **Minimum**: add the same `kill -0 "$watch_shell_pid" || break` guard to `_cmux_start_git_head_watch`.
2. **Make the guard actually hold**: PID alone is not a stable identity. Capture the parent's start time as well and compare both, or — more robustly — let the watcher inherit a pipe/fd tied to the parent shell and exit on EOF, which is immune to PID reuse.
3. **Best**: terminate the process group on pane close rather than relying on each watcher to notice. This shares a root cause with #1978 (workspace close does not terminate pty child processes).
4. Consider hoisting these watchers out of the shell entirely — one poller in the cmux app process, keyed by workspace, instead of two spawned shells per pane.

## Related

- #2962 — mentions orphaned `login → zsh → zsh` trees as a secondary observation; this is that path, isolated.
- #9843 — the same PR poll loop, from the rate-limit angle.
- #1978 — workspace close not terminating pty children.

lawrencecchen:
Fixes https://github.com/manaflow-ai/cmux/issues/10926.

The disowned watcher loops (zsh PR poll, zsh git HEAD watch, bash PR poll) guard parent liveness with a bare `kill -0 $watch_shell_pid`. macOS recycles PIDs within days on a busy machine, so once the recorded shell PID is reassigned to any live process the guard returns true forever and the watcher never exits. Combined with zshexit hooks not running on the untrapped-SIGHUP pane-close path, this accumulated 793 orphaned watchers holding 2.1 GB and spawning up to 208 concurrent `gh` processes over 20 days.

The fix is principled: it replaces the raw PID with a stable process identity. `_cmux_watcher_parent_start_time` records the parent's `/bin/ps -o lstart=` start time at watcher spawn (whitespace-normalized, since lstart pads single-digit days), and the shared per-iteration guard `_cmux_watcher_parent_alive` requires both PID liveness and a matching start time. A mismatch means the PID was recycled and the watcher exits; a failed ps also counts as parent-dead. Cost is one `/bin/ps` per second per watcher, the same subprocess cadence the loops already have. Residual risk: lstart granularity is 1 second, so a recycled PID whose new process has an identical start time string could still fool the guard; that requires the exact PID to be recycled within the same wall-clock second the original shell started, which is far outside observed reuse timescales.

The bash integration had the same pattern in its PR poll loop and gets the same helpers; its git probe is a one-shot, not a loop. The fish and nushell integrations have no disowned-watcher pattern and are unchanged. Commit 1 adds the failing regression test (`tests/test_issue_10926_watcher_pid_reuse_guard.py`, red without the helpers), commit 2 adds the fix (green); the test injects a live not-the-parent PID with a mismatched start time and asserts bounded-time exit, plus the parent-alive and parent-dead-unreused paths, for both shells, without touching any running cmux or `/tmp/cmux-debug.sock`.

---
## Summary by cubic
Fixes #10926 by pairing each disowned watcher's PID with the parent shell's start time, so a recycled PID no longer keeps the watcher alive — previously watchers accumulated (793 orphans / 2.1 GB after 20 days). Watchers now fail closed: a missing, malformed, or mismatched start time counts as parent-dead, and a watcher skips starting if the start time can't be captured.

**Bug Fixes**

- The identity comes from Darwin's kernel start time (via `sysctl kern.proc.pid.X`), falling back to a locale-independent `lstart` token, so an identically-timed PID recycle is effectively impossible.
- The `kill -0` liveness check runs every iteration; the start-time identity check runs every 30 iterations by default (`_CMUX_WATCHER_IDENTITY_INTERVAL`), adding no steady-state forks.
- The parent's start time is captured and cached once per shell lifetime, so a transient `ps` failure heals on the next watcher start.
- Applies to the zsh PR poll, zsh git HEAD watch, and bash PR poll; fish and nushell are unchanged.
- Regression tests — now wired into CI — cover parent-alive, parent-dead-unreused, recycled-PID, deferred-cadence, fail-closed, and locale-isolation paths against a scratch HOME and private socket, never a running cmux; existing orphans from before this fix are not cleaned up.

<sup>Written for commit b968841f72c69b4477463c109f7a2c91f842acc7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11035?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved background watcher reliability by detecting when a process ID has been reused.
  * Watchers now stop correctly when their original parent process exits or is replaced.
  * Improved watcher performance through periodic process-identity checks and frequent liveness checks.
  * Applied these safeguards consistently across Bash and Zsh integrations.

* **Tests**
  * Added regression coverage for live, terminated, and recycled parent processes across Bash and Zsh integrations.
  * Added the regression test to continuous integration checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1415,6 +1415,7 @@ jobs:
           python3 tests/test_issue_8093_ghostty_ssh_binary_path.py
           python3 tests/test_issue_6714_zsh_shim_noclobber.py
           python3 tests/test_issue_9356_bash_shim_noclobber.py
+          python3 tests/test_issue_10926_watcher_pid_reuse_guard.py
           python3 tests/test_issue_8953_zsh_prompt_wrap_guard.py
           python3 tests/test_nushell_shim_path_refront.py
           python3 tests/test_nushell_resume_command_dialect.py
--- a/Resources/shell-integration/cmux-bash-integration.bash
+++ b/Resources/shell-integration/cmux-bash-integration.bash
@@ -1433,14 +1433,13 @@ _cmux_watcher_parent_start_time() {
     kernel="$(/usr/sbin/sysctl -n "kern.proc.pid.$pid" 2>/dev/null | /usr/bin/od -An -tu4 2>/dev/null)"
     while read -r sec usec; do
         if [[ "$sec" =~ ^[0-9]+$ && "$usec" =~ ^[0-9]+$ && "$sec" -ge 1000000000 && "$sec" -le 3000000000 && "$usec" -lt 1000000 ]]; then
-            printf '%s%06d\n' "$sec" "$usec"
+            printf '%s\n' "$sec"
             return 0
         fi
     done < <(printf '%s\n' "$kernel" | awk '{ for (i=1;i<NF;i++) print $i, $(i+1) }')
     # Darwin's ps exposes process start time through `lstart`, which is a
     # locale-formatted string. Force the stable C locale and UTC timezone,
-    # then convert the five fields to a fixed-width numeric token before
-    # comparing.
+    # then convert the five fields to the same epoch-second token.
     raw="$(TZ=UTC LC_ALL=C /bin/ps -o lstart= -p "$pid" 2>/dev/null)" || return 1
     case "$raw" in *$'\n'*) return 1 ;; esac
     local -a words=()
@@ -1468,7 +1467,8 @@ _cmux_watcher_parent_start_time() {
         [0-9][0-9][0-9][0-9]) year="${words[4]}" ;;
         *) return 1 ;;
     esac
-    token="${year}${month}${day}${clock//:/}"
+    token="$(TZ=UTC LC_ALL=C /bin/date -j -u -f '%a %b %d %T %Y' "$raw" '+%s' 2>/dev/null)" || return 1
+    [[ "$token" =~ ^[0-9]+$ ]] || return 1
     _cmux_watcher_parent_identity_valid "$pid" "$token" || return 1
     printf '%s\n' "$token"
 }
@@ -1485,7 +1485,7 @@ _cmux_watcher_parent_identity_valid() {
     case "$identity" in
         ''|*[!0-9]*) return 1 ;;
     esac
-    (( ${#identity} == 16 || ${#identity} == 14 ))
+    (( ${#identity} >= 10 && ${#identity} <= 11 ))
 }
 
 _cmux_watcher_parent_alive() {
--- a/Resources/shell-integration/cmux-bash-integration.bash
+++ b/Resources/shell-integration/cmux-bash-integration.bash
@@ -1488,6 +1488,17 @@ _cmux_watcher_parent_identity_valid() {
     (( ${#identity} >= 10 && ${#identity} <= 11 ))
 }
 
+_cmux_watcher_parent_state_valid() {
+    local pid="${1:-}" state
+    state="$(LC_ALL=C /bin/ps -o state= -p "$pid" 2>/dev/null)" || return 1
+    state="${state#"${state%%[![:space:]]*}"}"
+    state="${state%%[[:space:]]*}"
+    case "$state" in
+        ''|Z*) return 1 ;;
+        *) return 0 ;;
+    esac
+}
+
 _cmux_watcher_parent_alive() {
     # $1 = parent PID, $2 = numeric start time recorded at watcher spawn. A
     # mismatch means the PID was recycled; a failed /bin/ps counts as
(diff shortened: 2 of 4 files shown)


teamleaderleo:
Reviewed: start-time identity, plus re-checking it every 30 iterations, closes the PID-reuse hole behind #10926, and both merged scripts pass `bash -n` / `zsh -n`. I merged main in. The only conflict was `ci.yml`, where main moved the test list into `tests/test-execution.toml`, so I registered `test_issue_10926_watcher_pid_reuse_guard.py` there (`macos-cli-no-socket`, next to 6714 and 9356). Auto-merge is on. This test has never run green on macOS, so this CI run is its first real check. The stacked follow-up #11161 can retarget to `main` once this lands. Thanks :)

PR #11062, opened 2026-08-28
Cloud terminals: manual-IO data path (attach --pipe-io relay + reconnecting pump)

lawrencecchen:
Cloud machine terminals switch from an exec attach pane (the full `cmux-tui attach` TUI running inside a local PTY) to Ghostty manual IO: the pane's surface parses daemon bytes fed by a new renderer-less relay, `cmux-tui attach --terminal <id> --pipe-io`, running against the machine link's local mux socket. Only cloud machine terminals change; local terminals, ssh workspaces, and remote tmux mirrors keep their paths.

## Rust relay (`attach --pipe-io`)

Extracted from the `feat-tui-manual-io` branch (#10742) and rebased on main: stdout carries the daemon replay then live PTY output, with a full reset (`ESC c` + erase-scrollback) before any replay that is not the relay's first output, because a replay replaces terminal state while a byte stream appends. stdin takes JSON lines `{"input":"<base64>"}` / `{"resize":{"cols":N,"rows":N}}`; stderr ends with one machine-readable exit reason. Exit 0 = terminal ended or embedder closed stdin (never respawn); exit 2 = daemon lost (respawn to resync from a fresh replay). On stream loss the relay probes the daemon once to distinguish a closed terminal from a daemon outage. The daemon-side vt remains the single reply authority, pinned by an e2e test asserting an inner DSR query is answered exactly once. The resource-boundary lint failure that blocked #10742 ('surface' in public help text) is fixed here, and the flag is documented in `spec/cli.md`.

## Swift pump, scoped to cloud

`TuiManualIOPump` owns one relay per cloud terminal pane: pumps relay stdout into the manual-mirror surface through the reviewed remote-output lane, forwards the surface's encoded input to relay stdin, and drives daemon-side sizing from applied resizes. Because the surface parses the daemon terminal's own byte stream, input encodings always match the daemon's mode state; the exec bridge's mode-mirroring bug class cannot exist on this path.

Reconnects: connecting → live → reconnecting(n) → live | ended | failed. Backoff 0.5s..30s cap; explained daemon-lost exits retry forever; five consecutive unexplained failures park in a failed overlay with manual Retry; a respawn resyncs by resetting the surface before the fresh replay; offline input is dropped, never queued. The pane reuses the cloud terminal reconnect overlay, and its Reconnect button skips the pump's remaining backoff. Panel close stops the relay (stdin EOF = clean detach; the daemon terminal survives in the machine's session); a detach transfer keeps the pump alive because the registry is keyed by the stable surface id.

Wiring: `CmuxTuiSurfaceProvider` decides once per open (fresh materialization and restored-pane re-projection share `openTerminalPane`), and the attach descriptor threads through `SurfacePaneFactory` and the control-surface layer (app-internal parameter, never parseable from socket params) into the workspace's existing terminal creation seams, so tab and split placements both work.

## Gating and version skew

`cloud.beta.terminalManualIO.enabled` (Settings > Beta Features > "Cloud Terminal Manual IO", default **on**) gates the path; off falls back to the exec attach pane. The bundled cmux-tui client comes from a rolling artifacts manifest, so an app can carry a client older than `--pipe-io`; a cached `--help` probe detects that and silently uses the exec pane instead of crash-looping the relay.

## Tests

- Rust e2e (`tests/cli.rs`): replay-on-reconnect, exit-reason classification, startup connect failure reports daemon-lost, resize reaches the daemon PTY, inner DSR query answered exactly once.
- Swift (`TuiManualIOPumpTests`): relay exit classification (including "exit 2 without a reason line is a usage error, not a retryable outage"), respawn decision, backoff pacing, stdin wire format, both relay argv forms (session and `--socket`), overlay presentation mapping, input channel pause/drop semantics, registry replace/remove.

Localization audit: new Settings strings (`settings.betaFeatures.cloudTerminalManualIO*`) and overlay strings (`tui.overlay.*`) are in `Resources/Localizable.xcstrings` with en and ja; the curated settings-search entry is registered. No other user-facing strings added.

Relation to #10742: this PR carries the relay and pump forward to main scoped to cloud attachments only; the local daemon-backed tab work in #10742/#10408 can rebase onto it and delete its copies.

## Dogfood round 2: resize and grid-desync fixes

Resizing felt laggy and pane vs PTY grids desynced. Two causes, two fixes:

**Ack-clocked latest-wins resize.** The pump forwarded every applied size sample immediately, and the relay applies each one as a synchronous `resize-surface` round trip on the same stdin thread that carries keystrokes — one divider drag on a cloud link queued dozens of stale sizes and stalled input behind them. The pump now keeps at most one resize in flight and remembers only the newest pending sample, clocked by the relay's existing per-resize `{"diag":{"resize":…}}` stderr line (2 s liveness timeout). stderr moved from a blocking drain to a streaming line reader; exit classification is unchanged. Local daemons (sub-ms acks) keep send-every-sample behavior; slow links converge on the final size after one round trip.

**Geometry authority follows user input.** Session restore recreates every workspace that ever viewed a terminal; every restored pane's relay claims geometry authority at attach, last claim wins, so a hidden duplicate frozen at a mid-layout grid could own the PTY size while the visible pane rendered its real grid — and the visible pane's resizes were recorded but never applied. The relay gains a third stdin verb, `{"claim":{"geometry":true}}` (older relays ignore unknown keys), and the pump sends it ahead of user input, throttled to once per 5 s per relay. The pane the user types in owns the PTY size. Verified live: a second 60x20 attachment shrank the PTY, one typed command in the pane restored 147x65. E2E `pipe_io_claim_line_reclaims_geometry_authority` covers steal → reclaim → post-reclaim resize.

Known trade-offs: input typed during an in-flight resize can wait up to one link round trip (same relay stdin); a pane resized but never typed in does not steal authority back until its first keystroke.

## Summary by CodeRabbit

* **New Features**
  * Added optional Cloud Terminal Manual IO for improved interactive rendering.
  * Added automatic relay reconnection, resize synchronization, and connection-status overlays.
  * Added terminal-specific `--pipe-io` attachment support.
* **Settings**
  * Added a beta-feature toggle and searchable settings entry, enabled by default.
* **Bug Fixes**
  * Cloud terminals now fall back to standard attachment when Manual IO is unavailable.
* **Documentation**
  * Documented the new command-line options and behavior.
* **Localization**
  * Added English and Japanese translations for the new settings and terminal status messages.


lawrencecchen:
Closing as superseded by merged PR https://github.com/manaflow-ai/cmux/pull/11523. That PR replaces nested cmux-tui manual-IO attach with native Ghostty manual I/O. This branch also retains unresolved relay lifecycle findings, so it should not be revived.

Issue #11040, opened 2026-08-28
cmux-tui: codex 0.150 interactive mode does not execute hooks.json commands, so codex agents never reach the agents view


lawrencecchen:
The cmux-tui agent pipeline (hooks -> journal -> roster reducer, https://github.com/manaflow-ai/cmux/pull/11002) works end to end with claude. Codex 0.150.1 parses `~/.codex/hooks.json` (it warns about clamping the SessionEnd hook timeout), the entries are enabled with refreshed `trusted_hash` values in `~/.codex/config.toml`, but no hook command executes in interactive mode: neither the installed `${CMUX_TUI_HOOK}` entries nor a temporary env-free probe entry appended to UserPromptSubmit ever ran (verified with file-logging probes; YOLO mode, `hooks = true`). Until codex actually executes them, codex sessions never appear in agents views.

Follow-ups to investigate: how codex 0.150 gates hook execution (per-hash trust confirmation flow? feature flag? event subset in interactive mode), whether hook subprocesses inherit the parent session environment (the installed commands resolve `${CMUX_TUI_HOOK:-:}` from env and silently no-op without it), and whether `agent hook install` rewriting hooks.json invalidates codex trust state. The installer may need a codex-specific activation step.

lawrencecchen:
Correction from the screen-detection work (https://github.com/manaflow-ai/cmux/pull/11068): in a clean environment during that build, codex 0.150.1 DID execute hooks.json commands interactively. It also loads project-level `.codex/hooks.json`, which the original probe did not account for. The earlier negative result is therefore suspect (likely an environment or config-scope problem, not a codex limitation). The fold arbitration in PR 11068 (hook > screen > socket, 30s staleness) makes the question non-blocking: codex gets state from screen detection when hooks are absent, and hooks take over when they fire. Keeping this open until someone reproduces the original failure or we close it as environment-specific.

PR #11068, opened 2026-08-28
cmux-tui: screen-detection agent source (journal events from parsed terminal state)

lawrencecchen:
TUI-parsed agent lifecycle states become cmux-tui journal events, exactly like hook events, so the journal-derived agent roster covers herdr-supported agents that expose no hooks (codex today: https://github.com/manaflow-ai/cmux/issues/11040). Builds on the journal reducer framework from https://github.com/manaflow-ai/cmux/pull/11002; until that PR merges, this diff shows its reducer commits too.

## How it works

- `crates/cmux-tui-core/src/screen_detect/manifest.rs` ports herdr's manifest engine (TOML rules: `id`, `state`, `priority`, `region` incl. `osc_title`/`bottom_non_empty_lines(N)`/`last_non_empty_above_prompt_box`, `regex`/`line_regex`/`contains`, `any`/`not` gate trees, `skip_state_update`, `aliases`, `min_engine_version`) from herdrdev/herdr@7b675f42 (Apache-2.0, attribution header kept). The 21 manifests vendored under `cmux-tui/vendor/herdr-manifests/` embed at compile time; no network, herdr's update endpoint is never called.
- A session-owned daemon thread (`screen_detect/scanner.rs`) samples every PTY's coalesced output revision every 100ms. Identity is resolved every tick from the foreground process-group leader's executable name matched against manifest id/aliases: an identity edge (spawn/swap/exit) evaluates the screen immediately, so a launched `codex` appears within one tick; output quiescence (300ms debounce) paces subsequent rescans. Transitions (never per-scan states) append `cmux_agent` journal events: `format cmux.agent-hook.v1`, `adapter {"id":"<manifest id>","version":1}`, `native_event "ScreenDetect"`, `normalized {"state":...}`.
- The roster fold reads the explicit state like the socket echo, with source `detected` and the adapter id as the agent. Arbitration in the fold: hook > screen > socket. A screen event may not overwrite a hook entry fresher than `STALE_HOOK_MS` (30s); a hook event always wins; socket loses to both (the direct socket-report commit re-asserts detected projections too). Exit: foreground process gone AND the entry is screen-established -> `done` event, roster removal. Reducer version bumps to 3 (snapshots re-fold from the journal head).
- Ordering (separate commit): agents views sort by (attention desc, updated_at desc), attention blocked=3 > working=2 > idle=1; latest transition first within a bucket. The dogfood branch additionally paints a status dot per row (blocked = error color, working = info color, idle = none).

## Trade-offs (all of them)

1. **First emission before stable state**: on spawn the row appears immediately with that instant's screen state (idle when no rule matches, including viewer/unknown screens). A mislabeled presence for under a second was chosen over a missing row for ten.
2. **No OSC progress capture**: the daemon does not track OSC 9;4, so the 4 vendored `osc_progress` rules (cursor, grok, qwen) never match; those agents rely on their screen/title rules.
3. **Shallow process identification**: the foreground leader's executable basename is matched against manifest id/aliases only. herdr's deep job scan (node/bun/python/shell wrapper argv digging, `opencode2`, versioned `muse-bin-*`, symlink canonicalization, full process-tree sweep) is not ported. Native agent binaries identify fine; an agent hidden behind a generic runtime leader name is not detected.
4. **Windows inert**: no foreground-process-name resolution on Windows, so screen detection emits nothing there.
5. **Screen exit removes only screen-established entries**: a stale hook entry is never removed by a screen `done`; it can only be overwritten by non-done screen states after 30s of hook silence. Socket entries are also not removed by screen exits.
6. **Idle seen-bit deferred**: herdr ranks idle-unseen above working; that needs frontend focus history, so ordering here is blocked > working > idle with recency. Noted as follow-up.
7. **Dots, not group headers**: attention buckets render as a per-row dot in the disclosure slot instead of section headers (no new localized strings, no burned rows in a narrow rail). Headers as a config option are a follow-up.
8. **Polling scanner**: one thread, 100ms cadence (an atomic read plus one process-group lookup per terminal per tick); viewport text is extracted only on quiescence or an identity edge. Chosen over per-terminal condvar waiter threads.
9. **STALE_HOOK_MS = 30s** is the chosen hook-staleness rule (journal-timestamp arithmetic, deterministic on replay).
10. **Viewer/unknown screens keep the prior state** (`skip_state_update` and unknown-state rules assert nothing) except for the initial presence emission (see 1).
11. **Reducer version bump** discards persisted roster snapshots once on upgrade; entries older than the retained journal rebuild lazily from the next events (fine for a live roster).
12. **Journal kind** for screen events is `agent.state.changed` / `agent.session.ended`; the fold dispatches on `native_event == "ScreenDetect"`, not on kind.
13. **CLI ordering unchanged**: `agent list` stays terminal-id sorted; the attention ordering applies to agents views.
14. **1Hz forced evaluation of never-quiet screens**: live dogfood found codex's working spinner redraws every ~100ms, so a working codex never passes the 300ms quiescence gate and detection froze at the spawn-time state. `observe_revision` now also evaluates when 1s has passed since the last evaluation, so busy screens sample at 1Hz. The cost is a small chance of matching a mid-redraw frame; herdr samples on a bare timer with no quiescence gate at all, so this is still stricter than upstream.
15. **Exit removals commit their done projection with source `hook`** (`apply_roster_delta` hardcodes it for removals, pre-existing shape); the live roster and views are correct, but `agent list` shows the dead row's source as hook even when screen detection ended it. Cosmetic, follow-up.
16. **codex 0.150.1 executed hooks in this environment**, contrary to https://github.com/manaflow-ai/cmux/issues/11040 as filed: a codex launched with the standard `~/.codex/hooks.json` emitted SessionStart/UserPromptSubmit/Stop, and the fold's arbitration kept the entry hook-owned with zero flapping while screen detection watched the same terminal. The live no-hooks demo therefore ran codex with a clean `CODEX_HOME` (no hooks.json), which produced a pure `source=detected` roster row.

## Verification

- Unit: all 21 vendored manifests parse; identify normalization (paths/aliases/case); rule priority + `any`/`not` gates + `line_regex`; region scoping incl. `osc_title`; real codex manifest classifying working/blocked/idle/viewer screens; malformed-manifest rejection; tracker debounce, edge-triggering, spawn/exit identity edges, idle presence; fold arbitration matrix (hook/screen/socket × fresh/stale) and exit removal (red-first commit); end-to-end scan -> journal -> roster -> `list_agents` plus journal replay determinism; hooks-win end to end; attention ordering.
- Suites on the Blacksmith testbox: `cargo test -p cmux-tui-core --lib` (1165 passed), `cargo test -p cmux-tui --bin cmux-tui -- agent sidebar screen` (132 passed; 144 on the dogfood branch).
- Hosted gate (`--filter screen_detect`, green on the exact head): https://github.com/manaflow-ai/cmux/actions/runs/33152512377 (this branch) and https://github.com/manaflow-ai/cmux/actions/runs/33151747668 (the sbsplit dogfood branch carrying the same commits plus ordering dots).
- Live tmux: real `codex` (no hooks fire) appearing via screen detection, claude still flowing via hooks, no hook/screen flapping.

## Summary by CodeRabbit

* **New Features**
  * Added automatic detection of agent activity states, including working, blocked, idle, and completed states.
  * Added support for identifying foreground agent processes and recognizing many agent tools.
  * Agent details and detected states are now preserved across sessions and replay.
  * Improved the Agents view by prioritizing attention-worthy and recently updated agents.
  * Agent adapter information is now included in session and event details.

* **Documentation**
  * Added licensing and maintenance documentation for bundled detection manifests.


lawrencecchen:
Closing as superseded by active userland replacement https://github.com/manaflow-ai/cmux/pull/11454. The replacement moves screen detection into a replaceable plugin boundary. This core branch is stale and has failed CLA/conflicting history.

PR #11091, opened 2026-08-28
iOS: reconnecting never exits the user's workspace or invalidates their current task

azooz2003-bit:
On iPhone (iroh, latest INTERNAL/NIGHTLY), the moment a reconnect began the mounted workspace detail was exited. Reproduced at the store level by the new regression tests (commit 1, red): recovery's redial teardown dropped every secondary Mac's `workspacesByMac` entry, which flips multi-Mac row-id scoping, changes every derived row id's raw value, remaps the selection, and re-keys the pushed navigation route (destroying the mounted detail); a failed first dial then re-entered the teardown with `foregroundMacDeviceID` already nil, so the retention filter keyed on the anonymous sentinel and wiped the whole list.

The fix is the general invariant the symptom kept violating: **connection state changes never terminate or invalidate the user's current context.**

- **Store:** `clearRemoteConnectionContext` no longer removes workspace rows. Teardown is a transport event; rows are removed only by authoritative events (healthy list reconcile, unpair/hide in the secondary reconcile pass, sign-out, team change), all of which already do their own explicit removal. All retained entries downgrade to `.unavailable` so the Computers dots and list chrome stay truthful, the teardown is idempotent (a second clear is a status no-op), and the per-Mac entry count stays stable so row-id scoping and route identity survive reconnects. The post-reconnect secondary refresh re-syncs the retained rows.
- **Navigation policy (defense-in-depth):** `WorkspaceShellCompactNavigationPolicy` takes `listIsAuthoritative`. While the list is reconnecting/disconnected, a cleared selection or a list hole keeps the mounted detail rendering its last-known snapshot (rename dialogs and sheets survive because the detail view identity survives); deletions confirmed by a healthy list still pop normally.

Commit 1 adds the failing store tests only (CI red); commit 2 adds the fix plus policy tests (CI green). Full `CmuxMobileWorkspace` suite passes (45/45); full `CmuxMobileShell` suite run included in the verification pass.

HIG: [Feedback](https://developer.apple.com/design/human-interface-guidelines/feedback) — status feedback should reach people "without having to take action or leave their current context"; only alerts may deliberately disrupt context. Transient connectivity is status, so it is shown in place (existing title spinner from #10821) instead of ejecting the user.

Follow-up (out of scope): genuine Mac removal while N→1 still flips row-id scoping and re-keys a mounted route; making scoping sticky would remove that last identity change, but it changes persisted-selection semantics and deserves its own pass.

---
## Summary by cubic
Fixes iOS reconnects so reconnecting never exits the user's workspace or invalidates their current task. Connection teardown no longer removes workspace rows or retargets the selection, and compact navigation keeps the mounted detail while the list is reconnecting or disconnected.

**Store**

- `clearRemoteConnectionContext` keeps all `workspacesByMac` rows, downgrading their status to `.unavailable` so the Computers dots and list chrome stay truthful.
- Rows are removed only by authoritative events: a healthy list reconcile, unpair/hide, sign-out, or team change.
- Retaining secondary entries means multi-Mac row-id scoping never flips mid-recovery, so derived row ids and the pushed navigation route keep their identity; a second teardown is now a status no-op.
- A genuine Mac removal while N→1 still flips row-id scoping and re-keys a mounted route; making scoping sticky is a follow-up.

**Navigation**

- `WorkspaceShellCompactNavigationPolicy` takes `listIsAuthoritative`; while the list is reconnecting or disconnected, a cleared selection or list hole keeps the mounted detail on its last-known snapshot.
- Deletions confirmed by a healthy list still pop normally.

<sup>Written for commit a0c781bdfca8fc3fadb4f7541cdd525d5a0cc967. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11091?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Workspace rows now remain visible during connection teardown and recovery.
  - Workspace selections and mounted detail views are preserved while reconnecting or disconnected.
  - Workspace statuses are updated to unavailable without removing retained rows.
  - Navigation now avoids popping detail routes until the workspace list is confirmed current.
- **Tests**
  - Added coverage for workspace retention, selection preservation, repeated teardown, and reconnect navigation behavior.

### DIFF
--- a/Packages/iOS/CmuxMobileShell/Sources/CmuxMobileShell/MobileShellComposite+WorkspaceListRecovery.swift
+++ b/Packages/iOS/CmuxMobileShell/Sources/CmuxMobileShell/MobileShellComposite+WorkspaceListRecovery.swift
@@ -123,10 +123,10 @@ extension MobileShellComposite {
             return
         }
         // Failed dials run their own cleanup with the default non-preserving
-        // teardown, dropping the secondary subscriptions preserved above (the
-        // foreground id is already nil, so that filter keeps only the
-        // anonymous key). Rebuild them so a failed foreground redial cannot
-        // strand healthy secondary Macs.
+        // teardown, cancelling the secondary subscriptions preserved above
+        // (their last-known rows stay, downgraded to unavailable). Rebuild the
+        // subscriptions so a failed foreground redial cannot strand healthy
+        // secondary Macs on stale rows.
         await refreshSecondaryMacWorkspaces()
     }
 
--- a/Packages/iOS/CmuxMobileShell/Sources/CmuxMobileShell/MobileShellComposite.swift
+++ b/Packages/iOS/CmuxMobileShell/Sources/CmuxMobileShell/MobileShellComposite.swift
@@ -10234,8 +10234,8 @@ public final class MobileShellComposite: MobileTerminalOutputSinking {
         connectionAttemptGeneration = UUID()
         // Capture the tagged foreground key BEFORE the identity clears below:
         // `foregroundMacKey` derives from `activeMacInstanceTag`, which
-        // `clearActiveConnectionContext()` nils, and the offline retention
-        // filter must keep the exact tagged entry.
+        // `clearActiveConnectionContext()` nils, and the offline status
+        // downgrade must hit the exact tagged entry.
         let offlineForegroundKey = foregroundMacKey
         focusedHandoffPreparedGenerations.removeAll()
         cancelRemoteOperationTasks()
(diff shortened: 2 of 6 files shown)


azooz2003-bit:
Superseded by https://github.com/manaflow-ai/cmux/pull/11873: same two commits rebuilt onto current main (this branch fell 622 commits behind and conflicts with the demonstration-mode retention added in #11289).

PR #11096, opened 2026-08-28
cmux-tui: configurable agents view sort and filters

lawrencecchen:
Stacked on https://github.com/manaflow-ai/cmux/pull/10966 (base branch of this PR) and merges https://github.com/manaflow-ai/cmux/pull/11068, whose commits (and PR 11002's) show in this diff until they merge. Matches the herdr agents-panel design: the priority order, seen semantics, and two-line layout are derived from herdr (https://github.com/herdrdev/herdr), Apache-2.0, commit 7b675f42af35, modified; see cmux-tui/ATTRIBUTIONS.md.

## What changes

- **Two-line agent rows by default**: line 1 is the state glyph plus the context (tab title, else the session/workspace subtitle), line 2 is the agent type in dim text. A view entry sets `"row_lines": 1` for the old compact single-line rows. Row spans, scrolling, selection, and mouse hits derive from one height table.
- **herdr priority order, exactly**: blocked(4) > idle-unseen(3) > working(2) > idle-seen(1) > unknown(0), most recent state change first inside a bucket. An idle agent counts as seen once its surface was this client's focused tab at or after the idle transition, tracked in a client-local focus-stamp map.
- **View header**: agents views draw the view title left and the sort mode ("priority", localized) right, like herdr.
- **State glyphs**: blocked = bold error-color dot, working = bold info-color dot, idle = dim open circle, matching the sort buckets.

## Trade-offs (all of them)

1. **Two-line rows halve the agents visible per screen**; that is why `row_lines: 1` exists.
2. **Seen is client-local presentation state**: two attached clients can order the same idle agent differently. That is correct for a per-viewer property and is deliberate; seen never touches the journal.
3. **Clock domains**: seen compares a client `SystemTime` stamp to the server's `updated_at_ms`. On remote sessions with skewed clocks an idle agent can be mis-bucketed until the next focus. Local sessions share one clock.
4. **Seen updates only while the sidebar renders** (stamps are taken in the projection path); a client that never draws the agents view marks nothing seen, which only affects its own ordering.
5. **The header consumes one rail line** in agents views and has no off switch (herdr parity; the row is also the sort-mode disclosure).
6. **Priority ordering now applies to workspace-scoped agents views too** (they previously kept tree order; all-scope kept pure recency). One ordering everywhere beats per-scope rules.
7. **Untitled agent tabs** show their session/workspace subtitle as the context line in two-line mode (the type is already line 2); single-line mode keeps the adapter-id name fallback.
8. **A roster row without an adapter id** (bare socket reports) falls back to its localized state word on line 2.

## Verification

- Testbox: `cargo test -p cmux-tui --bin cmux-tui -- agents_view_priority sidebar agent` (137 passed).
- Hosted gate (filter `agents_view_priority`) linked in comments when green; the merged screen-detection branch's own gate: https://github.com/manaflow-ai/cmux/actions/runs/33152512377.
- Live tmux dogfood on the refreshed `feat-tui-sbsplit-dogfood` (same head): top/bottom split config, two-line rows, dots, blocked-on-top ordering, unseen-idle above working dropping below after focusing the pane, exit removal.

---
## Summary by cubic
Reworks cmux-tui's agents view to match herdr and derives the live roster from the session journal instead of direct agent reports, so ended sessions disappear instead of remaining as done rows. The diff also carries the merged Computer Use, iOS discovery, relay rate-limit, and terminal/socket branches from the stacked base.

**Agents view**
- Default rows show context first and the adapter or state label below; `row_lines: 1` keeps compact rows.
- Priority is blocked > idle-unseen > working > idle-seen > unknown, with recent changes first in each group.
- Idle agents become seen when this client focuses their tab; clock skew can affect remote ordering until the next focus.
- Entries accept `sort` (`priority`, `recency`, `name`, `agent`, or `state`) and a validated `filter` object; pressing `s` cycles the client-local presentation sort.
- The fixed localized header shows the title and active sort mode, always showing `filtered` when a filter is active; invalid values warn and fall back without dropping the view.

**Roster and supporting changes**
- The roster is a journal reducer: hook, screen-detection, and direct-report events update terminal records, replayed events do not reapply projections, and ended sessions remove the row.
- Arbitration is hook > screen > socket, with durable watermarks, tombstones, and cursors fencing stale reports across restarts.
- Screen detection ports herdr's manifest engine (21 embedded Apache-2.0 manifests) and resolves foreground process identity every 100 ms so new agents appear without waiting for output quiescence.
- Relay 429 responses and journal lines now name the enforcement source; local backpressure failures use a `cooldown:` prefix.

<sup>Written for commit 4641817dd696779e319e5dbdf586cc401b200f97. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Sort and filter

Agents view entries now accept `sort` (`priority`, `recency`, `name`, `agent`, or `state`) and a validated `filter` object (`agent`, `state`, and `seen` criteria combined with AND). `priority` remains the default. Press `s` in a focused agents view to cycle the client-local presentation sort; the configured value remains the shared starting mode. The header shows the active localized mode and always shows `filtered` when a filter is active, including at narrow rail widths. Invalid values warn and fall back without dropping the view.


lawrencecchen:
Hosted gate green on the exact head (807efe691cb, filter agents_view_priority): https://github.com/manaflow-ai/cmux/actions/runs/33158421324. The refreshed feat-tui-sbsplit-dogfood points at the same commit and is the demo binary.

PR #11127, opened 2026-08-28
security(tui): fence and bound screen detection state

lawrencecchen:
Security follow-up for https://github.com/manaflow-ai/cmux/pull/11068.

- Fence detector work to the current surface lifecycle and identity.
- Bound journal prefixes and evaluation cadence.
- Use generic diagnostics, verify process identity, and keep projection indexes stable.

Trade-off: detector false negatives are accepted because screen detection is advisory, not an authorization decision. A stale or malformed source is ignored rather than allowed to drive an action.

---
## Summary by cubic
Hardens screen detection and agent roster folding so a process cannot spoof its identity and stale or out-of-order detection state cannot update projections. Process identity now comes from the kernel instead of process-controlled labels, and live agent roster folds are fenced, sequenced, and retried on snapshot failure. Detection stays advisory, so a missing or unverifiable source is ignored rather than trusted.

**Trust boundary**
- Linux process identity reads the kernel `/proc/<pid>/exe` link; interpreter-launched scripts recover their script path from the kernel's command-line record only for known interpreters, with the script's shebang verified to match, within a byte cap and regular-file check, and ambiguous option-led invocations rejected rather than guessed.
- A foreground identity change forces an immediate screen read; a failed read retries with bounded exponential backoff (100 ms up to a 2 s cap) instead of retrying every scan tick.
- Foreground identity lookups are paced per terminal and capped at 64 per scan with a rotating start index; a successful lookup resets the backoff window, and an unavailable lookup clears the cached identity so stale output cannot attach to a PID that exited or was reused.
- Screen-detected exits, including the terminal Done edge, journal under the state-change event kind with the normalized state instead of a session-ended kind, so the validator never infers a full hook session from a process sample.
- Background errors now report stable diagnostic codes instead of messages that could leak resource IDs, paths, or source errors.
- `tokio` is now a cross-platform dependency so signal handling lets the scanner halt cleanly on Windows.

**Fold consistency**
- Live agent roster folds and terminal retirement share a lifecycle fence so they cannot interleave.
- Folds replay the durable journal prefix in sequence order in bounded pages, matching restart replay when commits arrive out of order.
- A failed reducer snapshot write marks the cursor dirty and is retried on the next fold instead of silently advancing past the unsaved snapshot.
- Agent projections are read by primary key instead of scanning and sorting the whole table.

<sup>Written for commit 755221970ebf238e05d192d4418b73cb67d28d49. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11127?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lawrencecchen:
Fleet instruction update for head `755221970ebf238e05d192d4418b73cb67d28d49`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #11179, opened 2026-08-29
Require ownership evidence before attributing a same-TTY process

### LINKED ISSUE #11004: Memory attribution includes detached PPID-1 processes that merely share a cmux TTY
## Summary

A process that is no longer a descendant of cmux (reparented to PID 1) can retain a cmux terminal's controlling TTY. macOS's Force Quit / process-session accounting then presents that process's memory as cmux memory. The same ownership ambiguity is also present in cmux's own recursive memory diagnostics: `v2AnnotateTopSurface` currently promotes every PID sharing a surface TTY to a process-tree root, without requiring descendant, process-group, or explicit cmux-scope evidence.

This makes a detached REPL or daemon look like cmux is leaking/using the memory, and makes per-surface/workspace memory numbers unreliable.

## Reproduction

Observed on macOS with cmux 0.64.22:

1. Start cmux (the clean repro used PID `94862`).
2. From a cmux terminal, start a long-lived Python REPL/kernel and allocate memory:

   ```python
   big = list(range(50_000_000))
   len(big)
   ```

3. Let the kernel detach/reparent (the repro kernel was PID `68314`, PPID `1`) while it keeps the terminal open.
4. Compare the process relationships:

   ```sh
   ps -o pid,ppid,tty,rss,comm -p 68314
   ps -o tty= -p 68314
   ps -o tty= -p 67498   # cmux terminal's login process
   ```

   The kernel is a `python3` process with PPID 1, but has the same TTY as the cmux login process.
5. The repro's measurements were:

   ```text
   cmux phys_footprint: 1075 MB
   detached REPL RSS:   1939 MB (small repro)
   displayed/attributed: ~3014 MB

   Reported user case: 1075 MB cmux + ~13 GB REPL = ~14 GB
   ```

6. Open Force Quit Applications (or any session/TTY-based memory view). The detached kernel's memory is shown as part of cmux even though it is not in cmux's process tree.

## Code signal

On current main, `Sources/TerminalControllerTopSupport.swift` does this for each terminal surface:

```swift
let ttyPIDs = processSnapshot.pids(forTTYName: ttyName)
rootPIDs.formUnion(ttyPIDs)
surfacePIDs.formUnion(processSnapshot.expandedPIDs(rootPIDs: ttyPIDs))
```

TTY matching is useful for launchd/reparented cmux helpers, but treating the entire TTY population as roots also admits unrelated detached processes. The snapshot already records PPID, PGID/TPGID, and explicit `CMUX_*` scope, which should be used to distinguish proven ownership from a mere TTY collision.

## Expected behavior

- A same-TTY process with PPID 1 and no explicit cmux scope should not be silently attributed to a cmux surface/workspace.
- Recursive memory totals should include descendants and positively identified cmux-owned/reparented helpers, while putting ambiguous same-TTY processes in an explicit `unattributed`/warning bucket.
- Preserve valid attribution for cmux hook monitors and browser/WebKit roots that are intentionally reparented.
- If macOS Force Quit's own session aggregation cannot be controlled by cmux, document that limitation and make cmux's diagnostics clearly separate app footprint, proven child RSS, and same-TTY-but-unproven RSS.

## Acceptance criteria

- Add a regression fixture for an unrelated process with PPID 1 sharing a terminal TTY; it is excluded from the surface/workspace owner (or explicitly marked ambiguous), not counted as a proven child.
- Add/retain coverage for a positively scoped reparented cmux helper and a launchd-parented WebKit process.
- `cmux memory` and Task Manager report the ownership reason and do not inflate a workspace from a TTY collision.
- Validate the behavior on macOS 15/26 where feasible, and do not regress process teardown/agent hibernation, which also uses TTY and process-group evidence.

Related: #4364 (recursive memory diagnostics), #9069 (aggregate attribution), #6313 (runaway child memory), #10926 (orphaned TTY watchers), #7596 (memory audit).

jeonghi:
Fixes #11004.

## Problem

`v2AnnotateTopSurface` promotes **every** PID sharing a surface's TTY to a process-tree root:

```swift
let ttyPIDs = processSnapshot.pids(forTTYName: ttyName)
rootPIDs.formUnion(ttyPIDs)
surfacePIDs.formUnion(processSnapshot.expandedPIDs(rootPIDs: ttyPIDs))
```

A TTY is not proof of ownership. A process reparented to launchd keeps the controlling TTY of the terminal it was launched from, so a detached REPL or dev server gets summed into the surface's resources and presented as cmux memory — in `cmux memory`, the Task Manager, and every pane/workspace/window rollup above it.

I hit this from the reporting side: a `next dev` server orphaned to PPID 1 three days earlier was still holding a cmux TTY, and cmux's own memory number carried its 6.3 GB.

## Fix

`CmuxTopTTYOwnershipResolver` splits the TTY population using evidence the snapshot already records (`parentPID`, `processGroupID`, `cmuxSurfaceID`) but did not consult:

| reason | evidence |
|---|---|
| `cmux-process-scope` | explicit `CMUX_*` scope naming this surface |
| `surface-process-tree` | already proven via the cmux-scoped process tree |
| `tty-session-root` | parent is alive and off the TTY — the shell cmux forked |
| `tty-descendant` | descends from a proven owner |
| `tty-process-group` | process group led by a proven owner |
| `tty-collision` | shares only the TTY device |

Descendant and process-group evidence propagate to a fixpoint, so a chain is proven regardless of iteration order and a parent cycle still terminates.

Only proven PIDs become roots. Collisions land in `unattributed_tty_process_pids` with their own `unattributed_resources` summary, so the memory stays visible without being charged to the surface, and `tty_ownership_reasons` reports why each PID went where it did.

Two ordering details carry the "don't regress what already works" half of the issue:

- Explicit cmux scope is checked **before** the PPID-1 exclusion, which is what keeps intentionally reparented hook monitors attributed.
- The unattributed set is resolved **after** the webview roots are folded in, so a launchd-parented WebKit process is never mistaken for a collision.

Attribution only. `PaneMemoryGuardrail` and the agent-hibernation scope still use their existing TTY and process-group evidence, and teardown behavior is unchanged. The guardrail shares the same TTY-as-root shape and is worth a follow-up, but changing it changes runtime termination, which is outside this issue.

## Commits

Per `skills/cmux-testing`, the regression test ships first so CI shows it genuinely fails without the fix:

1. `test:` — the failing test only (CI red)
2. `fix:` — the fix (CI green)

## Tests

`cmuxTests/CmuxTopTTYCollisionAttributionTests.swift` drives the real `v2AnnotateTopWindows` path with five processes on one surface TTY:

| pid | ppid | | expected |
|---|---|---|---|
| `zsh` | app | | proven — entered the TTY from outside |
| `node` | zsh | | proven descendant |
| `python3` | 1 | detached, no scope, own pgroup | **collision** |
| `cmux` hook | 1 | scoped to the surface | proven |
| `rg` | 1 | pgroup led by the proven shell | proven |

It asserts the 1.9 GB `python3` is absent from `resources`, present in `unattributed_resources`, and that every PID carries the right ownership reason. Wired into `project.pbxproj`; `./scripts/lint-pbxproj-test-wiring.sh` and `./scripts/check-pbxproj.sh` pass.

## Verification I could and could not do

I could not build the app locally — no Zig and no GhosttyKit.xcframework on this machine, and building it was not a reasonable trade while diagnosing a memory-exhaustion problem. So I deliberately put the decision logic in a file that depends on Foundation alone and nothing else in the module, and compiled and ran it standalone against the same fixture plus edge cases: helper scoped to a *different* surface, children of an orphan (propagation must not escape through the orphan), `surfaceID == nil` snapshots, a pre-proven seed, a self-leading orphan, empty input, PIDs absent from the snapshot, and a parent cycle. 17/17 pass.

The three edited files pass `swiftc -parse`. **The Xcode target build and `cmuxTests` run still need CI** — I have verified the algorithm, not the integration compile.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11179"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790584917&installation_model_id=21920&pr_number=11179&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11179&signature=6a4fc1fb4b4db2e810ba5e5fd8e94c2cde030369b384f74f11c9027d24568ebd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes attribution so processes that only share a surface's TTY device are no longer counted as cmux memory. Previously every PID on a surface's TTY became a process-tree root, so a detached REPL or dev server reparented to launchd inflated that surface's memory; now only processes with ownership evidence are attributed, and the rest land in a separate unattributed bucket that stays visible but uncharged.

**Behavior details**
- TTY ownership is proven from `CMUX_*` scope, an already-proven tree, a live off-TTY parent that is itself cmux-owned, descent from a proven owner, or membership in a process group led by a proven owner.
- Proof resolves in linear time: parent and process-group indexes are built once, and each newly proven PID is expanded exactly once from a work queue.
- Collisions are reported via `unattributed_tty_process_pids` with their own `unattributed_resources` summary, and `tty_ownership_reasons` records why each PID was classified.
- Explicit scope outranks the PPID-1 exclusion, keeping intentionally reparented hook monitors attributed; unattributed resolution runs after webview roots so launchd-parented WebKit processes are not mistaken for collisions.
- Attribution only: the pane memory guardrail and agent-hibernation scope still use their existing TTY and process-group evidence.
- The regression test drives the real annotation path with six processes on one TTY, including a 1.9 GB detached process, a reparented scoped helper, a backgrounded job, and a process parented by a foreign live process that must land unattributed.

<sup>Written for commit 877f4854fc90f2e81950d2812efb1e4f821fef5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11179?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved TTY process attribution for surface monitoring.
  * Distinguishes surface-owned processes from unrelated processes sharing the same TTY.
  * Reports unattributed TTY processes and ownership reasons.
  * Provides more accurate resource summaries by excluding unattributed processes.

* **Bug Fixes**
  * Prevents detached or reparented processes from being incorrectly counted as surface-owned resources.

* **Tests**
  * Added coverage for TTY collision attribution, ownership reasons, and resource totals.

### DIFF
--- a/Sources/CmuxTopSnapshot.swift
+++ b/Sources/CmuxTopSnapshot.swift
@@ -264,9 +264,18 @@ final class CmuxTopProcessSnapshot: @unchecked Sendable {
     /// controlling TTY of the terminal it was launched from, so a detached REPL or dev
     /// server would otherwise be summed into the surface's memory.
     /// See https://github.com/manaflow-ai/cmux/issues/11004.
+    ///
+    /// - Parameters:
+    ///   - ttyPIDs: every PID sharing the surface's TTY device.
+    ///   - surfaceID: the surface being annotated, when it has an identifier.
+    ///   - cmuxOwnedPIDs: the window's app processes and their descendants, used as launch
+    ///     evidence for a process whose parent sits off the TTY.
+    ///   - provenPIDs: PIDs already proven through the cmux-scoped process tree.
+    /// - Returns: the proven set, the unattributed set, and the reason for each PID.
     func ttyOwnership(
         ttyPIDs: Set<Int>,
         surfaceID: UUID?,
+        cmuxOwnedPIDs: Set<Int>,
         provenPIDs: Set<Int> = []
     ) -> CmuxTopTTYOwnership {
         var ownershipProcesses: [Int: CmuxTopTTYOwnershipProcess] = [:]
@@ -280,7 +289,7 @@ final class CmuxTopProcessSnapshot: @unchecked Sendable {
                 cmuxSurfaceID: process.cmuxSurfaceID
             )
         }
-        return CmuxTopTTYOwnershipResolver.resolve(
+        return CmuxTopTTYOwnershipResolver(cmuxOwnedPIDs: cmuxOwnedPIDs).resolve(
             candidates: ttyPIDs,
             processes: ownershipProcesses,
             surfaceID: surfaceID,
--- a/Sources/CmuxTopSnapshot.swift
+++ b/Sources/CmuxTopSnapshot.swift
@@ -257,6 +257,37 @@ final class CmuxTopProcessSnapshot: @unchecked Sendable {
         Set(pidsByCMUXSurfaceID[surfaceID] ?? [])
     }
 
+    /// Splits a surface's TTY population into processes cmux can prove it owns and
+    /// processes that merely share the TTY device.
+    ///
+    /// Sharing a TTY is not ownership: a process reparented to launchd keeps the
+    /// controlling TTY of the terminal it was launched from, so a detached REPL or dev
+    /// server would otherwise be summed into the surface's memory.
+    /// See https://github.com/manaflow-ai/cmux/issues/11004.
+    func ttyOwnership(
+        ttyPIDs: Set<Int>,
+        surfaceID: UUID?,
+        provenPIDs: Set<Int> = []
+    ) -> CmuxTopTTYOwnership {
+        var ownershipProcesses: [Int: CmuxTopTTYOwnershipProcess] = [:]
+        ownershipProcesses.reserveCapacity(ttyPIDs.count)
+        for pid in ttyPIDs {
+            guard let process = processesByPID[pid] else { continue }
+            ownershipProcesses[pid] = CmuxTopTTYOwnershipProcess(
+                pid: process.pid,
+                parentPID: process.parentPID,
+                processGroupID: process.processGroupID,
+                cmuxSurfaceID: process.cmuxSurfaceID
+            )
+        }
+        return CmuxTopTTYOwnershipResolver.resolve(
+            candidates: ttyPIDs,
+            processes: ownershipProcesses,
+            surfaceID: surfaceID,
+            provenPIDs: provenPIDs
+        )
+    }
+
     func pids(forProcessGroupID processGroupID: Int) -> Set<Int> {
         Set(pidsByProcessGroupID[processGroupID] ?? [])
     }
--- a/Sources/CmuxTopTTYOwnership.swift
+++ b/Sources/CmuxTopTTYOwnership.swift
@@ -15,8 +15,8 @@ enum CmuxTopTTYOwnershipReason: String, Sendable, CaseIterable {
     /// The process was already proven through the surface's cmux-scoped process tree.
     case surfaceProcessTree = "surface-process-tree"
 
-    /// The process entered this TTY from outside it — its parent is alive and off the TTY,
-    /// which is how the shell cmux forked for the surface looks.
+    /// The process was launched onto this TTY by a cmux-owned process off the TTY, which
+    /// is how the shell cmux forks for a surface looks.
     case ttySessionRoot = "tty-session-root"
 
     /// The process descends from a proven owner on this TTY.
@@ -33,11 +33,17 @@ enum CmuxTopTTYOwnershipReason: String, Sendable, CaseIterable {
 
 /// The minimum a process has to expose for TTY ownership to be decidable.
 struct CmuxTopTTYOwnershipProcess: Sendable, Equatable {
+    /// The process identifier.
     let pid: Int
+    /// The parent process identifier. `1` means launchd, i.e. the process was reparented
+    /// and its parent link no longer says who started it.
     let parentPID: Int
+    /// The process group identifier, when the snapshot could read one.
     let processGroupID: Int?
+    /// The cmux surface named by the process's `CMUX_*` environment scope, when present.
     let cmuxSurfaceID: UUID?
 
+    /// Creates the process facts used to decide TTY ownership.
     init(pid: Int, parentPID: Int, processGroupID: Int?, cmuxSurfaceID: UUID?) {
         self.pid = pid
         self.parentPID = parentPID
@@ -69,8 +75,26 @@ struct CmuxTopTTYOwnership: Sendable {
     }
 }
 
-enum CmuxTopTTYOwnershipResolver {
+/// Decides which processes on a surface's TTY the surface actually owns.
+///
+/// Attribution fails closed: a PID is charged to the surface only when the snapshot
+/// carries positive evidence for it, so an unrelated detached workload that merely
+/// inherited the TTY is reported rather than summed into cmux's memory.
+struct CmuxTopTTYOwnershipResolver {
+    /// PIDs known to belong to cmux — the window's app processes and their descendants.
+    /// A process launched onto the TTY by one of these is a surface root; without this
+    /// evidence an off-TTY parent proves nothing.
+    private let cmuxOwnedPIDs: Set<Int>
+
+    /// Creates a resolver that trusts `cmuxOwnedPIDs` as launch evidence.
+    init(cmuxOwnedPIDs: Set<Int>) {
+        self.cmuxOwnedPIDs = cmuxOwnedPIDs
+    }
+
     /// Splits `candidates` into proven owners of `surfaceID` and same-TTY collisions.
+    ///
+    /// Runs in time linear in the candidate count: parent and process-group indexes are
+    /// built once, then each newly proven PID is visited exactly once from a work queue.
     ///
     /// - Parameters:
     ///   - candidates: every PID sharing the surface's TTY device.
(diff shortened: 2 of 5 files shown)


jeonghi:
Pushed `877f485` addressing the review. Also: **`main` does not currently compile against its own pinned `vendor/bonsplit` pointer** — details at the bottom, unrelated to this PR.

## Addressed

**Fail closed on `ttySessionRoot`** (correctness, `CmuxTopTTYOwnership.swift`). The finding is right and it was the same class of bug I opened the PR about, just via a different parent link. Proving a candidate because its parent is alive and off the TTY says nothing about whether that parent is cmux's — a detached kernel or daemon whose supervisor lives off the TTY was still promoted to a surface root. The window's `app_process_pids` and their descendants now travel down `v2AnnotateTopWindows → workspace → pane → surface` as launch evidence, and a candidate without it is a collision.

**Linear resolution** (performance). Also right: the fixpoint rebuilt `livePIDs.sorted()` on every pass, so a reverse-ordered descendant chain cost a pass per PID. Parent and process-group indexes are built once, and each newly proven PID is expanded exactly once from a work queue.

**Injectable resolver** (ambient global state). `CmuxTopTTYOwnershipResolver` is now a constructable struct taking its cmux-owned PID set at `init`, not a caseless enum around a static function.

**Docstrings.** Filled in across the new API and the test helpers.

## Not addressed: extracting to a SwiftPM package

I'd like a maintainer call here rather than doing it unilaterally, because the repo's own architecture guidance points the other way for a change this size.

`skills/cmux-architecture` says a package must be **whole-domain** — "One package owns a full domain... A boundary exists because more than one consumer needs the contents, or a build/test seam must exist" — and explicitly calls out splitting a domain into `CmuxAppearanceMath`-style units as "folder structure inside `CmuxAppearance`, not module structure". A 176-line TTY-ownership resolver is a unit of that shape, not a domain. Its only consumer is `CmuxTopSnapshot`, which is itself in the app target, so the extraction would either pull the whole process-attribution domain along or leave a one-consumer micro-package behind.

The mechanical cost isn't small either: pbxproj wiring in **both** `cmux` and `cmux-unit`, a workspace group entry validated by `check-workspace-package-groups.py`, and a package-local `Package.resolved` under `check-package-resolved-policy.py` — turning a bug fix into an architecture change.

The testability concern behind the rule is already met: the file imports only `Foundation`, defines its own process facts, and takes no snapshot dependency, which is exactly what let me exercise it standalone. Happy to extract it — as `CmuxTopProcessAttribution` covering the snapshot and attribution together, rather than the resolver alone — if you'd rather have it in this PR or a follow-up.

## Verification

I could not rely on CI: `ci.yml`'s `pull_request` trigger is path-filtered to the ghostty/GhosttyKit paths, so the macOS build and `cmuxTests` do not run on a PR that only touches `Sources/` and `cmuxTests/`. So I ran the real thing locally with the prebuilt GhosttyKit and the `cmux-unit` scheme CI uses.

**Green** (`877f485`):

```
✔ detachedSameTTYProcessIsNotCountedAsSurfaceMemory() passed
✔ collisionsAreReportedAsUnattributed() passed
✔ sameTTYProcessWithForeignOffTTYParentIsACollision() passed
✔ provenTTYOwnersCarryTheirAttributionReason() passed
✔ reparentedScopedHelperStaysAttributed() passed
✔ Test run with 5 tests in 1 suite passed
** TEST SUCCEEDED **
```

**Red** — same tests, `TerminalControllerTopSupport.swift` alone reverted to `3a76082`:

```
✘ Expectation failed: (int64(resources["resident_bytes"]) → 2654994432)
                   == (expectedBytes → 125829120)
✘ Expectation failed: (attributedPIDs → [4001, 4002, 4003, 4004, 4005, 4006])
                   == ([4001, 4002, 4004, 4005])
✘ Expectation failed: (surface["unattributed_resources"] → nil) as? [String: Any]
✘ Test run with 5 tests in 1 suite failed after 0.006 seconds with 8 issues
** TEST FAILED **
```

2.65 GB attributed instead of 120 MB, with the detached and foreign-parented PIDs both charged to the surface. The one test that passes in red (`reparentedScopedHelperStaysAttributed`) is the regression guard for scoped helpers, which the old code also attributed.

## Separate: `main` does not build against its pinned `vendor/bonsplit`

Hit while setting the build up, on `3a76082`, unrelated to this PR:

```
Sources/SessionDragSessionSource.swift:53:30: error: value of type
'TabDragTransferRegistration' has no member 'clearResidualCapability'
```

`vendor/bonsplit` is pinned at `18f3f8bc`, whose `TabDragTransferRegistration` exposes only `register(with:)` and `write(to:)`. `public func clearResidualCapability(from:)` exists on bonsplit `origin/main` (`c545256`), so the call site merged without the submodule pointer being bumped — the case `AGENTS.md` flags under **Submodule safety**. The path-filtered `ci.yml` above is presumably why no PR caught it.

I verified my change against a local bonsplit bump and **did not commit the pointer** — that's a maintainer decision and out of scope here. Happy to open a separate issue if it isn't already known.

PR #11190, opened 2026-08-30
vendor/bonsplit: re-advance pointer to origin/main (regressed in #11163)

austinywang:
## Summary
- #11163 moved the `vendor/bonsplit` submodule pointer back to `18f3f8b`, which predates bonsplit PR #223.
- #10804 (`Sources/SessionDragSessionSource.swift:53`) calls `TabDragTransferRegistration.clearResidualCapability(from:)` from that PR, so the app target has not compiled on `main` since (`ci.yml` on `main` red for the last 3 dispatches).
- Re-pin to bonsplit `origin/main` (`c545256`; `18f3f8b` is an ancestor, and it is already pushed).

## Test plan
- [ ] `ci.yml` dispatch on this branch green (macOS build + app-host tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11190"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790642026&installation_model_id=21920&pr_number=11190&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11190&signature=73f69e71d72e389311c24fabe3118440a3cae364b76251b8d4967135379074d5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
[...]

PR #11193, opened 2026-08-30
Fix main compile: duplicate exactEntry + processLiveness memberwise init

austinywang:
## Summary
`main` fails to build the app target for two remaining reasons (the third, the bonsplit pointer regression from #11163, was already restored by #11187); this PR fixes both so one merge turns the build lanes green:

1. **duplicate `exactEntry`** — #10658 and #7151 each added an identical `exactEntry(workspaceId:panelId:)` to `RestorableAgentSession` → `invalid redeclaration`. Keep one copy, fold the doc comments.
2. **`processLiveness` memberwise init** — #10658 declared `let processLiveness: RestorableAgentProcessLiveness = .unknown` on `AgentHibernationRecord` and passes `processLiveness:` at the record's only construction site (`AgentHibernationController+Records.swift:123`); a `let` with an initial value is excluded from the synthesized memberwise init → `extra argument 'processLiveness' in call`. Make it a defaulted `var`. No behavior change.

Note: submodule-pointer-only diffs short-circuit the Swift build lanes in `ci.yml`, which is why the now-closed #11190 showed green while main was unbuildable; this PR's Swift diff exercises the real build. The app-host unit-test shard failures are the known pre-existing storm on main and appear identically on every run.

## Test plan
- [x] `ci.yml` dispatch on this branch: `tests-build-and-lag` and `release-build` green (they fail on current main)

### DIFF
--- a/Sources/RestorableAgentSession.swift
+++ b/Sources/RestorableAgentSession.swift
@@ -1058,15 +1058,7 @@ struct RestorableAgentSessionIndex: Sendable {
     ///
     /// Security-sensitive callers use this instead of the compatibility lookup
     /// below so a stale workspace cannot adopt a same-panel entry from another
-    /// restored workspace.
-    func exactEntry(workspaceId: UUID, panelId: UUID) -> Entry? {
-        entriesByPanel[PanelKey(workspaceId: workspaceId, panelId: panelId)]
-    }
-
-    /// Returns only the process entry keyed by this exact workspace/panel pair.
-    ///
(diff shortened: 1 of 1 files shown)


austinywang:
ci.yml dispatch on this branch: [run 33335210826](https://github.com/manaflow-ai/cmux/actions/runs/33335210826) — `tests-build-and-lag`, `release-build`, and `swift-package-tests` all green (they fail on current main without these fixes). The only red lanes are the app-host unit-test shards and their `tests`/`ci-status` aggregates, which fail identically on every run today regardless of base — the known pre-existing storm.

PR #11250, opened 2026-08-31
vm: Blaxel retry, honest unsupported-op errors, snapshot 404 mapping, provider-agnostic quota refresh

lawrencecchen:
Hardens the Cloud VM control plane against the August Blaxel provisioning failures (18 of 19 real create attempts failed). Four fixes, each with unit tests; the snapshot-404 fix lands as a red-then-green commit pair.

**Blaxel control-plane retry** (`web/services/vms/drivers/blaxel.ts`): `blaxelFetch` was a single bare fetch with a 60s abort, so any transient 429/5xx or network blip failed the whole workflow. It now retries up to 4 attempts with full-jitter exponential backoff (250ms base, 4s cap) and honors `Retry-After` (capped 15s). 429 retries for every method because the server refused before doing work; 5xx and network errors retry only idempotent requests (GET, DELETE, HEAD, and this driver's fixed-content PUTs). POST is never replayed on 5xx or network failure because it may have executed: sandbox create keeps relying on its existing 409 name-collision loop (Blaxel exposes no idempotency-key mechanism on this API). Exhausted retries throw a distinct `BlaxelRetryExhaustedError`; non-retriable failures keep the exact historical `-> <status>` message shape the 409/404 matchers depend on.

**Snapshot 404 dispatch** (`web/services/vms/errors.ts`): `VmSnapshotNotFoundError` was in the `VmWorkflowError` union but missing from `vmWorkflowErrorTags`, so `runVmWorkflow` rethrew the raw FiberFailure and restore-of-unknown-snapshot returned 500 `vm_internal_error` instead of 404 `vm_snapshot_not_found`. The hand-kept tag array is replaced by a record checked with `satisfies Record<VmWorkflowError["_tag"], true>`, so the union and the dispatch table can no longer drift: a missing tag is now a compile error, and an extra tag is rejected too.

**Honest unsupported-op errors** (`web/services/vms/routeHelpers.ts`, `observability.ts`): Blaxel snapshot/restore throw `NotImplementedError`, which surfaced as 502 `vm_cloud_service_unavailable` `retryable:true`, telling users to retry an operation the provider will never perform. A deliberately narrow cause-chain check (driver `NotImplementedError`, or the gateway's "not supported by this provider" refusal) now maps to 501 `vm_operation_unsupported`, `retryable:false`, with per-phase copy that starts "Do not retry". The code is excluded from operator-fault alerting: a missing capability is a product limitation, not an incident, while every other 5xx still reports.

**Provider-agnostic quota refresh** (`web/services/vms/workflows.ts`): `refreshActiveLimitProviderStatuses` returned early for every non-Freestyle machine, so a stale `running` row blocked Blaxel creates until the 10-minute cron even though Blaxel has a status read and the cron already reconciles all providers. The lazy refresh on limit-exceeded now probes every candidate with a `providerVmId`, matching the cron. Candidates are `running` rows only, so the gateway's `"running"` fallback for a driver without `getStatus` stays a harmless no-op.

No entitlement numbers change (free 0, paid 5), no metering, no new product concepts.

## Trade-offs

- POSTs are not replayed on 5xx/network failure, so a create hitting a transient 500 still fails that attempt; the alternative (blind POST replay) risks duplicate sandboxes and double provisioning. The caller-level retry guidance (`vm_cloud_service_unavailable`, retryAfterSeconds) covers this path.
- The unsupported-op detector is string/name-based on the cause chain rather than a new typed error through the gateway. Threading a typed error would require widening every `VmProviderOperationError` signature across all drivers for one mapping; the narrow matcher (class name plus one exact gateway phrase) follows the existing `providerImageNotFound` pattern. Risk: a future provider message that happens to contain "not supported by this provider" would map to 501; the phrase is specific enough that this is unlikely.
- 501 keeps HTTP honesty but is >= 500, so it needed an explicit carve-out from operator-fault alerting; the alternative (409) would have kept alerting untouched at the cost of lying about the semantics.
- Older clients that special-case only `vm_cloud_service_unavailable` will show the generic error rendering for `vm_operation_unsupported`; the payload still carries `retryable:false` and the ui block, which current clients render.
- `blaxelFetch` and two retry helpers are now exported for tests (DI via an options bag, no module state); production call sites are unchanged.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11250"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790760221&installation_model_id=21920&pr_number=11250&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11250&signature=3a575c76ede959062458f507f830f44ff83ec9560a1ab6c5e16a6b84394f3032"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Hardens the Cloud VM control plane against the August Blaxel provisioning failures (18 of 19 creates failed) and fixes three error-mapping bugs. Unsupported operations now use a structured typed error, and retry exhaustion no longer leaks provider details or upstream bodies to callers.

**Blaxel control-plane retry**
- `blaxelFetch` retries transient failures up to 4 times with full-jitter backoff (250 ms base, 4 s cap), honoring `Retry-After` (capped 15 s), bounded by a wall-clock deadline covering all attempts and waits; 429s retry for every method, 5xx/network only for idempotent requests, and POST is never replayed.
- Exhausted retries throw a distinct `BlaxelRetryExhaustedError`; the public response reports `provider_retry_exhausted` with no URL or upstream body, which stays in operator logs only.

**Error mapping, localization, and quota refresh**
- Restore of an unknown snapshot now returns 404 `vm_snapshot_not_found`; the dispatch table is derived from the error union so missing tags are compile errors.
- Unsupported provider operations throw a structured `VmOperationUnsupportedError` from drivers and the gateway, mapping to 501 `vm_operation_unsupported`, `retryable:false`, with do-not-retry copy localized in 20 locale catalogs (locale negotiated from request headers, URL, cookie, or Accept-Language) and excluded from operator-fault alerting; message text no longer drives retry classification.
- The lazy quota refresh now probes every provider with a `providerVmId`, not just Freestyle, bounded to 200 candidates with concurrency capped at 10.

No entitlement numbers, metering, or product concepts change.

<sup>Written for commit 792af17f4346b3df7705bd1d933858b597bf53d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Reliability**
  * Added bounded automatic retries for temporary cloud-service failures, with cancellation, timeouts, and retry guidance.
  * Safely retries eligible requests while preventing duplicate operations.

* **Bug Fixes**
  * Snapshot-not-found errors now return accurate 404 responses.
  * Active VM limits refresh consistently across supported providers.

* **Error Handling**
  * Unsupported VM operations return non-retryable 501 responses with localized guidance.
  * Retry-exhausted failures provide safer responses without exposing provider details.
  * Expected provider limitations are no longer classified as operator faults.

* **Localization**
  * Added unsupported-operation guidance across supported languages.


lawrencecchen:
Both Vercel preview deployments wedged in pending for ~2h on head a1cabcd5 and never reported, blocking auto-merge. Pushed an empty commit to retrigger them; no content change, so the clean local autoreview of a1cabcd5 carries over.

PR #11258, opened 2026-08-31
fix(cmux-tui): write codex hook trust state so installed hooks execute

### LINKED ISSUE #11040: cmux-tui: codex 0.150 interactive mode does not execute hooks.json commands, so codex agents never reach the agents view

lawrencecchen:
## Root cause

codex 0.150 gates every unmanaged `hooks.json` handler behind per-hook trust. A handler runs only when the user config layer (`$CODEX_HOME/config.toml`) carries `hooks.state."<hooks.json path>:<event>:<group>:<handler>"` whose `trusted_hash` equals codex's normalized-identity hash: sha256 over sorted-key compact JSON of `{event_name, hooks:[normalized command handler]}` (codex-rs/hooks/src/engine/discovery.rs `hook_hash` + codex-rs/config/src/fingerprint.rs `version_for_toml` at rust-v0.150.1). The hash covers the normalized timeout, so SessionEnd hashes the clamped 3s value, not the configured 5s. An explicitly set `CODEX_HOME` is canonicalized before it lands in the key (codex-rs/utils/home-dir/src/lib.rs `find_codex_home`); the default `~/.codex` is not.

`agent hook install` wrote hooks.json but never any trust state, so codex parsed the file (it even printed the SessionEnd clamp warning the issue quotes) and silently skipped every handler as Untrusted. `codex exec` has no trust-review UI at all, and the interactive startup review is easy to dismiss, so codex sessions never produced journal events and never reached the agents view.

Disproven alternative theories from the issue: hook subprocesses DO inherit the parent environment (codex snapshots its own `std::env::vars_os()` at session start and `CMUX_TUI_HOOK` is not on the scrub list), the event names are correct, and the `hooks` feature defaults to enabled at 0.150.1, so no absolute-path command or extra feature key is needed.

Verified against the real 0.150.1 binary in an isolated `CODEX_HOME`: with the replicated `hooks.state` entries the installed commands execute (SessionStart, UserPromptSubmit, SessionEnd fired and reached the probe helper); with an identical hooks.json and no trust state nothing runs. A probe with a custom group ahead of the cmux group confirmed trust keys are positional (`...:session_start:1:0`).

## Fix

The installer now mirrors codex's hash exactly and writes the trust entries into `$CODEX_HOME/config.toml` next to hooks.json, keyed by the installed group's actual array position so pre-existing custom groups keep their indices. Edits are comment- and format-preserving via `toml_edit` (the same 0.22.27 the workspace's `toml 0.8` already vendors, so no new transitive code). Uninstall removes only entries whose `trusted_hash` matches a cmux-owned command hash, install refuses to touch a config whose `hooks` key is not a table (e.g. a stray `hooks = true`) instead of corrupting it, and `agent hook status` reports `partial` until config.toml trusts every installed handler.

Principled or hacky: principled with one accepted coupling. Writing trust on the user's behalf is what the install command means (the user asked cmux-tui to wire codex up), and it is the same state codex's own Trust-all review writes. The coupling is that the hash mirrors codex's normalization; if codex changes its scheme the hooks fall back to Untrusted (fail-safe: nothing executes, status shows partial), and the hardcoded binary-verified hash constants in the tests pin the algorithm against silent drift.

Uninstall tradeoff: uninstall never removes `[hooks]`/`[hooks.state]` tables, only the trust entries whose hashes prove cmux ownership, because empty-and-undecorated cannot distinguish an installer-created table from a user placeholder across processes. After install then uninstall on a config that never had `[hooks]`, an empty table skeleton remains; codex ignores it, and nothing is ever deleted that cmux did not provably write.

## Red/green

- red (test-only commit 21d83b4d14, fails because no trust state is written): https://github.com/manaflow-ai/cmux/actions/runs/33385345533
- green (fix commit 29395ebcbe): https://github.com/manaflow-ai/cmux/actions/runs/33385445589

Fixes https://github.com/manaflow-ai/cmux/issues/11040

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11258"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790766462&installation_model_id=21920&pr_number=11258&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11258&signature=09fc30b7d2c2269aa3bca10195a0e2d8e6e1265488eef83a6b4d786b1ff0e8a6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes codex hook installation so installed hooks execute under codex 0.150, which was silently skipping every handler without matching trust state in `config.toml`.

- `agent hook install` now mirrors codex's normalized-identity hash and writes `hooks.state` entries into `$CODEX_HOME/config.toml`, keyed by the installed handler's group and handler positions so pre-existing custom groups and foreign hooks keep their indices.
- Edits preserve comments and formatting via `toml_edit`, including inline `hooks = { ... }` tables; install refuses malformed `hooks` keys instead of corrupting the config.
- Uninstall deletes trust entries by the removed hook's positional key regardless of stored hash (so entries codex re-accepted after user edits are cleaned up too), with hash matching as fallback, and never removes tables so user-created commented tables stay byte-identical.
- Trust-state edits are preflighted before touching `hooks.json`; a failed config commit rolls `hooks.json` back, snapshot read errors fail loud, and concurrent config edits are preserved by re-rendering at commit time.
- Config reads are TOCTOU-free: opened once with `O_NONBLOCK`, regular file under 16 MiB required, read bounded, so swapped-in FIFOs or oversized files make install and uninstall fail closed and status report `partial`.
- Symlinked `config.toml` files are edited via their resolved target so the link survives; dangling links fail install.
- `agent hook status` hashes the entry actually on disk, so user edits to command or timeout demote it to `partial`.
- With `CODEX_HOME` explicitly set, trust keys use the canonicalized path so they match what codex computes.

Fixes https://github.com/manaflow-ai/cmux/issues/11040.

<sup>Written for commit 47be6ca7abc2174fec1a03144dc536ba5517a535. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11258?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex hook installation and removal across custom, normalized, and symlinked configuration paths.
  * Synchronized trusted hook settings so configured handlers are recognized accurately.
  * Codex status now reports partial or invalid trust configurations.
  * Added validation for malformed, oversized, non-regular, and unwritable configuration files.
  * Improved consistency during installation, uninstallation, and reinstallation, including rollback when updates cannot be completed safely.

### DIFF
--- a/cmux-tui/Cargo.lock
+++ b/cmux-tui/Cargo.lock
@@ -780,6 +780,7 @@ dependencies = [
  "terminput-crossterm",
  "tokio",
  "toml",
+ "toml_edit 0.22.27",
  "unicode-segmentation",
  "unicode-width",
  "url",
--- a/cmux-tui/Cargo.toml
+++ b/cmux-tui/Cargo.toml
@@ -82,6 +82,9 @@ axum = { version = "0.8", features = ["ws"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 toml = "0.8"
+# Same version toml 0.8 already pulls in; used for comment-preserving edits of
+# the user's codex config.toml when syncing hook trust state.
+toml_edit = "0.22"
 portable-pty = "0.9"
 crossterm = "0.29"
 crossbeam-channel = "0.5"
--- a/cmux-tui/crates/cmux-tui/Cargo.toml
+++ b/cmux-tui/crates/cmux-tui/Cargo.toml
@@ -33,6 +33,7 @@ unicode-segmentation.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 toml.workspace = true
+toml_edit.workspace = true
 base64.workspace = true
 getrandom.workspace = true
 glob.workspace = true
--- a/cmux-tui/crates/cmux-tui/src/agent_hook_install.rs
+++ b/cmux-tui/crates/cmux-tui/src/agent_hook_install.rs
@@ -1542,39 +1542,53 @@ fn resolve_codex_config_path(config_path: PathBuf) -> anyhow::Result<PathBuf> {
 
 /// Size gate shared by every config.toml consumer, checked against metadata
 /// before any read so an oversized file is rejected without loading it.
-fn codex_config_within_size_limit(config_path: &Path) -> anyhow::Result<bool> {
+/// Gate shared by every config.toml consumer, checked from metadata (of the
+/// symlink-resolved target) before any open: the file must be a regular file
+/// within the size limit, so a FIFO or other special file can never block a
+/// read and an oversized file is rejected without being loaded.
+fn check_codex_config_readable(config_path: &Path) -> anyhow::Result<()> {
     match fs::metadata(config_path) {
-        Ok(metadata) => Ok(metadata.len() <= MAX_CONFIG_BYTES),
-        Err(error) if error.kind() == io::ErrorKind::NotFound => Ok(true),
+        Ok(metadata) => {
+            anyhow::ensure!(
+                metadata.is_file(),
+                "{} is not a regular file; refusing to read it",
+                config_path.display()
+            );
+            anyhow::ensure!(
+                metadata.len() <= MAX_CONFIG_BYTES,
+                "{} exceeds 16 MiB",
+                config_path.display()
+            );
+            Ok(())
+        }
+        Err(error) if error.kind() == io::ErrorKind::NotFound => Ok(()),
         Err(error) => Err(error).with_context(|| format!("inspect {}", config_path.display())),
     }
 }
 
-fn read_codex_config_document(
-    config_path: &Path,
-) -> anyhow::Result<(Option<String>, toml_edit::DocumentMut)> {
-    anyhow::ensure!(
-        codex_config_within_size_limit(config_path)?,
-        "{} exceeds 16 MiB",
-        config_path.display()
-    );
-    let original = match fs::read(config_path) {
+fn read_codex_config_text(config_path: &Path) -> anyhow::Result<Option<String>> {
+    check_codex_config_readable(config_path)?;
+    match fs::read(config_path) {
         Ok(bytes) => {
             anyhow::ensure!(
                 bytes.len() as u64 <= MAX_CONFIG_BYTES,
                 "{} exceeds 16 MiB",
                 config_path.display()
             );
-            Some(
+            Ok(Some(
                 String::from_utf8(bytes)
                     .with_context(|| format!("{} is not valid UTF-8", config_path.display()))?,
-            )
+            ))
         }
-        Err(error) if error.kind() == io::ErrorKind::NotFound => None,
-        Err(error) => {
-            return Err(error).with_context(|| format!("read {}", config_path.display()));
-        }
-    };
+        Err(error) if error.kind() == io::ErrorKind::NotFound => Ok(None),
+        Err(error) => Err(error).with_context(|| format!("read {}", config_path.display())),
+    }
+}
+
+fn read_codex_config_document(
+    config_path: &Path,
+) -> anyhow::Result<(Option<String>, toml_edit::DocumentMut)> {
+    let original = read_codex_config_text(config_path)?;
     let document = original
         .as_deref()
         .unwrap_or_default()
@@ -1636,60 +1650,55 @@ fn codex_state_table_mut<'a>(
     ))
 }
 
+/// The validated inputs of one trust-state edit, kept alongside the rendered
+/// document so the commit can re-render against fresh file content when
+/// config.toml changed between the preflight and the commit.
+struct CodexTrustEdit {
+    config_path: PathBuf,
+    key_prefix: String,
+    desired: BTreeMap<String, String>,
+    owned_hashes: BTreeSet<String>,
+    install: bool,
+}
+
 /// A fully validated, pre-rendered config.toml trust-state write. Producing
 /// one performs every step that can fail for content reasons (read, parse,
 /// shape validation, rendering, target-directory writability), so the caller
 /// can sequence it BEFORE mutating hooks.json and commit it afterwards.
 struct PreparedCodexTrustWrite {
-    config_path: PathBuf,
+    edit: CodexTrustEdit,
+    original: Option<String>,
     updated: String,
     mode: u32,
 }
 
-/// Preflights the cmux-owned `hooks.state` trust-entry edit of the codex user
-/// config next to `hooks_path`, preserving unrelated keys, comments, and
-/// formatting. Returns `None` when the config already matches.
-fn prepare_codex_trust_state(
-    hooks_path: &Path,
-    from_env_override: bool,
-    root: &Map<String, Value>,
-    install: bool,
-) -> anyhow::Result<Option<PreparedCodexTrustWrite>> {
-    let parent = hooks_path.parent().context("codex hooks path has no parent")?;
-    // Resolve a symlinked config.toml to its target so the later atomic
-    // rename edits the real file instead of detaching the link.
-    let config_path = resolve_codex_config_path(parent.join(CODEX_CONFIG_FILE))?;
-    let key_hooks_path = codex_state_key_hooks_path(hooks_path, from_env_override);
-    let desired = if install {
-        codex_expected_trust_entries(&key_hooks_path, root)?
-    } else {
-        BTreeMap::new()
-    };
-    if !install && !config_path.exists() {
-        return Ok(None);
-    }
-    let (original, mut document) = read_codex_config_document(&config_path)?;
-    let owned_hashes = codex_owned_trust_hashes()?;
-    let key_prefix = format!("{}:", key_hooks_path.display());
+/// Reads the config and renders the trust edit against its current content,
+/// preserving unrelated keys, comments, and formatting. Returns the file's
+/// current text plus the rendered replacement, or `None` when the current
+/// content already satisfies the edit.
+fn render_codex_trust_document(
+    edit: &CodexTrustEdit,
+) -> anyhow::Result<(Option<String>, Option<String>)> {
+    let (original, mut document) = read_codex_config_document(&edit.config_path)?;
 
-    if let Some(state) = codex_state_table_mut(&mut document, &config_path, install)? {
+    if let Some(state) = codex_state_table_mut(&mut document, &edit.config_path, edit.install)? {
         let stale: Vec<String> = state
             .iter()
             .filter(|(key, entry)| {
-                key.starts_with(&key_prefix)
-                    && !desired.contains_key(*key)
+                key.starts_with(&edit.key_prefix)
+                    && !edit.desired.contains_key(*key)
                     && entry
                         .as_table_like()
                         .and_then(|entry| entry.get("trusted_hash"))
                         .and_then(toml_edit::Item::as_str)
-                        .is_some_and(|hash| owned_hashes.contains(hash))
+                        .is_some_and(|hash| edit.owned_hashes.contains(hash))
             })
             .map(|(key, _)| key.to_string())
             .collect();
         for key in stale {
             state.remove(&key);
         }
-        for (key, hash) in &desired {
+        for (key, hash) in &edit.desired {
             match state.get_mut(key).and_then(toml_edit::Item::as_table_like_mut) {
                 Some(entry) => {
                     entry.insert("trusted_hash", toml_edit::value(hash.clone()));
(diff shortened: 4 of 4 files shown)


lawrencecchen:
Addressed the review P1 in 9358543086: hooks.json was written before the config.toml trust edit, so a malformed config (invalid TOML, scalar `hooks`/`hooks.state`, oversized file) or an unwritable directory failed the install after leaving new untrusted hooks behind, and failed the uninstall after removing hooks while stale trust entries lingered.

The trust edit is now split into prepare and commit. Prepare does everything that can fail for content reasons (read, parse, shape validation, rendering the new document, a directory writability probe) and runs BEFORE hooks.json is touched; then hooks.json is written; then the pre-rendered config write commits. If that final write still fails, hooks.json is restored to its pre-operation bytes, best effort, with any restore failure chained into the error. New tests pin the contract: a scalar `hooks` key fails the install with hooks.json byte-identical (pre-existing custom file, and absent stays absent), and an unparseable config fails the uninstall with hooks.json untouched.

Hosted verification on head 9358543086, all 9 codex installer tests green on Linux and macOS: https://github.com/manaflow-ai/cmux/actions/runs/33387721819

lawrencecchen:
Addressed both follow-up findings in eccebd0af2.

1. Status-path size cap: every config.toml consumer now gates on metadata length before reading (`codex_config_within_size_limit`), and `agent hook status` reports the provider as partial for an oversized config instead of reading it, so a huge file or a symlink to one cannot stall the TUI.

2. Symlinked config.toml: the preflight resolves the link and applies every check, the mode lookup, the writability probe, and the final atomic write to the resolved TARGET, so the link survives and codex keeps editing the same file. A link that does not resolve fails the operation with a clear error before hooks.json is touched.

Pinned by tests: a symlinked config keeps its link across install, status, idempotent reinstall, and uninstall while the target file carries the trust entries; a dangling link refuses the install with hooks.json left absent; an oversized config demotes status to partial without being read.

Hosted verification on head eccebd0af2, all 12 codex installer tests green on Linux and macOS: https://github.com/manaflow-ai/cmux/actions/runs/33390641094

lawrencecchen:
Addressed both follow-up findings in bead308a46.

1. Handler-index trust keys: keys hardcoded handler index 0 and located only the cmux-owned group, so a foreign hook inserted before the cmux hook inside the same group made codex look up a nonexistent key and skip the hook while install and status reported success. `codex_owned_hook_position` now returns (group index, handler index) and the key uses both positions in install, uninstall cleanup, and status alike. Pinned by `codex_trust_keys_use_the_owned_hook_index_within_a_group`: with a foreign hook before ours in one group, status demotes to partial (codex would skip the hook), and a reinstall mints the key for the hook's actual position and restores installed.

2. Snapshot error swallowing: the pre-mutation hooks.json snapshot used `fs::read(path).ok()`, collapsing genuine read errors into None, so a rollback after a failed trust commit would have deleted a pre-existing hooks.json as if it had been absent. `snapshot_file` now propagates every error except NotFound and runs before any mutation, failing the operation up front. Pinned by `snapshot_file_propagates_non_notfound_read_errors` using a directory-in-place-of-file read error.

Hosted verification on head bead308a46: all 13 codex installer tests green on Linux and macOS (https://github.com/manaflow-ai/cmux/actions/runs/33392924629), and the snapshot unit test green on both (https://github.com/manaflow-ai/cmux/actions/runs/33392944899).

lawrencecchen:
Addressed both findings in c9ce2020f6.

1. P1 (FIFO hang): the metadata gate checked only length, so a FIFO at $CODEX_HOME/config.toml passed it and the following read blocked forever. `check_codex_config_readable` now requires `metadata.is_file()` (on the symlink-resolved target) before ANY open, in every config.toml consumer: install and uninstall fail closed in preflight with "not a regular file" and hooks.json untouched, and status reports partial. Pinned by `codex_operations_fail_closed_on_a_fifo_config_toml`: mkfifo with no writer, all three operations run on a worker thread behind a bounded `recv_timeout`, so a regression fails as a timeout panic instead of hanging the suite. No open ever touches the FIFO since the gate is stat-based.

2. P2 (preflight-to-commit window): the trust document was rendered at preflight and committed wholesale later, discarding any config.toml edit made in between. The render now lives in `render_codex_trust_document` over a kept `CodexTrustEdit` input set; the commit re-reads the file and compares against the preflight snapshot bytes, and on a mismatch re-renders against the fresh content (shape re-validated) and commits that — or, if the fresh content no longer validates, propagates the error so the caller's rollback restores hooks.json. Pinned by `codex_trust_commit_re_renders_after_a_concurrent_config_edit`: a concurrent edit injected between prepare and commit survives in the final file next to all 11 trust entries, and the invalid-fresh-content path errors with hooks.json rolled back.

Hosted verification on head c9ce2020f6: all 15 codex installer tests green on Linux and macOS (https://github.com/manaflow-ai/cmux/actions/runs/33396919778), snapshot unit test green on both (https://github.com/manaflow-ai/cmux/actions/runs/33396919907).

lawrencecchen:
Round 5 addressed in 1ec172aed4.

1. Inline hooks table: `hooks = { enabled = true }` passed the table-like check, but inserting an `Item::Table` into an inline table was silently discarded by `TableLike::insert` and the following `expect` panicked on valid input. `codex_state_table_mut` now matches the parent's representation (inserts an inline `state` value inside an inline `hooks` table, a regular table otherwise) and every former `expect` is a contexted error, so no config content can panic. Pinned by `codex_install_supports_an_inline_hooks_table` (install adds the 11 trust entries inside the inline table, uninstall returns it to `enabled = true` only).

2. Commented user tables: uninstall removed any empty `[hooks]`/`[hooks.state]`, including a user's commented placeholder, deleting their comments. Creation cannot be tracked across processes, so removal is now gated on empty AND undecorated (`item_is_removable_empty_table` + `decor_is_blank`; an unresolvable raw span counts as decoration, comments cannot exist inside inline tables so those check only their value decor). Pinned by `codex_uninstall_keeps_a_commented_user_state_table_byte_identical`: a commented empty `[hooks.state]` survives an install+uninstall cycle byte-identical.

3. Lock/versioned-write question, verified against the rust-v0.150.1 tag: codex has NO on-disk lock or version marker for config.toml, so there is no protocol to join.
- File writes go through `write_atomically` (codex-rs/utils/path-utils/src/lib.rs:122): `NamedTempFile::new_in(parent)` + `persist` (rename). No flock, no lock file, no version stamp.
- The direct edit path `apply_blocking_to_resolved_file` (codex-rs/core/src/config/edit.rs:738, write at :773) is a plain unlocked read-mutate-rename.
- The `expected_version` field on `config/batchWrite` is enforced inside the app-server process only (codex-rs/app-server/src/config_manager_service.rs:247-253): it compares a caller-supplied version string against the in-memory user layer's content fingerprint, then still finishes with the same unlocked atomic rename. It is an RPC-level optimistic check for app-server clients; an external process cannot join it, and its own compare-then-write has the identical terminal window.

Unilateral locking by cmux-tui would protect nothing, since codex would not honor the lock. The window is therefore inherent to editing this file from two processes; our commit already minimizes it (re-read + re-render on mismatch, c9ce2020f6) and the failure mode is fail-safe: a lost trust entry means the hook is Untrusted, codex refuses to run it, `agent hook status` reports partial, and reinstall heals.

Hosted verification on head 1ec172aed4: all 17 codex installer tests green on Linux and macOS: https://github.com/manaflow-ai/cmux/actions/runs/33400369465

lawrencecchen:
Round 6 addressed in 16412ce54f.

1. P1 (stat-to-read race): the metadata gate and the read were separate syscalls, so a concurrent swap to a FIFO or huge file between them defeated both protections. `read_codex_config_text` is now the single way every consumer (status, preflight, commit re-read) loads config.toml: open once with `O_NONBLOCK` on unix (no-op for regular files; stops a FIFO open from blocking on a missing writer), `fstat` the OPENED descriptor requiring a regular file within the 16 MiB cap, then a bounded `Read::take(MAX+1)` read from that same descriptor with an over-limit reject. The old stat-based helper is deleted, so there is exactly one read path. The existing FIFO and oversized tests now exercise this path.

2. P2 (status hashed the canonical shape): a user-edited command/timeout on the installed entry gives codex a different normalized-identity hash, so codex skips the hook as Modified, yet status kept comparing the stored trust hash against the canonical shape and reported installed. `codex_expected_trust_entries` now hashes the entry ACTUALLY present at the owned position, mirroring codex's full normalization: timeout defaults (600s, SessionEnd 1s) and the SessionEnd 3s clamp, matcher dropped for UserPromptSubmit/Stop, statusMessage, the Windows command override, and the additionalContextLimit event gate plus its 2500-token default filter. Install still rewrites entries to canonical form first, so the install-side trust writes are unchanged; only drift detection improves. The extended normalization was re-validated against the real codex 0.150.1 binary: a non-canonical entry (timeout 60, statusMessage, additionalContextLimit 100) trusted with the hash these rules produce was executed by codex.

Pinned by `codex_status_detects_a_user_edited_hook_entry`: after install, mutating the installed hook's timeout in hooks.json demotes status to partial; reinstall heals to installed.

Hosted verification on head 16412ce54f: all 18 codex installer tests green on Linux and macOS: https://github.com/manaflow-ai/cmux/actions/runs/33403527714

lawrencecchen:
Round 7 addressed in e23cd045b3.

Uninstall now follows the terminal rule: it removes ONLY the trust entries whose hashes prove cmux ownership and never removes `[hooks]`/`[hooks.state]` tables. The empty-and-undecorated heuristic (which still deleted a pre-existing empty undecorated placeholder, since table ownership cannot be proven across processes) is gone: `item_is_removable_empty_table`, `decor_is_blank`, and the removal branch are deleted.

Tradeoff, also stated in the PR body: after install then uninstall on a config that never had `[hooks]`, an empty `[hooks.state]` skeleton remains. codex ignores it, and nothing is ever deleted that cmux did not provably write.

Tests: the removes-entirely test became `codex_uninstall_removes_entries_but_never_tables` (expects the empty skeleton with no `trusted_hash` left); the commented-placeholder byte-identity test is unchanged and now trivially safe.

Hosted verification on head e23cd045b3: all 18 codex installer tests green on Linux and macOS: https://github.com/manaflow-ai/cmux/actions/runs/33406172402

lawrencecchen:
Round 8 addressed in 47be6ca7ab.

Uninstall previously matched trust entries against the canonical hash set only, but codex persists a NEW hash when a user reviews and accepts a modified cmux hook, so that entry survived uninstall at the cmux hook's positional key and would pollute a later hook installed at the same position.

Ownership is now keyed on the hooks.json relationship: `prepare_codex_trust_state` receives the PRE-rewrite hooks.json content, and `codex_owned_entry_keys` collects the positional key of every cmux-owned handler occurrence (proven by its command marker at its real position). Every such key that this operation removes from hooks.json has its trust entry deleted REGARDLESS of hash. The canonical-hash sweep stays as secondary cleanup for orphaned keys from prior layouts. Install passes its pre-rewrite root too, so a reinstall that shifts positions also clears a re-accepted entry stranded at the old position.

Pinned by `codex_uninstall_removes_a_re_accepted_trust_entry`: install, rewrite the stop entry's `trusted_hash` to an arbitrary value (simulating codex re-accepting a user-edited hook), add a user-owned entry at a foreign key, uninstall — the hook is gone from hooks.json, its trust entry is gone from config.toml, and the user's entry survives.

Hosted verification on head 47be6ca7ab: all 19 codex installer tests green on Linux and macOS: https://github.com/manaflow-ai/cmux/actions/runs/33409391835

PR #11259, opened 2026-08-31
cmux-tui: agent alerts, working-signal parity, agent wait, session resume (herdr round 2)

lawrencecchen:
herdr parity round for the agent pipeline, stacked on https://github.com/manaflow-ai/cmux/pull/11068 (base branch of this PR). Ported material is credited per file and in `cmux-tui/ATTRIBUTIONS.md`: herdr (https://github.com/herdrdev/herdr), Apache-2.0, pinned commit 7b675f42af35, modified by manaflow.

## A. Host-terminal alerts for agent transitions

Ported from herdr's `terminal_notify.rs`, improved: alerts are presentation, so the ATTACHED CLIENT (never the daemon) reacts to agent-changed events by emitting OSC 9 (Ghostty/iTerm2/WezTerm) or OSC 99 (kitty) plus a BEL through its own host terminal, tmux-passthrough-wrapped inside tmux. Config `notifications.agent_blocked` (default true; any entry into blocked) and `notifications.agent_idle` (default true; only when idle ends a working/blocked run). The focused terminal never alerts; edges are per surface.

## B. Working-signal parity

Flowing PTY output within 1.5s upgrades an otherwise idle screen read to working (herdr's flowing-output authority), and the tracker owes one re-evaluation when the window closes so the upgrade cannot stick. A passive OSC 9 tracker (ported from herdr's `pane/osc.rs`, trimmed to progress) feeds the four previously inert `osc_progress` manifest rules; capture is gated to agent panes. PTY activity only informs the SCREEN source; hook > screen > socket arbitration is untouched.

## C. `agent wait` (better than herdr)

`cmux agent wait (--state <s> | --any-of <a,b>) [--terminal <id>] [--timeout-ms <n>]` blocks until an agent reaches one of the states and prints the matching snapshot. herdr's `api/wait.rs` polls with regexes; ours is an event-driven subscription on agent projection commits (no polling), detached into the server's cancelable wait worker like `terminal.wait`. New `agent.wait` operation in the catalog.

## D. Journal-derived session resume

Ported herdr's `agent_resume.rs` planner, improved: session references come from the journal's hook-reported `agent_session_id`/`cwd` (no second persistence). `cmux agent resume` prints exact relaunch argv (`claude --resume <id>`, `codex resume <id>`, ...) for roster agents alive at the last shutdown, deduped by (agent, reference), with `agent_running` marking live sessions an apply must skip. References validate as bounded control-free data and ride as single argv elements, never shell text (hostile-input tests ported).

## Trade-offs (each with the strongest objection)

1. **No custom alert audio.** herdr ships done.mp3/request.mp3; their audio provenance is undocumented, so we rely on the host terminal's notification sound. Objection: herdr users get distinct per-transition sounds and we do not; answer: unverifiable provenance in an Apache repo is not worth vendoring, and the host sound covers the daily need.
2. **Alerts are client-local presentation.** Two attached clients can alert differently (focus differs). Objection: a daemon-side notifier would be one source of truth; answer: the daemon has no host terminal, and per-viewer focus suppression is the feature.
3. **tmux passthrough requires `allow-passthrough on`.** Off by default in tmux ≥3.3; alerts silently drop inside such tmux sessions. Same limitation as herdr.
4. **Activity window is 1.5s at a 100ms scan.** A paste into an idle agent's pane briefly reads working. Objection: false working states erode trust; answer: herdr accepts the same with pure PTY-activity authority, ours decays within ~2s and never overrides hooks or blocked screens.
5. **OSC progress capture costs one relaxed atomic load per output chunk on non-agent panes** and a byte scan on agent panes. Objection: hot-path scanning; answer: gating means shells never scan, herdr scans always-on.
6. **`agent wait` matches durable projections**, so a `done` from yesterday matches immediately. Objection: a caller may want "the NEXT transition"; answer: level-triggered waits compose better for coordination (no missed-edge races); next-transition semantics can be an opt-in flag later.
7. **Resume is an explicit command; no automatic respawn on daemon restore.** Objection (herdr's behavior): users expect sessions back without typing anything. Answer: cmux-tui terminals normally SURVIVE daemon restarts via terminal hosts, so auto-resume would fire mainly after crashes - exactly when surprise spawns (auth prompts, spend, wrong cwd) are least welcome. A `resume_auto` config key can gate a restore hook later; shipping the key without the hook would be dishonest.
8. **`agent resume` prints plans; apply is `cmux workspace current run -- <argv...>`.** One-command `--apply` is follow-up (it needs a multi-step CLI plan).
9. **Resume covers id-resumable agents**; gemini and amp have no known resume interface and are skipped explicitly.
10. **Two new catalog operations (127 total)**: spec, inventory, schema enum, markdown tables, and all seven SDK descriptors updated; both spec guards pass.
11. **tmux 3.4+ hides the outer terminal** (TERM_PROGRAM=tmux, TERM=tmux-256color in panes), which silences herdr's detection too; inside tmux we now default to passthrough-wrapped OSC 9 (every major terminal honors it, unknown ones ignore it). Objection: a kitty-outside-tmux user gets OSC 9 instead of OSC 99; answer: silence was worse, and kitty-in-tmux is detectable by nothing.
12. **macOS agent identification uses the kernel process name first** (pbi_name/pbi_comm), because launchers install version-named binaries (claude runs as `.../versions/2.1.251`) that defeat path-basename matching; the path remains the fallback for truncated names. Live dogfood found claude entirely invisible to screen detection without this.
13. **Merge note for the sidebar-splits world**: the splits branch has its own AgentChanged arm (projection invalidation); merging this PR there must fold the alert into that arm or the alert arm is unreachable (rustc warns). The dogfood branch carries the resolved merge.
14. **Pre-existing red on the base branch**: the upstream rebase of feat-tui-agent-screen-detection left ~19 hook-fence/projection tests failing (they fail at the base commit with none of this PR's changes); this PR neither causes nor fixes them, and its focused gates avoid that filter. Flagged to the branch owner.
15. **`agent wait` beyond 10s died with EAGAIN (os error 35).** The CLI caps socket reads at 10s except for long-poll wait operations; `agent.wait` was missing from that allowance (`cli/wire.rs response_read_timeout`), so the server kept waiting while the client's read timed out. Found live on macOS. Fixed with terminal.wait semantics: timeout_ms + 2s margin, unbounded wait means no client read deadline.
16. **Cross-adapter hooks clobbered the live roster owner.** The roster keys one agent per terminal and hooks always won, so a companion process of another adapter inside the same PTY (claude's codex plugin sidecar) drove the shared entry through its own 2-second lifecycle and its SessionEnd deleted the live BLOCKED claude; the static permission-prompt screen never re-established it. Found live: `agent list` returned empty while claude sat at a permission prompt. Reducer v4: a cross-adapter done never removes a live hook-owned entry, and takeover needs the owner's hook evidence stale (STALE_HOOK_MS). Objection: a genuinely new agent in the same terminal waits up to 30s to appear if the previous owner died silently; answer: the owner's own SessionEnd empties the slot instantly in every normal handoff, and 30s is the arbitration window screen detection already uses. Residual: two concurrent SAME-adapter sessions in one PTY still collide (one-agent-per-terminal is the model, as in herdr).
17. **Foreign OSC 7 pwd reports poisoned spawn cwd (sdet8 outage: every "Session operation failed").** Ghostty shell integration emits `file://<dhcp-name>/...` (e.g. `mac.lan`); `terminal_pwd_host_is_local` correctly rejects that host, but the terminal-host adoption snapshot (`snapshot_cwd`) preferred the raw OSC 7 report over the launch cwd, so after any daemon restart the adopted surface's SPAWN cwd became the URL. `local_cwd`'s fallback then returned it, every new tab inherited it into `PtyCommand::cwd`, and posix_spawn died with ENOENT. Fixed at five layers: the host snapshot converts before reporting; adoption converts on intake (hosts outlive daemon upgrades, old host binaries still send URLs); `local_cwd` converts its spawn-cwd fallback; `terminal.process.get` reports the converted cwd; and `platform::safe_spawn_cwd` degrades any unconvertible or missing requested cwd to the launch-default/home directory with a logged warning at both spawn sites, so a spawn can no longer fail over a directory. Host-locality decision: DHCP self-names like `mac.lan` stay REJECTED — the network controls those names, and an OSC 7 from an ssh'd remote shell must not be able to redirect local spawns; with the degradation in place, rejection costs nothing. Trade-off accepted: an explicitly requested missing `--cwd` now opens the terminal in the default directory with a warning instead of failing; the objection (scripts lose a loud error) loses to "one tab's weird pwd report must never make every new tab fail". Also: exited terminals kept as durable receipts no longer feed cwd inheritance. Investigated the reported "zombie running terminals with no host": the hosts were alive orphaned `__terminal-host` processes correctly re-adopted after daemon restarts (adoption reparents them away from the daemon, so they are invisible as daemon children); out-of-band host SIGKILL and restart reconciliation both converge to `exited` (verified live and already covered by `running_host_sigkill_detaches_exited_terminal_topology` and the `daemon_restart_prunes_*` tests).
18. **Unknown top-level config keys are now tolerate-and-warn, not fail-loud.** This round found two config defects: (a) this PR documented `notifications.*` but never taught `load_raw_config`'s hand-rolled `KNOWN` allowlist or its `section!` dispatch the key, so any config using the documented section was rejected; and (b) the rejection path discarded the ENTIRE file over one unknown key, silently replacing the user's layout with defaults (the only trace was a stderr line the TUI overdraws immediately). Chosen policy: parse known fields, skip each unknown top-level key, log it, and surface the first warning in the client status bar at startup and on reload; `Config.startup_warnings` carries it. The strongest expert objection is against tolerate-and-warn itself: a typoed section name (`notifcations`) now degrades that section to defaults instead of forcing an immediate fix, and a status-bar line is easier to miss than a startup error. Answer: fail-loud reproduces the exact defect class this round fixed, one key written by a newer cmux-tui (or one experimental key) blanks the whole layout on every older binary reading the same dotfile; the warning is now user-visible rather than log-only; and sections still reject unknown keys strictly inside themselves, so intra-section typos stay loud. Residual risk accepted: top-level typos degrade visibly-but-softly.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11259"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790768734&installation_model_id=21920&pr_number=11259&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11259&signature=2efca6e35b6adbe87f5b3fd6beef9aa2e7a7a9115651aa76abc020c3878c043e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Brings herdr's agent pipeline to cmux-tui: host-terminal alerts, working-signal parity, an event-driven `agent wait`, and journal-derived session resume.

- Attached clients alert on blocked and idle transitions via OSC 9/99 (focused terminal silent; configurable under `notifications`).
- Flowing PTY output within 1.5s promotes an idle read to working; passive OSC 9 capture feeds the previously inert `osc_progress` rules.
- `agent wait` subscribes to roster commits (no polling) and resolves immediately for already-satisfied states.
- `agent resume` prints exact relaunch argv from the journal, validated and deduped; no auto-respawn.
- macOS agent identification reads argv[0] via `KERN_PROCARGS2` first, so version-named launchers like `claude` still resolve.

**Bug Fixes**
- The `notifications` config section is now actually parsed; previously it was rejected as an unknown key and dropped the whole config.
- Unknown top-level config keys now skip individually instead of discarding the entire file; the first warning shows in the status bar.

Side effects: two new catalog operations and spec updates; alerts inside tmux use OSC 9 passthrough (the outer terminal is undetectable there), which requires `allow-passthrough on`.

<sup>Written for commit b007c222dd1089547d760ba2c5f32c5527a09219. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lawrencecchen:
Gates green on the exact head bfc54b0b7a8e: https://github.com/manaflow-ai/cmux/actions/runs/33389932614 (agent_signals) and https://github.com/manaflow-ai/cmux/actions/runs/33389935716 (screen_detect). The base branch carries ~19 pre-existing red hook-fence tests from its rebase (fail at the base commit without this PR); flagged in the PR body trade-off 11.

PR #11267, opened 2026-08-31
iOS Settings: per-Mac Connection section and explicit Show Missing Files toggle

azooz2003-bit:
Settings clarity round from dogfood feedback, three parts.

**Show Missing Files (Display).** The toggle said "Show missing files" with no hint of what feature it controls. Now titled Show Missing Files with a caption: "In a workspace's Files list, keep files that were deleted or moved instead of hiding them." (`includeMissing` in the artifact files listing.)

**Connection section.** The old section listed connected Macs with Focused/Ready labels plus one global "Active Transport" row; several Macs hold live connections at once and each dials its own configured method, and focused/ready is internal plumbing the user should not see. Each connected Mac is now a plain row whose trailing text is the transport that connection actually dialed (new `MobileMacConnectionSnapshot.routeKind` filled from the live `CmxAttachRoute`), navigating into that computer's detail. An All Computers row opens the Computers screen: the root presentation state machine allows the Settings sheet to swap to Computers in place, and the workspace-list host reuses the generalized settings-dismissal handoff.

**Networking restructure.** The Networking screen is scoped to app-wide relay configuration only: Relay Preference with managed relays, Custom Relays, the relay-policy refresh, Never Use Relays, and the DEBUG transport-mode section. Per-Mac concerns moved into each computer's detail: a Private Addresses section (that Mac's iroh path hints — add/edit/remove/enable pinned to its identity, reusing the existing editor) and the Connection Check section. The verbose connection log toggle, Share Verbose Log, Share Safe Report, and Clear Report folded into Settings' top-level Diagnostics section next to Share App/Network Log. `MobileIrohSettingsModel.observe` gained a quiet mode so embedded hosts don't record fake Networking-screen visits; `MobileIrohPrivateNetworksSection` is deleted. Localization keys orphaned by the removed status rows are left in the catalog intentionally (inert; removal would churn ~70 entries).

Localization: en + ja updated for the retitled toggle, caption, All Computers, section footer, the previously missing `mobile.settings.connection` header, and the new per-Mac Private Addresses strings; unused `connectionReady`/`connectionFocused` keys removed.

HIG pages consulted: Settings (toggle labels, footers for non-obvious effects, rows that navigate showing current value trailing).

Tests: `MobileRootPresentationStateTests` gains settings-to-computers swap + non-settings refusal; `MobileMacConnectionPoolTests` expectations carry `routeKind` (2/3 pass on fleet; `lateAnonymousIdentityRegistersFocusedConnection` fails identically on pure main on the same box — environmental, control-proven).

Sim-verified end to end on the isolated tag simulator: per-Mac Connection rows with trailing transport pushing the computer detail, All Computers swap, Show Missing Files caption, trimmed Networking screen, per-Mac Private Addresses + working Connection Check in the detail, and the folded Diagnostics section.

Dictionary:
- transport: the network path a phone-to-Mac connection dialed (Iroh, Tailscale, WebSocket).
- iroh path hints: extra per-Mac addresses the Iroh dialer may try before falling back to a relay.
- safe report: the privacy-scrubbed connection timeline export (no addresses, identities, or credentials).

## Summary by CodeRabbit

* **New Features**
  * Settings now lists connected Macs with their connection type and provides access to individual computer details.
  * Added an “All Computers” shortcut and smoother navigation from Settings to the Computers screen.
  * Computer details now support per-Mac private network addresses and connection checks.
  * Diagnostics now includes verbose logging, report sharing, and report clearing with confirmation.
  * Added relay-policy refresh controls.
* **Bug Fixes**
  * Improved transitions between Settings, Computers, and migration screens without unnecessarily dismissing the sheet.
* **Localization**
  * Added English and Japanese translations for new computer and settings controls.


azooz2003-bit:
Review triage for head c0c2f9f431:

Fixed: Computers now preempts the auto-connect migration introduction in the root presentation slot (regression test added), the All Computers row renders even with zero live connections, the private-address toggle/remove disable while a save is in flight, both embedded iroh-model hosts restart observation on reappearance and check cancellation after the detached log load, connection-row accessibility identifiers carry the app-instance identity, and the workspace list's two Computers entry paths share one helper.

Not fixed, with reasons: the connection row deliberately does not call `switchToMac` — surfacing focus roles in Settings was explicitly rejected in dogfood, and rows navigate to the computer detail instead. The Connection Check in the detail is the app-wide check relocated as-is; scoping its Mac-discovery stage to the row's computer needs a controller API change and is a follow-up, not a regression. The private-path editor disabling Save while the Mac is offline (`supportsPrivatePaths == false`) is pre-existing editor behavior identical to the removed Networking-screen path.

azooz2003-bit:
recheck

PR #11282, opened 2026-08-31
ios: terminal scroll edge effect via top overscan band

azooz2003-bit:
The iOS terminal now gets a scroll-edge treatment under the navigation bar: the surface extends under the (glass) bar and renders extra scrollback rows there — the new fork render-top-inset band — and a screen-anchored fade dissolves those rows into the terminal background as they approach the top, the way native lists read under Liquid Glass bars. The grid, PTY size, and keyboard system stay exactly where they were.

Depends on fork PR https://github.com/manaflow-ai/ghostty/pull/212 (`ghostty_surface_set_render_top_inset`): the drawable grows upward by the inset, the renderer fills the band with the rows directly above the viewport (clamped to history; empty on the alternate screen), and the shift folds into the existing pixel-scroll translation so every primitive composes correctly during scrolling. **Do not merge before the fork PR lands and the submodule pointer is an ancestor of fork `main`.**

cmux side: on iOS 26+ the terminal leaf ignores the top container safe area and the captured inset (read via `onGeometryChange` OUTSIDE the expansion — inside it reads 0) flows to `GhosttySurfaceView.setTopContentInset`; the viewport coordinator reserves the band so the grid starts exactly where it did before (`TerminalLetterboxGeometry` gains `topContentInset`); the render layer grows upward by the pixel-aligned applied inset so presents keep matching the drawable. The nav bar drops its opaque themed fill for system glass on the terminal surface only; pre-iOS-26 keeps the opaque bar and below-bar layout unchanged.

Why not the system scroll edge effect: UIKit renders it on the tracked content scroll view's OWN content subtree (the workspace list gets it because its cells live inside the table). The terminal's pixels are a Metal layer outside any scroll view, so no `setContentScrollView` association or stand-in can treat them — verified on-device (an on-window, bar-overlapping, frozen-mid-scroll stand-in renders nothing; the pill backdrops and the status-bar protection gradient are the only system treatments). The band's fade is therefore a `CAGradientLayer` in `GhosttySurfaceHostView`'s keyboard-invariant chrome space (the dock's home), sized by the live inset, colored from the live theme through `updateTerminalBackground`, and pinned to the screen so it holds while the render slides for the keyboard. It is a fade, not a glyph blur; a true progressive blur is not reachable for non-scroll-view content with public API.

Also fixes GhosttyKit source builds on fleet Macs without gettext: `-Di18n=false` in `ensure-ghosttykit.sh` (cmux never bundles the `.mo` output; runtime lookups are comptime-gated to the msgid).

Tests: fork render-state band tests + Size inset tests (zig, green); `TerminalLetterboxGeometry` band reservation tests and a viewport-coordinator placement test (swift, suite green). Localization audit: no new user-facing strings.

HIG: consulted the Human Interface Guidelines on scroll views and materials plus the `scrollEdgeEffectStyle`/`UIScrollEdgeEffect`/`UIScrollEdgeElementContainerInteraction` references (content extends under bars; the edge treatment provides bar legibility). The HIG page bodies did not load in this sandbox (title-only fetches); the design follows the API documentation semantics. Deliberate deviation: the fade approximates the system's progressive blur because the native effect cannot be applied to non-scroll-view content.

Verified on the isolated simulator `cmux-dev-sedge` (iPhone 17, iOS 27) and on-device: band rows render under the glass bar, track scrollback position continuously, clear on the alternate screen, and keep flowing under the bar through keyboard slides; the grid/PTY size is byte-identical to the pre-band layout. The round-5 screen-anchored fade is installed on-device for dogfood (its geometry is exercised by the same band path; visual confirmation pending dogfood).

## Summary by CodeRabbit

* **New Features**
  * Added iOS scroll-edge glass support for active terminal workspaces.
  * Terminal content can extend beneath the navigation bar and dock with smooth fade effects for readability.
  * Improved safe-area and viewport handling to preserve terminal layout while displaying overscan content.
  * Navigation chrome now adapts its appearance to terminal scroll-edge content.

* **Bug Fixes**
  * Kept browser and chat workspace behavior unchanged while improving terminal presentation.
  * Improved terminal rendering around navigation bars, dock areas, and keyboard transitions.


azooz2003-bit:
Re the mount-ordering finding on `terminalCapturedTopInset`: the transient zero-inset pass is absorbed by the existing viewport-report settle debounce — capacity reports only fire once the grid stops changing across frames (`pendingViewportReport`/`viewportReportSettleFrames`), which is the same mechanism that already gates keyboard-transition geometry churn. The band math keeps the grid invariant (bounds grow by exactly the reserved inset), so the transient never reaches the PTY unless it persists across the debounce window. Verified across six on-device rounds with no double negotiation or reflow flashes. The SwiftLint private-state nit is the established split-file extension pattern in this view (state accessed from WorkspaceDetailView+DerivedState.swift).

PR #11286, opened 2026-09-01
devbox image: sync cmux-bashrc body with the Blaxel template

lawrencecchen:
https://github.com/manaflow-ai/cmux/pull/11277 added ble.sh tput-cache seeding to Blaxel but left the shared devbox image behind. This PR restores the shell lockstep and bakes the seed caches into the E2B and Daytona Dockerfile plus the Freestyle replay. First PTY prompts no longer need to generate and print the cache update.

Current main also updated Blaxel tool pins in https://github.com/manaflow-ai/cmux/pull/11293. This branch now keeps all five agent pins and CUA driver 0.23.2 in lockstep across Blaxel, devbox, and Freestyle. The devbox image epoch is 2026-09-01-r1.

Validation: vm-devbox-image.test.ts passes 18/18, web typecheck is clean, and the extracted Freestyle devshell command passes bash -n. The earlier local amd64 image build proved all four test -s cache guards. Exact-head local autoreview is clean. The pre-existing non-atomic seed-copy behavior is tracked at https://github.com/manaflow-ai/cmux/issues/11348.


lawrencecchen:
recheck

PR #11319, opened 2026-09-01
vm: report-only orphan-volume and stuck-provisioning reporter

lawrencecchen:
Hourly cron (/api/cron/vm-reaper, :49) that reports, and only reports, the two cost-leak populations: orphaned machine-owned Blaxel volumes and VM rows stuck in provisioning. It deletes nothing and mutates no row state.

This is a deliberate descope. Two adversarial review rounds found five P1 defects each in the automated deletion and recovery machinery (deletion races against in-flight creates, unknown attachment states, stale observation reuse, unbounded scans, batch starvation). Safe automated deletion needs stable provider volume ids, observation epochs, and durable pagination cursors, designed as their own change. Until then, the reports give operators exact candidates for supervised cleanup.

Mechanics: bounded paginated provider scan with a partial-coverage flag, unknown attachment reported as unknown and never treated as free, usage events per candidate, and a cloud_vm_reaper_run PostHog summary per run (production-only, bounded 2s). maxDuration 300, CRON_SECRET-guarded.

Verification: bun run typecheck clean; vm-reaper suite 6 pass. Trade-off, stated: pre-existing orphan volumes keep billing until an operator acts on the reports; stuck rows are surfaced, not auto-recovered (the existing reconcile cron and the destroy-path fixes from PR 11251 stop new leaks).

Work was produced across interrupted codex runs and completed and reviewed by the session agent.

## Summary by CodeRabbit

* **New Features**
  * Added scheduled VM monitoring for stale provisioning records and orphaned storage volumes.
  * Added paginated volume inventory for improved coverage of large environments.
  * Added secure, privacy-preserving home-volume naming.
  * Added aggregate monitoring telemetry and JSON status reporting.
  * Monitoring now runs hourly on a scheduled basis.

* **Safety**
  * Monitoring is report-only and does not delete resources or modify VM records.
  * Added bounded scans, partial-coverage detection, duplicate-report prevention, and protected scheduled access.

### DIFF
--- a/web/app/api/cron/vm-reaper/route.ts
+++ b/web/app/api/cron/vm-reaper/route.ts
@@ -1,0 +1,24 @@
+import { captureVmReaperSummary } from "../../../../services/vms/observability";
+import { reapVmResources } from "../../../../services/vms/reaper";
+import { runVmWorkflow } from "../../../../services/vms/workflows";
+
+// Bounded batches keep a run well under this budget; the cap only guards a
+// hung provider call.
+export const maxDuration = 300;
+
+export async function GET(request: Request): Promise<Response> {
+  const secret = process.env.CRON_SECRET;
+  if (!secret || request.headers.get("authorization") !== `Bearer ${secret}`) {
+    return Response.json({ error: "unauthorized" }, { status: 401 });
+  }
+
+  try {
+    const summary = await runVmWorkflow(reapVmResources());
+    // Best effort and bounded; never turns a successful run into a 500.
+    await captureVmReaperSummary(summary);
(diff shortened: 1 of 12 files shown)


lawrencecchen:
Local autoreview's only remaining finding claims workflows.ts keeps an unused createHash import after the volume-naming move. False positive on this head: createHash is still used at web/services/vms/workflows.ts:2579 (token sha256 hashing), so the import is live; typecheck passes. The review otherwise reports the reaper as bounded and report-only with no actionable defects. Merging under the standing land-PRs directive.

PR #11341, opened 2026-09-01
iOS: constrain release-lane Macs to backend-served minimum versions

azooz2003-bit:
cmux iOS now refuses release-lane Macs older than a backend-served minimum, telling the user exactly which version to download.

**Backend.** `GET /api/mobile-mac-compat` (modeled on `/api/whats-new`: committed data literal in `web/data/mobile-mac-compat.ts`, module-load validation, ETag + CDN cache) maps iOS version tiers to per-channel Mac minimums. A tier is an inclusive `minIOSVersion` plus an optional inclusive `maxIOSVersion`: one open-ended tier captures every version from its minimum upward without listing each patch release, min+max scopes a bounded range, and min == max pinpoints one version. The initial tier starts at iOS 1.0.0 — covering the App Store lane (which ships as 1.0.0) and the 1.0.4 beta lane — and constrains them to the next stable release (`0.64.23`, i.e. anything newer than the current 0.64.22) and the next nightly (build counter > `3345650013201`, the newest nightly published at authoring time; nightly stamps are `<base>-nightly.<run id><attempt>`, so counters are monotonic).

**Fail-open by design.** An app version the server does not cover — below every tier, above the winning tier's bound, or served an empty entries list — gets NO Mac version limit at all: no limit beats accidentally admitting no Mac. Pinned by tests on both sides. A payload the client cannot fully parse is discarded (the previous policy stays), and the compiled-in `baked` fallback mirrors the committed web data, so devices enforce correctly before their first fetch and a bad remote edit cannot garble the constraint. Binaries built before this gate ignore the list entirely, so covering shipped versions is harmless.

**iOS enforcement.** `MobileMacCompatPolicy` (CmuxMobileShell) parses both version grammars, picks the tier for the running app version, and is enforced by `MobileShellComposite` at both authenticated-host admission points (connect route loop and `applyHostReportedIdentity`), alongside the existing instance-tag policy. Only release lanes are gated: `default` (and missing tags, which are pre-0.64.18 stable releases) against the stable minimum, `nightly` against the nightly minimum; dev/rc/staging lanes are exempt. A refused Mac surfaces the new `macAppVersionTooOld` pairing failure naming the Mac's version, the required version, and update guidance; the saved pairing stays intact and reconnects once the Mac updates.

**Fetch.** `MobileMacCompatCenter` mirrors `MobileWhatsNewCenter`: fetched at the app root, cached per API origin, pushed into the shell store.

**Onboarding.** The connect page now names the minimum for the running app version ("Requires cmux 0.64.23 or newer on your Mac.") in both the automatic and Tailscale bodies; the method cards drop their stale hardcoded 0.64.20/0.64.17 claims. HIG: consulted the Onboarding page (the page body did not render in my sandbox; applied its brevity principle: one short line, no extra step). Naming a companion app's minimum version deviates from "don't show requirements in onboarding" deliberately: the App Store cannot guarantee a separate Mac app's version.

**App Review note.** Once the next iOS build reaches App Review, the demo/review Mac must run cmux >= 0.64.23 or the reviewer will be refused; if that is ever a problem, editing `web/data/mobile-mac-compat.ts` (or serving an empty entries list) lifts the constraint remotely with no binary change.

**Dogfood.** DEBUG builds gate only under `CMUX_DEBUG_FORCE_MAC_COMPAT=1` (or launch arg `-CMUXDebugForceMacCompat YES`), evaluating the dev Mac by its reported version; fake an old Mac with `CMUX_DEBUG_MOBILE_APP_VERSION` on the Mac side.

**Localization audit.** New keys `mobile.pairing.macVersionTooOld.*` (4), `mobile.pairing.guidance.macVersionTooOld`, `mobile.onboarding.connect.bodyWithMinVersionFormat`, `mobile.onboarding.connect.tailscaleBodyWithMinVersionFormat`; updated `mobile.onboarding.connect.method.automaticDetail` / `tailscaleDetail`. All with en+ja in `ios/cmux/Resources/Localizable.xcstrings`.

**Tests.** `MobileMacCompatPolicyTests` (Swift Testing; runs in the existing test-ios `CmuxMobileShell` lane) covering stamp grammar, tier selection incl. bounded/pinpoint ranges, channel resolution, per-channel violations, payload decoding, fail-open semantics, baked fallback, and failure copy. `web/tests/mobile-mac-compat-route.test.ts` (17 bun tests) plus web typecheck.

## Summary by CodeRabbit

- **New Features**
  - Added Mac app version compatibility checks for iOS connections, with separate stable and nightly requirements.
  - Compatibility requirements can update remotely and remain available offline through cached or built-in policies.
  - Added clear connection guidance when the Mac app requires an update.

- **Improvements**
  - Onboarding displays the applicable minimum Mac version when required.
  - Simplified connection-method descriptions by removing outdated version-specific messaging.
  - Added localized messaging for Mac version compatibility failures.


azooz2003-bit:
Review findings addressed in 1afc7989d7 (deterministic cached-policy install at store construction, version-gate stash resolved in the single failure sink, served version grammar restricted to the client's 1-3 components, required nightly build named in refusal copy, versionless Tailscale fallback, shared remote-JSON loader).

Two findings are declined with reasons:

- Nightly minBuild digit-count (cubic P1): the premise is wrong. nightly.yml sets NIGHTLY_BUILD="${GITHUB_RUN_ID}${RUN_ATTEMPT}" and names the DMG cmux-nightly-macos-${NIGHTLY_BUILD}.dmg, so the published asset numbers (e.g. 3345650013201) are already the FULL counter including the 2-digit attempt — the same value stamped into CFBundleShortVersionString. The floor 3345650013202 compares like-for-like and correctly means "any nightly published after authoring".
- Empty-entries payload lifting the gate (cubic P2 / CodeRabbit Major): deliberate product decision. An app version the server does not cover must get NO Mac version limit; serving an empty list is the remote kill switch, and accidentally blocking every Mac is the failure mode we chose to avoid. Documented at MobileMacCompatPolicy.decode and pinned by tests on both sides.

azooz2003-bit:

azooz2003-bit:
recheck

PR #11342, opened 2026-09-01
iOS: deterministic App Store screenshot pipeline (listing parity)

azooz2003-bit:
Regenerating the App Store screenshots after UI changes was a manual, live-session pipeline (real paired Mac, real agent runs, hand-driven captures). This makes the whole set reproducible from in-app DEBUG preview fixtures, so any release can re-shoot the current UI deterministically.

**What the live listing has today** (app 6783338052, v1.0.0, en-US): 8 full-bleed iPhone 6.9" shots (Claude / Codex / workspaces / notifications / lock-screen inline reply / per-file diff / OpenCode / composer), 3 full-bleed portrait iPad 13" shots, and 2 July-era framed marketing pages on the legacy 12.9" slot.

**Changes**
- `SnapshotUITests` mirrors that roster 1:1: adds notifications-feed, per-file diff (`CMUX_UITEST_CHANGES_PREVIEW=diff`, the `Sources/SessionStore.swift` fixture the live shot came from), and Task Composer shots; numbering matches listing order; iPad captures are portrait full-bleed by default (`SNAPSHOT_IPAD_LANDSCAPE=1` restores landscape for framed compositions).
- `ScreenshotNotificationPresenter` gains a `reply` fixture mode (`CMUX_UITEST_NOTIFICATION_BANNER=reply`): merges a text-input Reply category into the registered set and schedules the agent-finished notification with a 6s fuse so the harness can lock the simulator and capture the lock-screen inline-reply UI. Strings localized (en + ja).
- `ios/fastlane/appstore-shots-plan.json`: truth manifest — display types, exact ASC pixel sizes, per-shot source captures and transforms, ASC's 10-per-type cap enforced by verify.
- `ios/scripts/appstore-shots.sh` + `appstore_shots_post.py`: orchestrator. `capture` dispatches the hosted `ios-screenshots.yml` run and downloads raw captures (capture never runs as local `xcodebuild test`, per repo policy); `lockshot` drives the lock-screen reply shot with simctl/axe/idb; `post` stages `final/<DISPLAY_TYPE>/` at exact ASC dims from a clean tree (iPhone Lanczos fit, iPad status-bar crop + 3:4 re-crop, framed 12.9 pages); `verify` structurally checks the staged set; `upload` resolves the plan-locale localization and previews with a real `asc --dry-run` (`--confirm` to execute, `--replace` for full-set refreshes); `asc-snapshot` snapshots the live listing.
- **Heals the broken iOS test lane**: `cmuxFeatureTests` had not compiled on main for days, which broke every `xcodebuild build test` of the cmux-ios scheme including this screenshots lane. Fixed four layers of rot: iroh broker test doubles missing `revokeStale`/`forgetMac`, a 2-arg `brokerFactory` closure (now 3-arg), the stale `TerminalRowCapacityFitTests` duplicate whose `fitFontSize`/`shouldRefit` API was deleted with terminal auto-zoom in #10616 (canonical tests live in `CmuxMobileTerminalKitTests`), and `SignOutFixture.deviceID`/actor-isolation fixes.
- Frame lane: titles remapped to the new roster in every locale (new screens get en+ja titles, others fall back to English); `propagate_locales` tolerates partial-language captures; empty titles compose without a header instead of crashing magick.

**Verified**
- CI run [33490590174](https://github.com/manaflow-ai/cmux/actions/runs/33490590174) is fully green end to end on this branch: build → 10 shots × iPhone 17 Pro Max + iPad Pro 13" (M5) → collection → locale fan-out → framing → artifacts.
- `post` + `verify` on the real CI captures stage all 13 finals at exact ASC dimensions (1320x2868 / 2064x2752 / 2048x2732); the diff shot is near-identical to the live listing (same fixture).
- `upload` dry-run against the live listing resolves the existing set ids and prints the exact append/replace plan; nothing was uploaded.
- Lock-screen shot: banner leg proven on an isolated iPhone 17 Pro Max simulator (matches the live shot's composition); the typed-reply leg needs a build containing the new `reply` category, so it stays best-effort until this merges (`post --skip-lockshot` stages the other 12).

**Known deviations from the live set** (called out in the plan): the iPad 13" page-1 sidebar+session split has no preview fixture yet (workspace list stands in); the 12.9" framed pages embed portrait captures because simulators cannot be rotated headlessly; session content is the rig's recorded transcripts, re-recordable via `ios/fastlane/frame_assets/record_sessions.sh` if exact live-content parity is wanted.

HIG: notification fixture uses the standard system notification + text-input action pattern (HIG "Notifications" page, https://developer.apple.com/design/human-interface-guidelines/notifications); no custom UI drawn.


azooz2003-bit:
Review findings addressed in b05f0baae7 and earlier commits:

- ASC localization is now resolved by the plan's locale, never `data[0]` (cubic P1, CodeRabbit locale-propagation); `post` stages captures only from the plan locale's directory.
- `post` clears `final/` before staging, so reruns and `--skip-lockshot` can never leave a stale screenshot for verify/upload to accept (cubic P1, CodeRabbit).
- `--device-type` maps from the API display-type key by dropping the `APP_` prefix; validated against the live listing with `asc screenshots upload --dry-run`, which resolves the existing set ids and prints the full delete+upload plan (cubic P1). `--replace` is an explicit opt-in; default remains append + dry-run preview.
- `capture --ref` rejects SHA values since run resolution is branch-based (cubic P1).
- `identify` works on convert-only ImageMagick installs (cubic P2).
- The reply fixture merges its category with existing registrations instead of replacing the set, and the reply field has the production placeholder (cubic P2, CodeRabbit); xcstrings entries carry `extractionState: manual` (CodeRabbit, cubic P3).
- The `testplan`+`only_testing` combination is NOT rejected by xcodebuild (run 33479247651 executed it), but `-testPlan` did break fastlane's screenshot collection (zero files vs the green run 32219546960), so the testplan experiment is reverted and the underlying compile rot in `cmuxFeatureTests` is fixed instead (cubic P1 partially refuted, root cause addressed).
- Not changed: the lockshot helper's fixed delays (CodeRabbit). It drives springboard's lock screen, which exposes no readiness signals to simctl/axe; this is a capture harness, not runtime app code, and it fails loudly by producing a wrong screenshot the operator inspects before staging.

azooz2003-bit:
Real-content follow-up pushed in d443eb1393e. Added appstore-shots-real.sh for a signed-in, paired simulator: it captures live Claude, Codex, OpenCode, workspace list, file diff, notifications, and Composer content; pastes and verifies the exact prompt; opens the real repository picker and fails on error/unavailable states; and captures the system inline-reply screen. post --capture-source real now rejects fixture raws and requires the real-marked lockshot. Local real set restaged and verify passed: iPhone 8/8 at 1320x2868, iPad 3/3 at 2064x2752, framed 2/2 at 2048x2732. Nothing was uploaded to ASC.

azooz2003-bit:
Follow-up beb8342e2b4 aligns the real lock-screen leg with the current ASC reference: light SpringBoard appearance, charging status glyph, live system notification, and expanded inline reply with the exact reply text. Re-staged real set and verify remain green at all required dimensions. No ASC upload.

azooz2003-bit:
Final capture hardening in 46f1b796678: the real driver falls back to SpringBoard's swipe-to-View path when HID long-press does not expose actions, clears between text-injection attempts, and verifies the inline reply AXValue exactly before saving. The captured light ASC-shaped reply was restaged and dimension verification passed.

azooz2003-bit:
Release integration added in 0d841e342c3. ios-screenshots.yml is now reusable and remains manually dispatchable. Every release.yml tag run calls it at the exact release ref with en-US + ja, uploads raw/composed screenshot artifacts, keeps ASC upload disabled, and gates build-sign-notarize on screenshot success. Both changed workflows pass actionlint.

azooz2003-bit:
Follow-up in bd74ec792f5: release and manual screenshot runs now capture all 9 localized app languages (en-US, de-DE, fr-FR, ar-SA, es-ES, zh-Hant, zh-Hans, ko, ja). The fan-out maps regional App Store slots to the matching localized raw and uses English only for listing locales without an app translation. It clears stale PNGs, validates source sets match en-US, and remains deterministic. Dispatched full-language validation run: https://github.com/manaflow-ai/cmux/actions/runs/33833265938.

PR #11344, opened 2026-09-01
billing: report-only team seat drift detection

lawrencecchen:
Detects and reports team seat drift without mutating anything. For every team-scope subscription row in the reconcile batch whose REMOTE status is active, the job reads the Stack roster (15s deadline-raced; the SDK's listUsers takes no AbortSignal, so the request cannot be cancelled — on Vercel the leak is bounded by the invocation lifetime) and, when member count disagrees with the Stripe quantity or the stored seats, emits one cmux_billing_team_seat_drift_detected event carrying member_count, stripe_quantity, and stored_seats with a unique per-detection insert id. Remote subscription metadata must agree with the local row's team identity, failing closed on drift.

This is a deliberate descope. Three review rounds found design defects in the mutating version (shared-cursor starvation of unreserved teams, pre-apply seat snapshot staleness, uncancellable roster reads compounding). Reporting first measures the underbilling population; enforcement (truing the Stripe quantity) follows as its own designed PR with a dedicated cursor once the drift data exists. Until then teams below their member count remain underbilled, stated plainly.

Verification: bun run typecheck clean; 24 tests pass across the reconcile and analytics suites, including assertions that no Stripe update and no seats write ever occur.

Work was produced across interrupted codex runs and completed and verified by the session agent.

### DIFF
--- a/web/services/analytics/stripeBilling.ts
+++ b/web/services/analytics/stripeBilling.ts
@@ -83,7 +83,10 @@ export async function captureBillingTeamSeatSync(
 ): Promise<void> {
   await captureBillingPayload({
     name: "cmux_billing_team_seats_synced",
-    insertId: `team-seat-sync:${input.subscriptionId}:${input.oldQuantity}:${input.newQuantity}`,
+    // A repeated legitimate transition (1->2->1->2) must not deduplicate the
+    // later event, so each sync gets a unique id; transport retries inside
+    // this call still share it because the payload is built once.
+    insertId: `team-seat-sync:${input.subscriptionId}:${globalThis.crypto.randomUUID()}`,
     subject: { scope: "team", stackTeamId: input.teamId },
     properties: {
       source: "billing_reconcile",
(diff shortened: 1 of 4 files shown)


lawrencecchen:
Parking this as a draft. Six review rounds across three design shapes (mutating sync, report-only detection, observe-before-apply) each produced new structural findings rooted in one cause: seat observation does not fit inside the shared reconcile loop and cursor. Latest round: the identity rule skips inactive subscriptions, status repair became dependent on an optional roster read, and the deadline breaker does not enforce its limit under concurrency. The next attempt should be a dedicated cron with its own cursor, a per-run Stack budget, and observation fully decoupled from repair. The branch keeps the working pieces: fail-closed identity checks, deadline-raced roster reads, unique drift events, and the test harness.

lawrencecchen:
Fleet instruction update for head `6dc41b42f49513e9ee0e3466c712501621727e66`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #11392, opened 2026-09-01
perf(cmux-tui): cache normalized sidebar file names

lawrencecchen:
Repeated filter keystrokes lower-case every directory entry on every pass, allocating one String per entry. Cache each entry’s normalized name when the directory is loaded, then reuse it for sorting and filtering.

Regression coverage preserves Unicode case-insensitive matching and visible names.

Checks: rustfmt --check on changed file; git diff --check. Rust tests/builds require hosted Blacksmith workflow.

---
## Summary by cubic
Speeds up sidebar filtering so case-insensitive matches no longer allocate a lowercase String per entry on every keystroke.

- Reuses one buffer for ASCII names, keeping the hot path linear; Unicode names fall back to per-entry lowercase.
- Adds regression tests for Unicode case-insensitive matching and long ASCII near-misses; visible entry names stay unchanged.

<sup>Written for commit 1712e362cc30868b6358d65a4f1e0d3b22682f7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11392?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved case-insensitive file filtering and sorting, including support for Unicode filenames.
  * Directory listings now provide more consistent filename matching and ordering.

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/sidebar_files/files.rs
+++ b/cmux-tui/crates/cmux-tui/src/sidebar_files/files.rs
@@ -67,12 +67,27 @@ fn compare_entries(left: &FileEntry, right: &FileEntry) -> Ordering {
 }
 
 pub fn filtered_indices(entries: &[FileEntry], query: &str) -> Vec<usize> {
-    let query = query.to_lowercase();
+    let normalized_query = query.to_lowercase();
     entries
         .iter()
         .enumerate()
-        .filter_map(|(index, entry)| entry.name.to_lowercase().contains(&query).then_some(index))
+        .filter_map(|(index, entry)| {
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Superseded by #11409, which explicitly rebases and replays this change on current main with the same files and commits. Closing the older divergent branch.

PR #11424, opened 2026-09-01
fix(cmux-tui): join machine action worker on shutdown

lawrencecchen:
MachineActionWorker previously dropped its thread handle while a provider action was active, detaching controller cleanup. Shutdown now signals cooperative stop and always joins the worker after the in-flight action returns. The regression test releases the blocked action before shutdown and verifies controller close.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11424"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790849078&installation_model_id=21920&pr_number=11424&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11424&signature=43d462a9751e170a614b6cf7efe9ed283d0c7aa33e955830b9cc19c3d2e21e09"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Fixes `MachineActionWorker` shutdown so controller cleanup can't be detached when a provider action is in flight. Shutdown previously dropped the worker handle while an action was active; it now signals cooperative stop and hands the handle to a reaper thread that joins the worker, keeping shutdown responsive. If the reaper thread can't start, the worker is joined synchronously instead of being dropped. The reaper handle is stored so a subsequent shutdown joins it once the action returns, and the owner's `Drop` joins it as well.

- Regression test confirms shutdown returns quickly, releases the action, verifies controller close, and calls shutdown again to join the reaper.

<sup>Written for commit c24da2a825134bf25744580f64e2eee000a9e3db. Summary will update on new commits.</sup>

[...]

PR #11416, opened 2026-09-01
fix(cmux-tui): cancel blocked machine completion sends

lawrencecchen:
Replace the 1 ms retry sleep used when forwarding machine-controller completions to the bounded app channel with the existing cancellation-aware send helper. Machine action shutdown now cancels the helper, so a full app queue wakes immediately and preserves the existing stop path.

Regression coverage uses two commits: test first, fix second.

Verification: hosted focused cmux-tui test dispatched with `./scripts/verify-cmux-tui-hosted.sh --filter canceling_machine_controller_completion_send_unblocks_when_queue_is_full`.

---
## Summary by cubic
Fixes a shutdown stall in the cmux-tui machine action worker by replacing the 1 ms retry sleep with the cancellation-aware send helper. When the app event queue is full, shutdown now wakes the blocked completion send immediately instead of waiting up to 1 ms per retry.

- Machine action shutdown cancels the send helper before stopping the worker.
- Added regression coverage for a full queue blocking completion sends until cancellation.

<sup>Written for commit 383bf69d50d05def6edad60bf694ff28daa02d96. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application shutdown responsiveness when event queues are full.
  * Prevented background actions from remaining blocked while waiting to deliver completion events.
  * Added cancellation handling to ensure workers stop promptly during shutdown.

PR #11479, opened 2026-09-01
cmux-tui: cancel completion sends and reap machine worker

lawrencecchen:
## Summary

- Replace the machine completion sender's 1ms retry loop with the existing cancellation-aware bounded send.
- Cancel that send during worker shutdown so a full app-event queue cannot retain a blocked worker.
- Keep the worker JoinHandle owned through an asynchronous reaper, join the reaper on owner drop, and synchronously join if reaper creation fails.

## Testing

- `rustfmt --edition 2024 --check cmux-tui/crates/cmux-tui/src/app.rs`
- `git diff --check`
- Local Cargo and Rust tests were not run per cmux-tui worktree policy.
- Hosted filters: `canceling_machine_controller_completion_send_unblocks_when_queue_is_full`; `machine_action_worker_shutdown_joins_after_cooperative_cancellation`.

## Related

- https://github.com/manaflow-ai/cmux/pull/11416
- https://github.com/manaflow-ai/cmux/pull/11424

Base is current `origin/main` at `6b425641ae4d474e77854da535442af2a0d0a475`.

---
## Summary by cubic
Fixes machine action worker shutdown so a full app-event queue no longer keeps the completion send blocked, and dropping the worker no longer waits for an in-flight provider call.

- Replaces the completion sender's 1ms retry loop with a cancellation-aware send, canceled during shutdown so the worker unblocks.
- Discards a queued command that is dequeued after shutdown, so a queued action no longer runs after stop.
- Spawns an asynchronous reaper to join the worker; joins synchronously when the reaper thread can't spawn, and joins a finished reaper on later shutdown calls.
- Dropping the worker detaches an unfinished reaper, so drop returns immediately and the provider call finishes in the background.
- Adds regression tests for cancellation unblocking, cooperative shutdown join, queued-command shutdown, and responsive drop.

<sup>Written for commit 7888a600338bd1790a736783ea46770789851ae8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11479?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown behavior for background operations, allowing them to stop cooperatively without blocking the interface.
  * Prevented hangs when completion queues are full or an operation is waiting for a provider response.
  * Ensured queued work is discarded after shutdown begins and cleanup completes safely when components are released.
  * Made shutdown safe to invoke repeatedly.

* **Tests**
  * Added coverage for cancellation, queue unblocking, repeated shutdown, discarded work, and non-blocking cleanup.

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/app.rs
+++ b/cmux-tui/crates/cmux-tui/src/app.rs
@@ -8641,7 +8641,8 @@ impl MachineActionWorker {
         self.sender.take();
         // The stop flag is cooperative: an in-flight provider action returns
         // at its transport deadline, then the worker observes stop and closes
-        // the controller. Join so provider cleanup cannot outlive the owner.
+        // the controller. Transfer the worker handle to the reaper so owner
+        // teardown does not wait for that provider deadline.
         if let Some(worker) = self.worker.take() {
             // Provider calls are synchronous and may remain blocked until
             // their transport deadline. Reap the owned handle asynchronously
@@ -8677,8 +8678,16 @@ impl MachineActionWorker {
 impl Drop for MachineActionWorker {
     fn drop(&mut self) {
         self.shutdown();
+        // The reaper owns the worker handle. Join only when it has already
+        // finished; dropping an unfinished reaper handle intentionally lets
+        // that cleanup thread continue without blocking owner teardown. The
+        // reaper still joins the worker and calls controller.close() after the
+        // provider's bounded transport call returns, so no worker handle is
+        // detached or leaked.
         if let Some(reaper) = self.reaper.take() {
-            let _ = reaper.join();
+            if reaper.is_finished() {
+                let _ = reaper.join();
+            }
         }
     }
 }
--- a/cmux-tui/crates/cmux-tui/src/app.rs
+++ b/cmux-tui/crates/cmux-tui/src/app.rs
@@ -8386,6 +8386,9 @@ impl MachineActionWorker {
                     if let Some(receive_hook) = receive_hook.as_ref() {
                         receive_hook();
                     }
+                    if worker_stop.load(Ordering::Acquire) {
+                        break;
+                    }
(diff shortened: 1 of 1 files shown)


lawrencecchen:
Closing after an exact review against current `main`.

The completion-send change is correct, but its value is small. The old loop checks `stop` before every 1 ms wait. It cannot retain TUI shutdown. The new cancellation path removes at most 1 ms of detached cleanup latency and avoids polling only while the event queue is full.

The reaper does not restore owner-scoped cleanup. `MachineActionWorker::drop` drops an unfinished reaper handle. The reaper then continues after its owner is gone and discards the worker join result. The fallback also joins the worker synchronously if reaper creation fails. A provider request can wait for 30 seconds, and an open request can wait for 5 minutes.

Tokio has the same boundary. A dropped `JoinSet` aborts tracked async tasks, but a started `spawn_blocking` task cannot be aborted. `CancellationToken` only signals cooperative cancellation. See https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html, https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html, and https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html.

A principled fix must make `MachineController::perform` interruptible, or give the owner a thread-safe handle that closes provider I/O. Shutdown can then cancel and join the worker without a reaper. The small completion-send cleanup can be a separate two-commit PR if it is still wanted.

No Rust build or test ran during this review.

PR #11480, opened 2026-09-01
perf(cmux-tui): index machine sidebar metadata

lawrencecchen:
## Summary
- index managed machine metadata once per sidebar frame
- preserve provider row order and first-match behavior for duplicate keys
- reduce machine rail metadata lookup from O(M²) to O(M) per frame

## Verification
- `rustfmt --edition 2024 --check cmux-tui/crates/cmux-tui/src/ui/sidebar.rs`
- `git diff --check`
- Cargo/Rust compilation and tests not run, per task constraint

The implementation follows the standard library HashMap lookup model: https://doc.rust-lang.org/std/collections/struct.HashMap.html
Ratatui rendering remains in the existing frame path: https://docs.rs/ratatui/latest/ratatui/widgets/trait.Widget.html

---
## Summary by cubic
Speeds up the `cmux-tui` machine sidebar by indexing managed machine metadata once per frame, replacing per-row scans with `HashMap` lookups and reducing frame cost from O(M²) to O(M). Machine rows still render in provider order, and duplicate keys keep first-match behavior; the stable key is cloned when building the index.

<sup>Written for commit 9a16961af1c0e2f5255aa1b0a2b96c3dbbee7220. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11480?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Performance**
  * Improved sidebar machine list rendering for faster lookups.
  * Ensured duplicate machine entries are handled consistently, displaying only the first occurrence.

### DIFF
--- a/cmux-tui/crates/cmux-tui/src/ui/sidebar.rs
+++ b/cmux-tui/crates/cmux-tui/src/ui/sidebar.rs
@@ -95,7 +95,7 @@ pub fn draw_machines(app: &mut App, frame: &mut Frame) {
     for managed in machine_ui.managed_machines() {
         // Keep the previous first-match behavior if malformed input repeats a
         // key, while normal snapshots remain one entry per stable key.
-        managed_machines.entry(managed.key).or_insert(managed);
+        managed_machines.entry(managed.key.clone()).or_insert(managed);
     }
     let rail_selection = machine_ui.rail_selection;
     let palette = rail::RailPalette::for_app(app, app.machine_sidebar_focused());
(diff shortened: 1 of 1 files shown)


lawrencecchen:
Closing because this PR has no benchmark or profile evidence that the machine-sidebar scan is a measured bottleneck. It builds and allocates a new HashMap on every rendered frame. The managed metadata already has a stable update boundary in [`MachineUiState::set_managed_machines`](https://github.com/manaflow-ai/cmux/blob/0247f51a1f30308df595606d0951c802ec038550/cmux-tui/crates/cmux-tui/src/machine.rs#L837-L847). If measured data later shows a real cost, keep a keyed index in `MachineUiState` and rebuild it when `set_managed_machines` receives a new snapshot, so frame rendering only performs lookups. No behavior fix is lost by closing this optimization-only patch.

Issue #7525, opened 2026-07-07
iOS: signed out on every app update; sign-in takes multiple attempts (session persistence + sign-in reliability)


austinywang:
## Symptoms (dogfood report, 2026-07-07)

- Every cmux iOS app update signs the user out. The session should persist across updates.
- Sign-in often takes ~3 attempts before the app actually reaches the signed-in state.
- Related: terminal connection flapping is tracked in https://github.com/manaflow-ai/cmux/issues/6311 and Stack auth on the attach hot path in https://github.com/manaflow-ai/cmux/issues/6151.

## Root causes found (code audit @ 7c5e7ba0e4)

Session persistence:
1. iOS uses the vendored Stack Auth SDK built-in `KeychainTokenStore` (`vendor/stack-auth-swift-sdk-prerelease/Sources/StackAuth/TokenStore.swift`), which sets no `kSecAttrService` and no `kSecAttrAccessGroup`. Tokens land in the default access group derived from the bundle id. Dev reload builds use a new bundle id per tag (`ios/scripts/reload.sh`: `BUNDLE_ID="dev.cmux.ios.$TAG_SLUG"`), so every new tag is a fresh keychain partition plus a fresh data container: guaranteed sign-out. No `keychain-access-groups` entitlement is pinned (commented out in `ios/Config/cmux.entitlements`), so no session sharing across builds is possible.
2. The hardened custom token stores in `Packages/Shared/CmuxAuthRuntime` (`KeychainStackTokenStore`, `FallbackTokenStore`) are wired only into macOS (`Sources/Auth/MacAuthComposition.swift`). iOS never got them.
3. `detectAuthProjectSwitch` (`ios/cmuxPackage/Sources/cmuxFeature/MobileAuthComposition.swift`, added in PR #7149) clears all persisted Stack tokens from a launch-time project-id comparison, including inferred defaults. Any dev/prod auth toggle across builds wipes the session even though keychain accounts are already project-scoped (`stack-auth-access-<projectId>`).

Sign-in reliability:
4. `SignInView` disables every sign-in control while `isRestoringSession` (launch with a stale `auth_has_tokens` flag and missing/undecodable `auth_cached_user`), so early taps do nothing (`Packages/iOS/CmuxMobileShellUI/Sources/CmuxMobileShellUI/SignInView.swift`).
[...]

PR #11564, opened 2026-09-02
iOS: launch with a cached session mounts the shell, not the sign-in loading screen

azooz2003-bit:
## Problem

Every launch of a signed-in iOS install showed the sign-in screen with the "Restoring session…" spinner until the Stack `/me` validation round trip finished (up to the 8s restore timeout), even on immediate relaunches.

## Cause

https://github.com/manaflow-ai/cmux/pull/11310 (commit "ios: keep sign-in visible during auth restore") routed the entire launch-restore window to `SignInView`: `MobileRootAuthGate.shouldShowSignIn` returned true whenever `isRestoringSession` was set, even with `stackAuthenticated == true`. A cached session always primes with `isRestoringSession: true` (`CMUXAuthState.primed`), so every warm launch parked on the loading sign-in surface. This contradicts the primed-state design: "A cached user with known tokens is treated as signed in immediately while runtime validation remains explicitly in progress. This lets the authenticated UI render."

## Fix

`shouldShowSignIn` now keeps sign-in only while actually signed out. A primed cached session mounts the workspace shell immediately; the shell's existing restoring inputs cover the validation window, and a definitive rejection still lands on sign-in when the coordinator clears the session.

The #11310 protections are preserved:
- onboarding still never presents during restore (`MobileOnboardingGate` unchanged), and
- while first-run onboarding is pending, a restoring launch stays on sign-in (new `onboardingPending` input) — mounting the pre-onboarding shell there would flash the add-device surface for a cached identity that may still be rejected.

## Commits

Two commits so CI proves the regression test: commit 1 adds the failing test only (red), commit 2 adds the fix (green). Verified locally: red without fix, 45/45 package tests green with it.

## Out of scope (noted during diagnosis)

- `reconnectStoredMacIfNeeded` still waits for restore to settle before the stored-Mac reconnect (intentional since #8299).
- A desynced `auth_has_tokens`/`auth_cached_user` pair (tokens in keychain, no cached user) still strands launch on sign-in until `/me` succeeds once; tracked by the pre-existing https://github.com/manaflow-ai/cmux/issues/7525.

## HIG

Checked https://developer.apple.com/design/human-interface-guidelines/launching (page body did not render in the sandboxed fetch; applied its core guidance: restore prior state and get people to content immediately, never park launch on a loading interstitial).

---
## Summary by cubic
Fixes a launch regression on iOS where every launch of a signed-in install showed the sign-in loading screen ("Restoring session…" spinner) until the Stack `/me` validation round trip finished, up to the 8s restore timeout.

`MobileRootAuthGate.shouldShowSignIn` now keeps sign-in on screen only while actually signed out. A primed cached session mounts the workspace shell immediately; the shell's existing restoring inputs cover the validation window, and a definitive rejection still lands on sign-in when the coordinator clears the session. The one exception: while first-run onboarding is pending, a restoring launch stays on sign-in so the pre-onboarding shell doesn't flash the add-device surface for a cached identity that may still be rejected. The old behavior, introduced in #11310, showed sign-in for the entire restore window.

Adds a regression test covering the primed cached session launch path.

<sup>Written for commit 9ecb02f97d7415fdf1b6493557c990a21d63e875. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11564?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iOS launch behavior for returning users with cached sessions, allowing the authenticated shell to appear while session validation is in progress.
  * Kept the sign-in screen visible when first-run onboarding remains incomplete.
  * Ensured attach-ticket users continue directly to the shell.

### DIFF
--- a/Packages/iOS/CmuxMobileShellUI/Sources/CmuxMobileShellUI/CMUXMobileRootView.swift
+++ b/Packages/iOS/CmuxMobileShellUI/Sources/CmuxMobileShellUI/CMUXMobileRootView.swift
@@ -496,7 +496,8 @@ struct CMUXMobileRootView: View {
         } else if MobileRootAuthGate.shouldShowSignIn(
             stackAuthenticated: authManager.isAuthenticated,
             attachTicketAuthenticated: hasActiveAttachTicketAuthentication,
-            isRestoringSession: authManager.isRestoringSession
+            isRestoringSession: authManager.isRestoringSession,
+            onboardingPending: isOnboardingPending
         ) {
             SignInView()
                 .transition(reduceMotion ? .opacity : .asymmetric(
@@ -828,6 +829,18 @@ struct CMUXMobileRootView: View {
             hasKnownPairedMac: store.hasKnownPairedMac,
             hasAccountMismatch: store.connectionRequiresReauth
         )
+    }
+
+    /// Whether first-run onboarding remains unfinished, i.e. it would present
+    /// once launch restore settles. While pending, a restoring launch stays on
+    /// the sign-in surface (see ``MobileRootAuthGate/shouldShowSignIn``);
+    /// once complete, a primed cached session mounts the shell immediately.
+    private var isOnboardingPending: Bool {
+        #if os(iOS)
+        return onboardingStore.progress != .complete
+        #else
+        return false
+        #endif
     }
 
     /// Whether first-run onboarding should present: only for a settled,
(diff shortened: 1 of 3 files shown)


azooz2003-bit:

azooz2003-bit:
recheck

PR #11567, opened 2026-09-02
fix(tui): bind socket lease to listener identity

lawrencecchen:
## Summary
- add a regression test for a pathname replaced before listener claim
- compare the pathname with the bound Unix listener fd before claiming
- fail closed during failed claim cleanup when the path is not this listener

## Verification
- `rustfmt --edition 2024`
- `git diff --check`
- Cargo/Rust builds and tests not run on the Mac per cmux policy

Parent PR: https://github.com/manaflow-ai/cmux/pull/11482

---
## Summary by cubic
Binds TUI socket leases to listener identity when available, preventing a replaced socket path from being removed during claim or stale cleanup. Unix and Windows keep claims usable when identity cannot be verified, but leave those publications in place instead of deleting them.

- `claim_bound` rejects a path that names a different Unix listener with `AddrNotAvailable`.
- Stale cleanup runs only after definitive `ConnectionRefused` or `NotFound` errors.
- Cleanup moves paths to atomic no-replace tombstones and restores them when their identity changes.
- Non-socket Unix paths use the same tombstone flow, while unsupported platforms fail closed.
- Adds regression coverage for replaced paths, stale-cleanup races, and unverified Unix and Windows listeners.

<sup>Written for commit e9f35b7a8d131adb54218dae9309e42c4c5e0d34. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11567?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmux-tui/crates/cmux-tui-core/src/platform.rs
+++ b/cmux-tui/crates/cmux-tui-core/src/platform.rs
@@ -54,6 +54,14 @@ pub mod transport {
         pub fn accept_with_wake(&self, wake: &ListenerWake) -> io::Result<Option<Box<dyn Stream>>> {
             self.inner.accept_with_wake(&wake.inner)
         }
+
+        /// Compare the pathname with this listener's bound object.
+        ///
+        /// `Some(false)` means the path names a different object. `None` means
+        /// this platform cannot expose a stable listener identity.
+        pub fn matches_path(&self, path: &Path) -> io::Result<Option<bool>> {
+            self.inner.matches_path(path)
+        }
     }
 
     impl ListenerWake {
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Superseded by #11521. The child PR retains the listener identity work and adds the accept-loop termination fix, so keeping both would duplicate the same changes. Reopen only if a unique requirement is identified.

PR #11586, opened 2026-09-02
Fix main CI after the Blaxel removal and non-root daemon landing

lawrencecchen:
`ci.yml` on `main` fails in `web-db-migrations`, `web-typecheck` (Web tests), `workflow-guard-tests`, and the gates that depend on them, since 4b4c0100e4 (non-root cloud VM terminals, pushed without a ci.yml run) and https://github.com/manaflow-ai/cmux/pull/11566. Reproduced on https://github.com/manaflow-ai/cmux/actions/runs/33600222487.

**Migration.** The remove-blaxel migration compares provider columns to the literal `'blaxel'`. On a fresh database drizzle applies every pending migration in one transaction, so the `ADD VALUE` from 20260820 is uncommitted and Postgres rejects it with `unsafe use of new value "blaxel" of enum type vm_provider`. Comparing `"provider"::text = 'blaxel'` avoids casting the literal to the enum. Verified: all migrations apply and re-apply on a fresh Postgres 16 and the rebuilt enum is `{e2b,freestyle,daytona}`. Behaviour on already-migrated databases (production) is unchanged.

**Supervisor grace helper (runtime fix).** `cmuxTuiDaemonCommand`'s grace-period helper is a background subshell (`sleep 2; kill -KILL daemon`) that the supervisor TERMs once the daemon has exited. dash, the Ubuntu `/bin/sh` on every devbox, forks that subshell with the parent's TERM trap still inherited and resets it afterwards, so a TERM that lands in that window is recorded as a pending trap and dropped. A daemon that exits promptly makes the supervisor send TERM inside that window, so every daemon restart on Linux waited out the full two-second grace period. Three supervisor tests timed out on CI and in an `oven/bun` Linux container; macOS `/bin/sh` is bash and does not race. The helper holds nothing but a sleep, so it is now KILLed. Principled: it removes the lost-signal race rather than widening timeouts. All 20 `vm-cmux-tui` tests pass in the Linux container (3.2 s) and on macOS.

**Deleted test.** `freestyle-cloud-shell-repair.test.ts` asserted source text of the legacy Freestyle arm (`cmuxd-remote`, `cmux-cloud-shell`) that #11566 removed. It was a source-shape test with nothing left to cover.

**Determinism gate.** The gate flagged the no-findmnt supervisor test for `sleep 100ms; expect(exitCode).toBeNull()`. The fake `mountpoint` now records the watcher's first direct check and the test waits for that event before asserting the supervisor is still up. `python3 scripts/check-test-determinism.py --strict` reports 0 active findings.

https://claude.ai/code/session_01H5V288HnYi8R7ciVie6Exq

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11586"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790924414&installation_model_id=21920&pr_number=11586&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11586&signature=461280f31b08c43246c771ce99353e93be897d44f80b89a4b307019d4d16dfd1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Restores main CI after the Blaxel removal, non-root daemon changes, and updated Cloud VM pricing. The daemon supervisor now polls for child exit, cancels its grace helper after reaping, and only force-kills a still-live child after two seconds, avoiding trap races, orphaned sleeps, and reused-PID kills.

- Casts provider columns to text in the Blaxel removal migration so fresh databases apply successfully without changing existing databases.
- Updates Freestyle repair and pricing tests for the current daemon commands and Unlimited Cloud VMs on Pro and Team plans.
- Makes the no-`findmnt` supervisor test wait for the watcher's first check, removing its strict determinism finding.

<sup>Written for commit a20cdbb782c519f9b4c89c8f5d4afc27a1e6b353. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved daemon shutdown handling so processes that exit during the grace period are detected promptly, preventing unnecessary signals and lingering background tasks.
  - Daemons that remain active beyond the grace period are now forcefully terminated more reliably.

- **Updates**
  - Updated the English pricing page to show **Unlimited** concurrent cloud VMs for both Pro and Team plans.


lawrencecchen:
Local review gate (codex gpt-5.6-sol, isolated) at head 4cf21d60ef: one remaining P2, asking to keep `freestyle-cloud-shell-repair.test.ts`. Assessed as not actionable: that file asserted source text of the legacy Freestyle arm (`cmuxd-remote serve --ws`, `CMUX_CLOUD_SHELL_PATH`, `readFreestyleCloudShellState`, `freestyleCloudShellSetupCommands`) that https://github.com/manaflow-ai/cmux/pull/11566 deleted; `grep -c` for those identifiers in `web/services/vms/drivers/freestyle.ts` returns 0. There is no remaining behavior to cover, and the test failed on main for exactly that reason. Earlier rounds' findings on the grace-period helper (orphaned sleep, unconditional KILL, pid reuse, helper cancellation) were each fixed in follow-up commits; 20/20 supervisor tests pass on macOS and in a Linux container at this head.

PR #11582, opened 2026-09-02
Make the Blaxel-removal migration apply on a fresh database

lawrencecchen:
Every fresh local dev database on `main` fails `bun dev` at migration time since https://github.com/manaflow-ai/cmux/pull/11566:

```
error: unsafe use of new value "blaxel" of enum type vm_provider
```

The removal migration's four `UPDATE ... WHERE provider = 'blaxel'` statements reference an enum value that an earlier migration adds in the same transaction, which Postgres forbids. Comparing `provider::text = 'blaxel'` sidesteps the enum lookup. Existing databases (staging, production) rewrite the same rows as before.

Found while dogfooding https://github.com/manaflow-ai/cmux/pull/11580 on a fresh worktree database.

https://claude.ai/code/session_01S7SVaoTDXVuXZKaN5QAwHc

---
## Summary by cubic
Fixes the Blaxel-removal migration so it applies on a fresh database instead of failing with Postgres' "unsafe use of new value" error. The UPDATE statements now compare the enum columns as text, which sidesteps the enum lookup and rewrites the same rows on existing databases.

<sup>Written for commit d29d60f99547ddd895a2623a0a8e5102fe3e4b4d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database migrations so provider updates complete successfully on both fresh and existing databases.
  - Prevented migration failures caused by PostgreSQL enum value comparisons.

PR #11587, opened 2026-09-02
fix(web): tolerate cloud_vms rows whose provider driver was retired

lawrencecchen:
Root cause of the "Cloud is unreachable" panel since ~06:00 UTC 2026-09-02.

https://github.com/manaflow-ai/cmux/pull/11566 removed the Blaxel driver and deployed to prod (`cmux-fuxilupdf`, sha `12f60191`). Production migrations are operator-run and the `remove_blaxel_vm_provider` migration has not been applied in staging or prod, so `cloud_vms` still holds rows with `provider = 'blaxel'` (7 running, across 6 accounts). `GET /api/vm` calls `vmCapabilitiesFor(entry.provider)` for every row, the registry throws `unknown VM provider: blaxel`, and the whole list returns 500. Axiom shows 364 `GET /api/vm` 500s on that sha and zero on the previous one. The Mac app maps any non-401/402 list failure to "Cloud is unreachable", so affected accounts lose every machine, not only the Blaxel one.

Three commits:

1. Failing test: 3 of 4 cases fail on main.
2. Code fix: a `cloud_vms` row whose provider is not in `PROVIDER_IDS` is treated as retired. `listUserVms` drops it, `requireUserVm` reports it as not found (stats/exec/attach/delete answer 404 instead of tripping the registry), and the status reconcile cron skips it without loading a driver. Driver removal and row cleanup are always two separate rollouts, so read paths must not assume every stored provider still has a driver.
3. Migration fix (still unapplied everywhere, so safe to amend): live Blaxel rows are marked `destroyed` with `failure_code = 'provider_retired'` before the enum rebuild, instead of being relabeled as running e2b machines. The `= 'blaxel'` comparisons now cast to text: on a fresh database the label is added in the same migrate transaction and Postgres rejects the enum literal ("unsafe use of new value"), which is why `bun run db:test` failed at this migration on main.

Verification: `bun test` on the vm unit suites, `bun run typecheck`, eslint clean. `db:test` applies the migration on a fresh Docker database (remaining failures in that suite are 5s timeouts that reproduce with the fix reverted, load average 17+). The full migration dry-ran against production in a rolled-back transaction: 8 rows retired, 17 relabeled, enum rebuilt to `{e2b,freestyle,daytona}`, nothing committed.

Operator follow-ups, not in this PR:
- Delete the 7 live sandboxes and 3 home volumes in the production Blaxel workspace (its key is a Vercel sensitive var; the local keys reach only the `manaflow` workspace, which holds 10 other DEPLOYED sandboxes).
- DONE 2026-09-02: migration from commit f75673ff43 applied to staging (07:47 UTC) and production (07:53 UTC). Prod: 8 Blaxel rows retired, enum rebuilt.

https://claude.ai/code/session_01XfdJ6PS9UE9yzsQdHAP5PF

## Summary by CodeRabbit

* **Bug Fixes**
  * Retired VM providers are now handled gracefully across VM listings, retrieval, and provider-status reconciliation.
  * Virtual machines associated with unregistered providers are treated as unavailable instead of triggering provider errors.
  * Added regression coverage to ensure retired-provider records do not invoke unavailable provider drivers.

### DIFF
--- a/web/db/migrations/20260901120000_remove_blaxel_vm_provider/migration.sql
+++ b/web/db/migrations/20260901120000_remove_blaxel_vm_provider/migration.sql
@@ -7,40 +7,16 @@
 -- old type, so a missed column aborts the whole transaction instead of silently
 -- leaving a split schema.
 --
--- 'blaxel' is compared as text on purpose: on a fresh database the label was
--- added by an earlier migration in the same transaction, and Postgres rejects
--- an enum literal for a label that is not yet committed ("unsafe use of new
--- value"). The text comparison reads the same rows without that lookup.
---
--- Blaxel machines are unreachable regardless: the driver is gone, so nothing
--- in the control plane can address them. Every Blaxel machine that is not
--- already destroyed is marked destroyed here, the same way markDestroyed does
--- it at runtime, with failure_code 'provider_retired' so the row says why.
--- Reads already hide destroyed rows, and a base whose active machine is
--- destroyed opens a fresh generation on the default provider, so those bases
--- heal on their next open. The sandboxes themselves must be deleted in the
--- Blaxel console; this migration only records that they are gone.
---
--- The rows are then rewritten to 'e2b' so the enum can be rebuilt without
--- dropping history (usage events, base generations). They are all destroyed
--- by then, so no read path will ever hand one to the e2b driver.
+-- Any row still naming a Blaxel machine is rewritten to 'e2b' first. Those
+-- machines are unreachable regardless (the driver is gone); this keeps history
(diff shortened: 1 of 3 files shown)


lawrencecchen:
Merged as 83fb126f16b1247ecac1c3b8ee69933bf2d49e3d with the read-path fix only. The migration rewrite commit (f75673ff431791407b534c7af630851ca17a327c) was reverted on this branch so the fix could ship gate-clean during the incident; its content stays in history at that sha. The migration change should land through https://github.com/manaflow-ai/cmux/pull/11582.

PR #11590, opened 2026-09-02
Cloud cleanup

theswerd:
Freestyle is now the only Cloud VM provider. `ProviderId` collapses to `"freestyle"`, so every provider switch loses its other arms and `assertNever` proves the registry exhaustive at one value.

## Removed
- `drivers/e2b.ts` and `drivers/daytona.ts`, their bake scripts, their provider tests, and `test-cloud-vm-ws-auth.ts` (it existed only to prove the two gateways' auth gates).
- The other arms of the driver registry, the create kill switch (`providerEnabledEnvKey`), and the image env selector (`providerBaseImageEnvKey`).
- The `@daytonaio/sdk` and `e2b` dependencies — nothing imports them now.
- 7 of 8 image-manifest entries, leaving the one Freestyle image.
- `verify-devbox-image.ts` keeps only its Freestyle branch; the inbound-firewall proof went with the driver it pinned.
- Every `CMUX_VM_E2B_ENABLED` / `E2B_API_KEY` / `E2B_CMUXD_WS_TEMPLATE` / `DAYTONA_*` reference in live code, tests, scripts and actionable docs.

## Migration
Rebuilds the `vm_provider` enum down to `('freestyle')` using the same shape as the Blaxel removal, rewriting surviving rows in all four provider columns first.

**Destroy any live machines on the removed providers before this runs** — afterwards nothing in the control plane can address them.

Validated against a real Postgres 17, not just written:
- all migrations apply in order;
- seeded rows across `cloud_vms`, `cloud_vm_usage_events`, `cloud_vm_bases` and `cloud_vm_base_generations` all rewrite to `freestyle`;
- the enum reduces to one value and `vm_provider_old` drops;
- new `'e2b'` inserts are rejected;
- **the `DROP TYPE` interlock was proven**: adding an unhandled `vm_provider` column makes the whole transaction abort with the enum untouched, which is the safety claim the migration comment makes.

## Tests
Provider-parameterized cases collapse to Freestyle. Three lost their premise entirely and were removed rather than reworded into something vacuous — worth a look, since this is real coverage going away:
- the cross-provider image-inference regression (the 2026-08-26 outage shape needs two providers to exist);
- the account-deletion "same provider id on different providers" case, rewritten as the dedupe rule it can still prove;
- the "off-only kill switch for a non-default provider" rule — Freestyle's flag is *required*, not recommended, so the rule has no instance left.

The env-audit suite keeps `e2b` as an env **value** on purpose: a deployed `CMUX_VM_DEFAULT_PROVIDER` still naming a removed provider is now a real stale-config shape the audit must fail on.

Fixed along the way: `tests/test_cloud_vm_attach_retry_script.py` asserted `load-dev-env.sh` round-trips `E2B_CMUXD_WS_TEMPLATE`, lines this branch deleted. There is no pytest lane in CI, so nothing would have caught it.

## Two rationales corrected rather than deleted
- The devbox Dockerfile restrictions began as one builder's parser limits; they stay because the Freestyle replay executes the same instructions.
- "Never installs docker" was justified by a platform limitation that no longer applies — Freestyle VMs *can* run Docker (nested virtualization). Kept as a deliberate image-scope choice, flagged in the test so someone can revisit it.

## Still contains the removed names, deliberately
- Two **applied** migrations. Editing applied migration SQL breaks replay and Drizzle's checksums; they record what the schema did.
- This branch's own migration — the `IN ('e2b', 'daytona')` literal is the code that rewrites the rows.
- `cmux-tui/crates/chatmux-relay`, a different subsystem with its own provider enum on a wire protocol. Unrelated to the Cloud VM registry; removing a serde variant there would be a breaking protocol change.

## Verification
- `tsc --noEmit` clean.
- `bun test tests/` → same **29 pre-existing failures as `main`, zero new**, confirmed by diffing failure lists before and after rather than eyeballing counts.
- Migration exercised against real Postgres 17 (above).
- `cmux-devbox-boot` and `load-dev-env.sh` parse; `CloudVMActionLauncher.swift`, `cmux.swift` and the touched test files parse; both message catalogs valid JSON.
- **Not built:** the Mac app. The Swift changes here are comment/string/validator edits only and pass `swiftc -parse`, but no tagged Xcode build ran.

https://claude.ai/code/session_01TyPnvTAscsTTA4XhHBUaE8

---
## Summary by cubic
Removes the E2B and Daytona Cloud VM providers, along with their drivers, bake scripts, provider tests, dependencies, and removed-provider env vars. Freestyle is now the only provider, so `ProviderId` collapses to `"freestyle"` and `assertNever` proves the registry exhaustive at one value.

**Migration**
- Rebuilds the `vm_provider` enum down to `('freestyle')`, rewriting surviving rows in all four provider columns first.
- Destroy any live E2B or Daytona machines before this runs; afterwards nothing in the control plane can address them.
- Validated against Postgres 17; the `DROP TYPE` interlock aborts the transaction if a provider-typed column is missed.

**Tests**
- Provider-parameterized cases collapse to Freestyle; three lost their premise and were removed.
- The env-audit suite keeps `e2b` as an env value so a stale `CMUX_VM_DEFAULT_PROVIDER` still fails the audit.
- Applied migrations, this branch's migration, and `cmux-tui/crates/chatmux-relay` still contain removed names deliberately.

<sup>Written for commit a639e1144bce3cd572db4be1520058ab4813ea7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11590?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Changes**
  - Cloud VM support is now limited to the Freestyle provider.
  - Existing Cloud VM records using retired providers are migrated to Freestyle.
  - Provider selection, validation, image resolution, smoke tests, and runtime configuration now accept Freestyle only.

- **Documentation**
  - Cloud VM setup, rollout, image, and daemon documentation now describes the Freestyle-only workflow.

- **Maintenance**
  - Retired provider integrations, build tools, dependencies, image entries, and related tests were removed.
  - Freestyle credential and environment variable guidance was updated.


theswerd:

lawrencecchen:
recheck

PR #11594, opened 2026-09-02
security: migrate CLA Assistant to guarded v3 (superseded by #11608)

lawrencecchen:
Superseded by [#11608](https://github.com/manaflow-ai/cmux/pull/11608). The policy guard and privileged policy must land in separate pull requests so the base-controlled guard can protect the policy change. This PR is retained only for traceability and will be closed after the replacement is ready.


lawrencecchen:
Superseded by #11608. The exact review correctly identified that a guard introduced in the same PR cannot protect that PR. The replacement lands the guard first, then the policy migration.

PR #11671, opened 2026-09-02
iOS: minimal Cloud VM section over in-process WireGuard (PR 4)

lawrencecchen:
PR 4 of docs/cloud-userspace-wireguard.md, stacked on the terminal-client PR (base feat-terminal-client-ws-wireguard). The minimal iOS surface for reaching Cloud VMs through the phone's in-process WireGuard tunnel, so a user gets terminals before installing any system VPN.

Two packages plus the app wiring.

CmuxMobileCloud (domain, no binary): the /api/vm client (list, tunnel enroll, cmux-remote attach, invitation approve) with request shapes mirroring the Mac's VMClient; the device identity (ios-<uuid> plus a Curve25519 WireGuard key, private half in the Keychain as AfterFirstUnlockThisDeviceOnly, only the public key sent, a locked store fails closed rather than re-minting); the wg-quick config builder that fills the server's blank PrivateKey and pins PersistentKeepalive 25; a CloudSessionController that enrolls and starts the tunnel on section appear, tears it down on disappear or background, and restarts on foreground while still visible (decision 5A, no timers); per-machine link with invitation approval polled on an injected Clock; and an output reducer that turns snapshot/output/resize/exit events into grid and byte writes, prefixing a re-snapshot with a full reset since the emulator has no reset API. 35 tests pass locally.

CmuxMobileCloudUI (views plus adapter): the Cloud section (machines list), a machine's terminal catalog with a New terminal action, and a terminal screen that feeds link bytes into the shared GhosttySurfaceView and sends typing and grid reports back. A CmuxTerminalClientKit adapter satisfies the domain's transport seams with the prebuilt Rust client.

App wiring: cmuxFeature depends on both packages, builds the CloudSessionController at the root scene (tokens through the auth coordinator, per-bundle Keychain identity, Application Support state dir), and injects it via the cloudSessionController environment key. The Computers screen shows a Cloud row (CloudEntryRow) that pushes the section; it renders only when a controller is composed.

HIG pages read: Lists and tables (https://developer.apple.com/design/human-interface-guidelines/lists-and-tables), Loading (https://developer.apple.com/design/human-interface-guidelines/loading), Navigation bars (https://developer.apple.com/design/human-interface-guidelines/navigation-bars). The Cloud entry is a disclosure row in the Computers inset-grouped list; the section, catalog, and terminal are pushed on that stack; loading states use inline progress rows.

Binary resolution: CmuxTerminalClient resolves through its local-xcframework override when CmuxTerminalClient.xcframework sits next to its Package.swift (the iroh-ffi pattern); the remote release pin stays PENDING until the xcframework workflow publishes it. Dev builds drop the one-off xcframework there. GhosttyKit comes from scripts/setup.sh as usual, so the UI package and app targets are verified by the tagged build rather than by a standalone swift build on a bare checkout.

No @codex review.


lawrencecchen:
Mac fleet instructions for head `f981d53ebe6a4306780a289427db7780cdc0f44c`. Planned tag: `pr-11671-f981d53e`; this is not yet a published build.

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-11671-f981d53e /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git f981d53ebe6a4306780a289427db7780cdc0f44c' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/11671 --source-digest f981d53ebe6a4306780a289427db7780cdc0f44c --cache-key cmux:pr-11671 --min-free-bytes 268435456000 --label cmux --label ram48)
```

PR #11655, opened 2026-09-02
ci: report ignored-only focused filters

lawrencecchen:
Focused cmux-tui verification currently reports a generic no-tests error when a filter selects only #[ignore] tests. Make the branch explicit by checking the ignored listing when the runnable listing is empty, so skipped matches are visible and cannot produce a false-green focused run.

No Rust build run locally per policy.

---
## Summary by cubic
Fixes the focused cmux-tui verification to report an explicit error when a filter matches only ignored tests, instead of a generic no-tests error.

<sup>Written for commit 2efa2d8d6c2f3a48baf62052b864c3f34e6216ae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #11739, opened 2026-09-02
ci: report ignored-only focused filters

lawrencecchen:
## Summary

- fail focused cmux-tui verification when the selector matches ignored tests but no normal tests
- add a regression guard for the workflow condition

## Validation

- `python3 -m pytest -q tests/test_tui_focused_filter_guard.py`
- `git diff --check`

This is a fresh current-main replacement for #11655. It keeps the original scope and has two commits so the regression test is visible before the fix.

---
## Summary by cubic
Fails focused cmux-tui CI verification when the test filter matches ignored tests but no normal tests. Previously the workflow errored only when the matched normal and ignored test lists were identical; now it also errors when the filter matches only ignored tests. Adds a regression test that reads the workflow file and pins the new condition in place.

<sup>Written for commit f0d91b8f7541d1c7fb3bc44b77587608c12ca093. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11739?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/cmux-tui.yml
+++ b/.github/workflows/cmux-tui.yml
@@ -489,7 +489,7 @@ jobs:
           ignored_listing="$RUNNER_TEMP/cmux-tui-focused-ignored-list.txt"
           cargo test --workspace --locked "$TEST_FILTER" -- --list --ignored > "$ignored_listing"
           grep -E ': test$' "$ignored_listing" | sort > "$ignored_names" || [[ "$?" -eq 1 ]]
-          if [[ -s "$normal_names" && -s "$ignored_names" ]] && cmp -s "$normal_names" "$ignored_names"; then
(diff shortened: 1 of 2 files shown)


lawrencecchen:
Closing as superseded by merged PR https://github.com/manaflow-ai/cmux/pull/11741, which fixes the ignored-only guard using the runnable-name difference.

Issue #4005, opened 2026-05-12
cmux claude wrapper blocks new `claude agents` (Agent View) dashboard


akira8900:
## Summary

The `claude` wrapper at `cmux.app/Contents/Resources/bin/claude` injects `--session-id <UUID>` for every subcommand not on its passthrough whitelist. This breaks the new **Agent View** feature shipped in Claude Code **v2.1.139** (released 2026-05-11), where `claude agents` is meant to open a full-terminal dashboard for managing background sessions.

Inside a cmux terminal, running `claude agents` is rewritten to roughly:

```
claude --session-id <uuid> --settings <hooks-json> agents
```

…which causes Claude Code to spin up a new session instead of taking over the terminal with the Agent View dashboard. The same dashboard works fine in a non-cmux terminal.

## Reproduction

1. Install Claude Code `>= 2.1.139` (`npm i -g @anthropic-ai/claude-code@latest`).
2. Open a cmux terminal (so `CMUX_SURFACE_ID` is set).
3. Run `claude agents`.
4. Expected: full-screen Agent View dashboard listing background sessions.
5. Actual: Claude starts a fresh interactive session (or prints a flat agent list).

Running with `CMUX_CLAUDE_HOOKS_DISABLED=1 claude agents` or invoking `~/.local/share/claude/versions/2.1.139/claude agents` directly works correctly, confirming the wrapper is the blocker.

## Affected commands

Likely the entire new Agent View shell surface ships subcommands the wrapper should pass through untouched:

- `claude agents`
- `claude attach <id>`
- `claude logs <id>`
- `claude stop <id>` / `claude kill <id>`
- `claude respawn <id>` / `claude respawn --all`
- `claude rm <id>`

Docs: https://code.claude.com/docs/en/agent-view

## Suggested fix

Extend the existing passthrough switch in `bin/claude`:

```bash
case "${1:-}" in
    mcp|config|api-key|rc|remote-control|agents|attach|logs|stop|kill|respawn|rm)
        clear_inherited_claude_auth_selection_env
        exec "$REAL_CLAUDE" "$@"
        ;;
esac
```

Local patch confirmed: after adding `agents` to the whitelist, the dashboard launches correctly inside a cmux terminal. Other Agent View subcommands (`attach`, `logs`, `stop`, `respawn`, `rm`) should likely be whitelisted for the same reason — they each operate on existing background sessions and don't need `--session-id` injection.

## Environment

- cmux: `0.64.4 (84)`
[...]

PR #11747, opened 2026-09-03
fix: pass claude stop/kill/logs/rm/respawn/gateway/import through the wrapper

counterposition:
Inside a cmux terminal, `claude stop <id>` opens a brand-new Claude Code session with the prompt "stop" instead of stopping the background session. Same for `kill`, `logs`, `rm`, `respawn`, `gateway`, and `import`. The wrapper only recognizes subcommands on a hardcoded allowlist; anything else gets `--session-id <fresh uuid> --settings ...` injected ahead of it, and the CLI then reads the command name as the `[prompt]` positional.

This is the same failure #4005 reported for `agents` and that 815724872a fixed for `attach`. Claude Code 2.1.x has since added the rest of the background-session family, and this PR adds all of them so the wrapper execs the real CLI with argv untouched. Hook settings are meaningless for these commands anyway: they act on a session that is already running in another process.

**Changes**
- `Resources/bin/cmux-claude-wrapper`: add the seven names to `claude_builtin_command_name`.
- `tests/test_claude_wrapper_hooks.py`: regression test asserting each passes through with no `--session-id`, `--settings`, or `NODE_OPTIONS` injection.

Two commits so CI shows the test failing before the fix. Verified against Claude Code 2.1.259 with `python3 tests/test_claude_wrapper_hooks.py`; every command in its `--help` output is now allowlisted.

Refs #4005

https://claude.ai/code/session_01MiwcMaECsbeFgUo7eG4zCx

---
## Summary by cubic
Fixes the cmux wrapper so `claude stop`, `kill`, `logs`, `rm`, `respawn`, `gateway`, and `import` pass through to the real CLI with argv untouched, instead of injecting a fresh `--session-id` that makes Claude Code treat the command name as a prompt and open a new interactive session. Adds regression tests for each command.

<sup>Written for commit c82832d37cc91f17663b37bb1c844694f667e19a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11747?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Claude Code background-session subcommands now pass through unchanged, without unwanted hook, session, or environment-setting injection.
  - Added support for `gateway`, `import`, `kill`, `logs`, `respawn`, `rm`, and `stop` invocations.

- **Tests**
  - Added regression coverage to verify correct handling of these subcommands.

### DIFF
--- a/Resources/bin/cmux-claude-wrapper
+++ b/Resources/bin/cmux-claude-wrapper
@@ -1240,9 +1240,9 @@ claude_passthrough_option_flag() {
 
 claude_builtin_command_name() {
     case "$1" in
-        agents|attach|auth|auto-mode|config|api-key|daemon|doctor|install|mcp|\
-        experimental-next|plugin|plugins|project|rc|remote-control|setup-token|\
-        ultrareview|update|upgrade)
+        agents|attach|auth|auto-mode|config|api-key|daemon|doctor|gateway|\
+        import|install|kill|logs|mcp|experimental-next|plugin|plugins|project|\
+        rc|remote-control|respawn|rm|setup-token|stop|ultrareview|update|upgrade)
             return 0
             ;;
     esac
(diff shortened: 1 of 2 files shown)


counterposition:

PR #11697, opened 2026-09-02
fix(cmux-tui): measure bench probes during in-flight creates

lawrencecchen:
Stacked on #11688.

bench interact previously joined every create worker before measuring same-connection typing, and it accepted the first render-state event from any surface. This change submits create batches and same-connection typing before response draining, holds workers while the separate probe runs, demultiplexes responses, and matches render-state to the requested surface.

Tests are test-first in two commits. Docs describe the in-flight probe boundary and list the bench scope. This PR does not change wait-stage semantics or the global input barrier.

Refs #11688 and #11347.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11697"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790955970&installation_model_id=21920&pr_number=11697&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11697&signature=4f5dab3239670a82c8b3d9aa7e1fa2c54e99363aa35b90ab00e189b280d66320"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Fixes `bench interact` so typing probes measure head-of-line blocking while create responses stay in flight, and consolidates every bounded wait into one `budgets` module with a `diag budgets` command to print it.

- Create batches submit fully before responses are read; a separate-connection probe runs while creates are in flight, and same-connection typing splits into interleaved and after-batch probes. Responses match by request ID, first-frame events match the requested surface, and bench counts are capped at 100,000 with clamped percentile ranks.
[...]

PR #11767, opened 2026-09-03
perf(cmux-tui): frontend never waits, per-target input ordering, delta apply, local drag preview (IX1)

lawrencecchen:
Frontend-only slice of the zero-wait interaction plan. No daemon (`cmux-tui-core`) code changes. Design: hq plans/cmux-tui-zero-wait-interaction.md (IX1).

Each item is prior-bad → now, with the mechanism and the test that pins it.

1. **Per-target input ordering.** Prior: every session mutation on the PTY input worker was a global barrier and ran inline on the worker thread, so a keystroke to an existing pane waited for a create round trip to a different pane (up to the 10 s request timeout). Now: mutations form their own FIFO lane on a separate thread and block only input addressed to the surfaces they destroy (close tab/pane/screen/workspace name them; creates name nothing, their input is deferred by semantic intent in the frontend as before). Principled: the ordering rule is stated by cause in the `pty_input` module doc. Tests: `slow_untargeted_session_mutation_does_not_delay_input_to_other_surfaces`, `targeted_session_mutation_blocks_later_input_to_its_target_behind_surface_operation`, `input_to_a_close_target_enqueued_before_the_close_runs_first`, `session_mutations_stay_in_enqueue_order`.
2. **`client-focus` off the draw thread.** Prior: first tree adoption inside the first `terminal.draw` called `client-focus` synchronously on remote sessions. Now: remote sessions ask on a worker; first draw uses the tree's default focus; the answer applies via `AppEvent::ClientFocusRestored` only while `client_focus_epoch` is unchanged (bumped on adoption and on user focus reports). Local sessions keep the synchronous path (no I/O). Tests: `remote_client_focus_restore_does_not_block_tree_adoption`, `restored_client_focus_yields_to_user_navigation`.
3. **Local split-drag preview, one commit on release.** Prior: each pointer motion enqueued a coalesced `set-split-ratio` and the divider moved after `layout-changed` triggered a refetch. Now: motion updates a client-local preview ratio applied during layout; release sends one `set-split-ratio` plus its settle barrier; Escape drops the preview and sends nothing; the committed preview holds until the tree carries the ratio or the commit is refused. Viewport column width drags keep their coalesced path. Tests: `split_drag_previews_locally_and_commits_once_on_release`, `escape_cancels_split_drag_without_a_request`, `committed_split_preview_holds_until_the_tree_carries_the_ratio`.
4. **Tree deltas applied to the cached tree.** Prior: the client subscribed to the coarse tree stream and refetched `list-workspaces` after every mutation. Now: it subscribes with `tree_events:"deltas"` and applies workspace added/closed/renamed/moved (exact next `workspace_revision` only), screen added/closed/renamed, and tab added/closed/renamed in stream order, emitting `MuxEvent::TreeDelta` so the frontend redraws from the cache; a gap, unknown parent, delta before any snapshot, `tree-changed`, or `layout-changed` keeps the refetch; a snapshot racing an applied delta marks the tree stale again. A committed remote mutation whose outcome the cache already shows settles as authoritative without a refetch. Kept on the refetch path on purpose: `pane-added`/`pane-closed`, because the Pane entity does not carry the split tree the mutation changed and the daemon emits no `layout-changed` to delta subscribers for a split. Also, the daemon today still emits `tree-changed` alongside most deltas, so the refetch still runs after them; deltas make the change visible one round trip earlier now, and the skip becomes effective once IX2 stops the redundant `tree-changed`. Tests: `tab_and_screen_deltas_apply_to_the_cached_tree_without_a_refetch`, `workspace_deltas_apply_only_the_exact_next_revision`, `pane_deltas_and_tree_changed_force_a_refetch`, `deltas_without_a_snapshot_baseline_force_a_refetch`, `subscription_requests_tree_deltas`, `remote_creation_skips_the_refetch_when_the_cached_tree_already_shows_it`.
5. **Lifecycle text and instant renames.** Prior: an attaching pane was an empty grid; an exited pane gave no sign; a rename showed only after the refetch. Now: one dim localized line, "starting" centered while the attach is outstanding (or before the mirror exists), "process exited" under the last frame once the surface is dead (exit cause is not carried by protocol v12 events, so only the state is named); submitted renames overlay the adopted tree until the authoritative name arrives or the rename is refused. Tests: `attaching_pane_shows_a_starting_line_instead_of_an_empty_grid`, `pane_lifecycle_text_prefers_exit_over_pending_attach`, `submitted_tab_rename_shows_at_once_and_yields_to_the_authoritative_name`, `refused_rename_overlay_drops_once_mutations_settle`.

**Hosted runs.** Focused (`attaching_pane_shows`, head bf85278): https://github.com/manaflow-ai/cmux/actions/runs/33703283056 passed. Full (head bf85278): https://github.com/manaflow-ai/cmux/actions/runs/33704944911 failed on: `clear_history_fallback_operations_remain_ordered` (Linux; relied on the old global barrier, fixed in the last commit and green 3/3 on the testbox), `final_browser_pointer_admission_accepts_exact_presented_frame` (Linux and macOS; also fails on `main` e341deb on the testbox, so not from this branch), `agent_hook_install::hermes_command_reaps_child_when_reaper_spawn_fails` (macOS; file untouched here), and the `x86_64-pc-windows-gnu` artifact job, which also failed in the other full runs of the night (33704596386, 33704249078). Testbox: full `cargo test -p cmux-tui` 1587 passed at 4 threads with only the two environment failures above; `cargo clippy -p cmux-tui --all-targets -D warnings` clean; `cargo fmt --check` clean.

**Measurement.** `bench interact` (IX0, PR #11697 head 5a7f2ab) drives the daemon over the raw protocol, so it does not exercise the TUI client paths this PR changes; the same-connection typing probe tracks daemon create latency in both builds by construction. Numbers are recorded for the environment, not as evidence for this PR. Before = main + IX0, after = this branch + IX0 (scratch merge, not pushed for review). Linux testbox, 30 creates: 1 client create.response p50/p99 4657/5148 ms before vs 708/1228 ms after, 8 clients 9350/26922 vs 9548/25994 ms. This Mac was under heavy load: the IX0 binary alone gave create.response p50 7250 ms, then 2494 ms on an immediate rerun; the merge binary gave 4793 ms between them. The IX1 evidence is the unit tests above (input to another surface flows while a create is blocked; drag motion sends zero requests; first draw does not wait on the focus round trip).

---
## Summary by cubic
Frontend slice of the zero-wait interaction plan (IX1), plus one `cmux-tui-core` change keeping daemon resource IDs out of the top 2^32 range the TUI reserves for client-local placeholders. Removes the round trips that made the TUI feel slow: input to one pane no longer waits on mutations to another, the first frame no longer waits on a focus query, split drags preview locally, tree deltas update the cache without a refetch, and lifecycle text, renames, and create placeholders show immediately.

**Input and first frame**
- Session mutations form their own FIFO lane and block only input to the surfaces they destroy; creates and renames block nothing.
- Input is deferred by target, never by global state: navigation acts immediately, creates are never deferred, and PTY input waits only for its own placeholder's create.
- `client-focus` is answered on a worker thread; the first draw uses the tree's default focus and applies the answer only while the user hasn't navigated.
- Split drags preview locally and send one `set-split-ratio` on release; Escape cancels without a request.
- New-pane, split, new-tab, and new-workspace render a client-local placeholder on keypress that shows a "starting" line and takes focus; the real pane occupies the same slot on resolution, and input deferred to a failed placeholder is retired, never replayed.

**Tree and lifecycle feedback**
- The client applies workspace/screen/tab deltas to the cached tree in stream order instead of refetching after every mutation; workspace moves preserve active workspace focus.
- Refetch stays for revision gaps, unknown parents, missing snapshots, `tree-changed`, `layout-changed`, and `pane-added`/`pane-closed`; the daemon still emits `tree-changed` alongside most deltas, so the skip becomes effective once IX2 stops that.
- Attaching panes show a "starting" line and exited panes show "process exited" instead of a blank grid.
- Submitted renames overlay the tree until the authoritative name arrives or the rename is refused.
- Failed mutations show the daemon's cause in the status line and where the pane would have appeared.

<sup>Written for commit 51b6ad0ea88daa21e6067ee5f850c251f4798247. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11767?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added live previews while dragging pane splits, with commit on release and cancellation via Escape.
  * Added temporary rename feedback and creation placeholders for workspaces, screens, tabs, and panes.
  * Added clear terminal status messages for starting, exited, or failed panes.
  * Improved remote session updates so visible changes appear faster and remain consistent during navigation.

* **Bug Fixes**
  * Improved input handling during surface operations, keeping unrelated panes responsive while targeted changes complete.
  * Failed operations now display the daemon’s reported cause.

### DIFF
--- a/cmux-tui/crates/cmux-tui-core/src/mux.rs
+++ b/cmux-tui/crates/cmux-tui-core/src/mux.rs
@@ -3543,7 +3543,31 @@ impl Mux {
     }
 
     fn next_id(&self) -> u64 {
-        self.next_id.fetch_add(1, Ordering::Relaxed)
+        // The TUI reserves the top 2^32 IDs for client-local placeholders.
+        // Keep daemon IDs in the disjoint low range, even if this counter is
+        // restored or wraps after an extremely long-lived process.
+        const MAX_DAEMON_ID: u64 = u64::MAX - (1 << 32);
+        loop {
+            let current = self.next_id.load(Ordering::Relaxed);
+            if current == 0 || current > MAX_DAEMON_ID {
+                if self
+                    .next_id
+                    .compare_exchange(current, 1, Ordering::Relaxed, Ordering::Relaxed)
+                    .is_ok()
+                {
+                    continue;
+                }
+                continue;
+            }
+            let next = if current == MAX_DAEMON_ID { 1 } else { current + 1 };
+            if self
+                .next_id
+                .compare_exchange(current, next, Ordering::Relaxed, Ordering::Relaxed)
+                .is_ok()
+            {
+                return current;
+            }
+        }
     }
 
     fn next_active_at(&self) -> u64 {
@@ -17568,6 +17592,16 @@ mod tests {
 
     fn test_mux() -> Arc<Mux> {
         Mux::new_for_test("test", SurfaceOptions::default())
+    }
+
+    #[test]
+    fn resource_ids_skip_the_tui_placeholder_namespace() {
+        const MAX_DAEMON_ID: u64 = u64::MAX - (1 << 32);
+        let mux = test_mux();
+        mux.next_id.store(MAX_DAEMON_ID, Ordering::Relaxed);
+        assert_eq!(mux.next_id(), MAX_DAEMON_ID);
+        assert_eq!(mux.next_id(), 1);
+        assert!(!((MAX_DAEMON_ID + 1)..=u64::MAX).contains(&mux.next_id()));
     }
 
     /// A machine resume reconnects every hosted terminal at once. Checkpoint
--- a/cmux-tui/crates/cmux-tui/src/app.rs
+++ b/cmux-tui/crates/cmux-tui/src/app.rs
@@ -13019,10 +13019,10 @@ impl App {
                 }
                 continue;
             }
-            if self.session.has_pending_mutations()
-                || self.session.remote_tree_is_stale()
-                || self.mux_recovery_generation.load(Ordering::Acquire) != 0
-            {
+            // Replay is gated by each input's own target (its semantic
+            // dependency below), never by pending mutations or a stale tree.
+            // The only global barrier is a daemon reconnect.
+            if self.mux_recovery_generation.load(Ordering::Acquire) != 0 {
                 return Ok(DeferredReplayOutcome {
                     action,
                     disposition: DeferredReplayDisposition::Blocked,
@@ -13123,9 +13123,12 @@ impl App {
     }
 
     fn pointer_route_is_globally_stale(&self) -> bool {
-        self.session.has_pending_pointer_mutations()
-            || self.session.remote_tree_is_stale()
-            || self.mux_recovery_generation.load(Ordering::Acquire) != 0
+        // Pending pointer mutations and a stale remote tree no longer stall
+        // pointer routes: the rendered frame already includes placeholders
+        // and the client's predicted layout, so a click routes against what
+        // the user sees (`input_waits_for_target`). Only a daemon reconnect or
+        // a frame-level route invalidation remains global.
+        self.mux_recovery_generation.load(Ordering::Acquire) != 0
             || self.pointer_route_phase.invalidates_all_pointer_routes()
     }
 
@@ -15590,10 +15593,7 @@ impl App {
             ) if !matches!(
                 &input,
                 TerminalInput::Mouse(MouseEvent { kind: MouseEventKind::Moved, .. })
-            ) && (self.session.has_pending_mutations()
-                || self.session.remote_tree_is_stale()
-                || self.mux_recovery_generation.load(Ordering::Acquire) != 0
-                || matches!(&input, TerminalInput::Mouse(mouse) if self.pointer_route_is_stale_for_mouse(mouse)))
+            ) && self.input_waits_for_target(&input)
                 && !self.input_can_update_pending_mutation(&input) =>
             {
                 return Ok(self.defer_input_with_sequence(
@@ -16651,8 +16651,20 @@ impl App {
             self.clear_finished_semantic_destinations();
             let destination_started = self.session.destination_mutation_started();
             let client_owned = self.input_is_client_owned(&input);
-            let semantic_dependency =
-                if client_owned { None } else { self.latest_semantic_destination_intent };
+            let destination = self.input_destination(&input);
+            // Input aimed at a placeholder depends on that placeholder's own
+            // create, not merely on the latest one.
+            let placeholder_intent = destination.filter(|id| is_placeholder_id(*id)).and_then(|id| {
+                self.create_placeholders
+                    .iter()
+                    .find(|placeholder| placeholder.surface() == id)
+                    .map(|placeholder| placeholder.intent)
+            });
+            let semantic_dependency = if client_owned {
+                None
+            } else {
+                placeholder_intent.or(self.latest_semantic_destination_intent)
+            };
             let semantic_result = self
                 .input_creates_session_destination(&input)
                 .then(|| self.allocate_semantic_destination_intent());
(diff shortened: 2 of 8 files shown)


lawrencecchen:
Rebased onto `origin/main` `8fa163da7c1` (head now `e94a4315ec6`). Conflicts were in `session/tree.rs` (main's fail-closed active-index parsing vs. my `parse_tab`/`parse_workspace` extraction: kept main's bodies, re-extracted on top), `session/remote.rs` (main's retired-surface agent fencing in `refresh_tree_inner` plus my delta-race fence: kept both; `RemoteTreeCache::replace` reindex uses main's `workspaces()`), and the test-module import list. Main also made `TreeView::workspaces` private behind `workspaces()/workspaces_mut()` with a location index, so the delta application, close-target lookup, and rename overlay now go through those accessors (extra commit). Testbox on the new head: `cargo test -p cmux-tui` 1653 passed, 5 failed, and all 5 (`final_browser_pointer_admission_accepts_exact_presented_frame`, `host_input_runtime_shutdown_joins_reader_before_returning`, `shift_triple_click_selects_a_complete_line`, `selected_wide_glyph_styles_both_grid_cells`, `wrapped_wide_selection_keeps_the_spacer_head_unselected`) fail identically on plain `main` `8fa163da7c1` on the same box. `cargo clippy -p cmux-tui --no-deps --all-targets -D warnings` reports the same 23 findings on main and on this head (none new; `cmux-tui-core` itself is clippy-red on main). `cargo fmt --check` clean. Hosted focused run on the new head: https://github.com/manaflow-ai/cmux/actions/runs/33708741731.

lawrencecchen:
Dogfood fix for the PTY-exhausted case (head now `73398f173c7`, one commit `feat(cmux-tui): show a create placeholder at once and the daemon's cause on failure`; the two items share the same `Failed` handler, so they landed together rather than as two commits).

1. **Cause, not label.** On `SessionMutationOutcome::Failed` the status line reads `<operation failed>: <cause>` (localized separator `operation_failed_with_cause`), with `remote command rejected: ` stripped and the text truncated to the status width with an ellipsis; the client log keeps the full daemon text. Test: `failed_mutation_status_shows_the_daemon_cause` (exact PTY-exhausted string, full and 48-column). Two existing tests that asserted the bare label now assert the cause form.
2. **Provisional pane.** Alt-n, split, new-tab, and new-workspace draw a client-local placeholder on the keypress, keyed by the mutation's semantic intent: it takes the daemon's slot (Zellij default distribution for Alt-n, requested direction at 0.5 for a split), shows the "starting" line, and takes focus so typing defers to it. The completion resolves it by identity; the real pane lands in the same slot. On `Failed` it shows `could not open: <cause>` where the pane would have been for `DURABLE_NOTICE_DISPLAY_DURATION`, then dissolves and focus returns to the prior pane; deferred input aimed at it is retired through the failed semantic destination and never replayed. On timeout it stays until the refetch decides. Placeholder ids live above `u64::MAX - 2^32`; attach, geometry claims, and focus reports skip them; session reset and cancellation clear them. Tests: `alt_n_shows_a_placeholder_pane_before_the_daemon_answers_and_resolves_it`, `refused_create_placeholder_shows_the_cause_then_dissolves_and_drops_typed_input`.

Testbox: `cargo test -p cmux-tui` 1654 passed, 7 failed of which 4 are the known main-on-box failures and 2 were the label assertions (fixed); clippy `--no-deps` findings identical to main (23, none in these files; `cmux-tui-core` has its own pre-existing clippy and rustfmt debt); `cargo fmt -p cmux-tui --check` clean. Hosted focused run on the final head: https://github.com/manaflow-ai/cmux/actions/runs/33713096646 (passed).

PR #11682, opened 2026-09-02
web: verify Stack access tokens locally on the iroh broker routes

lawrencecchen:
Follow-up to https://github.com/manaflow-ai/cmux/pull/11633. The iroh trust-broker routes carry roughly 100 to 140 requests per second from stable Macs, and every request verified the caller with a Stack `users/me` call. That kept Stack's project-wide rate limit exhausted and made every Stack-backed route (checkout, sign-in, devices, relay) fail at random. The circuit breaker in 11633 limits the damage; this removes the dependency.

The broker only needs the user id. The bearer token is an ES256 JWT that Stack signs with per-project keys published at `/api/v1/projects/<id>/.well-known/jwks.json`. The server now verifies it locally.

- `services/auth/stackAccessToken.ts`: `jose.jwtVerify` against Stack's remote JWKS. ES256 only; issuer, audience, and `project_id` pinned to our project; 60 s clock tolerance. Per-instance key cache (10 min max age, 30 s refresh cooldown on unknown `kid`, 3 s fetch timeout) so a cold Vercel instance pays one small fetch and a flood of forged tokens cannot turn into a flood of JWKS fetches. Every rejection and every key-set failure returns null.
- `verifyRequestIdentity` in `services/vms/auth.ts`: local verification first; anything not accepted falls back to the existing Stack path, which refreshes tokens and is gated by the throttle circuit. Cookie routes are untouched.
- `handleIrohRoute` uses `verifyRequestIdentity`. Devices, relay, VM, billing, and account routes keep the full Stack verification because they need teams and billing plan.
- `stackApiBaseURL` moved to `services/auth/stackApiBaseURL.ts` and is shared with the existing server-key PATCH helper, so a self-hosted Stack origin drives the JWKS URL too. No new exports on `app/lib/stack`, which 24 test files mock.

Verified against Stack itself, not only unit tests: a real access token from the dev project verifies with the real JWKS; a tampered signature and a wrong project return null. Warm verification takes well under a millisecond.

Trade-off, stated: a session revoked at Stack stays accepted on the iroh routes until its access token expires. Stack access tokens live one hour (`exp - iat = 3600` on a real token). Before this PR the same lag was 5 minutes via the native auth cache TTL. Pairing still requires pair grants, and registration mutations still pass the account-deletion tombstone check in the repository.

[...]

PR #11769, opened 2026-09-03
web: stop depending on client behavior for device-registry and relay load

lawrencecchen:
Stack Auth rate-limited our whole project because every authenticated request cost one `GET /api/v1/users/me`. The visible symptom was 429s on unrelated routes, including "Something went wrong starting checkout".

Every cmux build reports the same user agent, `cmux/102`, stable and nightly alike, so production traffic cannot be attributed to a version. This PR therefore assumes no client will ever update.

## Stop calling the auth provider per request

- `/api/relay/token` needs the account id and nothing else, so it verifies the access token locally against Stack's published keys.
- `GET` and `POST /api/devices` also need team membership, which the token does not carry. They read it from `stack_identity_snapshots`, a shared Postgres row refreshed from Stack at most once per TTL per user. Only a verification that listed every team may be stored, so a snapshot can never deny a team the user actually belongs to.
- `DELETE /api/devices` removes a row other team members can see, so it keeps live verification. It is user-initiated and rare.

Default TTL is `CMUX_STACK_IDENTITY_SNAPSHOT_TTL_MS`, one hour, matching the Stack access-token lifetime, so answering from a snapshot adds no staleness beyond accepting a locally verified token. Sign-out deletes the row; account deletion is enforced on read against the tombstone, so it takes effect immediately. Billing, VM mutations, account and admin routes keep verifying live.

Measured before the change: `/api/devices` plus `/api/relay/token` run at 64 req/s and spend 0.85 Stack calls each, about 4.8M calls a day. The iroh routes, already converted by #11682, run at 118 req/s and make essentially none.

## Stop paying for redundant work

The above protects Stack Auth on every client version but does not reduce what an un-updatable client costs us.

- `POST /api/devices` answers an unchanged registration without taking the per-team advisory lock or writing either row, when the only difference would be a presence timestamp refreshed within `CMUX_DEVICE_PRESENCE_TOUCH_INTERVAL_MS` (one minute). This is not a hypothetical: the Mac's dedupe compares route hints carrying observation timestamps, and `sanitizeServerPublishedRoutes` strips them before storing, so a shipped Mac re-registers byte-identical rows for as long as it runs. Any real difference, including route order, writes through unchanged.
- Native ingress 429s now carry `Retry-After`. The iroh broker client already in the field parses it into an account-scoped cooldown that survives relaunch, so it is the only backpressure signal that reaches builds we cannot update.

## Observability

Auth resolution and registration no-ops are stamped on the request span the tracer already emits (`cmux.auth.source`, `cmux.auth.provider_called`, `cmux.auth.snapshot_age_ms`, `cmux.devices.registration_no_op`). No new event stream and no extra ingest: the app-wide 2% head sample is ample at hundreds of requests per second. Queries are in `web/services/auth/README.md`.

## Migration

Adds `stack_identity_snapshots`. Run `bun run cloud-vm:migrate staging` and `bun run cloud-vm:migrate production` before merging.

## Tests

New: `tests/stack-identity-snapshot.test.ts`, `tests/vm-auth-snapshot.test.ts`, `tests/device-registration-no-op.test.ts`, plus a database-backed case proving an identical re-registration leaves `updated_at` and `last_seen_at` untouched while a changed route set still writes through. The full `tests/devices-route.test.ts` suite passes with `CMUX_DB_TEST=1` against Postgres 16 (27 tests). The migration SQL was applied to a container and the snapshot store round-tripped against it through the real Drizzle schema.

Client-side counterpart: #11774 bounds the retry loops that generate this traffic, for the users who do update.

## Summary by CodeRabbit

* **New Features**
  * Authentication can continue smoothly using recently verified account and team information.
  * Re-registering an unchanged device no longer updates its presence unnecessarily.
  * Throttled native requests now include a 60-second retry interval.

* **Bug Fixes**
  * Signing out promptly removes cached access to team-scoped device registry data.
  * Device registrations with changed routes or metadata continue to update immediately.

* **Tests**
  * Added coverage for authentication recovery, device registration behavior, snapshot freshness, sign-out handling, and retry responses.

### DIFF
--- a/web/app/api/account/route.ts
+++ b/web/app/api/account/route.ts
@@ -34,6 +34,7 @@ import {
   vaultUploadGrants,
   vaultUploadTombstones,
 } from "../../../db/schema";
+import { deleteIdentitySnapshot } from "../../../services/auth/identitySnapshot";
 import {
   ACTIVE_STRIPE_PRO_STATUSES,
   type ProMetadataJson,
@@ -670,6 +671,10 @@ async function markAccountDeletionTombstonePending(userId: string): Promise<Acco
           errorMessage: null,
         },
       });
+    // High-volume routes may answer from a stored identity snapshot without
+    // asking Stack. Drop it as soon as deletion is recorded so the tombstone
+    // takes effect on the next request rather than after the snapshot TTL.
+    await deleteIdentitySnapshot(userId);
     const legacySubrouterRetiredTenantIds = uniqueNonEmptyStrings(
       existing?.legacySubrouterRetiredTenantIds ?? [],
     );
--- a/web/app/api/account/route.ts
+++ b/web/app/api/account/route.ts
@@ -34,7 +34,6 @@ import {
   vaultUploadGrants,
   vaultUploadTombstones,
 } from "../../../db/schema";
-import { deleteIdentitySnapshot } from "../../../services/auth/identitySnapshot";
 import {
   ACTIVE_STRIPE_PRO_STATUSES,
   type ProMetadataJson,
@@ -671,10 +670,6 @@ async function markAccountDeletionTombstonePending(userId: string): Promise<Acco
           errorMessage: null,
         },
       });
-    // High-volume routes may answer from a stored identity snapshot without
-    // asking Stack. Drop it as soon as deletion is recorded so the tombstone
-    // takes effect on the next request rather than after the snapshot TTL.
-    await deleteIdentitySnapshot(userId);
     const legacySubrouterRetiredTenantIds = uniqueNonEmptyStrings(
       existing?.legacySubrouterRetiredTenantIds ?? [],
     );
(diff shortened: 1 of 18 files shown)


lawrencecchen:
Both review findings taken.

The no-op check now runs inside the transaction after the per-team advisory lock, so it compares against a row no concurrent registration can be mid-way through changing. The write it saves is the two upserts, which is the bulk of the cost.

Stale team membership is real and Stack sends no membership webhook to invalidate on, so the TTL is the bound. Default is now ten minutes instead of an hour. At ~4,000 active Macs that is under 7 Stack calls a second, against ~125 a second before this work. `DELETE /api/devices` was already live-verified.

Full `devices-route` suite passes against Postgres 16 with the moved check (27 tests, including one proving an identical re-registration leaves `updated_at` untouched while a changed route set writes through).

PR #11814, opened 2026-09-03
Cloud VM errors: terminal 404 for vendor-missing machines, vendor blame in telemetry, fail-fast client states

lawrencecchen:
## Why

A dev-server log showed `POST /api/vm/<id>/attach-endpoint` answering 502 `vm_cloud_service_unavailable` (`retryable: true`) because Freestyle returned `404 not found: vm <id>`. The machine no longer existed at the vendor, but cmux treated it as an outage, so the Mac sidebar and the CLI retried attach every cycle against a machine that would never come back, and the row stayed in the list. Not a dev-only artifact: the same mapping runs in production.

## Server (web/)

- `failWhenProviderVmMissing` wraps every provider call on an existing machine (attach, cmux-remote, enrollment approve, exec, stats, open-port, snapshot). A provider 404 is confirmed with one status probe, the row is marked `destroyed` (usage event `source: provider_missing`, model-plane revoke when wired), and the caller gets 404 `vm_not_found`, `retryable: false`, `details.reason: "provider_missing"`, `ui.title: "Machine no longer exists"`. An unconfirmed 404 (probe times out or says alive) keeps the original retryable error.
- Internal blame: `diagnostics` now carry `blame`, `fault` (`vendor` / `operator` / `user`), the vendor's raw `providerCode`, `providerStatus`, `providerPath`, `providerMessage`, `causeChain`, and one `internalReason`. Sentry title and the `cmux.observability.error` log line read `cloud VM vm_cloud_service_unavailable [vendor: freestyle] freestyle openCmuxRemote failed NOT_FOUND 404 /v5/vms/<id>: …`; tags `vm.fault`, `vm.blame`, `vm.provider_code`, `vm.provider_status`; spans get `cmux.vm.error_provider_*`. The response body stays vendor-free (leak test unchanged).
- Public copy uses cmux Cloud wording ("cmux Cloud machine", never "Cloud VM"). `openCmuxRemote` now maps to phase `attach` instead of `unknown`. The create kill switch answers from one `vmCreateDisabledResponse` instead of four copies.
- Docs: `web/services/vms/README.md` "Error contract" section.

## Mac app and CLI

See the second commit: structured `CloudVMErrorPayload` decoding in `VMClient`, terminal handling of `vm_not_found` (no 15 s backoff loop; the machine is dropped from the sidebar and the list refreshes), Reference id kept on every failure surface, Copy Details on the loading panel, honest waiting copy, CLI attach loop decides from the structured backend code and honors `retryAfterSeconds`.

## Tests

`bun test` on the vm-* suites (route auth, workflows, observability, kill switch, paywall, telemetry): green. New tests cover confirmed vs unconfirmed provider 404, the 404 route contract without vendor leaks, diagnostics extraction, fault classification, and the log/Sentry line.

---
## Summary by cubic
Stops clients from retrying forever against a machine the provider no longer has. A confirmed provider 404 now answers 404 `vm_not_found` with `retryable: false` and marks the machine destroyed, instead of the previous retryable 502 `vm_cloud_service_unavailable` that kept the row in the sidebar.

**Bug Fixes**
- The provider 404 is confirmed with one status probe; an unconfirmed 404 keeps the original retryable error.
- Client attach loops now stop on `vm_not_found` and refresh the list instead of backoff-retrying.

**Refactors**
- VM error telemetry now names the blamed dependency and the vendor's raw code, status, path, and message, without leaking them into the response body.
- Public error copy uses "cmux Cloud machine" consistently, and the create kill switch answers from one shared response.
- The VM list response drops `slug`, Go usage, and memory ladder fields; clients must stop reading them.

<sup>Written for commit ab2f0c5b4bb68f755228fb4dc069392d3d6140e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lawrencecchen:
Fleet instruction update for head `ab2f0c5b4bb68f755228fb4dc069392d3d6140e6`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #11846, opened 2026-09-03
fix(cmux-remote): bound LaneMux physical close

lawrencecchen:
Stacked on https://github.com/manaflow-ai/cmux/pull/11794.

LaneMuxLink::close runs physical FrameLink::close futures in a background cleanup task. A stalled provider left that task and its links retained forever after ReliableSession's timeout canceled its waiter. Bound the concurrent physical closes and publish a typed transport timeout so the cleanup task completes.

Regression history:
- 11815c370f7 test: bound lane mux physical close (fails before fix)
- ceaafe9b498 fix: bound lane mux physical close

Verification on Blacksmith Testbox tbx_01m1kgwy13qc2gvhbsyx0qk1me at ceaafe9b498802b600b6e6cb440861ba0aa9b3bc:
- New focused test passes.
- Test-only commit fails with aggregate close timeout.
- rustfmt check for link.rs passes.
- Full cmux-remote lib was previously run on equivalent parent content; it exposed an existing #11794 flaky test (`shutdown_rejects_send_waiting_for_queue_without_orphaning_completion`, second frame admission timeout).

Residual risk: physical providers that ignore cancellation inside `close` can still outlive the timeout if they block in non-async code; the trait has no force-close hook. Normal async providers are canceled when timeout drops their close future.

### DIFF
--- a/cmux-tui/Cargo.lock
+++ b/cmux-tui/Cargo.lock
@@ -687,6 +687,7 @@ dependencies = [
  "tokio",
  "tokio-rustls",
  "tokio-tungstenite",
+ "tokio-util",
  "tower",
  "tower-http",
  "url",
--- a/cmux-tui/crates/cmux-remote/Cargo.toml
+++ b/cmux-tui/crates/cmux-remote/Cargo.toml
@@ -41,6 +41,7 @@ socket2 = { workspace = true, optional = true }
 snow.workspace = true
 subtle.workspace = true
 tokio.workspace = true
+tokio-util.workspace = true
 tokio-tungstenite.workspace = true
 tower.workspace = true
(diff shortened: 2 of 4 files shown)


lawrencecchen:
Fleet instruction update for head `86d020d50692fe0ee9c9b611f4c2d8769c6911b4`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #11849, opened 2026-09-03
fix(python): reject unary machine stats follow mode

lawrencecchen:
## Summary

`machine_stats(follow=True)` previously discarded the caller's mode and sent a unary request. This stacked fix raises a clear `ValueError` and directs callers to `machine_stats_follow()`. The generator and committed Python output stay in sync. The new stream helper request annotation uses `object` instead of `Any`.

## Verification

- Regression commit `7b6e4ecde32` adds the behavior test and fails before the fix (`ValueError not raised`).
- Fix commit `d02a01b2608` makes the test pass.
- Python package tests: 110 passed.
- Python codegen check, bytecode compile, and `git diff --check`: passed.

This PR is intentionally stacked on `feat-cloud-machine-stats-event` because `main` does not yet contain the machine-stats command. The machine-stats freshness TTL remains a product/protocol decision and is not invented here.

---
## Summary by cubic
Fixes `machine_stats(follow=True)` previously discarding the caller's mode and sending a unary request; it now raises a `ValueError` directing callers to `machine_stats_follow()`.

- The codegen now emits the guard for any conditional stream method whose mode field is `follow`.
- Adds a regression test verifying no command is sent and the error is raised.
- Changes the request parameter annotation in `_open_command_stream_with_result` from `Any` to `object`.
- This PR is stacked on `feat-cloud-machine-stats-event` because `main` doesn't have the machine-stats command yet.

<sup>Written for commit 1bf574d4978d1c8826d02e65667a060a38cb2390. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cmux-tui/bindings/codegen/emit_python.py
+++ b/cmux-tui/bindings/codegen/emit_python.py
@@ -1024,6 +1024,14 @@ __all__ = [
             )
             request_name = self.models[f"commands/{wire_name}/request"].name
             conditional_stream = command.get("stream") is not None and command["stream"].get("mode_field") == "follow"
+            if conditional_stream:
+                mode_name = _snake(command["stream"]["mode_field"])
+                lines.extend(
+                    [
+                        f"        if {mode_name} is True:",
+                        f"            raise ValueError({_quote(f'{method_name}({mode_name}=True) requires {method_name}_follow()')})",
+                    ]
+                )
             arguments = []
             for name, _field in required + positional_optional + optional:
                 python_name = _snake(name)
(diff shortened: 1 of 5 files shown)


lawrencecchen:
Fleet instruction update for head `1bf574d4978d1c8826d02e65667a060a38cb2390`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #11901, opened 2026-09-04
feat(web): rebuild 404 page

lawrencecchen:
## Summary

Rebuilds the web 404 page from the supplied cmux reference instead of layering patches onto the previous implementation.

- Matches the reference page geometry: logo header, 404 heading, 1152px terminal shell, 40px title bar, 720px terminal body, CTA row, and footer.
- Uses one deterministic terminal transcript renderer, with no optional Atlas module or loading race.
- Keeps the terminal draggable with pointer input and arrow-key movement.
- Localizes every new 404 key across all 20 supported locales.
- Sends only locale and fixed action metadata to PostHog. No URL or pathname data is sent.
- Uses product-owned Community and Source destinations in recovery copy.
- Keeps the Mac CTA on the direct signed DMG URL with the HTML download attribute.

## Testing

- `bunx biome check` on all changed TSX and JSON files
- focused ESLint on all changed TSX files
- `bun run lint:complexity`
- `source scripts/load-dev-env.sh && bun run build`
- Playwright screenshot at 1434×1309 compared with the supplied reference
- Playwright behavior checks for keyboard dragging, direct download, Japanese rendering, and all 20 locale 404 routes

## Demo

Reference comparison is captured in the worktree artifact at `cmux-assets/pr11901/404.png`.

| Before (Desktop) | After (Desktop) |
|:---:|:---:|
| Before | After |

## Source reference

Original cmux terminal chrome supplied in the request


lawrencecchen:
Fleet instruction update for head `72e18dec30623ddcd2a0f4f7b0d12c02f9b791e9`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #11908, opened 2026-09-04
test(cmux-tui): stabilize full-gate assertions

lawrencecchen:
Repairs five test-only failures seen across full hosted TUI runs. The shutdown test now waits until the reader has submitted input, the browser pointer test supplies the rendered viewport size, the triple-click assertion uses the inclusive line endpoint, and wide-glyph tests inspect the completed frame buffer. The shared surface helper now runs sleep directly for 300 seconds, which halves child-process use and prevents full-suite fork pressure from closing panes. Product behavior is unchanged.

## Summary by CodeRabbit

* **Tests**
  * Updated terminal interface tests to wait reliably for input processing before shutdown.
  * Refined triple-click selection expectations for complete-line selection.
  * Made browser completion tests use a consistent terminal size.
  * Updated command fixtures for more direct and predictable execution.
  * Adjusted wide-glyph selection tests to inspect completed terminal output correctly.

### DIFF
--- a/.github/workflows/cmux-tui.yml
+++ b/.github/workflows/cmux-tui.yml
@@ -472,7 +472,9 @@ jobs:
         shell: bash
         run: |
           if [[ "$MODE" == "full" ]]; then
-            cargo test --workspace --exclude cmux-tui-core --locked
+            # Process-heavy CLI tests can exceed the runner's child-process
+            # limit when a high-core runner starts every test at once.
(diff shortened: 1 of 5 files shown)


lawrencecchen:
Closed as superseded. The test-only fixes and CLI assertion are already in merged PR https://github.com/manaflow-ai/cmux/pull/11911, and the workflow already uses --test-threads=2 to stay below the runner child-process limit. The remaining change in this PR raises the limit to 8, so it should not be merged.

PR #11960, opened 2026-09-04
Cloud VPN: make the tunnel usable without a terminal, and fix the MTU that silently stalled it

JacobZwang:
Four fixes found while dogfooding Cloud VMs over the WireGuard tunnel. Each was diagnosed against live machines.

## The MTU that stalled every connection

The provider config asks for `MTU = 1380`, but the path only carries ~1370: the VPC link is 1450 and WireGuard over an IPv6 endpoint costs 80 bytes (40 IPv6 + 8 UDP + 32 WireGuard).

The 10-byte overshoot is invisible until it is expensive. Handshake, ping and TCP SYN all fit, so the tunnel comes up, `wg show` reports a live peer, and the daemon port accepts connections — then the first large response stalls unsent forever. It surfaced as machines that connect and then fail every command with `transport error: Resource temporarily unavailable (os error 35)`, which reads as a broken daemon rather than a broken MTU.

Measured against a live machine:

```
1368 bytes → through          1372 bytes → dropped
daemon send queue: 12408 bytes stuck unsent
after clamping:    queue 0, snapshot returns, link connected
```

The config writer now takes the lower of the server's value and a safe ceiling. Too small only costs throughput. **The server-side value is still wrong and worth fixing at the source.**

## Turning the tunnel on without a terminal

Cloud machines have no public inbound port, so nothing reaches them until this Mac is on the private network — previously only via `cmux vpn up`, retyped after every reboot, with no sign in the UI that this was what the panel wanted.

The Machines panel now says so and offers to fix it. "Turn On" asks for an administrator once through the standard macOS dialog and hands the tunnel to a launchd job, so it comes up at every boot and the banner retires itself. A LaunchDaemon rather than the app holding the tunnel, because creating a utun and installing routes needs root and the shipping build has no NetworkExtension entitlement. The job watches the config too: the app rewrites it on re-enrollment, and a tunnel still carrying the previous peer reads as up while reaching nothing.

`cmux vpn install` / `uninstall` stay for people already in a terminal.

## `cmux vpn hosts` never worked

`/usr/sbin/install` does not exist on macOS; install(1) is in `/usr/bin`, and sudo resolves the path literally. Since the hosts sync runs after wg-quick, the tunnel was already up and only the `.internal` names were missing — so it read as a broken VPN. A stale block was the visible symptom: `/etc/hosts` still carried names from an earlier `10.17.26.0/24` network long after the machines had moved.

## Recycled addresses made a route permanently ambiguous

A VPC recycles addresses, so `ws://10.16.133.2:1337/` names a different machine once the previous one is destroyed. The known-daemon store kept the hint on every machine that had ever held the address, so after three the route failed with `multiple known daemons match this route; use --daemon FINGERPRINT` — unanswerable from the app, which has an address and no way to name which dead occupant it means. Pinning now takes the hint from the previous holder, and selection prefers the most recently used for stores written before that.

## Test plan

- [x] MTU clamp verified live: seeded a config with `MTU = 1380`, app rewrote it to `1360`
- [x] Confirmed the stall and its disappearance against a live machine (send queue 12408 → 0, snapshot returns)
- [x] `cmux vpn hosts` reaches sudo's prompt instead of "command not found"
- [x] Builds; new `vpn` verbs dispatch; banner preconditions hold locally
- [ ] The install itself needs an interactive administrator prompt, so it is unrun here
- [ ] cmux-tui changes are not compiled (those builds do not run on macOS)

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/11960"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791134969&installation_model_id=21920&pr_number=11960&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F11960&signature=30ae35fb99f8d22edb4203452b5213ce8d4d6d56a64e1a2bf3b68a988f3d0524"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes Cloud VPN issues found while dogfooding and makes joining the private network a setup step: tunnels now come up without a terminal, large transfers no longer stall silently, new machines are reachable immediately, `cmux vpn hosts` works on macOS, and recycled VPC addresses no longer leave routes ambiguous.

**New Features**
- The Machines panel shows a setup card when the tunnel is off; one button enrolls this Mac and installs a launchd job, so the tunnel comes up at boot and re-applies on config changes, and the card waits for the tunnel to come up before reporting done. The job strips `PreUp`/`PostUp`/`PreDown`/`PostDown` from its root-owned config copy so the user-writable source cannot hand root a command, and its identity derives from the tunnel's interface name so staging cannot replace the production job.
- A toolbar menu shows the tunnel's state and offers Reconnect and Turn Off, with Turn On available again after a teardown; the menu and card share one state that re-reads the system, so the panel agrees with the CLI and catches tunnels brought down outside the app, reporting "connected (until restart)" when no job backs the tunnel.
- `cmux vpn install` and `cmux vpn uninstall` build the same job from the CLI for terminal users.
- New machines send a gratuitous ARP before their route is handed out, so a machine that has only been talked *to* is reachable without waiting for it to speak first; telemetry records whether the announce ran.

**Bug Fixes**
- Clamps the tunnel MTU to 1360; the provider's 1380 overshoots the ~1370-byte path and stalled large transfers while handshake, ping, and TCP SYN succeeded. A malformed `MTU = 0` is floored at IPv6's minimum instead of being written through.
- `cmux vpn hosts` now calls `/usr/bin/install`; macOS has no `/usr/sbin/install`, so every sync failed after the tunnel was up and left stale `/etc/hosts` entries.
- Route hints now move to the daemon currently holding a recycled address, and selection prefers the most recently used daemon instead of erroring.

The server-side MTU is still 1380 at the source and is worth fixing there.

<sup>Written for commit 85fe5166ba54e1b611c2c3ee0471dc5aadffdd84. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/11960?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added `cmux vpn install` and `cmux vpn uninstall` commands for persistent VPN startup and configuration updates.
  * Added in-app controls to connect, reconnect, or disconnect from the Cloud VM private network.
  * Improved private-network connectivity when Cloud VMs are created, resumed, or reconnected.
* **Bug Fixes**
  * VPN tunnel configurations now enforce safe MTU limits for improved connectivity.
  * Improved remote daemon selection when route information is duplicated or stale.
  * Corrected system tool resolution used by VPN commands.

### DIFF
--- a/CLI/CMUXCLI+VPN.swift
+++ b/CLI/CMUXCLI+VPN.swift
@@ -21,6 +21,23 @@ import Foundation
 ///   response advertises `network_extension_available` so this command steers
 ///   without a new CLI release.
 extension CMUXCLI {
+    /// Absolute paths to the system tools `cmux vpn` shells out to.
+    ///
+    /// Named rather than inlined so a test can assert each one exists on the
+    /// machine. sudo resolves these literally, so a wrong directory is not
+    /// caught until runtime — and `hosts` runs only *after* the tunnel is up,
+    /// where a failure reads as "the VPN is broken" when the tunnel is fine
+    /// and only the `.internal` names are missing.
+    enum VPNTool {
+        /// macOS ships install(1) in /usr/bin; /usr/sbin/install does not exist.
+        static let install = "/usr/bin/install"
+        static let sudo = "/usr/bin/sudo"
+        static let dscacheutil = "/usr/bin/dscacheutil"
+        static let killall = "/usr/bin/killall"
+
+        static let all: [String] = [install, sudo, dscacheutil, killall]
+    }
+
     private static let wgQuickCandidates = [
         "/opt/homebrew/bin/wg-quick",
         "/usr/local/bin/wg-quick",
@@ -96,7 +113,7 @@ extension CMUXCLI {
         }
         if interfaceUp {
             let downStatus = runInteractiveProcess(
-                executablePath: "/usr/bin/sudo",
+                executablePath: VPNTool.sudo,
                 arguments: [wgQuick, "down", configPath]
             )
             guard downStatus == 0 else {
@@ -107,7 +124,7 @@ extension CMUXCLI {
             }
         }
         let status = runInteractiveProcess(
-            executablePath: "/usr/bin/sudo",
+            executablePath: VPNTool.sudo,
             arguments: [wgQuick, "up", configPath]
         )
         guard status == 0 else {
@@ -156,7 +173,7 @@ extension CMUXCLI {
             throw CLIError(message: "wg-quick is not installed. Install with: brew install wireguard-tools")
         }
         let status = runInteractiveProcess(
-            executablePath: "/usr/bin/sudo",
+            executablePath: VPNTool.sudo,
             arguments: [wgQuick, "down", configPath]
         )
         guard status == 0 else {
@@ -232,9 +249,9 @@ extension CMUXCLI {
         // instead of prompting a second time. Cache invalidation does not
         // require privilege and runs as separate argv-based commands below.
         let status = runInteractiveProcess(
-            executablePath: "/usr/bin/sudo",
+            executablePath: VPNTool.sudo,
             arguments: [
-                "/usr/sbin/install",
+                VPNTool.install,
                 "-o", "root",
                 "-g", "wheel",
                 "-m", "644",
@@ -245,8 +262,8 @@ extension CMUXCLI {
         guard status == 0 else {
             throw CLIError(message: "Updating /etc/hosts failed (exit \(status)). Check the output above.")
         }
-        _ = runInteractiveProcess(executablePath: "/usr/bin/dscacheutil", arguments: ["-flushcache"])
-        _ = runInteractiveProcess(executablePath: "/usr/bin/killall", arguments: ["-HUP", "mDNSResponder"])
+        _ = runInteractiveProcess(executablePath: VPNTool.dscacheutil, arguments: ["-flushcache"])
+        _ = runInteractiveProcess(executablePath: VPNTool.killall, arguments: ["-HUP", "mDNSResponder"])
         if jsonOutput {
             print(jsonString(["hosts_changed": true, "machine_count": entries.count]))
         } else if !quiet {
@@ -304,7 +321,7 @@ extension CMUXCLI {
            let configPath = status?["config_path"] as? String,
            let wgQuick = Self.firstExecutable(Self.wgQuickCandidates) {
             _ = runInteractiveProcess(
-                executablePath: "/usr/bin/sudo",
+                executablePath: VPNTool.sudo,
                 arguments: [wgQuick, "down", configPath]
             )
         }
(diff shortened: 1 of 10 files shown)


JacobZwang:
Worked through the review. Eight taken, two not — reasons below.

**Fixed**

- **Root execution on a user-writable config (critical).** Real, and the worst of these. The helper now copies the config into a root-owned file with `PreUp`/`PostUp`/`PreDown`/`PostDown` stripped and applies that. Verified against a config carrying all four, including a lowercase unspaced `  preup=`: none survive, keys/addresses/MTU untouched.
- **Job not scoped per deployment.** Label and paths now derive from the interface name, and uninstall asks the app which tunnel it is rather than guessing.
- **Duplicated autostart implementation.** One definition in `CmuxFoundation`, which the app and CLI both already import. They now differ only in how they obtain privileges.
- **Install blocked the main actor** while the authorization dialog was up — now a cancellable off-actor task returning only its result.
- **Raw `osascript` stderr in the UI** — goes to the debug log; the banner gets product copy.
- **`MTU = 0` written through** — floored at IPv6's minimum, falling back to the safe ceiling.
- **Homebrew leaking into a user-facing error** — replaced with product copy.
- **Unlocalized CLI errors** — now `String(localized:defaultValue:)`.

**Not taken**

- **`last_used_at_unix` as route ownership.** Agreed it doesn't *prove* ownership. But a wrong guess fails closed: `start_connected` rejects on key mismatch, so it cannot connect to the wrong daemon — the failure mode is a retry, not a mis-connection. The previous behavior failed *unconditionally* and asked the caller for `--daemon FINGERPRINT`, which the app cannot answer since it holds an address and no way to name which dead occupant it means. This is strictly better than that, and the accompanying hint-eviction means new stores never become ambiguous in the first place; this path only serves stores written before it. Probing candidates and transferring the hint after a key match is the right end state, but it is a larger change in a crate that cannot be built or tested on macOS, so I would rather not land it unverified.
- **Static namespace → injected owner.** `CmuxVPNAutostart` is now a constructable value type (`init(interfaceName:)`) whose members are pure string generation; the injectable seams the guideline is aimed at (filesystem, process) live in the callers. Restructuring further would add indirection without changing what is testable.

PR #11659, opened 2026-09-02
macOS: dial private-network Cloud VMs through one cmux-tui wg hub (no sudo VPN)

lawrencecchen:
PR 2 of the sequence in `docs/cloud-userspace-wireguard.md` (stacked on #11638). Base is `feat-ios-userspace-wireguard`, not `main`.

Cloud VMs sit on the owner's Freestyle VPC with no public port, so today the Mac reaches a machine's cmux-tui daemon only after `cmux vpn up` (sudo wg-quick). This makes the app dial private-network machines through one in-process WireGuard tunnel instead, with no sudo and no system interface.

The app spawns one `remote connect` sidecar per machine, and one WireGuard key supports one live session, so the sidecars cannot each own a tunnel. A single `cmux-tui wg hub` process (added in #11638) owns the app's tunnel and serves SOCKS5 on an owner-only unix socket; each sidecar that targets a private-network route gets `--wireguard-hub <socket>` when the bundled client advertises the `wireguard-hub` capability. Public hosts and older clients dial directly, exactly as before.

The hub has its own tunnel identity, `VMTunnelManager.Identity.app` (`mac-<uuid>-app`, `app.key`, `cmux-app.conf`), derived from the same per-install device id so a reinstall rotates both. It never shares the `cmux vpn up` key, so the two tunnels never fight over the server-side endpoint. `CloudWireGuardHub` owns the lifecycle behind injected spawner, readiness probe, and clock: lease-counted per link, stops 10 s after the last release, restarts on unexpected exit with bounded backoff, pinned for the `cmux vm tui` pane (a client the app cannot watch), stopped on sign-out, killed on app quit. `vm.tunnel_status` reports the hub under `app_tunnel`.

The route string and the daemon are unchanged; only the client decides how to reach the address. Hub readiness is the `hub-ready` JSON line the `wg hub` process prints when its listener accepts, consumed the same way `CloudMachineLink` consumes a sidecar's `connection-snapshot` line, raced against the process exit and a bounded timeout through one resolve-once value (a task group cannot be used: an AsyncStream `for await` ignores cancellation, so the group would block draining the reader after the winner is picked). The `routes` in that line are the authoritative membership set for the hub decision.

[...]

PR #11638, opened 2026-09-02
cmux-wg: in-process WireGuard transport for direct routes (PR 1)

lawrencecchen:
First PR of the sequence in docs/cloud-userspace-wireguard.md, which lets an iOS or macOS client reach a Cloud VM's private-network daemon with no system WireGuard interface, no root, and no VPN prompt.

This PR is the Rust foundation. No app or UI code yet.

New crate cmux-wg runs WireGuard entirely in process: boringtun's sans-IO noise::Tunn holds the session, a smoltcp interface and TCP sockets provide a userspace TCP/IP stack, and one tokio task drives UDP recv/send, encapsulation, timers, and per-connection bridges. WgConfig::parse_wg_quick reads a Freestyle-issued wg-quick file (Address, MTU, AllowedIPs, Endpoint, PersistentKeepalive, PresharedKey). WgNet::connect returns a WgStream that implements AsyncRead and AsyncWrite, so callers cannot tell it apart from a kernel socket. Decapsulated packets are dropped unless their source is inside the peer's AllowedIPs.

cmux-remote gains a Dialer seam on the direct WebSocket provider. OsTcpDialer is the default and preserves today's behavior exactly. WireGuardDialer dials through a shared WgNet when the target address is inside the tunnel's routes and falls back to the OS otherwise. SocksDialer dials SOCKS5 CONNECT over a Unix socket. The route, TLS for wss, the WebSocket upgrade, and the Noise handshake are unchanged in every case.

[...]

PR #12004, opened 2026-09-05
Require the exit-with-parent cmux-tui client capability at install time

lawrencecchen:
The cloud helpers the app spawns (`remote connect`, `wg hub`) pass `--exit-with-parent` so a crashed or replaced app cannot leave a stale hub behind. This makes that a probed capability (`exit-with-parent` in `remote-probe --json`) and requires it wherever a client is bundled: `scripts/reload.sh`, `nightly.yml`, `release.yml`, `reload-build.yml`, next to the existing `wireguard-hub` guard. Shell tests cover the install script and the workflow greps.

Two parser tests are kept from the earlier revision of this branch: a heterogeneous graph row array is rejected without a crash, and snapshot rows order without the tuple-cast trap. They pass against main's parser, so the branch no longer carries its own parser rewrite; main fixed the trap in https://github.com/manaflow-ai/cmux/pull/12034.

Removed from this branch: the Go proof under `Prototypes/freestyle-wg-probe`. The in-process WireGuard transport it demonstrated ships in `cmux-wg` and the app's `wg hub` (https://github.com/manaflow-ai/cmux/pull/11638, https://github.com/manaflow-ai/cmux/pull/11659, https://github.com/manaflow-ai/cmux/pull/12042). The last revision that still carried it is 94a280b3a3c.

Nightly keeps `--max-fallback 5` on the client resolver: with the exact-commit form a merge would fail the next nightly until the cmux-tui artifacts for that commit publish.

https://claude.ai/code/session_01TPvGNpenmACZrUXehq47EJ

---

> [!NOTE]
> **Low Risk**
> Mostly capability probing and install-time guards; behavior change is rejecting older clients missing `exit-with-parent`, which is intentional until artifacts publish.
> 
> **Overview**
> Adds **`exit-with-parent`** to the capabilities advertised by `remote-probe --json`, so the app can tell whether bundled cloud helpers will die with their parent instead of leaving stale `remote connect` / `wg hub` processes behind.
> 
> **Install and CI gates** now require both `wireguard-hub` and `exit-with-parent` wherever `cmux-tui` is bundled or probed: `scripts/reload.sh`, nightly/release/reload-build workflows, and the fast-build probe Python checks. Shell tests (`test_install_cmux_tui_client.sh`, `test_nightly_universal_build.sh`) were updated to match.
> 
> Also adds a **Swift regression test** that snapshots with mixed-type `workspaces` rows are rejected without crashing the parser.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 968d2ca1123c89f689c90953e8df3730fe394ad4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lawrencecchen:
@JacobZwang could you evaluate this approach?

This is a greenfield CLI proof for Freestyle Cloud userspace WireGuard. It selects an authenticated VM, creates temporary cloud resources, runs WireGuard inside the process, and opens a temporary private PTY shell. It does not create a system VPN, TUN interface, or OS route. The main question is whether this is the right transport proof before integrating the path into cmux-tui's existing remote terminal session.

Please focus on:

- Whether the userspace WireGuard and temporary firewall design matches the intended subrouter model.
- Whether the cleanup and token-expiration handling is sufficient.
- Whether the next step should move the terminal path into cmux-tui instead of extending this custom bridge.

PR #12061, opened 2026-09-06
Fix tmux compatibility read fan-out rate limiting

### LINKED ISSUE #11803: Named-agent spawn fails with rate_limited: __tmux-compat display-message fans out 9 read-plane calls (burst=4) and self-drains its per-connection limiter
> **Correction (root cause updated).** My initial write-up blamed `pollingMethods` classification of `system.identify`/`window.current`. Runtime measurement disproved that: `cmux identify` (same method, direct) passes while `tmux display-message` (via `__tmux-compat`) fails, and 6/6 calls at 3s intervals fail even though each is a fresh process/connection. The real cause is below — a single `__tmux-compat` command self-drains its own per-connection token bucket.

### Summary

Spawning a **named** teammate from a `claude-teams` / `codex-teams` surface fails:

```
Error: Could not determine current tmux pane/window
```

because the `__tmux-compat` pane-identity command is rate-limited:

```
$ tmux display-message -p '#{session_name}:#{window_index}.#{pane_index}'
Error: rate_limited: Polling rate limited for this connection
```

### Root cause

A single `__tmux-compat display-message` invocation makes **9 read-plane (polling) calls on one `SocketClient` connection** via `tmuxFormatContext()` (`CLI/cmux.swift`). `ControlClientRateLimiter` is a per-connection token bucket with `burst = 4` (`Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Server/ControlClientRateLimiter.swift`), so the 5th call onward is rejected with `rate_limited` — the command **self-drains its own bucket within one invocation**.

Static enumeration of the polling-method calls made in one `tmuxFormatContext()` path:

| function | polling-method socket calls |
| --- | --- |
| `tmuxFormatContext` (direct) | `surface.current`, `pane.list`, `surface.list` |
| `resolveWorkspaceId` | `workspace.list`, `window.list`, `workspace.list`, `workspace.list`, `workspace.current` |
| `tmuxWorkspaceItems` | `workspace.list` |
| **total** | **9 calls, all in `pollingMethods`** |

Note `resolveWorkspaceId` calls `workspace.list` **3 times** in one invocation.

This explains the observations:
- `cmux identify` / `cmux top --processes` pass (1 call each ≤ burst).
- `tmux has-session` passes (no polling method).
- `tmux display-message` fails **every time**, including 6/6 at 3s intervals — each fresh connection still self-drains (9 > 4), so refill/backoff never helps.

### Reproduction

Static (no build):
```bash
C=<cmux checkout>
grep -n 'sendV2(method:' "$C/CLI/cmux.swift" | grep -E 'workspace\.|surface\.|pane\.|window\.'
```
Runtime:
```bash
tmux display-message -p '#{session_name}:#{window_index}.#{pane_index}'   # rate_limited
tmux has-session                                                          # passes
```

### Impact

Named subagent panes can't be spawned. The only workaround is `ALLOW_UNNAMED_AGENT=1`, which routes subagents to invisible in-process background tasks and disables subagent-pane observability.

### Fix direction

The per-connection polling bucket is meant to throttle spammy external polling, but here a **single** CLI command legitimately fans out 9 read-plane calls on one connection. Candidates:
1. **Reduce the fan-out.** `resolveWorkspaceId` calls `workspace.list` 3× per invocation — de-duplicating within one command drops 9 → 7 (still > 4, but a clear, isolated efficiency fix). Serving `tmuxFormatContext`'s context from a single aggregate read would drop it to ~1.
2. **Don't subject a single CLI command's internal read fan-out to the per-connection polling bucket** (e.g., serve fresh published read-plane snapshots without consuming a token — the 9 methods are all in `readSnapshotMethods`), or size the bucket to a single command's legitimate fan-out (sustained rate still capped by refill).

(a) snapshot-serving helps only if all 9 distinct methods are warm; (c) a dedicated connection does **not** help (a fresh connection still makes 9 calls — confirmed by the 6/6 failures).

Happy to send a PR — starting with the `resolveWorkspaceId` `workspace.list` de-duplication.

### Related
- #9917 — same symptom (agent-teams pane resolution breaking) but different root cause (phantom `dock:global` pane → "Pane has no surface to target").

### Environment
- macOS (Darwin 25.6), Apple Silicon, local build off `main` @ `68716b552b`

austinywang:
Fixes https://github.com/manaflow-ai/cmux/issues/11803

## Summary
- Add a behavior regression covering the nine read-plane calls made by `tmuxFormatContext`.
- Follow with a separate fix commit that admits the complete legitimate one-command fan-out while preserving the refill-based sustained polling limit.

## Trade-offs
- The per-connection burst increases only enough to cover the documented tmux compatibility read fan-out (`4 → 9`); the refill interval remains unchanged, so sustained polling is still throttled.
- This keeps the limiter at the control-socket request boundary rather than adding a tmux-specific bypass or weakening read-plane classification.
- A command-scoped batch protocol would be a broader redesign; the measured fan-out-sized burst is the smallest change that fixes the existing protocol without changing its security boundary.

## Verification
- Test-only regression commit: `4ab7795782`.
- Fix commit: `e67004ad9d`.
- `git diff --check` passes.
- GitHub package CI ran `CmuxControlSocket` with 432 tests across 55 suites; all passed.
- Tagged Debug app verification passed both exact probes through its isolated tmux shim:
  - `tmux display-message -p '#{session_name}:#{window_index}.#{pane_index}'` → `cmux:0.0`, exit 0.
  - `tmux has-session` → exit 0.
- Cloud reload was attempted first but backend provisioning failed because `cmux-dev-backend-1` was not resolvable; the required tagged local fallback build succeeded.

## CI note
- The dispatched full CI run's package tests pass. Its unrelated app-host shards fail on pre-existing missing `SurfaceCatalog` members, and the warning-budget gate reports the pre-existing `Sources/TerminalController.swift` `var` warning; neither failure is in this change.

### DIFF
--- a/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Server/ControlClientRateLimiter.swift
+++ b/Packages/macOS/CmuxControlSocket/Sources/CmuxControlSocket/Server/ControlClientRateLimiter.swift
@@ -10,6 +10,8 @@ public actor ControlClientRateLimiter {
     /// Token-bucket tuning values.
     public struct Configuration: Sendable, Equatable {
         /// Number of polling requests admitted immediately for a new client.
+        /// The default covers the current tmux compatibility read fan-out;
+        /// refill still bounds sustained polling.
         public let burst: Int
         /// Nanoseconds required to refill one token.
         public let refillIntervalNanoseconds: UInt64
@@ -19,7 +21,7 @@ public actor ControlClientRateLimiter {
         /// Values are clamped to a positive burst and a non-zero interval so
         /// the limiter cannot accidentally become an unbounded admission path.
         public init(
-            burst: Int = 4,
+            burst: Int = 9,
             refillIntervalNanoseconds: UInt64 = 100_000_000
         ) {
             self.burst = max(1, burst)
--- a/Packages/macOS/CmuxControlSocket/Tests/CmuxControlSocketTests/ControlPlaneConcurrencyTests.swift
+++ b/Packages/macOS/CmuxControlSocket/Tests/CmuxControlSocketTests/ControlPlaneConcurrencyTests.swift
@@ -157,6 +157,34 @@ struct ControlPlaneConcurrencyTests {
         #expect(await limiter.admit(method: "system.top") == .allowed)
     }
 
+    @Test func pollingLimiterAdmitsTmuxCompatReadFanoutBeforeRefill() async {
+        let clock = TestMonotonicClock()
+        let limiter = ControlClientRateLimiter(now: { clock.now })
+        let tmuxFormatContextMethods = [
+            "surface.current",
+            "pane.list",
+            "surface.list",
+            "workspace.list",
+            "window.list",
+            "workspace.list",
+            "workspace.list",
+            "workspace.current",
+            "workspace.list",
+        ]
+
+        for method in tmuxFormatContextMethods {
+            #expect(await limiter.admit(method: method) == .allowed)
+        }
+        guard case .limited(let retryAfter) = await limiter.admit(method: "surface.list") else {
+            Issue.record("the complete tmux compatibility read fan-out should fit in one burst")
+            return
+        }
+        #expect(retryAfter > 0)
+
+        clock.advance(by: 100_000_000)
+        #expect(await limiter.admit(method: "surface.list") == .allowed)
+    }
+
     @Test func readSnapshotPublishesAtomicallyAndKeysByMethodAndParams() {
         let store = ControlReadSnapshotStore()
         let initial = ControlCallResult.ok(.object(["generation": .int(1)]))


austinywang:
## Review audit

Re-checked against `HEAD` `e67004ad9d865df08496cac072b48e40cf2d6df1`.

| Comment ID | Author | File:line | Ask | Disposition | Commit SHA |
| --- | --- | --- | --- | --- | --- |
| `IC_kwDORDHQWM8AAAABS28gSA` | `coderabbitai` | — | Recent review reported no actionable comments. | `already-fixed` | `e67004ad9d865df08496cac072b48e40cf2d6df1` |
| — | GitHub review threads | — | GraphQL review-thread and submitted-review enumeration returned no review asks. | `already-fixed` | `e67004ad9d865df08496cac072b48e40cf2d6df1` |

PR #12117, opened 2026-09-08
Validate the devbox image pin and report image drift

lawrencecchen:
Merging an edit under `web/services/vms/images/devbox/` reaches no machine until a snapshot is baked and the manifest bump lands, and nothing said so. Two agents promoting at once is worse: a promotion rewrites twelve entries in one file, so they collide there, and the merge a human or agent writes by hand has no bake behind it.

Each default entry already records the commit it was baked from, so all of this is derivable with no schema change and no provider credential.

## `--pin`: states no bake can fix

- **A ladder assembled from two bakes.** Valid JSON, one default per kind and size, and sm running a different image from md. This is the exact shape of a hand-resolved `manifest.json` conflict.
- **A default baked off main.** `CMUX_BAKE_ALLOW_BRANCH=1` records a commit that squash merging never puts on main and that disappears when the branch is deleted, so the image's lineage stops being verifiable.

Safe to make a required check, which is why the `pull_request` trigger loses its path filter here: a required check that never runs blocks a PR forever, so the job now always reports and decides internally what to run.

## Drift: source ahead of the pinned image

Reports the image inputs that changed since the bake, and the promote commands. Expected right after an image source PR merges, and only a bake clears it, so this reports and never gates.

## Both states are live on main today

The defaults were baked from `39bfc41fad`, which exists only on `feat-vm-guest-trust-dead-code`; main carries that work as squash commit `dfb0a6fef7`. Main is also three image inputs ahead of that bake (`Dockerfile`, `cmux-bashrc`, `build-devbox-freestyle.ts`), so the Option+Backspace fix from #12099 is merged and on no machine. This workflow is not in the required set, so it reports both without blocking anyone; the honest fix is a bake from main.

Also adds `skills/cmux-backend/references/devbox-image-deploys.md`: what ships on merge (the manifest) versus what ships on a bake, the rule that a manifest conflict is re-promoted rather than hand-resolved (`devbox:promote --image <snapshot-id>` adopts a snapshot you already baked), and the fact that a bake is not reproducible, so every promotion carries unrelated agent and apt upgrades that the manifest diff is the place to review.

Follow-ups worth deciding separately: making this check required, and a `push: main` job that bakes, verifies and opens the manifest PR by itself.

---

> [!NOTE]
> **Medium Risk**
> Changes VM image CI and manifest validation; `--pin` can fail builds if enabled as required, and full drift checks depend on complete git history in Actions.
> 
> **Overview**
> Adds **`devbox:drift:check`** so merged devbox image source edits cannot silently outpace what VMs actually boot. It compares each default manifest entry’s `repoCommit` to current image inputs (devbox templates, desktop files, `build-devbox-freestyle.ts`) and lists changed paths, with a separate **`--pin`** mode for manifest states baking cannot fix (mixed bakes per kind after a bad `manifest.json` merge, or defaults baked from commits not on main).
> 
> The **Cloud VM image contract** workflow now runs on **every PR** (path filter removed so a required check cannot skip forever), uses **`fetch-depth: 0`** for historical file reads, runs **`--pin`** as a validating step, adds **`vm-image-drift.test.ts`**, and runs the full drift report as an informational step (workflow remains optional for merges). **`skills/cmux-backend/references/devbox-image-deploys.md`** documents merge vs bake shipping, promote flow, drift, and parallel-promotion conflicts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d25b8f7e990421b5ef7ffd3366ff41077db739e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/cloud-vm-image-contract.yml
+++ b/.github/workflows/cloud-vm-image-contract.yml
@@ -49,6 +49,9 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+          # The drift check reads the image inputs at each default entry's
+          # bake commit, which a shallow clone does not carry.
+          fetch-depth: 0
 
       - name: Setup Bun
         uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
@@ -62,4 +65,11 @@ jobs:
         run: bun run devbox:manifest:check
 
       - name: Test image contracts
-        run: bun test tests/vm-image-manifest.test.ts tests/vm-image-sizes.test.ts tests/vm-devbox-image.test.ts
+        run: bun test tests/vm-image-manifest.test.ts tests/vm-image-sizes.test.ts tests/vm-devbox-image.test.ts tests/vm-image-drift.test.ts
+
+      # A merged edit to the image source ships nothing until a snapshot is
+      # baked and the manifest bump lands. This step is the only place that
+      # says so. It does not gate the merge (the workflow is not required):
+      # source and image are allowed to move apart, but never silently.
+      - name: Report devbox image drift
+        run: bun run devbox:drift:check
--- a/.github/workflows/cloud-vm-image-contract.yml
+++ b/.github/workflows/cloud-vm-image-contract.yml
@@ -1,20 +1,10 @@
 name: Cloud VM image contract
 
 on:
+  # No path filter on pull_request: a required check that never runs blocks a
+  # PR forever, so the job always reports and decides internally what to run.
   pull_request:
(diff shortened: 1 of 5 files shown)


lawrencecchen:
Fleet instruction update for head `5ada303be2d7f286566263e72578911f32a164d4`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #12130, opened 2026-09-08
Speed up devbox promotion: parallel ladder, and wait for the guest not the clock

lawrencecchen:
A promotion took 784s. It now takes 221s, measured end to end on the same source and account, with two changes.

**Parallel ladder.** The six sizes are independent (boot the master, resize, snapshot, boot the result) and nearly all of that time is waiting, so they run together behind `--concurrency` (default 6, `1` restores serial). Verify overlaps them for the same reason: it only reads the master snapshot, and the manifest is still written only if both pass.

**Wait for the guest, not the clock.** Each size spent 60 of its ~81 seconds in two hardcoded 30-second sleeps, both immediately after a poll that had already proven the daemon was up, and verify spent another 60 the same way. The derive now retries the WebSocket smoke, which is the readiness signal it was sleeping for, and polls the daemon probe it ran afterwards anyway. Verify polls for the identity files and requires the digest to repeat once, so a machine mid-mint is never compared against another. These are strictly more robust than a fixed guess: a slow guest is waited for, up to a 90 to 120 second budget, instead of being assumed ready at 30.

| phase | before | parallel only | + polls |
| --- | --- | --- | --- |
| bake | 199s | 205s | 188s |
| verify | 131s | overlapped | overlapped |
| derive 6 sizes | 453s | 88s | 33s |
| **end to end** | **784s** | **335s** | **221s** |

Per size, seconds:

| size | serial | parallel | + polls |
| --- | --- | --- | --- |
| sm (2 vCPU, 3.8 GiB) | 37 | 37 | 9 |
| md (4, 7.8) | 79 | 81 | 22 |
| lg (8, 15.8) | 81 | 80 | 24 |
| lgx (12, 23.8) | 81 | 81 | 23 |
| xl (16, 31.8) | 83 | 81 | 26 |
| 2xl (32, 63.8) | 88 | 88 | 32 |

Also `apt-get clean` before the snapshot. The bake already dropped `/var/lib/apt/lists`, but every image still shipped 215 MiB of downloaded `.deb` files: a machine now boots with 5.8 GiB used instead of 6.0 GiB on a 16 GiB root.

**What this rules out**, verified against the API rather than the comment in the code: baking once on the largest shape and shrinking. Freestyle refuses on every axis, `resize is grow-only: requested 2 vCPU / 4096 MiB / 131072 MiB is below the current 32 vCPU / 65536 MiB / 131072 MiB on at least one axis`, and the same for storage alone. The ladder must be derived upward from the smallest bake, which is what it does.

**Also measured, and not worth doing:** the image is 5.8 GiB, of which `/usr` is 3.8 and `/opt` is 2.0 (1.5 of that `/opt/freestyle`, the provider's own, and 437 MiB Chrome). There is no `.cargo` or `.rustup` to reclaim, because the image installs a prebuilt cmux-tui rather than building it. After the apt cache there is nothing else of size that is ours.

Every test snapshot these measurements created was deleted from the account.

## Summary by CodeRabbit

- **Improvements**
  - Devbox image verification and size derivation now run concurrently, with labeled progress output.
  - Devbox size derivation supports configurable worker concurrency and preserves requested output order.
  - Resize, post-boot, and identity checks now wait for readiness through polling rather than fixed delays.
  - Image promotion validates both verification and size derivation before updating manifests.
  - Cleanup now clears cached package archives in addition to existing caches.
  - Identity validation now waits for stable machine identity data within a defined timeout.

### DIFF
--- a/web/scripts/build-devbox-freestyle.ts
+++ b/web/scripts/build-devbox-freestyle.ts
@@ -513,7 +513,7 @@ try {
   // lack the placeholder key approval. Drop them: each machine's first shell
   // seeds its own from the env. (The static codex config those shells wrote
   // is the same bytes on every machine and stays; the verifier checks it.)
-  await step("clean", `rm -rf /var/lib/apt/lists/* /root/.npm/_cacache ${WORK_HOME}/.npm/_cacache 2>/dev/null; rm -f /root/.claude.json ${WORK_HOME}/.claude.json; ${devboxJournalResetCommand}; sync; true`);
+  await step("clean", `apt-get clean; rm -rf /var/lib/apt/lists/* /root/.npm/_cacache ${WORK_HOME}/.npm/_cacache 2>/dev/null; rm -f /root/.claude.json ${WORK_HOME}/.claude.json; ${devboxJournalResetCommand}; sync; true`);
   await step("no-stale-claude-seed", `test ! -e /root/.claude.json && test ! -e ${WORK_HOME}/.claude.json && echo no-stale-claude-seed`);
 } catch (error) {
   console.error(`bake failed: ${String(error)}`);
--- a/web/scripts/derive-devbox-sizes.ts
+++ b/web/scripts/derive-devbox-sizes.ts
@@ -11,6 +11,12 @@
  * Usage:
  *   FREESTYLE_API_KEY=... bun scripts/derive-devbox-sizes.ts <master-snapshot-id> <slug-prefix>
  *       [--sizes sm,md,lg,lgx,xl,2xl] [--out <json>] [--replace-slug]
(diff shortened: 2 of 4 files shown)


lawrencecchen:
Fleet instruction update for head `ebe5c1d0490c1091a86dc28103d39b3331582f5b`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #3699, opened 2026-05-07
Fix #3463: store NODE_OPTIONS guard file under $HOME instead of $TMPDIR

psh4607:
## Summary

Store the Claude `NODE_OPTIONS` restore module under persistent user storage so it survives macOS cleanup of `$TMPDIR` between login sessions or reboots.

Both current Claude launch entrypoints now prefer `$HOME/.claude/cmux/cmux-claude-node-options` and fall back to `${TMPDIR:-/tmp}` only when `HOME` is unavailable.

## Current implementation

- `Resources/bin/cmux-claude-wrapper` — uses the persistent HOME-based guard directory.
- `CLI/cmux.swift` — applies the same location in the `claude-teams` path, keeping both entrypoints consistent.
- `tests/test_claude_wrapper_hooks.py` — sandboxes HOME and covers guard placement, invalid HOME, and TMPDIR cleanup.
- `tests/test_cli_claude_teams_env.py` — verifies the Swift CLI environment uses the same persistent location and fails open for unusable HOME.

The obsolete `Resources/bin/claude` path was deleted on `main` and was not restored during conflict resolution.

## Verification

- Tagged Debug app build succeeded with Zig 0.15.2.
- Both Python regression files parse successfully.
- GitHub reports the refreshed head as mergeable.
- Tests are left to CI per repository policy.

Closes #3463

PR #3278, opened 2026-04-29
fix: move NODE_OPTIONS restore guard out of $TMPDIR to user cache dir

flyinghanger:
## Summary

Move the `restore-node-options.cjs` guard out of `$TMPDIR` to a persistent **no-whitespace** cache path so it survives macOS `periodic` cleanup of `/var/folders/.../T/`, while avoiding the `Application Support` space breakage that sank #2983.

Fixes the recurring `Cannot find module '/var/folders/.../T/cmux-claude-node-options/restore-node-options.cjs'` failure originally reported in #2983 and contributes to the broader NODE_OPTIONS-injection issues tracked in #3059.

## The bug

`Resources/bin/claude`, `CLI/cmux.swift` and `daemon/remote/cmd/cmuxd-remote/agent_launch.go` all write the same restore shim under `${TMPDIR}/cmux-claude-node-options/` (or `os.TempDir()` / `NSTemporaryDirectory()` equivalents), then inject `NODE_OPTIONS=--require=<that path>` for the lifetime of the Claude session. macOS `com.apple.periodic-daily` reaps that path after ~3 days of no access, so every node child process from a long-running session crashes on module resolution.

## Why not `Application Support` (#2983)

Greptile flagged the previous attempt: `~/Library/Application Support` contains a space, and node splits `NODE_OPTIONS` on whitespace, so `--require=.../Application Support/...` is truncated at the first space and `MODULE_NOT_FOUND`s on **every** launch — replacing a 3-day-latent bug with an immediate one.

## New path layout (no whitespace)

| Source | New location |
|---|---|
| `XDG_CACHE_HOME` set | `\$XDG_CACHE_HOME/cmux/cmux-claude-node-options/` |
| macOS (default) | `~/Library/Caches/com.cmuxterm.app/cmux-claude-node-options/` |
| Linux (default) | `~/.cache/cmux/cmux-claude-node-options/` |

`Library/Caches` is the Apple-recommended location for regenerable runtime data and is **not** swept by `com.apple.periodic-daily` like `\$TMPDIR` is. The Go daemon uses `os.UserCacheDir()`, which resolves to the same destinations per platform.

## Files

- `Resources/bin/claude` — `ensure_node_options_restore_module()` now selects an XDG/macOS/Linux-specific cache dir.
- `CLI/cmux.swift` — `createClaudeNodeOptionsRestoreModule()` now uses `FileManager.url(for: .cachesDirectory, ...)` + bundle ID.
[...]

Issue #7028, opened 2026-06-28
Per-pane PATH shim (`cmux-cli-shims/<id>/claude`) reaped from `$TMPDIR` after ~3 idle days → long-lived pane silently runs `claude` unwrapped, no hooks fire


mrshu:
## Summary

cmux installs a per-pane PATH shim at `$TMPDIR/cmux-cli-shims/<PANEL_ID>/claude` (a small script that `exec`s `cmux-claude-wrapper "$@"`) and prepends that dir to the pane's `PATH`, so typing `claude` gets the hook-injecting wrapper (`--settings <hooks>` + `--session-id` + `CMUX_CLAUDE_HOOK_*`).

macOS's periodic temp cleaner (`dirhelper`) deletes files under the per-user `$TMPDIR` (`/var/folders/.../T/`) that haven't been **accessed in ~3 days**, leaving the now-empty shim directory behind. In a **long-lived pane** that hasn't launched `claude` in 3+ days, the shim file is reaped but the (empty) shim dir is still first on `PATH`. A subsequently hand-typed `claude` then falls through to the real binary (`~/.local/bin/claude`) and runs **unwrapped** — so it injects no hooks, emits no `agent.hook.*`/`set_status`, and the session is invisible in the sidebar (and to anything consuming `cmux events`) for its entire life.

This is the same root cause as #3463 (macOS reaping cmux's `$TMPDIR` artifacts) but for the **PATH shim** rather than the `restore-node-options.cjs` file, and the consequence is strictly worse: #3463 is a noisy-but-non-blocking error, this is **silent and blocking** (no hooks at all, no error shown).

## Environment

- cmux 0.64.16 (96)
- macOS 15.7.7
- Claude Code 2.1.195
- Shell: fish (irrelevant to the bug — `PATH` ordering is correct; see below)

## Symptom

A pane's Claude session runs normally but never appears as Running/Idle/Needs in the sidebar, and `cmux events` shows zero `agent.hook.*` / `set_status` for its `workspace_id`, even though the agent is actively working.

## Root cause / evidence

The silent session's env is intact and PATH ordering is correct — the shim dir is **first**, ahead of `~/.local/bin`:

```
CMUX_CLAUDE_WRAPPER_SHIM=/var/folders/.../T/cmux-cli-shims/F0B5A305-.../claude
CMUX_CLAUDE_WRAPPER_SHIM_ROOT=/var/folders/.../T/cmux-cli-shims/F0B5A305-...
[...]

PR #12208, opened 2026-09-09
Keep the NODE_OPTIONS restore module outside the swept temp directory

### LINKED ISSUE #3463: NODE_OPTIONS --require path in $TMPDIR becomes stale after macOS temp cleanup
## Summary

cmux's `claude` wrapper script creates `restore-node-options.cjs` in `$TMPDIR/cmux-claude-node-options/`, then sets `NODE_OPTIONS=--require /path/to/restore-node-options.cjs` before launching Claude Code.

macOS periodically cleans up `$TMPDIR` (`/var/folders/...`) between sessions. When Claude Code spawns child Node.js processes, they inherit the `NODE_OPTIONS` environment variable, which still references the now-deleted `.cjs` file — resulting in a noisy (but non-blocking) error on session stop.

## Error message

```
Cannot find module '/var/folders/fw/8t708g857plftvkpw2v1yhmh0000gn/T/cmux-claude-node-options/restore-node-options.cjs'
```

## Steps to reproduce

1. Launch Claude Code via cmux
2. End the session (or reboot Mac, which clears `$TMPDIR`)
3. Launch Claude Code via cmux again
4. On session stop, the error appears in the child Node process

## Root cause

```bash
# In /Applications/cmux.app/Contents/Resources/bin/claude:
guard_dir="${TMPDIR:-/tmp}/cmux-claude-node-options"
guard_path="$guard_dir/restore-node-options.cjs"
# ... writes restore-node-options.cjs to $TMPDIR ...
export NODE_OPTIONS="--require $guard_path ..."
```

`$TMPDIR` on macOS resolves to `/var/folders/...`, which is session-scoped and cleaned between logins/reboots. After cleanup, `guard_path` no longer exists, but the `--require` path stays in `NODE_OPTIONS` and is inherited by all child Node processes.

## Suggested fix

Store `restore-node-options.cjs` in a persistent location instead of `$TMPDIR`:

```bash
# Option A: use ~/.claude/cmux/ (persists across sessions)
guard_dir="$HOME/.claude/cmux"

# Option B: recreate the file if missing (guard at Node require time)
# wrap in try/catch or check existence before --require
```

## Environment

- macOS (Apple Silicon)
- Node.js v24.14.0
- cmux v0.63.2

ibrahimhajjaj:
Fixes #3463.

The restore shim is written to `$TMPDIR` but referenced by `NODE_OPTIONS` for the life of the session. macOS sweeps that directory, and once the file is gone every node process in the session dies at preload, reporting little more than a version string. It is not limited to the wrapped agent: anything that inherits the variable is affected, and a failed CLI write can look like a successful one, which is how I found it.

There are earlier attempts at this and I do not want to add noise, so for the record: #3699 has green checks but has gone stale and now conflicts with main; #12067 conflicts and its `tests` check is failing; #11270 and #3278 also have failing checks. This branch is cut from today's main and `python3 tests/test_claude_wrapper_hooks.py` passes. Happy for it to be closed in favour of any of those if one gets rebased.

The change: move the shim to `$HOME/.cmux/node-options`, matching the `$HOME/.claude` and `$HOME/.subrouter` paths already used in this file, with a `CMUX_NODE_OPTIONS_DIR` override following the existing `CMUX_CUA_STATE_DIR` style.

Not `~/Library/Application Support/cmux`: node splits NODE_OPTIONS on whitespace, so a `--require=` under a path containing a space breaks every node process instead. I tried that first and the suite caught it, so there is a comment to stop it being moved back. Might be worth checking against the failing runs on the other branches.

Two existing tests pin the old path, which is the part that looks like it catches people:

- `test_live_socket_tmpdir_failure_skips_node_options_injection` becomes `..._guard_dir_failure_...` and forces the failure through `CMUX_NODE_OPTIONS_DIR`, since an unwritable `TMPDIR` no longer reaches the guard dir. Same invariant: when the module cannot be written, no `--require` is injected.
- `test_live_socket_stale_mktemp_literal_does_not_warn` points the override at its own temp dir so it still exercises the real path.
- `run_wrapper` gains an `extra_env` parameter.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12208"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791531542&installation_model_id=21920&pr_number=12208&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12208&signature=24dbbfd68c397686ed3e5edb4bab31c7df12c9b37aa1b1be4049b4832df80ce8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Changes where `NODE_OPTIONS` preload files live for every Node descendant in a cmux Claude session; a bad path or permissions could skip injection or break startup, but behavior on write failure stays “no inject.”
> 
> **Overview**
> Fixes session-wide Node failures when macOS sweeps temp files while `NODE_OPTIONS` still points at cmux’s `--require` restore shim.
> 
> The restore module is written to **`~/.cmux/cmux-node-options`** (override **`CMUX_NODE_OPTIONS_DIR`**) in both **`cmux-claude-wrapper`** and **`createClaudeNodeOptionsRestoreModule`** in Swift, with comments that the path must stay **space-free** (Node splits `NODE_OPTIONS` on whitespace) and still include **`/cmux-`** so existing path recognition keeps working.
> 
> Tests gain **`extra_env`** on **`run_wrapper`**, rename the unwritable-temp case to guard-dir failure via **`CMUX_NODE_OPTIONS_DIR`**, and point the stale-`mktemp` test at the new directory name.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25968a260b211bce91c403baeb3cf316becf1ade. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Moves the NODE_OPTIONS restore shim from `$TMPDIR` to `$HOME/.cmux/cmux-node-options` (overridable via `CMUX_NODE_OPTIONS_DIR`) so macOS temp sweeps no longer delete the preload file and break every node process that inherits `NODE_OPTIONS` for the session. The path stays space-free because node splits `NODE_OPTIONS` on whitespace.

- Aligns the Swift CLI's shim writer with the wrapper's guard dir and keeps the `/cmux-` path component so `isCmuxNodeOptionsRestoreModulePath` still recognizes the shim.
- Tests gain an `extra_env` parameter on `run_wrapper`; the unwritable-dir injection test now forces failure via `CMUX_NODE_OPTIONS_DIR`, and the stale mktemp literal test points at the new location.

<sup>Written for commit 25968a260b211bce91c403baeb3cf316becf1ade. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Node.js startup protection files are now stored in a dedicated, persistent directory rather than temporary storage.
  - This prevents temporary-directory cleanup from removing required files and causing Node.js startup failures.
  - The storage location can be customized with the `CMUX_NODE_OPTIONS_DIR` environment variable.

- **Tests**
  - Added coverage for custom guard-directory configuration and failures affecting that directory.

### DIFF
--- a/CLI/cmux.swift
+++ b/CLI/cmux.swift
@@ -23703,16 +23703,19 @@ struct CMUXCLI {
     }
 
     private func createClaudeNodeOptionsRestoreModule() throws -> URL {
-        let rawTemporaryDirectory = ProcessInfo.processInfo.environment["TMPDIR"]?
+        // Must match the guard_dir in Resources/bin/cmux-claude-wrapper. NODE_OPTIONS
+        // outlives the session, so this cannot live under TMPDIR, and the path has to
+        // keep "/cmux-" for isCmuxNodeOptionsRestoreModulePath to recognise it.
+        let overrideDirectory = ProcessInfo.processInfo.environment["CMUX_NODE_OPTIONS_DIR"]?
             .trimmingCharacters(in: .whitespacesAndNewlines)
-        let temporaryDirectory: String
-        if let rawTemporaryDirectory, !rawTemporaryDirectory.isEmpty {
-            temporaryDirectory = rawTemporaryDirectory
+        let root: URL
+        if let overrideDirectory, !overrideDirectory.isEmpty {
+            root = URL(fileURLWithPath: overrideDirectory, isDirectory: true)
         } else {
-            temporaryDirectory = NSTemporaryDirectory()
+            root = FileManager.default.homeDirectoryForCurrentUser
+                .appendingPathComponent(".cmux", isDirectory: true)
+                .appendingPathComponent("cmux-node-options", isDirectory: true)
         }
-        let root = URL(fileURLWithPath: temporaryDirectory, isDirectory: true)
-            .appendingPathComponent("cmux-claude-node-options", isDirectory: true)
         try FileManager.default.createDirectory(at: root, withIntermediateDirectories: true, attributes: nil)
         let restoreModuleURL = root.appendingPathComponent("restore-node-options.cjs", isDirectory: false)
         try writeShimIfChanged(Self.claudeNodeOptionsRestoreModule, to: restoreModuleURL)
--- a/Resources/bin/cmux-claude-wrapper
+++ b/Resources/bin/cmux-claude-wrapper
@@ -851,9 +851,13 @@ ensure_node_options_restore_module() {
     # NODE_OPTIONS is inherited by every descendant for the life of the session, so
     # the module it points at has to outlive the session. TMPDIR is swept, and once
     # the file goes missing node refuses to start any script at all.
-    # Keep this path free of spaces: node splits NODE_OPTIONS on whitespace, so a
-    # --require= under "Application Support" breaks every node process instead.
-    local guard_dir="${CMUX_NODE_OPTIONS_DIR:-${HOME}/.cmux/node-options}"
+    # Two constraints on this path, both load-bearing:
+    #   - no spaces, because node splits NODE_OPTIONS on whitespace, so a
+    #     --require= under "Application Support" breaks every node process instead
+    #   - it must still contain "/cmux-", which is how
+    #     isCmuxNodeOptionsRestoreModulePath recognises our own shim when stripping
+    #     injected flags on restore
+    local guard_dir="${CMUX_NODE_OPTIONS_DIR:-${HOME}/.cmux/cmux-node-options}"
     local guard_path="$guard_dir/restore-node-options.cjs"
     mkdir -p "$guard_dir" || return 1
     local temp_path
--- a/Resources/bin/cmux-claude-wrapper
+++ b/Resources/bin/cmux-claude-wrapper
@@ -848,8 +848,12 @@ exec_real_claude_passthrough() {
 REAL_CLAUDE="$(find_real_claude)" || { echo "Error: claude not found in PATH" >&2; exit 127; }
 
 ensure_node_options_restore_module() {
-    local guard_dir="${TMPDIR:-/tmp}"
-    guard_dir="${guard_dir%/}/cmux-claude-node-options"
+    # NODE_OPTIONS is inherited by every descendant for the life of the session, so
+    # the module it points at has to outlive the session. TMPDIR is swept, and once
+    # the file goes missing node refuses to start any script at all.
+    # Keep this path free of spaces: node splits NODE_OPTIONS on whitespace, so a
+    # --require= under "Application Support" breaks every node process instead.
+    local guard_dir="${CMUX_NODE_OPTIONS_DIR:-${HOME}/.cmux/node-options}"
     local guard_path="$guard_dir/restore-node-options.cjs"
     mkdir -p "$guard_dir" || return 1
     local temp_path
--- a/tests/test_claude_wrapper_hooks.py
+++ b/tests/test_claude_wrapper_hooks.py
@@ -68,6 +68,7 @@ def run_wrapper(
     hooks_disabled: bool = False,
     setup_sandbox=None,
     process_timeout: float | None = None,
+    extra_env: dict[str, str] | None = None,
 ) -> tuple[int, list[str], list[str], str, str, str, str, str, str, str]:
     with tempfile.TemporaryDirectory(prefix="cmux-claude-wrapper-test-") as td:
         tmp = Path(td)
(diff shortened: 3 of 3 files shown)


ibrahimhajjaj:

ibrahimhajjaj:
recheck

ibrahimhajjaj:
Both correct, thanks. Confirmed each in the code and pushed 25968a2.

1. `isCmuxNodeOptionsRestoreModulePath` matches on `contains("/cmux-")`, and `$HOME/.cmux/node-options/...` does not contain that substring, so the sanitizer would have stopped recognising cmux's own shim and left the injected `--require` and heap cap in `NODE_OPTIONS` on restore. Default is now `$HOME/.cmux/cmux-node-options`, which keeps the substring, so the matcher and its tests are untouched. Comment added at both ends so the prefix does not get tidied away later.

2. `createClaudeNodeOptionsRestoreModule` in `CLI/cmux.swift` was still building its root from `TMPDIR`. It now resolves the same directory as the wrapper and honours `CMUX_NODE_OPTIONS_DIR`, so there is one path rather than two.

The shell suite still passes and the Swift file parses. I cannot build the macOS target locally, so the Swift side is relying on CI.

RuslanTar:
Another data point in favour of getting this merged. Reproduced on cmux 0.64.25 (latest), macOS 27.0 (26A428).

**Timeline on one machine**

- 09-19 19:06: the wrapper creates `$TMPDIR/cmux-claude-node-options/restore-node-options.cjs`.
- Every later launch hits the `cmp -s` branch in `ensure_node_options_restore_module`, so the file is never rewritten. Reads by node's preload don't update its atime either: after hundreds of hook runs the atime still equals the birth time.
- 09-23 03:35: `com.apple.bsd.dirhelper` (`CLEAN_FILES_OLDER_THAN_DAYS=3`) removes it on its first run after the file turned 3 days old.
- 03:36–04:17: every node hook in every open Claude session fails with `Cannot find module '…/restore-node-options.cjs'` (about 40 failures across Stop and PreToolUse hooks).
- 04:17: the next wrapper launch writes the file again and the errors stop.

So it isn't a rare edge case. Any machine that keeps a Claude session open overnight hits it on a regular cycle, about every 4 days.

**Why it matters beyond noise:** Claude Code treats a hook that crashes before running as a non-blocking error. Every node-based PreToolUse guard (secret-read guards, write/path guards) therefore **fails open** until something launches `claude` again. A missing preload silently switches off user safety hooks.

**On the fix:** the approach in this PR looks right to me: a durable, space-free path, plus the `CMUX_NODE_OPTIONS_DIR` override. One thing to watch during rollout: sessions started before the upgrade keep `--require=<old $TMPDIR path>` in their environment until they exit. Those can still break once the old file is swept, unless the new wrapper keeps writing (or at least touching) the old path for a transition period.

#7028 is the same bug class for the per-pane `cmux-cli-shims/<surface>/claude` shim. That one fails differently: the pane silently runs `claude` unwrapped with no hooks. It might be worth moving both out of `$TMPDIR` in one pass.

PR #12210, opened 2026-09-09
Fix terminal colors and pane clipping

### LINKED ISSUE #12204: Inactive terminal pane dimming washes out colors in mixed split layouts
## Problem

In a workspace with terminal, browser, and Dock panes, unfocused terminal panes are covered by a strong light/gray wash while the browser and Dock keep normal contrast. Terminal palettes and TUI status colors become pastel and low-contrast, and the different surface types no longer look like parts of one coherent workspace.

This is especially conspicuous with a light terminal theme. The current terminal appearance path defaults `unfocused-split-opacity` to `0.7`, which resolves to a 30% background-colored overlay on inactive terminal panes.

## Reproduction

1. Use a light terminal theme.
2. Create a workspace with two terminal splits and a browser pane (a visible Dock terminal makes the inconsistency even clearer).
3. Focus the browser pane.
4. Compare the inactive terminal content with the active browser and Dock content.

## Actual

- Inactive terminal panes look bleached/disabled rather than merely unfocused.
- ANSI/TUI colors lose saturation and text contrast.
- Browser and Dock surfaces remain full-contrast, so focus styling is inconsistent across a mixed layout.

## Expected

Inactive panes should remain comfortably readable and preserve the terminal palette. Focus should still be obvious, but via a subtle, theme-aware treatment that feels consistent across terminal, browser, and Dock surfaces.

Please investigate the product-level treatment rather than only tweaking one constant:

- Reconsider the default strength/compositing of the inactive terminal overlay, particularly for light backgrounds.
- Keep explicit Ghostty `unfocused-split-opacity` configuration working as intended.
- Decide whether the default should rely more on the focused-pane border/accent instead of washing pane contents.
- Check light, dark, transparent, terminal-only, and mixed terminal/browser/Dock layouts.
- Add focused coverage for appearance resolution and inactive-surface presentation where practical.

Related but not duplicate: #6709 / #6899 concern adding an obvious focused-pane cue; #3272 concerns surface-tab styling rather than pane content.

austinywang:
Fixes #12204. Inactive terminals now keep full contrast by default, while explicit Ghostty dimming settings continue to work. Fresh-install adaptive colors use Catppuccin Latte/Mocha, including the fallback when theme files are unavailable. Terminal surfaces stay clipped to their panes during resizing and portal rebinding, with terminal dimensions derived from the actual scroll viewport.

This includes follow-up #12295, an authoritative-snapshot correction to the existing wheel-scroll test fixture, and independently specified fallback palette coverage. The fallback regression was committed before its fix and failed for both light and dark on the pre-fix commit.

Validation:
- Tagged app rebuilt and installed for pushed HEAD `24125c7e4a`.
- Xcode Release terminal-core suite: all 313 tests in 55 suites passed (exit 0).
- macOS 15 GhosttySurfaceOverlayTests: https://github.com/manaflow-ai/cmux/actions/runs/34558113554 (test and summary steps passed).
- The original wheel test failed identically on the parent commit and the original follow-up head; its fixture now supplies the same authoritative response as the production input path.
- Fleet runtime check: three terminals and a focused browser survived 12 window resizes and workspace switching. Terminal output and live geometry remained intact; inspected before/after screenshots show containment and full inactive-pane contrast.
- Built app contains both managed theme resources, matching their tracked contents. Package-lock policy, test wiring, and whitespace checks pass.
- Localization audit: no new user-facing strings; existing theme names and resources are reused.

Production requires the normal macOS app release. No backend deployment, database migration, secret, or remote feature-flag change is needed. Managed defaults apply only when adaptive defaults are enabled and the user's effective Ghostty configuration has no directives. Explicit opacity remains authoritative; a fill-only setting retains Ghostty's normal dimming behavior.

### DIFF
--- a/Packages/macOS/CmuxTerminalCore/Sources/CmuxTerminalCore/Config/GhosttyConfig.swift
+++ b/Packages/macOS/CmuxTerminalCore/Sources/CmuxTerminalCore/Config/GhosttyConfig.swift
@@ -59,13 +59,12 @@ public struct GhosttyConfig {
     public var command: String?
     /// The scrollback limit. Ghostty measures this in bytes, not lines.
     public var scrollbackLimit: Int = 50_000_000
-    /// The opacity (0...1) applied to unfocused split panes.
-    public var unfocusedSplitOpacity: Double = 0.7
-    /// The fill color for the unfocused-split overlay, or `nil` to use the
-    /// background color.
+    /// The opacity (0...1) applied to unfocused split panes; the cmux default keeps terminal content at full contrast.
+    public var unfocusedSplitOpacity: Double = 1.0
+    private var hasUnfocusedSplitOpacityDirective = false
+    /// The unfocused-split overlay fill, or `nil` to use the background color.
     public var unfocusedSplitFill: NSColor?
-    /// The explicit split-divider color, or `nil` to derive one from the
-    /// background.
+    /// The split-divider color, or `nil` to derive one from the background.
     public var splitDividerColor: NSColor?
 
     // Colors (from theme or config)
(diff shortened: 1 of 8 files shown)


austinywang:

austinywang:
recheck

PR #12463, opened 2026-09-13
Drop an unreadable claude --settings path instead of failing the launch

danielraffel:
## Problem

`cmux restore claude <id>` can hard-fail with:

```
cmux: --settings merge error: ENOENT: no such file or directory, open '/var/folders/.../T/subrouter-claude-settings-3294281412/settings.json'
cmux: warning: --settings merge failed; your --settings was ignored
Error: Settings file not found: /var/folders/.../subrouter-claude-settings-3294281412/settings.json
```

The warning is not true. The wrapper collects user `--settings` values into `CMUX_USER_SETTINGS` and builds `CMUX_FILTERED_ARGS` with them removed, but it only swaps them in on the **success** path:

- success: `HOOKS_JSON="$CMUX_MERGED_SETTINGS"; set -- "${CMUX_FILTERED_ARGS[@]}"`
- failure: prints the warning and **never runs `set --`**

So `$@` still carries the unreadable path, and the final `exec "$REAL_CLAUDE" --settings "$HOOKS_JSON" "$@"` hands Claude **two** `--settings` flags. Claude ≥ 2.1.169 is last-wins (as the wrapper's own comment documents), so the dead path wins and Claude refuses to start.

This is reachable for any launcher that hands Claude a transient settings file. Subrouter's `sr claude proxy` writes one under `$TMPDIR/subrouter-claude-settings-<rand>/` and deletes it when `sr` exits, so every restore of such a session hits it — a reboot is not required.

## Change

`Resources/bin/cmux-claude-wrapper`: an unreadable user `--settings` path is dropped rather than forwarded, so exactly one `--settings` (cmux's hooks) reaches Claude and the warning becomes accurate. A readable settings file is still merged exactly as before.

## Testing

Two-commit regression pattern. `tests/test_claude_wrapper_dead_settings_path.py` (new, wired into `.github/workflows/ci.yml`) drives the real wrapper with a fake `claude` that records its argv, across five scenarios: restore replay, the `--settings=` equals form, a mix of a dead path and valid inline JSON, a tilde path, and the no-Node fallback.

At the test-only commit it **exits 1** and reproduces the literal `Error: Settings file not found`. At the fix commit it **exits 0**. `swift test --package-path Packages/macOS/CMUXAgentLaunch` stays green (399 tests / 55 suites), and the sibling wrapper suites still pass.

Follow-up PR stacks on this one to stop recording dead paths in restore plans at all, and to resume Subrouter-routed Claude sessions through the launcher.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12463"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791864272&installation_model_id=21920&pr_number=12463&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12463&signature=72e99d2141389cf49eb4830e7341705417ddcb6b4f7254390cec4d7b5ed30dce"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes `cmux restore claude` hard-failing when a captured `--settings` path no longer exists, so restored sessions launch with cmux hooks instead of Claude refusing to start. An unreadable `--settings` path is now dropped with a warning only for cmux-owned restore replays; readable settings are still merged as before.

**Bug Fixes**
- Unreadable `--settings` paths are dropped only on cmux-owned restore replays; interactive launches still fail closed so an invalid settings file can't silently weaken Claude's policy.
- The regression test drives the real wrapper across seven scenarios, including the interactive fail-closed contract and the no-Node fallback.
- CI now runs the new test.

<sup>Written for commit 43f5d7d4a1772f9b045c08b3f3ccdebccb096b52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Claude launches no longer forward unreadable or nonexistent settings-file paths.
  * Valid inline settings and readable settings files continue to work.
  * Built-in hook settings, prompts, resume arguments, and other launch options are preserved.
  * Clear warnings are shown when invalid settings paths are skipped.

* **Tests**
  * Added regression coverage for separate and equals-form settings, tilde paths, mixed inputs, and environments without Node.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1611,6 +1611,7 @@ jobs:
           CMUX_CLI_BIN="$CLI_BIN" python3 tests/test_cli_socket_autodiscovery.py
           python3 tests/test_codex_wrapper_resume_hooks.py
           python3 tests/test_claude_wrapper_hooks.py
+          python3 tests/test_claude_wrapper_dead_settings_path.py
           python3 tests/test_codex_wrapper_computer_use_mcp.py
           python3 tests/test_cmux_cua_mcp_smoke.py
           python3 tests/test_cmux_cua_build_cache_safety.py
--- a/Resources/bin/cmux-claude-wrapper
+++ b/Resources/bin/cmux-claude-wrapper
@@ -1632,9 +1632,18 @@ unset CMUX_GENERATED_HOOKS_JSON
 # does not rely on the CLI's multi-flag behavior. Requires `node`; if node is
 # missing or the merge fails, preserve the user's original settings and disable
 # cmux hooks for that launch rather than passing an ambiguous pair of flags.
-# A --settings path that is not readable is dropped up front with a warning:
-# Claude would refuse to start on it, so forwarding it (with or without cmux's
-# hooks) can only fail the launch.
+# An unreadable --settings path is dropped up front with a warning ONLY when
+# cmux owns this launch as a restore replay. There the path came from a launch
+# record cmux itself captured, the launcher that created it (for example
+# `sr claude proxy`) has already deleted it, and forwarding it can only fail a
+# restore the user asked for.
+#
+# An ordinary interactive launch keeps failing closed. A settings file can
+# define permission rules and hooks, so silently starting Claude WITHOUT the
+# policy the user named -- because of a typo, a permission change, or an
+# unmounted volume -- would turn an invalid configuration into a quietly
+# less-restricted session. Claude refusing to start is the correct outcome
+# there, and it is the user's own argument to fix.
 unset CMUX_USER_SETTINGS_B64
 CMUX_FILTERED_ARGS=()
 CMUX_USER_SETTINGS=()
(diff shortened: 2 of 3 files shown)


danielraffel:
Thanks — both findings were valid and are fixed in `43f5d7d`.

**Limit missing-settings removal to restore replays** (P1): you are right, and this was the more important bug of the two. Dropping an unreadable `--settings` on every launch turned fail-closed configuration handling into fail-open: a typo, a permission change, or an unmounted volume would have started Claude *without* the permission rules and hooks the user named.

The drop is now restricted to a launch cmux owns as a restore replay (`cmux_claude_app_restore_owned == 1`), where the path came from a record cmux itself captured and the launcher that created it has already deleted it. An ordinary interactive launch forwards the path and lets Claude refuse to start, which is the correct outcome and is the user's own argument to fix. `test_interactive_launch_keeps_failing_closed_on_an_unreadable_settings_file` pins that contract; the restore-shaped scenarios now opt in explicitly via `CMUX_AGENT_RESTORE_LAUNCH`.

**Require both warning markers in the equals-form test** (minor): confirmed — `DEAD_SETTINGS_PATH not in stderr or "warning" in stderr` also passed when neither value appeared. It now requires both.

Verification: the regression test still exits 1 at the test-only commit reproducing the literal `Error: Settings file not found`, and exits 0 at the fix commit. `swift test --package-path Packages/macOS/CMUXAgentLaunch` is green, as are the sibling wrapper suites.

PR #12464, opened 2026-09-13
Restore Subrouter-routed claude sessions through sr claude proxy --resume

danielraffel:
> **Stacks on #12463.** The first two commits here are that PR; review the top four. Once #12463 merges this diff shrinks to those.

## Problem

`sr claude proxy` hands Claude a private per-launch `--settings` file under `$TMPDIR/subrouter-claude-settings-<rand>/` and deletes it when `sr` exits. cmux captures the pre-injection argv into `CMUX_AGENT_LAUNCH_ARGV_B64` and replays it on restore, so the recorded path is always dead.

#12463 stops that from being fatal, but the restored session is still wrong: it gets only `ANTHROPIC_BASE_URL` and `CLAUDE_CONFIG_DIR` via `CMUX_PRESERVE_CLAUDE_AUTH_SELECTION_ENV`. The `ANTHROPIC_AUTH_TOKEN` and the `X-Subrouter-Agent` / account-ID headers lived **only** in the deleted file, so account selection, pooling and failover are all bypassed. Re-invoking the launcher regenerates them — which is exactly how Codex restores already behave.

## Changes

**Stop recording dead paths** (`ClaudeRestoreSettingsPathFilter.swift`, `AgentRestoreReadableFileResolver.swift`): a `--settings` path whose file is unreadable is dropped when the restore plan is built. Plan time rather than capture time deliberately — a captured path is valid when captured, and a user settings file that merely moved must not be forgotten; plan time is the only point where a replayed path can be checked against the filesystem it will actually run on. Inline JSON, empty values, a dangling `--settings`, and anything after `--` are untouched.

**Route the resume** (`SubrouterClaudeResumeRouting.swift`): a Claude restore is rewritten to `sr claude proxy --resume <id>` only when the record carries an **agreeing pair** of markers:

1. `SUBROUTER_CLAUDE_RESUME_COMMAND` — exported by `sr` (manaflow-ai/subrouter#327), exact text only.
2. `CMUX_AGENT_LAUNCH_SUBROUTER_CLAUDE_RESUME_COMMAND` — minted by `cmux-claude-wrapper`, which first `unset`s any inherited copy and re-exports only when the argv still carries an **existing** `subrouter-claude-settings-*/settings.json`.

Marker 1 alone is insufficient because it leaks to every descendant — a plain `claude` run inside an `sr claude` session inherits it. The bound copy is minted per launch against physical evidence only `sr` creates, and the wrapper is the last process that can see that file: the settings merge removes user `--settings` before the argv is captured.

**Provenance never reads `ANTHROPIC_BASE_URL`**, so a local `http://127.0.0.1:31415` pool and a hosted `https://sr.cmux.com` pool behave identically (both are parameterized test cases). A local base URL by itself is explicitly not provenance.

**Fails closed.** Marker-only, bound-only, disagreeing pair, non-exact text (`--restore`, `; rm -rf /`, an absolute path, `cx`, uppercase, empty), cmux's own launchers, `.direct` mode, codex records, and `sr` not resolvable on the restore `PATH` all leave today's plan byte-for-byte unchanged.

## Testing

Two-commit regression pattern. New: `SubrouterClaudeRestoreRoutingTests.swift` (15 planner/policy tests), `SubrouterClaudeResumeRoutingTests.swift` (7 unit tests), `ClaudeRestoreSettingsPathTests.swift` (7 tests), `tests/test_claude_wrapper_subrouter_resume_marker.py` (8 wrapper scenarios, wired into CI).

At the test-only commit the routing suite fails 15 tests; at the fix commit `swift test --package-path Packages/macOS/CMUXAgentLaunch` is green at **421 tests / 58 suites**, all wrapper suites pass, `xcodebuild -scheme cmux-cli` succeeds, and `scripts/lint-pbxproj-test-wiring.sh` reports ok.

End-to-end, a synthetic hook record driven through the same `AgentRestorePlanner` the CLI uses yields `["sr","claude","proxy","--resume","<id>","--model","opus","--permission-mode","acceptEdits"]` with the stale auth env removed and `CMUX_AGENT_RESTORE_LAUNCH` / `CMUX_CUSTOM_CLAUDE_PATH` retained; the unproven and no-`sr`-on-PATH scenarios keep the existing replay.

**Not verified:** no real `cmux restore claude <id>` was executed — the CLI needs the app socket and would have targeted a live surface. Planner output was exercised through the identical API path instead. The in-app typed-restore path (`AgentResumeCommandBuilder`) and `.forkAgent` are untouched.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12464"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791864297&installation_model_id=21920&pr_number=12464&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12464&signature=d4bcae739f5a086be526b1472f1a29a0d8c398604ad65e58895d2e8371dc9061"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Restores Subrouter-routed Claude sessions via `sr claude proxy --resume` instead of replaying a captured argv whose settings file was deleted, so account selection, pooling, and failover work again. Restore planning and the Claude wrapper also drop `--settings` paths that no longer exist, preventing "Settings file not found" launch failures.

**Bug Fixes**
- The restore planner checks `--settings` readability at plan time and drops missing files in both spellings; inline JSON, empty values, dangling options, and anything after `--` stay.
- The wrapper drops an unreadable `--settings` with a warning only for cmux-owned restore replays and stops re-capturing it; interactive launches still fail closed so permission and hook settings can't be silently lost.

**New Features**
- A resume record with an agreeing `SUBROUTER_CLAUDE_RESUME_COMMAND` / `CMUX_AGENT_LAUNCH_SUBROUTER_CLAUDE_RESUME_COMMAND` pair plans `sr claude proxy --resume <id>` and clears captured auth selection for the launcher to regenerate.
- The wrapper mints the launch-bound marker only when the argv still carries an existing `subrouter-claude-settings-*/settings.json` and clears inherited copies.
- Provenance relies only on the marker pair; marker-only, disagreeing or non-exact text, cmux launchers, direct mode, codex records, or `sr` missing on the restore PATH all keep the existing replay.

<sup>Written for commit 798c555ef98871912b88f6664d5d33cdc30dc1a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12464?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Claude sessions launched through Subrouter can resume through the same routing path.
  - Restore and fork operations preserve valid settings files while removing inaccessible paths.
  - Subrouter-specific settings and environment details are handled safely during Claude restoration.

- **Bug Fixes**
  - Prevented invalid settings paths from causing Claude launches and restores to fail.
  - Added warnings when inaccessible settings files are skipped.
  - Improved resume-marker validation to avoid incorrect routing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1620,6 +1620,7 @@ jobs:
           python3 tests/test_claude_wrapper_mutual_shim_loop.py
           python3 tests/test_claude_wrapper_shim_root_survives_tmpdir_change.py
           python3 tests/test_claude_wrapper_user_binary_resolution.py
+          python3 tests/test_claude_wrapper_subrouter_resume_marker.py
           python3 tests/test_claude_teams_test_utils.py
           CMUX_CLI_BIN="$CLI_BIN" python3 tests/test_cli_codex_teams_informational.py
           CMUX_CLI_BIN="$CLI_BIN" python3 tests/test_cli_claude_teams_fallback_path.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1611,6 +1611,7 @@ jobs:
           CMUX_CLI_BIN="$CLI_BIN" python3 tests/test_cli_socket_autodiscovery.py
           python3 tests/test_codex_wrapper_resume_hooks.py
           python3 tests/test_claude_wrapper_hooks.py
+          python3 tests/test_claude_wrapper_dead_settings_path.py
           python3 tests/test_codex_wrapper_computer_use_mcp.py
           python3 tests/test_cmux_cua_mcp_smoke.py
           python3 tests/test_cmux_cua_build_cache_safety.py
--- a/CLI/CMUXCLI+Fork.swift
+++ b/CLI/CMUXCLI+Fork.swift
@@ -178,7 +178,8 @@ extension CMUXCLI {
             observedPermissionMode: record.permissionMode
         )
         guard let invocation = AgentRestorePlanner(
-            executableFileResolver: AgentRestoreExecutableFileResolver()
+            executableFileResolver: AgentRestoreExecutableFileResolver(),
+            readableFileResolver: AgentRestoreReadableFileResolver()
         ).invocation(
             for: request,
             ambientEnvironment: processEnvironment
--- a/CLI/CMUXCLI+Restore.swift
+++ b/CLI/CMUXCLI+Restore.swift
@@ -205,7 +205,8 @@ extension CMUXCLI {
             observedPermissionMode: record.permissionMode
         )
         guard let invocation = AgentRestorePlanner(
-            executableFileResolver: AgentRestoreExecutableFileResolver()
+            executableFileResolver: AgentRestoreExecutableFileResolver(),
+            readableFileResolver: AgentRestoreReadableFileResolver()
         ).invocation(
             for: request,
             ambientEnvironment: processEnvironment
--- a/Packages/macOS/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentLaunchEnvironmentPolicy.swift
+++ b/Packages/macOS/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentLaunchEnvironmentPolicy.swift
@@ -172,6 +172,15 @@ public struct AgentLaunchEnvironmentPolicy: Sendable {
            let path = normalizedValue(env["PATH"]) {
             selected["PATH"] = path
         }
+        if normalizedKind == "claude" {
+            // Subrouter's resume marker and the wrapper's launch-bound copy are
+            // exact command text, never a URL or credential. They cross into
+            // the durable restore record only as an agreeing pair, so a marker
+            // inherited from an ancestor `sr claude` session proves nothing.
+            selected.merge(SubrouterClaudeResumeRouting().capturedEnvironment(in: env)) { _, marker in
+                marker
+            }
+        }
         return selected
     }
 
--- a/Packages/macOS/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentRestorePlanner.swift
+++ b/Packages/macOS/CMUXAgentLaunch/Sources/CMUXAgentLaunch/AgentRestorePlanner.swift
@@ -14,19 +14,37 @@ public struct AgentRestorePlanner: Sendable {
     ]
 
     private let isExecutableFile: @Sendable (String) -> Bool
+    private let isReadableFile: @Sendable (String) -> Bool
 
     /// Creates a restore planner.
     ///
-    /// - Parameter isExecutableFile: Executable-path lookup used for optional wrapper shims.
-    public init(isExecutableFile: @escaping @Sendable (String) -> Bool) {
+    /// - Parameters:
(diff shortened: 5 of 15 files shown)


danielraffel:
Thanks — the wrapper-side findings raised on #12463 are fixed and this branch carries them (it stacks on #12463). Two findings here I have **not** fixed, and I would rather say so plainly than paper over them.

**Route Subrouter restores through the remote command builder** (P1): you are right, this is a real gap. `Sources/Workspace+AgentLifecycle.swift` calls `resumeStartupInput(useLocalRestoreVerb: false)`, which renders `AgentResumeCommandBuilder` output on the remote host and never enters the CLI-only `AgentRestorePlanner` path this PR changes. So a remote workspace snapshot of a Subrouter-routed Claude session is not routed and keeps today's replay. That is a scope limitation, not a regression — the remote path behaves exactly as it does on `main` — but it does mean the fix is local-restore only. I did not extend it because the two paths build commands differently and I would rather you say whether routing belongs in `AgentResumeCommandBuilder` too, or whether the local-only scope is acceptable for now with a documented limitation.

**Replace path-shape attestation with authoritative launcher provenance** (major): the mechanism is described accurately — a nested plain Claude launch inside an `sr` session can inherit the exact marker and supply a `--settings` matching the Subrouter path shape that still exists, and the bound copy is then exported.

Worth being explicit about what that does and does not buy an attacker, because I do not think a stronger check is reachable here. Within one user account this was never a cryptographic boundary: anything the wrapper can verify (a nonce in the settings file, a path shape, an inherited variable) a same-user process can also read and reproduce. What bounds the impact is the consequence, not the check — the marker text is validated against two exact strings, so the worst outcome is that a session is restored by running `sr claude proxy --resume <id>`, a fixed credential-free command, instead of `claude --resume <id>`. There is no argument injection and no credential exposure, and the routing fails closed on anything unproven.

If you would rather have authoritative provenance anyway, the shape I would reach for is `sr` writing the binding itself (for example registering the resume command through the cmux control socket at launch) so the wrapper is not the attesting party at all. That is a larger change and a cross-repo contract, so I would want your direction before building it.

Issue #12486, opened 2026-09-13
Cloud link selector failure leaves blank terminal


austinywang:
A cmux Cloud terminal can still end up as a blank pane when the cmux-tui link exits with status 1 because its tab selector cannot be resolved. The visible error is similar to: `cmux-tui link exited with status 1: {"code":"selector.not_found","details":{"scope":"tab","selector":"tab_…"},"message":"no tab matches …"}`. The Cloud tree may still show the workspace/terminal while the terminal surface has no prompt, replay, cursor, or state overlay.

Repro:
1. Use a cmux Cloud machine with multiple workspaces/terminals.
2. Open or switch to a Cloud workspace while its terminal link is being created or restored.
3. Observe the cmux-tui selector.not_found link error and the resulting blank terminal pane.

Expected:
- Link creation resolves the live tab identity and does not target a stale tab selector.
- If the selector is genuinely gone, the terminal retries or shows an explicit connecting/reconnecting/error state instead of a permanently blank pane.
- Logs include the workspace, terminal, tab selector, and lifecycle transition needed to diagnose the race.

Related: #12376 (previous blank Cloud terminal follow-up), #9424 (selector resolution falling back incorrectly). Screenshot/report from 2026-09-12.

PR #12508, opened 2026-09-13
fix: preserve Cloud projection identity during nightly restore

### LINKED ISSUE #7867: Nightly session restore fails for Codex: pane returns as bare shell, resume binding never persisted (0.64.17-nightly)
### What happened

Session restoration on the **nightly** channel failed to bring back a running **Codex** session. Reported by the user on the cmux Mac mini dogfood machine on 2026-07-10 (~15:44 local), verbatim:

> nightly session restoration just failed for me for codex

The Codex pane did not come back as a resumed Codex session; affected nightly workspaces show bare `~` shells and at least one dead panel (`ttyName: null`) instead.

### Environment

- cmux NIGHTLY `0.64.17-nightly.2906518677101` (bundle `com.cmuxterm.app.nightly`)
- Most recent nightly relaunch: `2026-07-10T03:37:08Z` (Jul 9, 20:37 local), pid 40128, still running at report time
- macOS Darwin 25.5.0, machine user `cmux`
- Stable cmux `0.64.17` (`com.cmuxterm.app`) running **side-by-side** on the same machine/user — relevant because parts of `~/Library/Application Support/cmux/` are shared between the two instances (see below)
- Note: the report came ~19h after that relaunch. If the failure the user observed happened today, it may have gone through Reopen Previous Session rather than app relaunch; both funnel into session restoration and both need checking.

### Evidence collected from the machine (2026-07-10)

1. **The pre-relaunch snapshot contains nothing that could resume Codex.** `~/Library/Application Support/cmux/session-com.cmuxterm.app.nightly-previous.json` (rotated at the Jul 9 20:37 relaunch) records 17 workspaces / 21 terminal panels. Every terminal panel is persisted as a bare shell (`title: "~"`, `workingDirectory: /Users/cmux`) except two `✳ Claude Code` panes. No panel carries any agent/resume information.
2. **The session-file schema has no agent fields at all.** Panels persist only `type`, `title`, `ttyName`, `stableSurfaceId`, `terminal.workingDirectory`, `terminal.isRemoteTerminal` (+ browser state). There is no `launchCommand` / resume-binding field, so agent resume must depend entirely on the out-of-band binding store keyed by `ttyName` (#7750).
3. **`ttyName` keys are demonstrably unreliable in these snapshots.** In the *current* nightly session file, `ttys011` is recorded on three different panels across three workspaces; the previous file has it on two. macOS recycles tty names, and two cmux instances (stable + nightly) allocate from the same pool.
4. **Shared un-namespaced state across instances.** Session files are per-bundle-id, but `~/Library/Application Support/cmux/search.db` (Session Index) is shared by stable and nightly. If resume bindings (or anything keyed by `ttyName`) live in shared state, the two instances can stomp each other.
5. **The restore path emits zero diagnostics.** `~/Library/Logs/cmux/startup-com.cmuxterm.app.nightly.log` contains detailed `app.init.*` / `appDelegate.*` events for the Jul 9 relaunch but **no** events matching restore/resume/binding/codex (`grep -icE "restor|resume|codex|agent|binding"` → 0). Restoration failures currently cannot be triaged from logs at all.
6. Current live nightly session shows a workspace (`A6D48E98`) with a panel whose `ttyName` is `null` and generic `Terminal` titles — consistent with restore producing dead/blank terminals.

### Hypotheses to root-cause (in likelihood order, to be verified)

- **Capture-side loss:** the codex pane's resume binding was never persisted (or was overwritten) before/at quit, so restore had nothing to resume — consistent with evidence 1–4.
- **Restore-side binding mismatch:** binding exists but lookup by recycled/duplicated `ttyName` misses or misroutes it (#7750, #7391), possibly aggravated by the stable instance sharing the tty namespace and shared state.
- **Codex-side failure:** restore *did* run `codex resume <id>` and codex failed ("No saved session found", #6209; sqlite lock, #5557) and the pane fell back to a bare shell — then the next periodic save persisted the bare shell, erasing the evidence.

### Related issues

#6209 (codex `resume` "No saved session found" → bare shell), #3499 (codex resume ID lost after restart), #7750 (bindings keyed on ephemeral ttyName scramble across relaunch), #7391 (claude resume binding written to wrong workspace after restored TTY cache drift), #5802 (autoResumeAgentSessions gate never resumes), #5557 (codex resume sqlite lock), #6194 (claude twin of #6209), #4782 (multiple agent sessions not restored).

### Definition of done

- Root cause identified for how a live Codex session's restoration is lost across a nightly relaunch (capture vs restore vs codex-side), with the failing path covered by a regression test committed before the fix.
- Fix makes Codex sessions restore reliably across relaunch, including with a second cmux instance (stable + nightly) running under the same user.
- Session-restore decisions (bindings found/missing, resume command issued, per-panel outcome) emit events into the startup log sink so the next report like this is triageable from logs.

austinywang:
## Problem
Nightly session restore could lose Cloud terminal projection identity while a provider was reconnecting. The workspace and tab stayed visible, but an autosave serialized only resolved projections; a later restore then produced a bare shell or a blank pane. This is the same lifecycle seen in #12486 when `cmux-tui` returned `selector.not_found` for a stale `tab_*` selector.

## Change
- Keep pending restored remote projections in a panel keyed store until their provider resource is published.
- Include pending identities in session projection capture, preserve one owner across replacement and workspace moves, and clear them on panel or machine teardown.
- Add durable nightly breadcrumbs for restore staging/commit, projection capture/assignment, and Cloud link start/success/failure.

This keeps terminal identity separate from ephemeral tab placement and prevents a partial restore from being saved as an anonymous shell.

## Validation
- `python3 scripts/swift_file_length_budget.py`
- `git diff --check`
- `./scripts/setup.sh`
- `swiftc -frontend -parse` on changed Swift files
- `xcodebuild ... -scheme cmux-unit ... -only-testing:cmuxTests/CloudWorkspaceLiveProjectionTests test` reached compilation but is blocked by the existing environment symbol mismatch `GHOSTTY_ACTION_OPEN_URL_KIND_OSC8` in `Sources/GhosttyTerminalView.swift`.
- Tagged `./scripts/reload.sh --tag issue-7867-nightly-restore-followup` was started without launching; it was interrupted after the shared Xcode host remained in a long dependency/build queue. No app was launched.

Fixes #7867
Related to #12486

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restoration of cloud projections received before their resources become available.
  * Pending projections now persist correctly through restore and autosave operations.
  * Prevented outdated projections from reappearing after projections are replaced, moved, closed, or removed.
  * Improved handling of pending projections when moving them between workspaces.

* **Reliability**
  * Added more complete startup and cloud-connection activity tracking to improve troubleshooting of connection and restoration issues.


austinywang:
CI validation note: the integration test job reached the app build and failed on the pre-existing `GHOSTTY_ACTION_OPEN_URL_KIND_OSC8` symbol error in `Sources/GhosttyTerminalView.swift`, before running the selected suites. The local tagged reload reached the same shared Xcode build queue and was interrupted after no patch diagnostics appeared. Repository guards and Swift parse/budget checks pass.

austinywang:
Merged origin/main and resolved the SurfaceCatalog conflict. The branch now retains main’s cloud projection index invalidation while using the panel-keyed pending restore store; the merge commit is 3dec5d2657ad56fea127b9ae0dee77a08d597c5c.

austinywang:
Resolved both CodeRabbit inline findings in commit `afa94789bb`: preflight link attempts now emit `cloud.link.start` and classified `cloud.link.failed` breadcrumbs without mutating retry state, and pending projection removal now invalidates the index and notifies observers even when no live projection exists. The nested pending-record merge now uses a single panel-ID Set.

PR #12731, opened 2026-09-16
fix(web): pause HMR while a page is in back-forward cache

lawrencecchen:
Next dev App Router HMR left its WebSocket alive while a page entered the back-forward cache. On restore, stale socket callbacks could race a new page lifecycle and surface a WebSocket failure. The patched client pauses on `pagehide`, closes the old socket, ignores stale callbacks, and reconnects once on persisted `pageshow`; normal network disconnect retries remain unchanged.

Validation:

- Six lifecycle tests pass for both installed Next module formats, covering pagehide close, stale callback suppression, one restore, retry cancellation, and genuine network retries.
- Real Chromium new-headless mode on the direct dev backend tag `hmrpg` restored the same document from back-forward cache. Evidence reports a persisted pageshow, one closed old HMR socket, one open replacement, no not-restored reasons, and no WebSocket console error: `out/hmr-page-lifecycle/evidence/bfcache-result.json`.
- Preview: https://cmux-gsuyi7z7l-manaflow.vercel.app (Vercel returned HTTP 200 through protection-aware curl).

The live older backend on port 4405 was not modified. Merge this PR before that build consumes the patch.


lawrencecchen:
Fleet instruction update for head `5fc580f147a18f8c1f305a0934bae5d2ed2bdf04`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #12750, opened 2026-09-16
Repair iOS archive compilation after pairing update

azooz2003-bit:
Latest main fails every iOS archive after the pairing update: the root still calls the removed What's New compatibility method, the disconnected toolbar uses the retired device-ID API, and fitted sheet sizing is unguarded on the iOS 17 deployment target.

Remove the obsolete What's New calls while retaining the store policy refresh, restore pairing identities in the disconnected toolbar, and apply fitted sheet sizing only on iOS 18 and newer. iOS 17 retains its existing custom sheet heights. These fixes follow the current APIs and deployment target without changing the compatibility policy.

Validation: reproduced all six compiler diagnostics on clean main (`ff8f866bb4a`) with a fleet Release archive, then successfully archived the same official `com.cmux.app` target with these changes. `git diff --check` passes. Manual export and signing pass the official identity, production origin, push entitlement, purchase-link and symbol checks. Apple `altool --validate-app` reports `VERIFY SUCCEEDED with no errors`. A Release iPad Air 11-inch M3 simulator build installs and starts. Full UI verification remains blocked by the remote computer-use screenshot permission setup. No upload or App Review submission was performed from this branch.

Checked Apple's [sheet guidance](https://developer.apple.com/design/human-interface-guidelines/sheets); the change retains the existing sheet layout and adds an availability fallback.


lawrencecchen:
Fleet instruction update for head `d8ef5561f513b229ec4e67c76974a04a2a6e114a`: this PR is classified `ios`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

Issue #12680, opened 2026-09-15
Regression in 0.64.23: dropped screenshot inserts a deleted clipboard path instead of attaching


kgwoo:
### cmux version

0.64.23 (build 103)

### macOS version

26.6.2

### Mac chip

Apple Silicon (M5 Pro)

### Installation method

Direct download (DMG)

### Can you reproduce this on cmux NIGHTLY?

I could not test NIGHTLY

### Bug description

Dragging a screenshot thumbnail into a pane running Claude Code inserts `/var/folders/…/T/clipboard-….png` as plain text instead of attaching the image. The file behind that path is deleted right after it is inserted, so Claude Code cannot find the image.

This started with 0.64.23. Downgrading to 0.64.22 fixes it. The same Claude Code versions (2.1.270 to 2.1.272) attach images normally in another terminal.

### Expected behavior

The dropped screenshot is attached as `[Image #N]`, as in 0.64.22.

### Steps to reproduce

1. Run `claude` in a cmux pane.
2. Take a screenshot with Cmd+Shift+4.
3. Drag the screenshot thumbnail into the pane.
4. The prompt shows the plain `clipboard-*.png` path, and that file no longer exists.

### Shell and environment

zsh, Claude Code 2.1.272

### Relevant logs or crash reports

Watching `$TMPDIR` on a Debug build of main (`6f118af63b`) with `debug.terminal.simulate_file_drop`:

```
CREATED clipboard-2026-09-15-115237-D889ACB8.png
DELETED clipboard-2026-09-15-115237-D889ACB8.png (73 ms)
```

### Screenshots or screen recordings

Before/after videos are in #12670.

### Additional context

Fix: #12670. #10359 copies transient drag images into an owned `clipboard-*.png`, and #9266 deletes that copy in `executePreparedImageTransfer` right after `sendText` returns.

austinywang:
Another independent report of the same regression (via email, gururaghav.sd@gmail.com):

- cmux 0.64.23 (build 103, commit 211b8bb35), macOS 26.4.1, M4, Claude Code 2.1.274
- Dragging the floating screenshot preview types `/var/folders/<…>/T/clipboard-2026-09-17-143801-9077D4B7.png`; reading it moments later fails, file gone from $TMPDIR
- Same drag into Warp attaches the image fine
- Workarounds that work: Ctrl+V (Claude Code reads the clipboard directly), or dragging the saved file from Finder instead of the floating preview
[...]

PR #12752, opened 2026-09-16
Preserve image attachments when dropping or pasting into terminals

### LINKED ISSUE #12684: Regression in 0.64.23: dropping or pasting an image into the terminal inserts a `clipboard-*.png` path that does not exist on disk
## Summary

Since 0.64.23, dropping an image that has no backing file (for example the macOS screenshot floating thumbnail) into a terminal surface inserts a path like:

```
/var/folders/4c/…/T/clipboard-2026-09-15-150955-EFC7CF6E.png
```

The file is never present on disk. Anything that consumes the pasted path (Claude Code in my case) sees a plain string pointing at a missing file. Pasting an image from the clipboard with Cmd+V produces the same dead path.

0.64.22 does not have this problem: the same drop inserts the real temporary file written by screencaptureui, which exists, and the consumer works.

## Environment

- cmux 0.64.23 (build 103), installed via Sparkle auto-update on 2026-09-14
- macOS 26.6.2 (25G83), Apple Silicon
- Consumer in the terminal: Claude Code 2.1.272 (same result with 2.1.271)

## Steps to reproduce

1. Run any program that reads pasted paths in a cmux terminal (Claude Code works, but `cat > /dev/null` then inspecting the typed text is enough).
2. Take a screenshot with Cmd+Shift+4 and, while the thumbnail is still showing in the corner, drag it onto the terminal.
3. Look at the inserted text, then check whether the file exists:

```
ls -la "<the inserted path>"
```

Expected: the path exists, or the drop inserts the source file's own path.

Actual: `No such file or directory`. I searched `$TMPDIR`, `~/Library/Containers/com.cmuxterm.app`, `~/Library/Caches`, and any `*.png` modified in the last 30 minutes under `/var/folders`: nothing was written at any point.

Same result with Cmd+V when the clipboard holds an image.

## Comparison with 0.64.22

Rolled back to 0.64.22 (dmg from the GitHub release). The same drop inserts:

```
/var/folders/4c/…/T/TemporaryItems/NSIRD_screencaptureui_8kATwV/Capture d'écran 2026-09-15 à 15.26.42.png
```

That file exists and Claude Code turns it into an image attachment as expected.

## Notes

- The 0.64.23 binary contains a new `CmuxCloudImagePaste` module (`CloudImagePasteCoordinator`, `TerminalImagePasteWriting`, the string "Paste the file path into the terminal."). It looks like the local terminal drop now goes through that path and the temporary file write either does not happen or the file is removed before the path is handed to the terminal.
- Release notes for 0.64.23 mention "Fix terminal file drops being ghosted after pane teardown", "Fix file drag hit-test amplification" and "Fix file drag-and-drop in Dock sessions", which touch the same area.
- The `cmux-claude-wrapper` shim is not involved: it only injects hooks settings.

Happy to test a nightly build.

austinywang:
## Summary

Dropped or pasted image files must remain available while Claude Code or Codex reads them asynchronously. Keep cmux-owned local image files until the existing app-exit cleanup instead of deleting them when terminal text delivery finishes.

A copied image can also advertise a source URL on the clipboard. Resolve its image/rich-text payload before that auxiliary URL so image bytes accompanied by a folder URL, web URL, or missing file path become an image attachment. Ordinary Finder file/folder pastes retain their file-URL identity and remote-upload routing; visible rich text keeps its existing priority. Terminal drag routes now carry the original pasteboard through the pane/overlay adapters, so those paths cannot discard the image bytes before shared preparation. Existing backing files still take priority over drag thumbnails.

## Validation

- The original test-only lifetime commit failed both drop/paste cases; the lifetime fix passed both on the hosted macOS lane.
- Separate test-only and fix commits prove the paste regression: all three auxiliary-URL cases fail [before](https://github.com/manaflow-ai/cmux/actions/runs/35160017708); all seven lifetime/priority/control cases pass [after](https://github.com/manaflow-ai/cmux/actions/runs/35160089753).
- Expanded the suite to drag mode and a live PTY consumer. It checks bracketed-paste boundaries and reads the delivered image from disk through direct and routed drops. [Current drag validation](https://github.com/manaflow-ai/cmux/actions/runs/35162985452) is running.
- Tagged fleet build at `5395d8562bcdf8a5644ab23257e457297c20abe6` succeeded, installed, launched, and responded through its tag-bound CLI. The subsequent commit changes only CI runner selection. Full CI validation is still in progress.
- Corrected concrete inherited CI failures in runner selection, isolated iOS lane-version fixtures, Dock audit coverage, and the test flag fixture. Replaced fixed-delay assertions with causal completion signals; the transport installer exposes an internal wait for its existing task so its cancellation test can wait for completion. The determinism gate passes with zero exemptions. The runner guard, 68 workflow-security tests, and iOS lane-identity tests pass.
- Swift file-length budgets, Package.resolved policy, and diff whitespace checks pass.
- Localization audit: no user-facing strings changed.

Fixes #12684. Related: #12680.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved paste handling so copied images and rich text take priority over accompanying URLs.
  - Preserved pasted image files after insertion, preventing them from becoming unavailable unexpectedly.
  - Maintained correct handling for text, image, file, folder, and URL pastes, including drag-and-drop operations.

- **Tests**
  - Added coverage for image-transfer persistence, content-priority rules, and Finder paste behavior.
  - Improved test synchronization for more reliable results.

- **Chores**
  - Updated release delivery automation to use the configured build runner.


kgwoo:
Love cmux and would like to help where I can. I hit this one myself and put up
a fix in #12670 on 09-15, with a regression test and before/after videos. Looks
like the same change as here, and it's currently green on CI.

Either way works - just flagging it in case it's useful.

PR #12754, opened 2026-09-16
Preserve existing Mac identity across the v2 upgrade

azooz2003-bit:
A Mac upgrading to v2 published its new installation ID into the older account directory, so existing iOS BETA builds saved a second Nightly computer. Preserve the original physical Mac ID and exact build namespace in that compatibility publication while retaining the separate v2 installation ID and team key. Replies to older iOS sessions use the same original ID.

Already affected Macs prove ownership of their exact former slot, revoke that binding, and retry registration under the original ID. Repairs never match display names or combine Stable and Nightly. A failed or interrupted repair can retry safely.

This also restores the v2 Mac runtime after the explicit-pairing merge reintroduced the older runtime. Pairing still requires the user's setting; compatibility remains a separate service under the v2 owner.

Validation:
- Fleet package tests cover first upgrade, existing duplicates, restart, Stable isolation, denied revocation, and lost revocation replies.
- A real production test account reproduced two computers, then verified one original ID with the v2 key after repair and restart. Probe-owned bindings were revoked afterward.
- Actual older BETA and modern INTERNAL simulator checks are in progress, including saved-row repair, workspace use, and restart. Merge is held for recorded verification and focused app tests.

## Summary by CodeRabbit

* **New Features**
  * Introduced v2 pairing flow with pairing opt-in, pairing readiness status, and IROH-only ready states.
  * Added support for preserving physical device identities during Mac upgrades and repairing duplicate compatibility registrations.
  * Added visibility into local socket addresses and pending port changes.
  * Added simulator-stream capacity handling and improved device-specific connection status reporting.

* **Changes**
  * Pairing no longer generates Tailscale attach tickets or refreshes pairing codes when selecting an app target.
  * Connection handling now uses first-frame timeouts and reports capacity errors without unnecessarily closing connections.


lawrencecchen:
Merging on Lawrence's directive to unbreak the Nightly macOS build on main (https://github.com/manaflow-ai/cmux/issues/12736). Head 1cd4dd984 already passed a full Nightly build on this branch (https://github.com/manaflow-ai/cmux/actions/runs/35074430809). Recorded iOS verification can continue as a follow-up.

PR #12768, opened 2026-09-16
Give each awaited ownership scan its own wait budget

lawrencecchen:
Fixes the startup fallback where restored agent panes show a bare shell plus:

> cmux could not verify whether this agent session is already running, so it did not resume the session automatically. Run 'cmux restore --surface' to resume it here.

**Cause.** `SharedLiveAgentIndex.indexForOwnershipDecision()` charged one 10 s wall-clock budget to the whole request and re-checked that clock even after a wait succeeded. On a Mac with many live agents a restore routinely needs two scans (the one already running when session restore asked, then the request's own scan once a hook lands mid-flight), and main-actor latency during a large session restore counts against the same clock. Every agent pane then failed closed and never resumed automatically.

**Evidence.** Reproduced on a Mac with 168 workspaces, about 90 live agent sessions, 5,800 processes and load average 55. The DEBUG timing log added here measured the live-agent scan on that Mac at 8.8 s cold and 2.0 to 3.7 s warm, so two scans never fit one budget. The process walk itself is not the cost (under 0.3 s for all 5,800 processes, measured with a standalone probe); a `sample` of the loader thread puts most of the time in Claude transcript resolution (`claudeProjectDirName`, `transcriptPathInAnyProject`), which is a separate optimization and out of scope here.

**Fix.** The budget applies per awaited scan (still 10 s), capped at four scans per request, and an index returned by a successful wait is never discarded on the clock afterwards. A loader that does not finish inside one budget still fails closed. DEBUG builds log each ownership decision (`agentIndex.ownership`) and reload duration (`agentIndex.reload`) so the next report can be attributed instead of guessed at.

Principled: it fixes the time accounting rather than raising a constant. Residual risk: worst-case blank-pane time during startup admission is now up to two scan budgets instead of one, and a cold scan near 10 s on a loaded Mac can still fail closed; the loader hot spot above is the follow-up for that.

Commit 1 adds the failing test, commit 2 the fix.


lawrencecchen:
Fleet build instructions for this PR, head `c4cd3d6c4227c23d0359e6b7cfb21e19f9d570cd`:

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-12768-c4cd3d6c /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git c4cd3d6c4227c23d0359e6b7cfb21e19f9d570cd' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/12768 --source-digest c4cd3d6c4227c23d0359e6b7cfb21e19f9d570cd --cache-key cmux:pr-12768 --min-free-bytes 268435456000 --label cmux --label ram48)
```

The job survives disconnects. Do not resubmit after a wait timeout; rerun `cmux-ci wait` with the same ID. The artifact receipt records worker, queue/build/package/upload times, cache state, and disk before/after cleanup. The build is exact-head and does not include uncommitted edits.

PR #12766, opened 2026-09-16
Revert the portal-owned resize phase from #12662

lawrencecchen:
Reverts https://github.com/manaflow-ai/cmux/pull/12662, which shipped in 0.64.24 with a PR body that still said dogfood failed.

With the resize phase active, `TerminalSurface.updateSize` and `reapplyAssignedGrid` refuse every renderer and PTY size write while a window or divider drag is in progress, and only the drag-end pass clears the phase. On 0.64.24 panes stop tracking the drag and show stale blocks and missing cells (user recording from 2026-09-16, build 0.64.24 (104)). The corruption in https://github.com/manaflow-ai/cmux/issues/12657 is reported on 0.64.23 too, so that PR did not fix it either; it needs a separate root cause.

The pure move of `makeNSView`/`updateNSView`/`dismantleNSView` into `GhosttyTerminalView+Representable.swift` is kept, since it changed no code. Everything else returns to the 0.64.23 behavior.

`scripts/swift_file_length_budget.py` reports three files back at their pre-#12662 length (+2, +3, +10 lines). Those lines are the restored originals, not new code.

Build and preflight evidence follows in a comment.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12766"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792146015&installation_model_id=21920&pr_number=12766&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12766&signature=04ff5107b91cf835239fb00cd1fd71bc88f29c1f954e14d187b70b320c8691ad"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
[...]

Issue #12681, opened 2026-09-15
Terminal text layout corrupt after updating to 0.64.23


rizkidarmawan21:
## Summary
After updating to 0.64.23, terminal rendering becomes severely corrupted while running Codex. Long lines are split into individual or short fragments; text appears pushed to the far right edge; and multiple `Jump to bottom` overlays can remain visible. This makes terminal output difficult to read and interact with.

## Environment
- cmux: 0.64.23 (build 103)
- macOS: 15.7.9 (24G830)
- Hardware: MacBook Air, Apple M2

## Observed behavior
- Normal terminal/Codex output no longer lays out as continuous wrapped lines.
- Portions of text render one character or tiny fragment per line at the right edge.
- `Jump to bottom` overlays duplicate/remain on screen.
- A separate affected pane shows similarly broken short fragments and clipping.

## Expected behavior
Terminal output should wrap normally at pane width, with no detached right-edge glyphs or persistent duplicate scroll overlays.

## Notes
Started immediately after latest cmux update. I have screenshots, but am not attaching them because displayed terminal output may contain confidential context.

I can provide a sanitized reproduction or additional diagnostics if useful.

<img width="1478" height="1390" alt="Image" src="https://github.com/user-attachments/assets/963eb77b-f758-4dd4-80fd-37678d64dcf8" />

<img width="2048" height="215" alt="Image" src="https://github.com/user-attachments/assets/6d3157b7-ef24-44b4-ba8f-5531a14b02bd" />

<img width="1196" height="1050" alt="Image" src="https://github.com/user-attachments/assets/aa00c11e-f924-441e-a500-5cda6ef2e420" />

austinywang:
0.64.24 includes merged PR #12662 (resize publication across pane/restore transitions), but this issue's text-corruption symptom is not explicitly covered by the release notes and remains open for verification. Please retest the exact Codex narrow-pane/glyph-fragment repro on 0.64.24; if it persists, this is a residual rendering bug rather than the already-fixed size-publication path.

austinywang:
0.64.24 includes merged PR #12662 (resize publication across pane/restore transitions), but this issue's text-corruption symptom is not explicitly covered by the release notes and remains open for verification. Please retest the exact Codex narrow-pane/glyph-fragment repro on 0.64.24; if it persists, this is a residual rendering bug rather than the already-fixed size-publication path.

austinywang:
[...]

PR #12769, opened 2026-09-16
Make committed pane geometry the only source of terminal size

### LINKED ISSUE #12657: Regression: Claude Code TUI garbles after terminal resize
### Summary

Claude Code's terminal TUI can become visually garbled after the terminal or pane is resized. Codex TUIs may show the same failure, but Claude Code is the clearest and most frequent repro.

This appears to be a regression in/around cmux 0.64.22. The rendered frame can contain duplicated, stale, miswrapped, or otherwise out-of-place fragments after the resize settles. Restarting the CLI or forcing a full redraw may temporarily recover the surface.

### Steps to reproduce

1. Open a local terminal surface running Claude Code.
2. While the TUI is active (especially during streaming or with a multi-line prompt/output), resize the window or drag a pane divider narrower and wider.
3. Observe the TUI during and after the resize.
4. Repeat with Codex if needed.

### Expected behavior

The terminal grid and PTY receive the final dimensions, and the TUI redraws once into a coherent frame with correct wrapping, cursor placement, and content ordering.

### Actual behavior

After resizing, Claude Code can render a garbled frame: duplicated or stale rows, fragments from multiple widths, incorrect wrapping, or cursor/input content at the wrong position. The issue is especially reproducible with Claude Code; Codex appears affected less consistently.

### Regression

- First observed around cmux 0.64.22.
- Related historical reports: #8229 (0.64.19 narrow reflow duplicated content) and #8285 (Claude Code TUI stays stuck at a narrowed width after widening).

### Environment

- cmux: 0.64.22 (reported regression)
- macOS / hardware: not yet captured
- CLI: Claude Code; Codex also suspected

lawrencecchen:
Fixes https://github.com/manaflow-ai/cmux/issues/12657 and https://github.com/manaflow-ai/cmux/issues/12681 at the root. Builds on the revert in https://github.com/manaflow-ai/cmux/pull/12766.

The terminal grid and PTY size used to derive from whatever `NSView.bounds` a hosted view had when AppKit called `layout()`, and five generations of flags (#11323, #11528, #11530 then #11887, #12662) tried to infer after the fact whether that frame was final. Every layout transition could leak a placeholder size to the PTY, and TUIs reflowed to it permanently.

Now the size is a value the pane host owns. `TerminalPaneGeometry` (size, backing scale, phase `interactive` or `settled`) is committed to `TerminalSurface.commitPaneGeometry` by the window portal and by nothing else while the view is portal-owned. The portal commits from exactly two places: a drag tick, where every frame is on screen, and the settled pass, where a forced layout moved nothing or the interaction owner reported its end. A bind seed, a hidden entry, a detach, a collapsed frame while leaving a workspace, and a mid-layout clip have no path to the PTY. When the bounded settlement retry runs out while layout is still moving, nothing is published; the next external change schedules the pass that commits. Runtime surface creation for a visible pane waits for the first commit so the PTY's first window size is the pane's real size; a pane nobody is showing starts at the default size and reflows once on reveal. Views outside a portal keep upstream behavior and size from their own bounds.

Deleted: the portal settlement flags on the view, the tab-drag deferral, the per-view `updateSurfaceSize` gate chain, and the view-side end-of-live-resize bypass. `TerminalSurface.updateSize` is no longer public.

Tests. `TerminalWindowPortalCommittedGeometryTests` covers bind-before-settle, hidden entries, hide-after-commit, drag ticks then settle, and runtime creation waiting for the commit. The lifecycle test class now installs a debug seam for `Workspace.portalRenderingEnabled`: since #12414 that authority reports every workspace-less test surface as hidden, which had silenced the whole portal suite in every test host (13 of 37 failing on the base branch, hosted run https://github.com/manaflow-ai/cmux/actions/runs/35159380397). With the seam, `cmuxTests/TerminalWindowPortalLifecycleTests` is 41 of 41 on a fleet mini (`cmux-unit`, `cmux-austin-mini-0`); the base branch with the same seam fails one of its 37.

Preflight on tags `cpg1`, `cpg2`, `cpg3`: the debug log shows every size apply with `caller=commitPaneGeometry`, runtime creation waits (`surface.create.wait` then `surface.create.resume`) and starts at the committed size, no grid below 40 columns across split creation and pane resize bursts, and the rendered wrap matches the divider position.

Known limits. Window-edge live resize is not driven by the socket and is unverified end to end. The 800 by 600 initial size for hidden panes is the Ghostty embedded default; committing the first size through the surface config is a follow-up in the Ghostty fork. Cmd+Shift+D still shows one frame where the old terminal paints under the new pane's tab bar, because the portal and SwiftUI commit in different transactions; that class needs the AppKit pane tree, not this PR.

## Summary by CodeRabbit

* **New Features**
  * Terminal pane sizing is coordinated through committed portal geometry.
  * Interactive resizing updates panes responsively, while settled geometry is applied reliably afterward.
  * Terminal runtime creation waits for a valid pane size when required.
  * Geometry values validate dimensions and account for display scaling.

* **Bug Fixes**
  * Improved handling of hidden panes and display-scale changes.
  * Reduced redundant geometry synchronization during portal updates.
  * Restored scrollback now preserves prompt row boundaries correctly.
  * Improved reliability when panes appear, resize, or reconnect during layout changes.


austinywang:
Scope coordination with #12796: that branch currently carries an adaptation of this PR's committed-pane implementation (30c271d9f8, d1ec959444, 6d8aa7b93c), now explicitly attributed in its description. It is the same candidate geometry path, not a separate Claude/Codex fix, and must land only once.

I am validating #12681 against split/ruler, hidden reveal/restore, interactive resize, and hibernation transitions. Review-driven repairs on #12796 cover the clip viewport/scroller gutter, repeated batch hierarchy work, pending publication acknowledgement, and keeping settled commits out of active drags. I also converted the geometry tests into a separately runnable Swift Testing suite; earlier semantic CI alone did not execute these tests. Geometry remains a candidate explanation for the private screenshot, not a verified diagnosis. No screenshot or private support content has been published. Neither PR should be treated as reproduction-verified yet.

austinywang:
Follow-up from #12796: I validated the retry-exhaustion concern with a real AppKit anchor that changes width during the forced layout pass. The regression-only behavior at `cbda7d4ba9` failed in hosted run https://github.com/manaflow-ai/cmux/actions/runs/35167612575: committed geometry changed from 520×280 to 280×280 and `needsSettledCommit` cleared despite that pass moving the hierarchy.

`2f36e55f22` removes the forced commit on exhaustion, shares settlement handling between bind and external-layout paths, and uses your `e4d0cf26ff` stability observation (forced layout changed nothing) without an interaction-end override flag. It also preserves the no-redundant-reconcile gate for divider as well as native window resizing. These follow-up contributions are attributed in #12796. Fixed focused CI is pending, and this is still not proof that the original private screenshot is fixed. Both PRs remain overlapping candidates that must land only once.

azooz2003-bit:
A/B datum from a tagged fleet build of this head (`fe7ad6a5c69`): the geometry legs pass (reveal fires WINCH, `stty size` / emulator grid / pane frame converge exactly on split, divider resize, and hidden→reveal), but a restored agent still garbles on its first post-restore resize. lldb on the resumed codex shows thread mask `0xFBF6E027` — SIGWINCH blocked at the `cmux restore` execve, so committed geometry can publish the right size while the agent can never learn about later ones. The exec-site mask reset in https://github.com/manaflow-ai/cmux/pull/12796 covers that path; the two changes are complementary, and #12681-class reports need both.

lawrencecchen:
Fleet build instructions for this PR, head `fe7ad6a5c69ee795a2c20914df515a85676ec6ab`:

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-12769-fe7ad6a5 /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git fe7ad6a5c69ee795a2c20914df515a85676ec6ab' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/12769 --source-digest fe7ad6a5c69ee795a2c20914df515a85676ec6ab --cache-key cmux:pr-12769 --min-free-bytes 268435456000 --label cmux --label ram48)
```

The job survives disconnects. Do not resubmit after a wait timeout; rerun `cmux-ci wait` with the same ID. The artifact receipt records worker, queue/build/package/upload times, cache state, and disk before/after cleanup. The build is exact-head and does not include uncommitted edits.

PR #12778, opened 2026-09-16
Fix process-scope ownership of unrelated concurrent forks

lawrencecchen:
## Problem
Concurrent Unix forks briefly hold every parent process-scope marker. The tracker treated close-on-exec copies as ownership, so one hook's cleanup could claim an unrelated command. A second lifetime bug let an expired command drop its final marker descriptor while asynchronous cleanup still retained the inode identity; inode reuse could then make that inactive registration claim and kill a new command.

## Change
Require explicit descriptor inheritance. Linux matches inode, mount ID, and close-on-exec state from one bounded `fdinfo` snapshot; macOS checks `PROC_FP_CLEXEC` with vnode identity. The tracker now shares ownership of the marker descriptor until every scan releases the registration, reserving its inode throughout cleanup. Existing process-group, environment, and lineage cleanup remain in place.

Includes the existing formatter output and the WireGuard-peer/notify-alias fixture repairs from the Cloud branch so complete verification can run.

## Verification
Separate regression and fix commits prove both bugs: a paused pre-exec fork must carry only its explicitly inherited marker, and a paused final scan must retain its marker after the scope drops and its child is reaped. Both regressions fail deterministically before their fixes; all five selected Linux ownership/cleanup tests pass afterward.

Full parallel Linux binary runs with the lifetime fix passed 1,666/1,666 tests four times. Hermes reaper handling passed in all five runs; one run failed only the separate mouse-capture fixture. `cargo fmt --all -- --check` and all five selected marker/cleanup tests pass.

The complete Linux workspace run remains red in unrelated tests: exited-terminal input (`keep_on_exit...`), legacy closed-terminal journal repair (`startup_repairs_legacy_terminal_close_dangling_resource_rows`), and proxy-listener shutdown (`injects_into_html_and_honors_the_opt_outs`). All marker tests and the complete binary suite passed in that run.

Exact-head hosted verification: https://github.com/manaflow-ai/cmux/actions/runs/35137451841. Both hosted Linux and macOS pass the complete binary suite, including Hermes, then fail the same exited-terminal input fixture. These remaining full-suite failures are not claimed as resolved by the marker repair.


lawrencecchen:
Fleet instruction update for head `2046c65e4453f72cf3a597f0b2e2ee245556ace3`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #12822, opened 2026-09-17
Fix Codex writer ownership during quit and restore

### LINKED ISSUE #12805: Restored Codex pane opens read-only: "This conversation is open in another app" after quit and reopen
Seen on cmux 0.64.24-nightly.3513963328901 [5994f5e] after quit and reopen. A restored Codex pane shows:

```
🔒  This conversation is open in another app          R to Retry
    Close it there and press R to continue here.
```

The text is Codex's external-writer notice (`codex-rs/tui/src/chatwidget/rendering.rs`). Codex holds an OS `flock` on `~/.codex/thread-writer-locks/<thread>.lock` for the life of the process that owns the thread. `codex resume <id>` against a held lock gets `thread <id> already has an active writer` and falls back to a read-only view with this banner.

How cmux triggers it:

1. On quit, `applicationShouldTerminate` / `applicationWillTerminate` (`Sources/AppDelegate.swift`) persist the session index and kill cmux's own helpers, but never signal or wait for agent pty children. They get SIGHUP only implicitly when the pty master closes as the process exits, and cmux does not wait for exit. `AgentHibernationController+ProcessTermination.swift` already has a SIGTERM→SIGKILL terminate-and-wait path, but quit does not use it.
2. On relaunch, restore types ` cmux restore codex <id>`, which execs `codex resume <id>` (`CMUXAgentLaunch/AgentResumeArgv.swift`). The only guard is cmux's own PID/argv/hook-store scan (`liveSessionOwner`, `agent.restore.admit`). When that scan cannot bind the still-alive old codex PID to the session (Codex hook-store incompleteness is explicitly modeled as `hasUnboundedCodexIncompleteness`), admission returns `.admitted` and the new `codex resume` execs into a held lock.
3. cmux never checks the Codex writer lock itself. No hits for `thread-writer-locks`, `active writer`, or `another app` in `Sources/`, `CLI/`, `Packages/`.

Existing tests cover a live owner suppressing autoresume only for grok/amp, and the quit→restore cycle test uses a dead prior process. Nothing covers Codex relaunch while the old process is alive.

Proposed fix:

- Quit: terminate-and-wait agent pty children via the existing hibernation termination path and return `.terminateLater` until they exit (bounded).
- Admission: add a Codex-specific `LOCK_EX|LOCK_NB` probe of `~/.codex/thread-writer-locks/<thread>.lock` in `TerminalController+AgentRestoreAdmission.swift` and the plan-time gates in `DockSplitStore+SessionRestore.swift` / `Workspace+AgentLifecycle.swift`. A held lock means "live owner", regardless of what the PID scan says.
- CLI: `RestoreAdmissionRetryPolicy` waits briefly for the lock to clear instead of exec-ing into a read-only view.

Note: the same banner appears when the thread is genuinely open in the ChatGPT desktop app or another `codex app-server`, so the probe must report who holds it rather than kill it.

austinywang:
Closes #12805.

After quit/reopen, a surviving Codex process could still hold the thread's writer lock while cmux launched `codex resume`, leaving the restored pane silently read-only. Restore now waits briefly for the writer to exit and otherwise leaves a recoverable diagnostic instead of starting a second writer.

- Quit uses the existing asynchronous termination flow: load a fresh index, persist the pre-signal session snapshot, then terminate proven local Codex process groups. The existing hibernation signaling/exit machinery revalidates process birth identity, process group, kernel TTY, and cmux scope. The new gate additionally requires ancestry under this exact app and a still-owned local panel. Groups get five seconds after SIGTERM, then at most one second after validated SIGKILL; the normal quit watchdog bounds the overall snapshot/cleanup phase to ten seconds. Its signal-phase timeout retains the pre-signal snapshot.
- `CMUXAgentLaunch` owns one nonblocking writer-lock probe. It handles persistent files from older Codex versions and the current `.coordination.lock` protocol, validates UUID/path/inode evidence, and never creates, deletes, or retains a provider lock. The protocol was checked against [Codex's writer-lock implementation](https://github.com/openai/codex/blob/5e636ea760d1c821a9162af652a14225fdbd8a0f/codex-rs/rollout/src/writer_lock.rs).
- CLI restore retries lock contention for up to five seconds **before** admission/binding claims, then checks again without waiting immediately before exec. This also protects a newer CLI paired with an older app that does not advertise admission support. Vault uses the same restore path and carries the indexed Codex account root into the actual child environment; captured verification homes are likewise pinned into the launch account.
- No new socket methods, remote-relay allowlist entries, or routing parameters. Existing app ownership/admission checks remain intact. A new plan-time lock gate would suppress the command that needs to wait, and the app's ambient account need not match the child account, so the authoritative check belongs at CLI execution.

Validation:

- Test-only commit `e111ee3c348` precedes the fix. Its original executable regression harness fails **5/7** cases on the old code. The final harness against that same production revision fails **7/9** behavioral assertions; it passes **9/9** with this change. It compiles the production planner and CLI preflight/exec files with disposable homes, real kernel flocks, and a fake executable. Coverage includes retained external ownership, release during CLI retry, the admission-capability bypass, account isolation, exact argv/cwd/environment, legacy commands, and remote providers. No real Codex session is used.
- `arch -arm64 swift test --package-path Packages/macOS/CMUXAgentLaunch --jobs 4`: **403 tests / 56 suites passed**, with no compiler warnings. New tests cover coordination contention, stale/missing files, invalid/nonregular/symlink paths, bounded retry, literal effective homes, and account pinning.
- Added app-host ancestry/PID-reuse/orphan/quit-timeout and Vault account tests, wired to the Xcode test target and the non-tolerant focused CI lane. These app-host tests were **not run locally**.
- Swift syntax parsing, `python3 scripts/swift_file_length_budget.py`, project normalization, test wiring, shard selector validation, Package.resolved policy, workspace grouping, and whitespace checks passed. The clone's length checker derives budgets from Git; its requested length TSV is absent. Every new Swift file is below 500 lines; the largest is 349. Neither budget TSV was changed.
- Localization audit: five new CLI keys translated in all nine supported macOS locales, matching format placeholders; catalog parity and localization tests pass. No web UI messages changed.

Material trade-offs and remaining verification:

- A preflight is advisory. A writer starting after the final check can still win Codex's atomic acquisition; eliminating that last race requires provider cooperation. cmux must release its probe before exec so it does not become the conflicting writer itself.
- Quit signaling is intentionally limited to proven local Codex trees. External, orphaned, remote, ambiguous, and incomplete process scopes are not killed based on a lock observation. Other agents retain their existing quit contract, including Claude's separate transcript-protection requirements. If cleanup cannot prove ownership or finish in time, the launch guard provides the safe fallback.
- Process inspection reports candidate PIDs/executable names with the exact lock inode open. An open descriptor is not proof of flock ownership, so those diagnostics never authorize termination or guessed focus.
- The synchronous CLI reuses the existing bounded admission backoff with an injectable sleeper: flock has no release-notification API, and this is an exec-based process boundary, not UI polling. The final post-claim check has no retry delay.
- Legacy local shell commands need a literal absolute inline `CODEX_HOME`; ambiguous commands fail closed with manual recovery guidance because login-shell startup can change accounts. Explicit remote Codex launches do not probe local locks.
- Per task instructions, no dev app was built or launched, and no local `xcodebuild` tests were run. Full app compilation and the new app-host tests remain for hosted CI; UI dogfood requires a later authorized build. This PR does not claim a verified quit/reopen UI run.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12822"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792199970&installation_model_id=21920&pr_number=12822&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12822&signature=56360ca7dc0302fa2101d23b0c54c96cff60e2c894a22a0483f9031a7f8fb68a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes a race where a surviving Codex process could hold the thread's writer lock after quit/reopen, leaving the restored pane silently read-only.

Quit now terminates proven local Codex process groups after persisting the pre-signal session snapshot, bounded by the existing quit watchdog. Restore probes the Codex writer lock before admission and again immediately before exec, retrying for up to five seconds on contention. The probe is advisory; Codex remains the final authority if another writer starts after the final check.

Session entries carry the indexed Codex home into the child environment so resume uses the same account that owns the transcript. Quit signaling is limited to proven local Codex trees; external, orphaned, and remote processes are never killed based on a lock observation.

**Migration**
- Legacy local shell commands require a literal absolute inline `CODEX_HOME`; ambiguous commands fail closed with manual recovery guidance.

<sup>Written for commit f4a4d875b85adeecd3a8ab94cfe674c5aed08d41. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12822?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lawrencecchen:
Mac fleet instructions for head `f4a4d875b85adeecd3a8ab94cfe674c5aed08d41`. Planned tag: `pr-12822-f4a4d875`; this is not yet a published build.

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-12822-f4a4d875 /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git f4a4d875b85adeecd3a8ab94cfe674c5aed08d41' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/12822 --source-digest f4a4d875b85adeecd3a8ab94cfe674c5aed08d41 --cache-key cmux:pr-12822 --min-free-bytes 268435456000 --label cmux --label ram48)
```

Issue #3051, opened 2026-04-20
Scrollbar appearance causes PTY resize oscillation (flicker in TUI apps)


kensave:
## Problem

When content overflows the terminal viewport, the scrollbar appears and `FocusableTerminalView` triggers a PTY column recalculation. This causes a resize oscillation that makes TUI applications (e.g. Kiro, Claude Code) flash:

```
1. Content overflows → scroller appears
2. SwiftTerm recalculates: columns = (frame.width - scrollerWidth) / cellWidth
3. PTY resize (SIGWINCH) → TUI reflows at narrower width
4. Content may no longer overflow → scroller hides
5. SwiftTerm recalculates: columns = frame.width / cellWidth  
6. PTY resize (SIGWINCH) → TUI reflows at wider width
7. Goto 1
```

Observed as a 2-column width oscillation (~60ms cycle) in terminal debug logs.

## Root Cause

In `Sources/TerminalView.swift`, `FocusableTerminalView.layout()` sets the terminal to the full container bounds:

```swift
override func layout() {
    super.layout()
    if let tv = terminalView, bounds.size.width > 0, bounds.size.height > 0 {
        tv.setFrameSize(bounds.size)  // ← full width, no scroller reservation
        setupScrollerTracking(in: tv)
    }
}
```

The scroller is a child of the terminal view. When it becomes visible, SwiftTerm internally subtracts the scroller width from the available area, reducing the column count and firing `sizeChanged`.

## Proposed Fix

Always reserve the scroller width in the terminal frame so the PTY column count is stable regardless of scroller visibility:

```swift
override func layout() {
    super.layout()
    if let tv = terminalView, bounds.size.width > 0, bounds.size.height > 0 {
        let scrollerWidth = NSScroller.scrollerWidth(for: .regular, scrollerStyle: .overlay)
        let terminalSize = NSSize(
            width: bounds.size.width - scrollerWidth,
            height: bounds.size.height
        )
        tv.setFrameSize(terminalSize)
        setupScrollerTracking(in: tv)
    }
}
```

This permanently reserves ~15px for the scroller. The terminal gets a stable column count, and TUI apps never see width oscillation.

Alternative: use `.legacy` scroller style which always shows the scroller track (no toggle), or set `scroller.scrollerStyle = .overlay` and prevent SwiftTerm from including it in column calculations.

## Reproduction

1. Run any TUI app that renders near-full-width content (e.g. `kiro`, `claude`)
[...]

PR #12903, opened 2026-09-18
Fix Cloud Codex garble and flicker after pane resize: keep the legacy scroller gutter constant

### LINKED ISSUE #12885: Cloud workspace: Codex TUI garbles and flickers after resizing its pane
## Summary
In a cmux Cloud workspace, a pane running a Codex session becomes badly garbled and keeps flickering after the pane is resized. The Codex TUI is drawn over the neighbouring browser pane's content (Google Calendar in the screenshots), rows overlap, fragments of earlier frames stay on screen, and stray dark blocks appear across the pane.

## Steps to reproduce
1. Open a cmux Cloud workspace (remote machine) and split it so a terminal pane sits next to a browser pane.
2. Start `codex` in the cloud terminal pane.
3. Drag the divider to resize the Codex pane.

## Observed
- Codex output garbles immediately and keeps flickering / repainting.
- The browser pane's page shows through the Codex frame (the terminal background is no longer painted opaque).
- Fragments of previous frames remain (for example a stale `Working (610h 46m 23s ...` line and a partial `rking`), and dark rectangular blocks appear at random positions.

## Expected
The Codex TUI re-lays out cleanly at the new size with an opaque terminal background, the same as a local terminal pane.

## Notes
- Reported by @austinywang on 2026-09-17 with three screenshots (handed to the fixing agent; can be attached here on request).
- Possibly related: #5299 (cmux ssh: terminal corrupts on column grow), #8285 (one-way width resize), #3051 (PTY resize oscillation / flicker), and the closed local-layout issue #12681 with its follow-up PRs #12845 and #12769.

austinywang:
Closes #12885

## Root cause

A Cloud terminal pane mirrors a remote cmux-tui PTY into a local Ghostty surface. Every remote `resized` frame is applied as a reset (`ESC c` + `CSI 3 J`, which empties local scrollback) followed by a replay that carries the history back. The hosted scroll view showed its vertical scroller only while the surface had scrollback, and under the **legacy scroller style** (a mouse attached with "Show scroll bars: Automatically", or "Always") AppKit reserves a 17 pt gutter for a shown scroller, and the terminal grid is derived from the remaining content width.

So each replay flipped the scroller twice and moved the grid by two columns each way; every move was a new size report to the remote PTY; the remote answered each report with another resize and another replay. A closed loop: Codex received `SIGWINCH` about twelve times a second with alternating widths, and its inline-viewport redraws landed in a grid of the other width, which is the garble, the stale fragments, the two-cell spill at the left edge of the composer band in the report's screenshots, and the endless flicker. The loop ran until a timing accident broke it.

The see-through background in the screenshots is not part of the defect: Austin's Ghostty config sets `background-opacity = 0.7`, and the "browser pane" behind the Codex frame is a separate Chrome window (its tab strip, site-info icon and bookmarks bar are visible, and the calendar's red current-time line runs across both cmux panes).

Upstream Ghostty avoids this class by forcing the overlay scroller style. cmux deliberately honors the system style (#10632), which is kept; the fix makes the scroller's *layout* contribution constant instead.

## Fix

`TerminalScrollBarPresencePolicy` (new, in `CmuxTerminalCore/Scrollbar`, with an AppKit-free `TerminalScrollerStyle` input) owns scroller presence: with the legacy style the scroller stays present so the gutter is constant and the grid depends only on the pane; with the overlay style, which reserves nothing, it still hides while nothing can scroll, so it never covers the rightmost column of a full-screen app. `GhosttySurfaceScrollView.shouldShowTerminalScrollBar()` maps AppKit's `scrollerStyle` onto it and calls the policy. This also removes the one-shot reflow of a local pane when its first row scrolls off (#3051) for legacy scrollers.

## Reproduction and evidence

A real Cloud machine is not reachable from the permitted tagged Debug build (it targets the dev backend, where `vm ls` answers `cloud_disabled`), so the repro used the closest faithful equivalent: the app's real Cloud mirror pipeline (`CloudTuiManualMirrorSession` + the manual-mirror `TerminalSurface`) attached through a local-only DEBUG socket hook to a real `cmux-tui` daemon (the prebuilt binary bundled in the installed app, same commit as this checkout) running the real Codex CLI (0.146.0), behind a Unix-socket proxy adding 40 ms each way. The hook and the diagnostic log lines were never committed.

Before (main), after a stepped divider resize with no further input, the instrumented session log shows the local grid alternating 79 ↔ 81 columns on its own, one `resize-attached-view` request per remote replay, 12 per second, 285 PTY resizes and 195 replays in 25 s, until it stopped by chance:

```
16:45:40.011 size.apply cols=81 rows=37 … inFlight=81x37
16:45:40.083 size.apply cols=79 rows=37 … inFlight=81x37
16:45:40.090 size.apply cols=81 rows=37 …
16:45:40.166 size.apply cols=79 rows=37 …
16:45:40.261 size.apply cols=81 rows=37 …
16:45:40.341 size.apply cols=79 rows=37 …
… (12 resize.send per second for 20 s, no user input)
```

After (this branch), the same stepped resize: 13 requests during the interaction, then none, and Codex re-laid out cleanly at the new width (full-width composer band, re-wrapped tip and warning lines, no fragments or blocks). The before state is the log above plus the three screenshots and the 1.7 s screen recording from the report, in which the Cloud pane goes blank and repaints with stray bands during the drag; in my repro Codex was idle, so its frames between replays looked coherent even while the loop ran (the visible garble needs Codex to be mid-redraw, as in the report). The after-state screenshots from the tagged build are on the dogfood Mac (`/var/folders/.../cmux-screenshots/fixed-after-working_*.png`) and are described in the handoff; GitHub's API cannot attach images from the CLI, so they are not embedded here.

## Regression test

`cmuxTests/TerminalScrollBarGutterStabilityTests` hosts a real `GhosttySurfaceScrollView` in an offscreen window, pins the legacy scroller style on its `NSScrollView` (reached through the public view hierarchy, so production code carries no test seam), publishes Ghostty scrollbar packets the way the runtime does (history present → emptied → refilled), and asserts the terminal surface keeps the same content width throughout. It fails on main (the width moves by the gutter: CI run on the test-only commit reported `afterReset → 640.0 == withHistory → 625.0`) and passes with the fix. A second test asserts the overlay style never changes the width. The policy table itself is covered by `TerminalScrollBarPresencePolicyTests` in the package. Commit 1 adds the tests (CI red), commit 2 the fix (CI green).

## Carried main fixes

Two fixes from `main` are carried on this branch so its CI can run; each drops out of this diff once its own PR lands.

- `Sources/Cloud/CloudMachineLinkManager.swift` (+ `PhonePushClient.swift`): main stopped compiling when #12739 made the cmux-tui capability probe optional and #12741 added call sites against the old signature. Same commits as #12913.
- `tests/test_ios_appstore_lane_identity.py` and `.github/scripts/install-app-store-provisioning-profile.sh`: `workflow-guard-tests` has been red on every `main` run since #12935, which stopped overriding `PRODUCT_BUNDLE_IDENTIFIER` on the archive command (it also renamed the notification extension) in favor of the per-target `CMUX_APP_BUNDLE_IDENTIFIER` / `CMUX_HOST_BUNDLE_IDENTIFIER`, and added an extension-profile step to the installer. The guard's fake `xcodebuild` still read the retired variable, so every lane failed the origin verifier with `unexpected release bundle identifier '<absent>'`; and the installer's new certificate-fingerprint pipeline aborted silently under `pipefail` when the certificate could not be read (or hashed empty input). Because `linux-preflight` gates on that job, every macOS job on every PR was skipped. Same commits as #12999.

## Trade-offs, stated

- Legacy-style users now see an (empty, disabled) scroller track in panes with no scrollback, the way Terminal.app behaves. This is the price of a grid that does not depend on content; the alternative (forcing overlay like upstream) would revert #10632.
- The loop is closed at its root (the grid no longer depends on the terminal's own content), but the mirror still applies a `resized` replay into whatever grid the pane has at that moment, so during a fast drag over a slow link the content can look briefly mis-wrapped until the last replay lands (about one round trip after the drag ends). The complete answer is a remote-authoritative grid pin for Cloud mirrors (the tmux mirror already has one), which needs the local grid change to be ordered ahead of the replay bytes in the Ghostty fork; out of scope here.
- `CloudTuiManualMirrorSession.reconcileRemoteGrid()` re-reports the pane's grid whenever the remote's grid differs, which the cmux-tui spec forbids for changes made by another client; two visible projections of one terminal could fight. Not reproduced, left for a follow-up.
- No user-facing strings were added or changed (localization audit: nothing to localize; the new file has comments only).
- The `swift-file-length-budget` and warning budgets are untouched: the hosted view file shrinks by 7 lines (the new logic lives in a new 50-line package file), no new warnings.


austinywang:
## Review audit (re-checked against HEAD 083149925c)

No inline review threads were opened. Top-level bot findings:

| comment | author | file:line | ask | disposition | commit |
| --- | --- | --- | --- | --- | --- |
| CodeRabbit pre-merge check (summary comment, first review) | coderabbitai | `Sources/GhosttyTerminalView.swift` (former `setScrollerStyleForTesting`) | No test or debug seam in production source | fix: the test now pins the scroller style on the scroll view through the public view hierarchy; the DEBUG member is gone | 1df631dc0a |
| CodeRabbit pre-merge check (summary comment, second review) | coderabbitai | `Sources/TerminalScrollBarPresencePolicy.swift` | Move the pure presence decision into `CmuxTerminalCore` with an AppKit-free style input | fix: `TerminalScrollBarPresencePolicy` + `TerminalScrollerStyle` now live in `CmuxTerminalCore/Scrollbar` with their own package test; the view maps `NSScrollView.scrollerStyle` onto it | 1df631dc0a / 34709b0991 |
| CodeRabbit pre-merge check (summary comment, current) | coderabbitai | `Sources/Cloud/CloudMachineLinkManager.swift` | Out-of-scope refactor | already-fixed elsewhere: this is main's compile fix from #12913, carried here so the app builds; it leaves this diff once #12913 merges (documented in the description) | 71a857a40a, 8a81273c27 |
| CodeRabbit pre-merge check (summary comment, all reviews) | coderabbitai | test helpers in `cmuxTests/TerminalScrollBarGutterStabilityTests.swift` | Docstring coverage below 80% | disagree: every touched production function is documented; the remaining undocumented functions are `@Test` bodies whose display names are their documentation | — |
| cubic, Vercel Agent Review, Greptile | bots | — | no findings (neutral) | no change needed | — |

CI proof: test-only commit 1df631dc0a fails on the intended assertion ([run 35293563713](https://github.com/manaflow-ai/cmux/actions/runs/35293563713): `afterReset → 640.0 == withHistory → 625.0`), fix commit 34709b0991 passes ([run 35293566357](https://github.com/manaflow-ai/cmux/actions/runs/35293566357)), and the current head passes ([run 35296856256](https://github.com/manaflow-ai/cmux/actions/runs/35296856256)). Required checks are green.

PR #12931, opened 2026-09-18
Report paste delivery as sent or queued on the terminal socket API

### LINKED ISSUE #12930: API: allow non-queueing literal text delivery — terminal.paste ack conflates delivered and queued
### Problem or motivation

I'm building a macOS app that routes screenshots and prompts into running AI coding sessions (Claude Code, Codex). I'd like to support cmux as a **first-class native transport** through the documented socket API — not through the tmux compatibility shim.

The app holds a hard safety contract for every terminal it injects into:

1. **Never write to a surface that isn't live and ready.** The check has to fail *before* any text is written, not after.
2. **Never report a write as delivered when it was only accepted for later delivery.**
3. **Send literal text and Return as two separate operations**, so a user's prompt is never submitted unless they explicitly chose to submit it.

On **v0.64.22 (build 102)** I couldn't find a native method that satisfies (1) and (2) together — even for a live, in-window surface. Every candidate either queues silently or returns an acknowledgement that can't distinguish delivery from queueing.

What I measured (tagged source at `v0.64.22`):

| Candidate | Observed | Why it doesn't meet the contract |
| --- | --- | --- |
| `surface.send_text` | Raw recorder received CR for an input LF; forwards `sendInputResult`, which permits queueing | Multiline text isn't literal. A queued reply arrives after the side effect is already accepted. |
| `terminal.input` / `mobile.terminal.input` | `TerminalController.swift:14740` forwards to the same `sendInputResult`; `.queued` is a *success* with a queued flag | Same raw-newline and queue limitations. No target-readiness fence. |
| `terminal.paste` / `mobile.terminal.paste` with `submit_key: none` | **Closest.** Correct UTF-8 and literal LF inside bracketed paste. Handler at `TerminalController.swift:14929` consumes `TerminalPanel.sendText` | `TerminalPanel.swift:695` forwards the boolean from `TerminalSurface+Input.swift:73`, which means delivered **or** queued. The response doesn't distinguish them, and no atomic ready/no-queue parameter is consumed. |
| `surface.action` / `tab.action` | Action dispatch read in `TerminalController+ControlSystemContext2.swift` | Tab management (rename/pin/move/reload/duplicate/create/close), not an arbitrary literal-paste operation. |
| Health check before the write | Native readback worked; health returned `in_window` | An earlier observation doesn't establish readiness at a later write. `in_window` is attachment state, not a transaction fence. |

So today the only way to use cmux natively would be to send text and *hope* it wasn't queued — which is exactly the failure mode the contract exists to prevent.

### Proposed solution

Either of these would unblock a native integration. They're independent — one is enough.

**Option A — an explicit no-queue request (preferred).** An optional parameter on `terminal.paste` / `mobile.terminal.paste`:

```json
{
  "method": "terminal.paste",
  "params": { "text": "...", "submit_key": "none", "require_ready": true }
}
```

With `require_ready: true`, if the target surface can't accept the write right now, return a typed error (e.g. `surface_not_ready`) and **write nothing**, instead of queueing. This is preferable because it lets the caller fail closed before any side effect exists.

**Option B — a truthful delivery outcome.** Keep current behavior, but make the response distinguish the two cases that `TerminalSurface+Input.swift:73` already knows apart internally:

```json
{ "result": { "delivery": "delivered" } }
{ "result": { "delivery": "queued" } }
```

Option B is still sufficient: a caller that gets `queued` can treat it as a failure and never send the Return key, then surface an honest partial-delivery result to the user.

Either option keeps the existing default behavior intact for callers who don't opt in.

### Alternatives considered

- **`surface.send_text` / `terminal.input`** — ruled out: LF is converted to CR, so literal multiline text can't be sent without submitting, and `.queued` is returned as a success.
- **Health preflight (`surface-health`, `in_window`) then write** — ruled out: this is time-of-check-to-time-of-use. The source identifies `in_window` as attachment state rather than a transaction fence, so an earlier observation doesn't establish readiness at the later write.
- **Read-back after the write** — I already do read-back (`read-screen` / read-text works well), and it's a good oracle for the *positive* case. But read-back can't un-send text. The contract requires rejecting *before* the write when readiness is uncertain, and no amount of post-hoc reading provides that.
- **tmux compatibility shim** — deliberately excluded. I want cmux surfaces handled as native cmux surfaces with their own identity and capability facts, not masqueraded as tmux. Falling back to the shim would also misreport which transport the user is actually on.
- **Ship cmux as detect-only** (list sessions, disable delivery) — the current fallback, but it means cmux users see their sessions and can't actually use them, which is the worst of both.

### Additional context

**Version measured:** cmux v0.64.22 (build 102), macOS 26.5.1. Measurements were taken against a disposable automation-mode app instance with owned teardown; no input was sent to any existing user terminal.

**Related but distinct:** #1472 covers background workspaces whose PTY is never created (`Surface not ready` on read-screen/send-key), with PR #6236 in flight. That's about surfaces which have no live PTY at all. This request is about a **live, in-window surface** where the write is accepted but the acknowledgement can't tell the caller whether it landed. Fixing #1472 wouldn't resolve this one.

**What already works really well, for context —** I don't want this to read as only complaints:

- `terminal.paste` with `submit_key: none` delivers correct UTF-8 and literal LF inside bracketed paste. The literal-text half is exactly right; only the acknowledgement is ambiguous.
- `agent.resolve_delivery_target` is excellent for this use case. With `pid_resolution: controlling_tty` it correctly mapped two inert processes with *identical* titles and cwd to their distinct surfaces, returned `not_found` for a nonexistent PID without falling back to a presentation target, and correctly rejected an ambiguous TTY mapping. That last behavior is the right call and is what makes safe targeting possible at all.
- Separate text and key operations, plus native read-back, are the right primitives. The missing piece is just a way to say "don't queue this".

Happy to test a patch against my integration and report back, or to open a PR if you can point me at the preferred shape (Option A vs B).

chriskr7:
Closes #12930.

## Problem

`TerminalSurface.sendText` takes two structurally distinct branches and collapses both into `true`:

```swift
private func sendTextAfterExplicitInput(_ data: Data) -> Bool {
    ...
    guard surface != nil else {
        guard allowsRuntimeSurfaceCreation() else { return false }
        let queued = enqueuePendingSocketInput(.pasteText(data))   // ← queued
        if queued { requestInputDemandSurfaceStartIfNeeded(); didAcceptExplicitInput() }
        return queued
    }
    ...
    writeTextData(data, to: liveSurface)                           // ← delivered
    didAcceptExplicitInput()
    return true
}
```

Its own doc comment says so — "Whether the text was delivered **or** queued". A `terminal.paste` client therefore cannot tell whether the text is on screen or still in the pending-input queue. That matters for any caller that must not act on text it cannot see: in my case a tool that pastes a prompt into an agent session and must never press Return on text it has not confirmed, and must never tell the user it delivered something that has not landed.

The information is not missing — it is discarded at the return.

## Approach

The surface layer already solves this twice, so this follows the existing convention rather than introducing a new one:

| Path | Typed result |
| --- | --- |
| `sendInputResult` | `InputSendResult` |
| `sendNamedKey` | `NamedKeySendResult` |
| `sendText` | **`Bool`** ← the odd one out |

`InputSendResult` ("The outcome of sending text input to a surface") already has exactly the right cases, so this reuses it rather than adding a type. If you would rather have a distinct `TextSendResult`, splitting it out is mechanical — say the word.

## Changes

- `TerminalSurface.sendTextResult(_:) -> InputSendResult` — each branch maps to `.sent`, `.queued`, `.surfaceUnavailable`, `.inputQueueFull` or `.processExited`
- `TerminalPanel.sendTextResult(_:)` and `ControlTerminalSocketTarget.sendTextResult(_:)` — mirroring the `sendInputResult` / `sendNamedKeyResult` pairs already alongside them
- `terminal.paste` / `mobile.terminal.paste` gain a `delivery` field: `"delivered"` or `"queued"`

## Compatibility

Deliberately additive:

- `sendText` stays, still returns `Bool`, now a one-line shim over `.accepted`. **No existing caller changes** — the `TextBoxSubmitSurfaceControlling` conformance included.
- Every `Bool` result is preserved branch for branch, including the empty-text early return (`.sent.accepted == true`) and the enqueue-failure case (`.inputQueueFull.accepted == false`).
- The paste error code stays `surface_unavailable`. The specific cause is added under `data.reason` instead, mirroring the detail the submit-key path in the same handler already reports.
- `delivery` is a new response field; clients that ignore it are unaffected.

## Verification

```
$ swift build                                          # Packages/macOS/CmuxTerminal
Build complete! (11.57s)

$ swift test --filter TerminalSurfaceExplicitInput
✔ Test run with 18 tests in 1 suite passed after 0.076 seconds.

$ xcodebuild -project cmux.xcodeproj -scheme cmux \
    -configuration Debug -destination "platform=macOS" build   # app target
** BUILD SUCCEEDED **
```

New test in `TerminalSurfaceExplicitInputTests` pins the actual distinction on a cold surface, where `sendText` answers `true` for both branches:

```swift
#expect(fixture.surface.sendTextResult("hello") == .queued)
#expect(fixture.surface.sendText("hello"))
```

`emptyInputDoesNotNotifyThePaneHost` gains the matching empty-input leg for the new entry point, alongside the existing ones.

Happy to adjust the shape — a distinct `TextSendResult`, a different field name than `delivery`, or an opt-in `require_ready` parameter that rejects instead of queueing (Option A in #12930, which I would still find more useful for fail-closed callers, but which is a behavior change rather than a reporting one).

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12931"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792300172&installation_model_id=21920&pr_number=12931&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12931&signature=4254ab134675f572b68363010dbfaa30038badcb9313513b938b7155e1304355"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Closes #12930. Reports whether pasted text was delivered to the live surface or queued for a cold one. `sendText` collapsed both branches into `true`, so `terminal.paste` clients couldn't tell if text was on screen. Adds `sendTextResult(_:)` returning the existing `InputSendResult` and reports it as a new `delivery` response field.

- `sendText` stays as a one-line shim, so no existing callers change.
- The error code remains `surface_unavailable`; the specific cause now rides in `data.reason`.

<sup>Written for commit 25d943499c80455cccf0e8b2d9a4369956c9b315. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12931?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Terminal paste responses now indicate whether text was delivered immediately or queued for delivery.
  - Paste responses include machine-readable delivery status and failure reasons.
  - Existing boolean paste APIs continue to report whether the request was accepted.

- **Bug Fixes**
  - Improved distinction between unavailable surfaces, exited processes, full input queues, and text queued while a terminal starts.
  - Empty input continues to be handled successfully without triggering unnecessary notifications.


chriskr7:

chriskr7:
One more request, from building the client side against this: **please also advertise the outcome as a `system.capabilities` token**, not only as a response field.

### Why a response field alone cannot be used safely

My caller's rule is to refuse a write when it cannot know the write landed. To decide that from the *response* shape, I would have to send a paste and look at what comes back — which is the write I was trying to avoid. It is circular: discovering whether it is safe to send requires sending.

So the decision has to be answerable **before** the first write, and the only pre-act channel is the handshake.

### Measured against v0.64.25 (build 106)

`system.capabilities` already returns a named token list, and it is clearly the intended place for this kind of negotiation — 51 tokens today, including `terminal.input.ordered.v1`, `terminal.render_grid.verified_replay.v1` and `terminal.replay.v1`. None of them says anything about a write's outcome.

And the current `terminal.paste` reply, against a live disposable surface with `submit_key: "none"`:

```json
{"ok": true, "result": {"surface_id": "…", "workspace_id": "…", "submitted": false}}
```

A follow-up `surface.read_text` *did* show the pasted marker, so delivery and read-back both work — the acknowledgement simply cannot say which of the two branches ran.

### The ask

Alongside the `delivery` field this PR adds, advertise something like:

```
terminal.paste.delivery.v1
```

Then a client can gate on the handshake, fail closed on releases that do not have it, and start permitting writes the moment a user updates — with no client release in between. Without the token, every client has to ship an update that knows which cmux versions to trust, which is exactly the coupling capability negotiation exists to avoid.

Happy to add it to this PR if you point me at where the token list is assembled.

chriskr7:
Measured this end to end against a disposable **v0.64.25 (build 106)** instance rather than reading it off the source, and the asymmetry shows up directly in the wire replies. Same socket, same session, two adjacent calls:

```jsonc
// surface.send_key  {"surface_id": "<id>", "key": "enter"}
{"ok": true, "result": {"queued": false, "surface_id": "…", "surface_ref": "surface:2", …}}

// terminal.paste    {"surface_id": "<id>", "text": "…", "submit_key": "none"}
{"ok": true, "result": {"submitted": false, "surface_id": "…", "workspace_id": "…"}}
```

**cmux already reports delivery truthfully for keys, today, in the protocol.** `surface.send_key` returns `queued`, which is `NamedKeySendResult`'s `.sent` / `.queued` reaching the client. `terminal.paste` returns `submitted` — which is about the submit key, not about whether the text arrived — and nothing else.

So this PR is not asking for a new idea. It is asking for the text path to answer the question the key path already answers on the same socket.

A couple of other things the same run confirmed, in case they are useful:

- **Delivery and read-back both work.** A `submit_key: "none"` paste carrying `한글-テスト-🙂` and a newline came back intact through `surface.read_text`, and `surface.send_key enter` then executed the line as its own act. The primitives are right; only the acknowledgement is silent.
- **A stale surface fails closed.** Pasting to a closed surface's UUID answers `{"ok": false, "error": {"code": "not_found"}}`, which is exactly what a careful client wants.
- **A fresh split is listed before it can act** — `in_window: true`, `render_health: "not_started"`, and no `tty` yet. That is issue #1472's shape from the client side, and it is the case where a queued paste is most likely and least visible.

Still happy to add the `terminal.paste.delivery.v1` capability token alongside the field if you want it in this PR — as noted above, without a handshake token a client cannot gate on the field without first performing the write it is trying to decide about.

Issue #12448, opened 2026-09-13
cmux Cloud: new terminal shortcuts do not inherit the active remote working directory


austinywang:
## Report

In the **cmux Cloud product**, creating a new terminal with a keyboard shortcut does not inherit the active terminal's current remote directory. It should inherit that directory by default.

User report:

> one of the issues with cloud is that when i do the shortcuts of making a new terminal and stuff it doesn't seem to inherit my previous directory... by default it should do that though...

Clarification:

> this is for the cmux cloud product

The supplied screenshot shows a Cloud workspace named `GOD WORKSPACE`, an existing terminal showing a project directory in its agent status line, and a newly opened lower terminal showing `cmux@cmux in ~`.

## Reproduction

1. Open a terminal in a cmux Cloud workspace.
2. Change to a remote directory other than the remote home directory.
3. With that terminal active, use a keyboard shortcut to create another terminal/tab or split in the same Cloud workspace.
4. Check the new terminal's current directory with `pwd`.

The report does not identify a specific key combination; cover the supported new-terminal and split shortcuts.

## Actual

The new Cloud terminal starts in the remote home directory rather than inheriting the active terminal's current remote directory.

## Expected

By default, the newly created terminal starts in the active source terminal's current remote directory on the same Cloud machine. In a workspace with multiple terminals at different directories, inheritance follows the terminal that initiated creation.

## Related

- https://github.com/manaflow-ai/cmux/issues/12048: similar symptom in `cmux ssh`; this report specifically concerns cmux Cloud.
- https://github.com/manaflow-ai/cmux/issues/10756: earlier cmux-tui headless terminal default-directory fix; verify whether a related path is involved.
- https://github.com/manaflow-ai/cmux/issues/12347: Cloud directory provenance in sidebar metadata.

PR #12968, opened 2026-09-19
Cloud shortcuts preserve cwd through native layout creation

austinywang:
## Problem
Cloud shortcuts that create a tab or split can take the native cmux-tui layout path when the source projection has a remote tab ID. That path skipped the live foreground-process cwd query added for the fallback path, so a shell that had changed directories could still create the new remote terminal in its spawn directory.

## Change
Pass the source terminal's live cwd through the layout-creation protocol and emit it as `cwd` for both `pane.split` and `pane.run`. Empty values keep the daemon's existing fallback behavior, while non-blank paths preserve significant whitespace. The live process query runs behind a nonisolated concurrent helper so the main actor only awaits its result.

This is a follow-up to #12448 / #12450 and keeps the local and SSH creation paths unchanged.

## Validation
- Added command-level regression coverage for Cloud native split and tab creation, including cwd forwarding, blank-value handling, and path whitespace preservation.
- `python3 scripts/swift_file_length_budget.py` passes.
- `git diff --check` passes.
- CodeRabbit review passes, including Swift actor/concurrency checks.
- The targeted Xcode test run was attempted but stopped at package resolution while other local builds held the shared build service; no local test result was produced.
- GitHub CI reaches the workflow guard, which currently fails on a missing temporary `ExportOptions.plist` fixture before macOS compilation; dependent Linux preflight/tests fail from that guard result.

## Summary by CodeRabbit

* **New Features**
  * New terminals created from an existing remote terminal now open in that terminal’s current working directory.
  * Terminal creation supports specifying a working directory, including for split panes and new tabs.
  * Existing terminal creation behavior remains unchanged when no working directory is provided.

* **Bug Fixes**
  * Improved working-directory detection and handling for terminal resources, including graceful fallback when unavailable.


lawrencecchen:
Mac fleet instructions for head `eb153ad0e839590e4beb3f14db0de4075815d0c7`. Planned tag: `pr-12968-eb153ad0`; this is not yet a published build.

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-12968-eb153ad0 /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git eb153ad0e839590e4beb3f14db0de4075815d0c7' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/12968 --source-digest eb153ad0e839590e4beb3f14db0de4075815d0c7 --cache-key cmux:pr-12968 --min-free-bytes 268435456000 --label cmux --label ram48)
```

PR #12971, opened 2026-09-19
reload: add explicit build-only validation mode

teamleaderleo:
## Reviewer summary

Adds an explicit build-only reload mode for compiling and checking a tagged build without replacing the running app.

## Checks

y cubic
Adds a `--build-only` mode to `scripts/reload.sh` so tagged builds compile and validate without touching the running app, `cmuxd`, socket, or tag state. Also fixes the pre-build cleanup so it no longer removes the tagged bundle a running app executes from.

- Requires `--tag`; fails if combined with `--launch`.
- Skips tag cleanup, app/daemon termination, log-path writes, and reload-state publication; the active runtime stays on its previous revision.
- Stages the new bundle separately and removes the temporary artifact after validation.
- Plain `--tag` still replaces the same-tag runtime by default.
- Adds a CI regression test for tagged-bundle preservation and documents the mode in `CLAUDE.md` and the tagged-builds reference.

<sup>Written for commit 591d924e67133bfabd19234a6aff90f73bac2eb7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12971?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a `--build-only` option to validate a tagged build without replacing or stopping the currently running app.
  - Build-only validation keeps the active app, daemon, and tag state unchanged while staging the new bundle separately.
  - Normal reload behavior remains unchanged.

- **Documentation**
  - Added usage guidance and examples for the new build-only workflow.
  - Clarified that the option cannot be combined with launch behavior.

- **Tests**
  - Added automated coverage to verify that build-only validation preserves the active tagged app.


teamleaderleo:
Follow-up polish after self-review:

- build-only now skips stale CLI/tag cleanup, so it cannot delete or rewrite existing discovery state before the build;
- it no longer updates `/tmp/cmux-<tag>` or `/tmp/cmux-last-debug-log-path`;
- successful validation removes its temporary staged app bundle, while failure cleanup remains intact;
- the success output avoids advertising paths that have already been removed.

Validation: `bash -n scripts/reload.sh`, invalid `--build-only --launch` guard, `git diff --check`, and `node scripts/lib/reload-shim.test.mjs` (11/11). A full tagged Xcode build remains unavailable in this environment.

teamleaderleo:
Verified on the local Mac mini after initializing the pinned submodules:

- `CMUX_DEV_BACKEND_MODE=local CMUX_DEV_BACKEND_URL=https://cmux-dev-backend-1.tail137216.ts.net:4170/ ./scripts/reload.sh --tag codex-build-only-local --build-only`
- full Debug Xcode build completed successfully in 980s;
- build-only completion reported the running app, cmuxd, and tag state unchanged;
- temporary staged artifact was removed on success;
- worktree remains clean.

teamleaderleo:
Performance measurements on the same local Mac:

- Baseline (`9c2ba78`, normal tagged reload, isolated DerivedData): **1320.69s** wall time.
- PR (`8f413003`, first full build-only validation): **980s** wall time.
- PR warm build-only rerun with the same DerivedData cache: **59.51s** wall time.

The cold numbers are not a claimed compile-speed regression/improvement: they were run at different times under different concurrent machine load, and build-only skips runtime teardown/publication by design. The useful result is that the build-only path preserves the expensive compile behavior while avoiding runtime-state churn; its cached validation pass completed in about one minute.

teamleaderleo:
Rebased onto current `main` and added two commits: `11092810ca` (failing test) and `591d924e67` (fix).

**Bug found while validating this with the real `scripts/reload.sh`:** `--build-only` deleted the existing tagged bundle (`Build/Products/Debug/cmux DEV <tag>.app`) and never recreated it, while reporting that the running app was left unchanged. Before every build, `cleanup_incomplete_xcodebuild_outputs` removes `XCODEBUILD_TAG_APP_PATH`. That is fine for a normal reload, which replaces the bundle afterwards, but build-only skips the replacement. That bundle is the one a running tagged app executes from. The same function also runs from the failure trap, so a failed build-only build would have removed it too.

**Fix:** the cleanup keeps the tagged bundle when `BUILD_ONLY=1` and still clears this run's own outputs. The new test runs the real functions extracted from `reload.sh` and also pins that a normal reload still clears a stale tagged bundle.

**Validation (M5 MacBook Air, real `reload.sh`, tag built beforehand):** state snapshot before vs after `--build-only` following a genuine Swift edit is identical: tagged bundle present with the same executable inodes and CDHash, tag state files unchanged, no staged `.reload-*.app` left behind; the build took 51.8 s. `--build-only --launch` refuses in under a second and changes nothing. A plain reload as a control does change the snapshot, so the method can see a replacement.

**Not validated:** I could not launch a tagged app here (`--launch` needs `~/.secrets/cmuxterm-dev.env`), so the claims about a live process, `cmuxd` and socket ownership were checked only with no app running. To get past the shared-backend gate outside cmuxterm-hq, the run used this branch merged locally with #12973.

teamleaderleo:
Continued in https://github.com/manaflow-ai/cmux/pull/13127 (same commits, in-org branch `feat/explicit-build-only`) so its macOS jobs run on the repository's configured runners. Push further changes to `feat/explicit-build-only` on manaflow-ai/cmux, not to the fork.

PR #12976, opened 2026-09-19
build: stabilize project resources for incremental builds

teamleaderleo:
## Summary

Two malformed resource entries in `project.pbxproj` make Xcode produce a different project description on every load, so it re-plans every build. Both are repaired, and both are needed (fixing only the first still gave 5 descriptions in 5 loads):

- `Resources/bin/open` belonged to no group, so Xcode invented a `Recovered References` group with a new id each load. It is now in the Resources group.
- The English `cloud-agent-skill.md` was not under `en.lproj/` like its Japanese variant. It moves there; the coverage script and tests follow.

**Behavior change:** on `main` Xcode drops that variant group and ships the English file unlocalized, so the Japanese file is never selected. With this change, Japanese-preferring users get `ja.lproj/cloud-agent-skill.md`. The literal `resourceURL/cloud-agent-skill.md` fallback in `CloudAgentSkillLauncher.skillMarkdown` no longer matches anything.

## Testing

Parent commit `9c2ba78be4` against this PR, separate worktrees and DerivedData, builds alternated, `./scripts/reload.sh --tag <tag> --prod-auth`. M4 Pro, Xcode 26.3, on a machine other agents were also building on. The edit is one comment line in one Swift file.

| | project descriptions, repeated loads | no-op build | one-file edit | build descriptions written |
| --- | ---: | ---: | ---: | ---: |
| parent commit | 3 in 3 | 35 s, 34 s | 53 s, 55 s | one per build |
| this PR | 1 in 7 | 26 s, 27 s | 50 s, 51 s | 0 |

Commands for the project-description check are in the first comment.

**Not explained:** after the PR worktree sat idle for ~25 minutes, Xcode wrote a new build description anyway, twice, with the project description unchanged. The old and new descriptions differ only in the order of compiled `<lang>.lproj/Localizable.strings` entries. After the first gap the next one-file edit recompiled the whole `cmux` target (608 s); after the second it did not (52 s). The parent commit re-plans on every build, so it cannot show this either way. Something besides the project file invalidates the build description, and this PR does not address it.

Passed: `tests/test_cloud_vm_skill_coverage.py`, `tests.test_normalize_pbxproj`, `./scripts/check-pbxproj.sh`. `cmuxTests` has not been run against the relocated resource, and no app test job ran on this PR's CI.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12976?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792371457&installation_model_id=21920&pr_number=12976&ref=codesmith_pr_footer&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12976&signature=bc4e71e620e15a4a86e5580b027808f0f4d111b55e9a7782a0bf934731f71751"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes two malformed Xcode project resource entries so incremental builds reuse a stable build description instead of synthesizing project references during project loading. No-op builds in the malformed state dropped from about 97–101s to 72–75s.

- Attaches the orphaned `Resources/bin/open` file reference to the Resources group.
- Moves the English `cloud-agent-skill.md` into `Resources/en.lproj/` and updates the project file path plus the direct coverage/test references.

<sup>Written for commit 129fac969d0e3d7d5f1086d4534743f665639a23. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12976?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Updated references to the bundled Cloud Agent skill documentation to use its English localization path.
  - Ensured the documentation is correctly included and discoverable within the app bundle.
  - Content remains unchanged; only its bundled location was updated.

- **Tests**
  - Updated validation messages and coverage checks to reflect the localized documentation location.
  - Confirmed the bundled documentation remains available at its updated path.


teamleaderleo:
## Reproducing the project-description check

Results are in the PR description. These are the commands.

Setup: a clean worktree with `vendor/bonsplit` and `ghostty` populated and a prebuilt `GhosttyKit.xcframework` present; without the xcframework, project loading fails with `local binary target 'GhosttyKit' … does not contain a binary artifact`.

Counting distinct project descriptions (~10 s per load after the first):

```bash
for i in 1 2 3; do
  xcodebuild -showBuildSettings -project cmux.xcodeproj -scheme cmux -configuration Debug \
    -destination 'platform=macOS' -derivedDataPath /tmp/pif-dd >/dev/null
done
ls /tmp/pif-dd/Build/Intermediates.noindex/XCBuildData/PIFCache/project | grep '^PROJECT@'
```

One `PROJECT@…` entry means every load produced the same description. `main` at `e91468b846` gave three entries for three loads with the same content hash and a different `mod` value each time.

Diffing two dumps directly:

```bash
xcodebuild -project cmux.xcodeproj -dumpPIF /tmp/pif-1.json -clonedSourcePackagesDirPath /tmp/sp -skipPackagePluginValidation
xcodebuild -project cmux.xcodeproj -dumpPIF /tmp/pif-2.json -clonedSourcePackagesDirPath /tmp/sp -skipPackagePluginValidation
shasum /tmp/pif-1.json /tmp/pif-2.json
```

On `main` the dumps differ at two leaf ids, and every other difference is a signature derived from them: the synthesized `Recovered References` group at `/groupTree/children[12]` in the project PIF, and `/buildPhases[2]/buildFiles[3]/guid` in the `cmux` target PIF. With this PR the dumps are byte-identical.

teamleaderleo:
The `xcodebuild -list` check that could not run in that environment does run if the submodule contents are supplied without touching the checkout's gitlinks. `vendor/bonsplit` and `ghostty` are identical between `e91468b846` and the fix commit, so rsyncing them into the worktree (rather than `git submodule update`, which repoints the shared gitdirs) is enough; a prebuilt `GhosttyKit.xcframework` is also required, otherwise `-list` fails with `local binary target 'GhosttyKit' … does not contain a binary artifact`.

With that, the claim can be measured without building at all. `xcodebuild -showBuildSettings -project cmux.xcodeproj -scheme cmux -configuration Debug -destination 'platform=macOS' -derivedDataPath <throwaway>` populates `<DD>/Build/Intermediates.noindex/XCBuildData/PIFCache/`, and counting distinct `PROJECT@` keys across repeated loads shows the synthesis directly. ~10 s per load after the first.

| tree | malformed entries left | loads | distinct `PROJECT@` keys |
| --- | --- | ---: | ---: |
| `main` at `e91468b846` | orphan + variant group | 3 | **3** |
| orphan attached only | variant group | 5 | **5** |
| both repaired | none | **7** | **1** |

Stable key across all 7 loads: `PROJECT@v11_mod=3ec5fa1ab24d62338dd66ac030843a46_hash=b8a004fb84f0cbdf2516cef41548d515`. On `main`, three loads gave three different `mod` values with the same content hash, and `project.pbxproj` mtime and size unchanged (`1789779661 2155464`) — load nondeterminism, not a file change.

Two things this pins down:

**Both repairs are load-bearing.** Attaching the orphan alone leaves it unstable (5 loads, 5 keys), so the `en.lproj` move cannot be dropped to avoid relocating the resource in the bundle.

**They churn different guids in different PIFs.** The orphan churns the synthesized `"Recovered References"` group at `/groupTree/children[12]` in the project PIF — the file reference's own guid is stable, only the invented group's is not. The variant group churns `/buildPhases[2]/buildFiles[3]/guid` in the `cmux` target PIF, and reaches the project PIF indirectly because targets are referenced by hash (`/targets[0]` flips between `TARGET@v11_hash=14a04858…` and `TARGET@v11_hash=e3515b32…`).

Sensitivity was checked rather than assumed: reintroducing a *different* malformed entry on top of the repair — a duplicate `PBXBuildFile` id, which exists on a downstream branch — makes the count churn again (3 loads, 3 keys). So one key across seven loads means no residual synthesis, not merely "fewer anomalies".

---

Separate from build time, and possibly worth its own line in the description: **Xcode is not preserving the variant group at all today.** In the PIF built from `main`, `cloud-agent-skill.md` appears as a plain `{"type": "file", "path": "cloud-agent-skill.md", "fileType": "net.daringfireball.markdown"}` under `Resources` — the `PBXVariantGroup` is gone and the `ja` child with it. The Japanese localization of that resource is currently not being treated as a localization. This PR's path repair is what restores it.

teamleaderleo:
Clarification on the performance evidence: the 97–101s to 72–75s wall-clock measurement came from the fork experiment with one additional malformed project object. On upstream, this PR has verified the important precondition—project descriptions are unstable across repeated `main` loads and byte-identical across repeated loads with this fix—but upstream wall-clock timing is still pending. The measured fork timing should not be read as a universal upstream speedup claim.

Validation on the current PR checkout passes:
- `python3 tests/test_cloud_vm_skill_coverage.py`
- `python3 -m unittest tests.test_normalize_pbxproj` (5 tests)
- `./scripts/check-pbxproj.sh`

PR #12994, opened 2026-09-19
fix: close control socket recovery races

teamleaderleo:
## Problem

The control socket could miss a path deletion or replacement during watcher setup. A stale listener could then keep running, or a later configuration update could change permissions on a path it no longer owned.

## Change

The path monitor now performs its first ownership check from the DispatchSource registration callback. The callback is tied to the watcher lifecycle and uses the existing generation and inode checks.

Configuration updates also stop and fail closed when the listener no longer owns its configured path.

## Testing

- `swift test --package-path Packages/macOS/CmuxControlSocket --filter SocketControlServerPathMonitorTests`
- Covers unlink before monitor registration.
- Covers deletion, replacement inode preservation, repeated recovery, and stale generations.
- `git diff --check`

## Demo Video

Not applicable. This is a socket recovery fix with no user-facing visual flow.

## Review Trigger

```text
@codex review
@coderabbitai review
@greptile-apps review
@cubic-dev-ai review
```

## Checklist

- [x] I added regression coverage for the reported races
- [x] I verified the current review findings against the current branch
- [ ] Human review completed

Tact #79: https://github.com/teamleaderleo/Tact/issues/79

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved local socket listener reliability when its socket path is moved, replaced, or no longer owned.
  - Listeners now stop safely when they cannot reapply required permissions, helping prevent stale or invalid connections.
  - Improved monitoring during startup to detect socket-path changes without a timing gap.
  - Prevented recovery callbacks from outdated monitoring sessions from affecting the active listener.
  - Preserved replacement sockets and regular files when configuration recovery fails.


teamleaderleo:
Replaced by https://github.com/manaflow-ai/cmux/pull/13229: same commits, head branch moved into the org.

PR #12997, opened 2026-09-19
Fix Option dead-key composition in terminal

### LINKED ISSUE #12947: Regression (#12343, v0.64.23+): Option dead-key accent composition silently swallowed, not sent to terminal
### cmux version

0.64.24 (also present in 0.64.23 and 0.64.25)

### macOS version

macOS 27

### Mac chip

Apple Silicon (M1/M2/M3/M4)

### Installation method

Built from source

I could not test NIGHTLY

### Bug description

On a U.S. keyboard layout, Option+E is the standard macOS dead key for acute
accent (Option+E then E → é). This works correctly in every other app
(TextEdit, Safari, Notes, real Terminal.app) and worked in cmux before this
regression.

Since #12343 (merged 2026-09-11, first shipped in v0.64.23), pressing
Option+E in a cmux terminal surface produces **no visible effect at all**
(no accent marker, nothing sent to the shell), and the following E keystroke
then produces a plain, unaccented "e", as if the dead-key state was never
entered. This is not the "claimed key sent to terminal as Alt+e" trade-off
described in the PR body; the keystroke appears to be swallowed entirely.

I also tested with the right-hand Option key specifically, since the PR
description states "Unclaimed/right-side Option composition remains
available through the existing sided modifier translation", same result,
no accent, on both Option keys.

This affects every dead-key accent (é, è, à, ç, ã, ñ, î, ü, etc.) for French on the standard U.S.
layout dead-key accents rather than switching input sources, making cmux
unusable for typing accented text in the terminal.

### Expected behavior

Option+E then E should compose "é" as it does system-wide, matching the
behavior restored by #4382 before #12343's macos-option-as-alt change
removed the dead-key preservation path.

### Steps to reproduce

1. Set keyboard input source to "U.S." (Apple's standard layout).
2. Focus a cmux terminal surface.
3. Press Option+E, release, then press E.
4. Observe: no accent marker appears on Option+E, and E alone produces a
   plain unaccented "e".
5. Compare with any other app (TextEdit, Terminal.app): Option+E shows the
   pending accent marker, then E produces "é".

### Shell and environment

_No response_

```text

```

_No response_

### Additional context

Related: #1653, #3981 (printable Option keys, different case, not dead
keys), #4377, #4382 (original dead-key fix), #5223 (press-and-hold, also
different from dead keys), #12343 (this regression's likely source).

Workaround: pinning to v0.64.22 (last release before #12343) restores
correct dead-key composition.

austinywang:
Fixes #12947.

Option dead-key events (for example Option+E on a US layout) must reach AppKit with the original Option modifier so the system can enter and complete dead-key composition. The terminal still uses Ghostty's translated event for ordinary Option-as-Alt input. Added regression coverage to ensure dead-key events preserve Option and are not emitted as translated terminal text.

Validation: `git diff --check`. Full Xcode validation was unavailable in this environment because `xcodebuild` is not installed.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/12997?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792376332&installation_model_id=21920&pr_number=12997&ref=codesmith_pr_footer&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F12997&signature=44b5d2050bd986a0632c743642c5424c2143624c60fd2e48ce230a18356c2b3e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes Option dead-key composition in the terminal so AppKit receives dead-key events (like Option+E on a US layout) with the original Option modifier and can complete the composition. Previously, these events were replaced with Ghostty's translated event, which prevented dead-key composition from starting and caused the following key to be swallowed. Ordinary Option-as-Alt input is unchanged, and regression tests now cover the dead-key path. Fixes #12947.

<sup>Written for commit 456ba2a9e5deb3b7b78dbaed94cd956ac962a7a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/12997?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Option-based dead-key input so accented characters and other compositions are processed correctly by macOS.
  * Prevented dead-key events from being incorrectly translated or forwarded as raw terminal input.

* **Tests**
  * Updated input handling coverage to verify that dead-key composition remains with macOS and does not produce unintended terminal output.


lawrencecchen:
Mac fleet instructions for head `456ba2a9e5deb3b7b78dbaed94cd956ac962a7a7`. Planned tag: `pr-12997-456ba2a9`; this is not yet a published build.

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-12997-456ba2a9 /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git 456ba2a9e5deb3b7b78dbaed94cd956ac962a7a7' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/12997 --source-digest 456ba2a9e5deb3b7b78dbaed94cd956ac962a7a7 --cache-key cmux:pr-12997 --min-free-bytes 268435456000 --label cmux --label ram48)
```

teamleaderleo:
Reviewed, and I think this needs one change before it lands.

`isDeadKey` checks the original, Option-bearing flags. So on a layout where Option+e/u/n/i/` are dead keys, those keys now always go to AppKit composition, even when `macos-option-as-alt` explicitly claims Option. Users who set `macos-option-as-alt = true` would lose M-u, M-n and friends in readline and emacs. It also flips the assertion in main's `DeadKeyCompositionRegressionTests` that #12343 added on purpose.

The root cause of #12947 looks like Ghostty's `detectOptionAsAlt()`, which makes Option-as-Alt the default on US and US-International layouts when the setting is unset. #12343 made cmux follow that. The narrower fix: keep the original event for composition only when `macos-option-as-alt` is unset (auto-detected), never when the user set it. Whether to diverge from Ghostty's US default at all is a product call.

The conflict with main is only `cmuxTests/CJKIMEInputTests.swift` (main wrapped those tests in `AppContextSerialGate`), and the red checks are inherited from main.

PR #12999, opened 2026-09-19
Fix workflow-guard-tests: App Store lane guard follows the per-target bundle id contract

austinywang:
## Why

`workflow-guard-tests` has failed on every `main` run since #12935, and because `linux-preflight` gates on it, every macOS job on every PR is skipped. The step `Validate iOS App Store lane identity` fails:

```
error: production iOS artifact has unexpected release bundle identifier '<absent>'
error: archive runtime origins are not production-only; refusing to export or upload
FAIL: beta export-only lane succeeds with fake Apple tools
FAIL: beta archive command stamps the beta bundle id
FileNotFoundError: .../beta-upload-test/ExportOptions.plist
```

#12935 stopped passing `PRODUCT_BUNDLE_IDENTIFIER=` on the archive command line (the global override also renamed the notification extension) and now stamps `CMUX_APP_BUNDLE_IDENTIFIER` / `CMUX_HOST_BUNDLE_IDENTIFIER`, which `ios/Config/Release.xcconfig` maps onto each target. It also requires `IOS_APPSTORE_EXTENSION_PROVISIONING_PROFILE_NAME` for a manual App Store export and added an extension-profile step to `install-app-store-provisioning-profile.sh`.

## Changes

**Guard test (`tests/test_ios_appstore_lane_identity.py`)**
- The fake `xcodebuild` derives the archived bundle id from `CMUX_APP_BUNDLE_IDENTIFIER=` (falling back to `PRODUCT_BUNDLE_IDENTIFIER=`), mirroring the xcconfig.
- The beta and App Store lane tests assert the new contract: app id and host id stamped, and no global `PRODUCT_BUNDLE_IDENTIFIER=` override on the archive command (the regression #12935 fixed).
- `_base_env` supplies the extension profile name; the App Store test checks the export options map `com.cmux.app.NotificationService` to it.
- The installer tests supply the extension profile secret the workflows pass (`IOS_APPSTORE_EXTENSION_PROVISIONING_PROFILE_BASE64`) and assert the resolved extension profile name is exported.

**Installer (`.github/scripts/install-app-store-provisioning-profile.sh`)** — one real bug surfaced by the tests: `resolve_expected_cert_fingerprint` pipes `security find-certificate` through `openssl x509` inside a command substitution, so under `set -eo pipefail` an unreadable certificate aborts the whole installer silently, and when only the `x509` step fails it records the SHA-256 of empty input as the expected fingerprint. It now resolves the fingerprint in two steps and leaves it empty on failure, which `validate_extension_profile` already treats as "skip the certificate check". No behavior change when the certificate is readable.

## Verification

- On `main` (`9c2ba78be4`), `python3 tests/test_ios_appstore_lane_identity.py` fails as above.
- With the first commit, CI's guard step gets past both lanes and fails on the two installer tests; with the second, both installer tests pass locally (run in isolation). The full suite is what this PR's `workflow-guard-tests` job runs.


lawrencecchen:
Fleet instruction update for head `5e06e43b1eefe44acb84d375717c35e3ca2bced3`: this PR is classified `other`. No macOS build tag is claimed. The current controller app recipe does not establish iOS/test readiness; that requires the appropriate validated recipe. Use `cmux-ci` for supported jobs, retain the returned ID and receipt, and wait on the same ID after any timeout. Do not use retired maclease allocation or post credentials. Exact-head tags will be posted only after the applicable build succeeds.

PR #13002, opened 2026-09-19
Cloud workspace rename: sidebar row, title bar and Cloud tree always show the accepted daemon name

### LINKED ISSUE #12986: Cloud workspace rename: left sidebar and top bar keep the old name while the Cloud tree shows the new one (three surfaces must always agree)
## Problem

Renaming a cmux Cloud workspace does not update every place the workspace name is shown. In the attached session the right-side Cloud tree shows the renamed workspace (under machine `austin` → Workspaces the row reads `||`, with its tab `Queue issues and cre…` beneath it), while the left workspace sidebar row for that same workspace still reads **Cloud VM** and the window top bar / workspace title above the panes also still reads **Cloud VM**.

So the three surfaces disagree:

| Surface | Shows |
|---|---|
| Right Cloud tree (machine → Workspaces) | new name (`\|\|`) |
| Left workspace sidebar row | old name (`Cloud VM`) |
| Top bar / workspace title | old name (`Cloud VM`) |

The left sidebar and top bar update only sometimes; the right Cloud tree updates reliably. The user's words:

> rename the workspace doesn't update properly in the left sidebar and the top bar sometimes but does it for the right sidebar.. why are they all different, they must agree at all times

This is the same class of drift as #12677 (left sidebar `Cloud VM` vs Cloud tree `workspace-1`), #12499 (cloud rename authority) and #12458 (sidebar/layout name drift), all closed. The rename path still lets the surfaces diverge, so those fixes did not make one authority feed every view.

## Reproduction

1. Open a Cloud VM workspace in the macOS app (nightly / current main) with the Cloud right sidebar visible.
2. Rename the workspace (Cmd+R / Cmd+Shift+R on the workspace, or rename from the sidebar row, or `cmux rename-workspace`).
3. Compare the workspace name in three places: the left workspace sidebar row, the top bar / workspace title above the panes, and the workspace row under the machine in the right Cloud tree.
4. Repeat a few times, including renaming again, switching workspaces and back, and letting a Cloud snapshot/refresh arrive.

### Actual

The right Cloud tree shows the new name while the left sidebar row and/or the top bar keep the old name (`Cloud VM`). It is intermittent: sometimes the left sidebar and top bar update, sometimes they do not.

### Expected

- A cloud workspace has exactly one authoritative name. The left sidebar row, the top bar / workspace title, and the right Cloud tree row all render that one value and change together, immediately, on every rename path (keyboard, sidebar row, Cloud tree, CLI/socket).
- No surface may cache its own copy of the title that can go stale; every view derives from the same model property (or the same observable projection of the remote state), and every rename path writes to that same authority.
- The three views agree at all times: right after the rename, after switching workspaces and back, after a Cloud snapshot/refresh/reconnect, and after app restore.
- If a rename is rejected by the remote side, all three surfaces revert together with the same name; no surface may show a name the others do not.
- Regression test that renames a cloud workspace through each entry point and asserts the sidebar row title, the workspace/top-bar title, and the Cloud tree node title are identical (before and after a simulated remote refresh), plus an audit of every other place a cloud workspace name is displayed (window title, tab strip, notifications, workspace picker) so they read from the same authority.

## Environment

- cmux macOS app (development build from main, 2026-09-18), Cloud Machines enabled.
- Machine `austin` (starred) with 10 issue workspaces plus the renamed one; a second machine `keen-olive-puffin` also present.

austinywang:
Closes #12986

## Root cause

A Cloud workspace's name is painted in three places that read two different things:

| Surface | Reads |
| --- | --- |
| Left workspace sidebar row (`SidebarWorkspaceSnapshotFactory`) | local `Workspace.title` |
| Title bar / `NSWindow.title` / toolbar label (`TabManager.resolvedWorkspaceDisplayTitle`) | local `Workspace.title` |
| Cloud tree workspace row (`CloudTreeNode` → `SurfaceRemoteWorkspace.name`) | the accepted daemon graph (`CloudVMState`) |

By design (#12504) cloud names are never speculative: local rename paths only submit the name to the daemon, and the local title is written when the accepted graph comes back, through `CloudWorkspaceRenameService.reconcileRemoteWorkspaceName` with `.remote` provenance. That single write is what keeps surfaces 1+2 equal to surface 3.

83bfdf9834 ("Preserve user workspace titles during binding") widened the guard in that function from *"an equal confirmation keeps user provenance"* to *"any user-owned local title returns before the graph name is applied"*. So once a Cloud workspace ever carried a `.user` title — a rename typed while `cmux vm new` was still discovering the remote identity, `cmux vm new --title`, a manifest written by a build without remote provenance (the placeholder `Cloud VM` decodes as user-owned), or a rollback after a failed write — the accepted daemon name never reached the local title again. Every later rename (Cloud tree "Rename…", another client, and even the sidebar / Cmd+Shift+R / `cmux rename-workspace` paths, because they wait for the graph too) updated the Cloud tree and left the sidebar row and the title bar on the old name. That is the "sometimes" in the report: workspaces without user provenance converged, workspaces with it were permanently stale.

The guard was meant to protect the creation race (a name typed before the remote identity binds must not be clobbered by an older snapshot). That window is already protected by the unacknowledged intent: `bindCloudWorkspace` submits the typed name before reconciling, and `pendingCloudRenameName` covers the queued/in-flight write and the daemon receipt until the graph includes it. Provenance was never the right key.

## Fix

`Sources/Surfaces/CloudWorkspaceRenameService+Reconciliation.swift`: the accepted graph is the one authority for a cloud workspace's name and the local title is its projection. Only an unacknowledged local intent outranks a snapshot; an equal confirmation still keeps user provenance; any differing graph name is applied as the remote title. No surface keeps its own copy: the sidebar row, title bar, window title, toolbar label, notifications, palette and session restore all read the one local projection, and the Cloud tree reads the graph the projection is written from, in the same `publish` transaction.

The 1/30 s title-bar coalescer (`NotificationBurstCoalescer`) was checked and left alone: it runs the latest action and reads live state at flush time, so it cannot drop the final value.

## Tests (two commits: red, then green)

`cmuxTests/CloudWorkspaceRenameSurfaceParityTests.swift` (wired into `cmux.xcodeproj`, `lint-pbxproj-test-wiring.sh` green) reads all three surfaces — `SidebarWorkspaceSnapshotFactory.makeSnapshot().title`, `TabManager.resolvedWorkspaceDisplayTitle(for:)`, and the Cloud tree `.workspace` node's `name` — through the existing `CloudNameAuthorityFixture` / in-memory daemon, and asserts they are identical:

- after a Cloud tree rename (`catalog.renameRemoteWorkspace`, the exact call behind "Rename…") of a workspace that previously had a `.user` title, then after a remote refresh, then after renaming again from the tree and locally;
- after each local rename path on a user-titled workspace: `TabManager.setCustomTitle` (inline rename, Cmd+Shift+R, palette, context menu) and the `workspace.rename` socket command (`cmux rename-workspace` / `cmux workspace rename`);
- after a rename from another client (new accepted graph, no local intent), through a workspace switch and a refresh;
- after a session restore from a current manifest, a manifest without the remote marker, and a pre-provenance manifest, while the daemon renamed the workspace in the meantime;
- a rejected rename (local and Cloud tree) leaves all three on the accepted name;
- a creation-time rename outranks an older snapshot only while its intent is pending, and lands on all three once acknowledged.

The existing `CloudInitialWorkspaceNamingTests` creation-race test now pins the pending intent (and an older snapshot arriving inside that window) instead of the provenance guard. Because the sharded app-host suite tolerates Swift Testing failures, the new suite is also added to the non-tolerant focused lane in `ci.yml` (shard 6), like the other regressions that must stay red-capable.

Test evidence (filled in below as runs complete):

- test-only commit d1632a4d80, expected red on the new suite: https://github.com/manaflow-ai/cmux/actions/runs/35420608021
- fix commit 3c05b14484, expected green with the suite executed: https://github.com/manaflow-ai/cmux/actions/runs/35420610640

## Rename paths and surfaces verified in the tagged app

Tagged dev build `12986-cloud-rename-sync`, built on the self-hosted Tart runner from the fix commit 3c05b14484 (sources identical; the build script only strips the two `.env.example` files): https://github.com/manaflow-ai/cmux/actions/runs/35420563241. The PR head 1412afed8f is that commit plus the maintainer's merge of main (#12903 only); every file this PR touches is byte-identical between the two.

Dogfood checklist against a real Cloud VM (pending; each step compares the left sidebar row, the title bar and the Cloud tree row):

- [ ] sidebar inline rename
- [ ] Cmd+Shift+R
- [ ] `cmux rename-workspace`
- [ ] Cloud tree "Rename…"
- [ ] after switching workspaces
- [ ] after a refresh

## Localization

No user-facing strings added or changed (code, comments, tests and a CI step only); nothing to localize. Audit: `rg` over the changed Swift files for new bare English in UI calls found none.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/13002?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792377265&installation_model_id=21920&pr_number=13002&ref=codesmith_pr_footer&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F13002&signature=6b0db013259eddfe766303847be13655caf051bef8a0e7f8710bd7a39ae38aa3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes #12986 by making the accepted daemon name the single authority for a Cloud workspace’s display name, so the sidebar row, title bar, and Cloud tree converge after every rename instead of leaving local surfaces stuck on an old user-owned title.

**Bug Fixes**
- `reconcileRemoteWorkspaceName` no longer lets a stale user-owned title block the accepted daemon name.
- Only an unacknowledged local rename can outrank an older snapshot; equal confirmations keep user provenance.
- Rejected renames leave every surface on the accepted name.

**Tests**
- Adds `CloudWorkspaceRenameSurfaceParityTests` covering all rename paths, refresh, workspace switch, session restore, and the creation race.
- The existing creation-race test now pins the pending rename intent instead of relying on provenance.
- The suite runs in the non-tolerant CI lane so Swift Testing failures can't be masked.

<sup>Written for commit f04b50348847ca42517fa95a9c5747f1c9b180b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/13002?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Cloud workspace names now remain consistent across the sidebar, title bar, and Cloud workspace list after renames, refreshes, workspace switches, and session restoration.
  * Accepted remote workspace names now correctly replace conflicting local titles.
  * Rename requests rejected by the service no longer leave inconsistent names displayed.
  * Titles chosen during workspace creation remain visible until the rename is acknowledged, then synchronize with the accepted cloud name.

* **Tests**
  * Added regression coverage for workspace rename consistency across supported rename flows.


lawrencecchen:
Mac fleet instructions for head `f04b50348847ca42517fa95a9c5747f1c9b180b3`. Planned tag: `pr-13002-f04b5034`; this is not yet a published build.

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-13002-f04b5034 /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git f04b50348847ca42517fa95a9c5747f1c9b180b3' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/13002 --source-digest f04b50348847ca42517fa95a9c5747f1c9b180b3 --cache-key cmux:pr-13002 --min-free-bytes 268435456000 --label cmux --label ram48)
```

PR #12929, opened 2026-09-18
Fix non-glass overlay hosting that disrupts Minimal Mode chrome

austinywang:
Fixes #12917.

The non-glass overlay wrapper introduced in 0.64.25 replaced `window.contentView` and changed SwiftUI's sizing contract, allowing overlays to disrupt Minimal Mode sidebar and titlebar chrome. Generic overlays now stay beside the existing content view in `NSThemeFrame`. Browser portals use a SwiftUI-owned AppKit anchor inside the content hierarchy, preserving the macOS 27 native-hover hosting requirement without wrapping the SwiftUI root.

The resolver caches the browser anchor per window. Mounting that anchor notifies existing browser portals to reconcile their installation target. Pointer routing converts between AppKit's superview coordinates and host-local coordinates once. No setting, shortcut, Ghostty option, timer, or user-facing string is added.

Review findings are addressed in the current head, and all three actionable threads are resolved:

- [Explicit AppKit host deinitializer](https://github.com/manaflow-ai/cmux/pull/12929#discussion_r4044420898): added in `f62d74dd2a`.
- [Shared fallback and cached resolver](https://github.com/manaflow-ai/cmux/pull/12929#discussion_r4044420913): implemented in `f62d74dd2a` and `b65f998713`.
- [Reconcile portals when the SwiftUI host mounts](https://github.com/manaflow-ai/cmux/pull/12929#discussion_r4044652949): mount notification and registry observer implemented in `45b352e868` through `8f6b32baa6`.

Validation for pushed head `33ad23e685c88fa661e2df8f8e78dad65fdbb30b`:

- **Dev build available:** [pr-12929-33ad23e6](http://127.0.0.1:17320/pr-12929-33ad23e6). Fleet job `ae095aac31318513be1c6804` succeeded, published the exact artifact, and confirmed workspace cleanup. The restored app's bundle ID, embedded commit, and recursive code signature were verified; the app was launched. Austin has confirmed that the app works.
- **Focused overlay tests passed:** [WindowOverlayChromeTests](https://github.com/manaflow-ai/cmux/actions/runs/35417005240) executed all three tests on this SHA, including glass/non-glass geometry, content ownership, resizing, and flipped-parent pointer routing. The regression tests were committed before the fix.
[...]

Issue #12917, opened 2026-09-18
0.64.25: workspace/pane tab bar disappears after overlay wrap (macOS 27, Minimal Mode, no window glass)


jaeyongjaykim:
## Summary

After updating from **0.64.24 → 0.64.25**, the tab bar disappears. Session state still has the workspaces and the left sidebar marked visible, so this looks like a chrome/layering regression rather than a collapsed-sidebar or lost-session bug.

This started on the 0.64.25 Sparkle update. The published changelog for 0.64.25 does not mention tab chrome.

## Environment

- **cmux:** 0.64.25 (106) `b685a275c` (was 0.64.24)
- **OS:** macOS 27.0 (build 26A428), Apple Silicon
- **Install:** Homebrew cask / Sparkle stable channel
- **Window:** fills the display (`1800×1129` at origin `0,0`; native-fullscreen sized)
- **Presentation:** Minimal Mode (`workspacePresentationMode = minimal`)
- **Window glass:** off
  - `sidebarBlendMode = withinWindow`
  - `bgGlassEnabled` unset / false
- **Sidebar experiment:** `cmux.flags.remote.sidebar-appkit-list-experiment = 1`
- **Session:** left sidebar `isVisible: true`, width `240`, selection `tabs`; 6 workspaces still present; each workspace has a single pane surface (`isFullWidthTabMode = false`)

## Regression window

- **Last good:** 0.64.24
- **Broken:** 0.64.25
- Likely commit: https://github.com/manaflow-ai/cmux/pull/12683 (`c11cad52`, “Investigate macOS 27 native browser hover”), merged the same day as the 0.64.25 tag.

`main` still has this overlay-hosting change. I have not seen a 0.64.26 chrome fix.

## What I think happened

PR #12683 changed overlay installation when window glass is **not** installed.

**0.64.24** (`WindowContentOverlayTargetResolver`):

```swift
// overlays sit in NSThemeFrame, beside window.contentView
return WindowContentOverlayInstallationTarget(container: themeFrame, reference: contentView)
```

**0.64.25:**

```swift
// wrap the whole content view so overlays live *inside* it
root = WindowContentOverlayRootView(contentView: contentView)
root.install(in: window) // window.contentView = wrapper
```

[...]

Issue #6506, opened 2026-06-20
Browser panes stop loading after long uptime — all panes fail at once (shared WebKit networking process wedges)


austinywang:
## Summary

After cmux runs for a long time, **all browser panes stop loading pages at once** — they spin or show nothing. Chrome and the macOS shell (`curl`, DNS) keep working fine at the same moment. Restarting cmux fixes it, and on restart every pane reloads simultaneously.

## Symptoms

- Pages refuse to load in cmux browser panes after long uptime ("sometimes").
- Failure hits **all panes at once**, not a single site.
- External Chrome and `curl`/DNS from the shell are unaffected during the same window.
- A cmux restart clears it; all panes re-fetch together right after launch.

## Diagnosis

Browser panes are WKWebView (Apple WebKit). Each open pane keeps a live `com.apple.WebKit.WebContent` process, and they all funnel through a shared per-app `com.apple.WebKit.Networking` process. With 100+ workspaces each holding a browser pane, there are **170+ WebContent processes** alive at once, all behind essentially one networking process. When that shared networking process wedges (stalled/leaked connections, an fd/connection ceiling, or a network transition over long uptime), **every pane loses loading simultaneously** — while Chrome and the CLI keep working because they have independent network stacks.

This was **not** memory pressure when observed: 61% RAM free, 0 swap, no jetsam kills.

Likely trigger: a network transition during long uptime. This machine resolves DNS through Tailscale MagicDNS (`100.100.100.100`); WKWebView's networking process is prone to holding stale connections across interface/route changes.

## Evidence (captured live, right after a restart)

- cmux had just been restarted. First `ps` snapshot caught the app at **48% CPU / running**; seconds later **0% CPU / idle** — that spike was every pane re-navigating at once after launch (the "all browsers loading on restart" the reporter saw).
- WebKit footprint at the time: **171 `WebContent` + 8 `Networking` processes**, ~5.5 GB RSS across WebKit processes.
- No CFNetwork/WebKit error lines surfaced in the unified log for the pre-restart window — the restart wiped the live state — so this is diagnosed from the architecture plus the reload spike rather than a single logged error.

## Proposed fix

- **Suspend/unload browser panes that aren't visible**, reload on focus, so 100+ panes don't pin live WebContent processes and share one networking process indefinitely. (Same architectural root as #4188 — retained WebKit growth.)
[...]

PR #13003, opened 2026-09-19
Fix browser refresh losing failed navigation requests

### LINKED ISSUE #12998: 0.64.25: in-app browser refresh is unreliable after update
## Problem

A Discord report from salex.oops (2026-09-17) says that after updating to cmux 0.64.25, page refresh in the in-app browser became unreliable and is blocking work.

Verbatim report:

> Latest cmux broke two things I rely on: the top window buttons disappeared (I can’t find any setting to restore them), and page refresh in the in-app browser is unreliable. This is blocking my work. Please help me roll back to the previous version and pin it so auto-update doesn’t override it.

The missing top controls are tracked separately in #12917. This issue isolates the browser behavior so it can be reproduced and fixed independently.

## Known context

- Last known good release for the paired report: 0.64.24
- First reported broken release: 0.64.25
- 0.64.25 merged #12683, which changed non-glass browser overlay hosting and portal hierarchy.
- Existing related browser lifecycle reports: #1132, #4387, #6506, #8478, #9561, and #6253.
- The 0.64.24 → 0.64.25 diff does not change the basic BrowserPanel reload handlers, so the failure may be in portal hosting, focus/input routing, WebKit lifecycle, or cache/navigation state.

## Investigation needed

Capture the exact failure mode from the reporter or a focused repro:

1. Does clicking the reload button do nothing, reload stale content, or reload and then revert?
2. Does the same happen with View → Reload, Cmd+R, Cmd+Shift+R, and the reload button context menu's Hard Refresh?
3. Does it occur after workspace switching, pane/tab switching, fullscreen changes, or a WebKit load failure?
4. Does navigating to about:blank and back recover the surface?
5. Record cmux version/build, macOS version, install channel, URL type (ordinary web, localhost, or Cloud), and whether Minimal Mode is enabled.

## Expected

Reloading the active in-app browser surface should initiate a real navigation and display the current page reliably. A cache-bypassing hard refresh should remain available for stale local/dev content.

## Acceptance criteria

- Identify the concrete trigger and distinguish button, keyboard, portal lifecycle, WebKit networking, and cache failures.
- Add a regression test for the reproduced failure.
- Fix the root cause with the smallest scoped change.
- Preserve existing soft reload and hard refresh semantics.
- Document any remaining WebKit or environment limitation if the report cannot be reproduced.

austinywang:
A manual refresh after a provisional navigation failure can drop the failed request's method, headers, and body, or reload the previous document before the generated error page commits. Browser reload actions now share the existing failed-request replay policy with the CLI. Replayable requests are preserved, and an unreplayable upload is never silently changed into a GET.

Closes #12998

### Evidence and scope

Reproduced on macOS 26.4.1 (25E253), NIGHTLY `0.64.25-nightly.3540722443201` / `9c2ba78` (the branch's baseline): a loopback fixture received `POST /submit` with `payload=preserve-me`, closed the connection to cause a provisional failure, and then received `GET /submit` when the toolbar Reload button was clicked. The omnibar still named `/submit`; the generated error document was `about:blank`.

This establishes a refresh failure in scope, **not the exact Discord symptom or a newly introduced 0.64.25 defect**. The v0.64.24 → 0.64.25 diff leaves the reload handlers unchanged. #12683 changes non-glass content/overlay ownership and can affect presentation independently of navigation; the separate hosting/chrome fix remains in #12929 / #12917. This PR makes no portal or titlebar change.

| Investigation | Result |
| --- | --- |
| Reload button, Cmd+R, View → Reload Page | Ordinary page reload worked on the running baseline. The manual failure path ignored request metadata that CLI reload already used. |
| Cmd+Shift+R and context-menu Hard Refresh | Both issued real network requests on the baseline. Healthy pages retain `reloadFromOrigin()`; failed-request hard refresh uses the existing `.reloadIgnoringLocalCacheData` policy. No cache or cookie clearing. |
| Workspace and pane-tab switching | An edited input survived each round-trip, with no extra fixture request. Regression coverage checks portal hide/anchor replacement/rebind retains the WebView and the next explicit reload reaches that instance. |
| Provisional failure | Ordinary connection-refused GET recovered after the server started. Failed POST refresh changed to GET; tests cover both committed error-page and still-exposed previous-URL states. |
| Discarded/replaced WebViews and long uptime | Existing explicit discard/termination recovery and callback generation fencing are retained. No evidence established a new failure here; no automatic replacement/reload or networking-process reset is introduced. The multi-hour #6506 wedge is not reproduced. |

The existing replay limit is deliberate: if WebKit omits a POST body, or the request uses a stream/oversized body, refresh stays on the error page instead of inventing a GET. The user must return to the form and submit again. Retained requests replay intact, while ordinary soft refresh preserves the request's cache policy and normal document/history behavior.

### Validation

- Test-only commit `66e3fad233` precedes fix commit `7243f877fd`.
- Hosted `BrowserFailedNavigationReloadTests`: final pass [35424101802](https://github.com/manaflow-ai/cmux/actions/runs/35424101802) ran the final SHA and passed all 18 generated focused cases on macOS 15.7.9. Earlier run [35422946185](https://github.com/manaflow-ai/cmux/actions/runs/35422946185) was a transient hosted build-phase failure before execution; the corrected rerun is the authoritative result.
- Existing `BrowserReliabilityRegressionUITests`: app built, but [all 5 tests failed](https://github.com/manaflow-ai/cmux/actions/runs/35417733601) at the shared fixture’s `workspace.create` socket call, before browser navigation. This remains an explicit UI automation verification limit.
- `python3 scripts/swift_file_length_budget.py`, project normalization, `./scripts/check-pbxproj.sh`, `./scripts/lint-pbxproj-test-wiring.sh`, and `git diff --check` pass. No budget TSV edits.
- Localization audit: no new product strings; `python3 scripts/localization_catalog.py check` passes (6 catalogs, 9 locales, zero parity errors).
- Final tagged Blacksmith build: failed closed after its 1200-second queue timeout; no local app compile, retired fleet lease, or controller fallback.

Remaining verification limits: macOS 27, the actual Sparkle update, the Discord reporter's unknown URL/request, and a long-uptime networking wedge. The final dev app is retained for Austin's dogfood; no merge is requested or performed by this change.

Final tagged build: requested from HQ `reload-cloud.sh --builder blacksmith --direct-backend` at final SHA `f710eff744b3d1573afc4b48a13e3742eedb1625`. Blacksmith run `35420022856` remained queued for the full 1200-second fail-closed window and was cancelled; the script refused both fleet and local fallback because the mandated manifest paths were `/nonexistent/hosts.json` and `RELOAD_CLOUD_FALLBACK_LOCAL=0`. No tagged artifact or opener URL was produced.


lawrencecchen:
Mac fleet instructions for head `f710eff744b3d1573afc4b48a13e3742eedb1625`. Planned tag: `pr-13003-f710eff7`; this is not yet a published build.

```bash
JOB_JSON=$(~/.local/bin/cmux-ci submit --kind cmux --command 'CMUX_FLEET_BUILD_TAG=pr-13003-f710eff7 /Users/Shared/cmux-build-fleet/recipes/cmux.sh https://github.com/manaflow-ai/cmux.git f710eff744b3d1573afc4b48a13e3742eedb1625' --artifact artifacts/cmux.app.zip --workspace https://github.com/manaflow-ai/cmux/pull/13003 --source-digest f710eff744b3d1573afc4b48a13e3742eedb1625 --cache-key cmux:pr-13003 --min-free-bytes 268435456000 --label cmux --label ram48)
```

PR #13031, opened 2026-09-19
Portal: drop the settled-commit mark when a hosted terminal view is hidden

### LINKED ISSUE #13027: Side browser garbles terminal bottom lines and prompt
## Confirmed customer report

Muhammad Abdur Rahman AdnaN reports that when the side browser is opened, the bottom lines of the terminal and the prompt writing section get messed up. Screenshots attached (two).

## Evidence

```
Mine bottom lines and prompt writing section gets messed when side browser is opened.
```

- Reporter: Muhammad Abdur Rahman AdnaN

Source report: Discord Manaflow #general, 2026-09-18

## Expected

Opening the side browser should not corrupt terminal rendering. The prompt and bottom lines should stay readable while the side browser is open.

aliyansajid:
Refs #13027. **This does not fix #13027**, and the code path it touches is not
even exercised by opening a side browser (evidence below). It fixes a separate,
real defect in the portal's settlement budget, found while investigating, with
its own regression test.

## What this fixes

`WindowTerminalPortal` marks a pane entry `needsSettledCommit` when its frame
changes, and `markNeedsSettledCommit` refills the four-pass convergence budget
only when the entry was not already pending — that is what makes a "new
settlement episode".

`commitSettledPaneGeometries` skips any entry whose hosted view is hidden or
detached, and does not clear the mark when it skips, so a mark outlives the
frame it described:

1. The portal hides a hosted view (transient or tiny frame during layout churn).
2. The entry stays `visibleInUI && needsSettledCommit`, so
   `finishGeometrySynchronization` keeps spending retries on passes that can
   never commit, down to zero.
3. On reveal, the stale mark makes `markNeedsSettledCommit` treat it as the same
   episode, so the budget is not refilled.
4. With no retries left nothing is scheduled, and the revealed frame is
   published only if that one pass happens to find the hierarchy already at rest.

Hiding a portal-hosted view now drops the settled commit it was owed, at every
site where the portal hides one. Every hidden-to-shown transition marks the
entry again, so nothing is lost, and hidden entries stop draining the budget.

Two commits per the repo's regression policy: the test alone first (red), then
the fix.

## Dogfood evidence, and why this is not the #13027 fix

I built a tagged dev app and drove it through the debug CLI.

**The resize path is healthy.** One settled commit per split, correct grid, rows
unchanged, nothing spurious:

```
surface.size.apply caller=commitPaneGeometry grid=153x50->76x50  (browser opened)
surface.size.apply caller=commitPaneGeometry grid=76x50->153x50  (browser closed)
```

**This PR's code path never runs during a browser split.** In the whole session's
debug log:

```
portal.hidden ... value=1  (hide)    : 0 occurrences
portal.hidden ... value=0  (reveal)  : 0 occurrences
portal.hidden.deferKeep              : 15 occurrences
```

Every transient-geometry event took the `deferKeep` branch, which deliberately
keeps the view visible. The hide this PR modifies never fired, so the change
cannot affect the side-browser scenario either way.

**I could not reproduce the reported garbling.** With a live TUI that redraws on
`SIGWINCH` and anchors a prompt box at the bottom, opening the side browser
produces a clean repaint at the new size — no stale rows, no duplicated
fragments, no corrupted prompt box.

### A measurement trap for anyone else investigating

`cmux read-screen` reports **logical** lines, not physical rows. At 76 columns a
freshly printed 150-character line is still reported as one 150-character line.
So "the grid still holds rows wider than the terminal" cannot be established with
`read-screen | awk '{print length($0)}'` — it reports the same width whether or
not the content was reflowed. I initially drew the wrong conclusion from exactly
that measurement. Screenshots or physical-row instrumentation are required.

I am leaving this PR open because the settlement-budget defect is real and
tested, but it must not be merged as a fix for #13027, and that issue should
stay open.


aliyansajid:
Closing this. I opened it based on reading the portal code rather than on a
reproduction, and once I actually ran a tagged build it turned out the code path
I changed never even fires when you open a side browser (0 `portal.hidden`
events in the whole session, all transient geometry took the `deferKeep`
branch). So this was never going to fix #13027.

Leaving the branch up in case the settlement budget thing is worth picking up
separately, but it shouldn't sit in the review queue looking like a fix for
something it isn't.

Issue #13019, opened 2026-09-19
Team picker: account popover to switch teams, create a team, and open Settings (teams share cloud workspaces)


austinywang:
## Problem

The team picker in the macOS app is not good enough to use teams at all: there is no way to switch between the teams the signed-in user belongs to, and no way to create a team from the app. Teams exist so that people in the same team share cloud workspaces and machines with each other; without switching and creating teams that cannot happen.

The user's words:

> is there an issue working on team picker.. currently our picker right now is not good. we don't have the ability to switch or create teams? we should be able to enable this.. and the idea is that people in teams can share cloud workspaces with each other ideally.... that is the whole point... so let's get to work on this and make sure that this is working

## Desired UI

Reference: the Claude Code account menu (screenshot in the report). A popover anchored at the sidebar footer account row (avatar initials + display name, with the current team as the second line), containing:

- Header: avatar, display name, current team name (and plan if we have it).
- Divider.
- The user's teams, one row each, with the active team marked; selecting a row switches the active team.
- A `Create team…` row that asks for a name, creates the team, and switches to it.
- Divider.
- `Settings` with its `⌘,` shortcut hint (opens the app settings; the account/team settings tab if one exists).

Omit the other rows from the reference (Usage / weekly limit, Show pet, Invite a friend, Log out): the popover is the team switcher plus settings, "just have the different teams in some sort of UI that looks like this, ideally also put settings in here too". Sign-out can stay where it is today. The footer row itself shows avatar + name + current team so the active team is always visible.

## Expected behavior

- Switching the team re-scopes everything cloud-related to that team: the Cloud sidebar (machines, workspaces, terminals), machine creation, `cmux vm`/cloud CLI calls made through the app, and any team-bound credentials. Open cloud workspaces from the previous team are closed or clearly marked; nothing from another team leaks into the current view.
- Every member of a team sees the same cloud machines and workspaces for that team; a machine or workspace created by one member is visible to and usable by the others. If today's data model scopes cloud resources per user, this issue includes moving them to team scope (with membership checks on every read and mutation in the backend, not only in the client).
- Creating a team creates it in the backend (Stack Auth teams) with the creator as a member and makes it the active team; the picker refreshes without a restart.
- The active team is remembered across launches and is consistent between the app, the CLI, and the web dashboard (same selected-team source of truth; no separate per-surface copies that can drift).
[...]

PR #13032, opened 2026-09-19
Team picker: switch teams, create a team, and open Settings from the account popover

aliyansajid:
Part of #13019.

## Against the issue's checklist

**Covered here**

- Popover header: avatar, display name, current team, and the plan badge when
  the account is Pro.
- One row per team in backend order, checkmark on the active one. Selecting a
  row switches; selecting the already-active row dismisses without re-scoping.
- `Create Team…` opens an inline form in the same popover. Creates in the
  backend with the signed-in user as creator, makes it active, and the picker
  reflects it without a restart. Failures are shown in place.
- `Settings` row, opening Settings at the Account tab.
- Opening the popover refreshes the team list, so teams gained or lost
  elsewhere are right when you look at it.
- Active team persists across launches and never points at a team you are not
  in: an unknown selection falls back to a team you belong to, on every refresh
  rather than only at sign-in.
- Configurable keyboard shortcut (`toggleAccountTeamMenu`), so it is editable in
  Settings and settable from `~/.config/cmux/cmux.json`.
- Localized across all nine supported locales.
- The four behavior tests the issue asks for.

**Not covered, deliberately**

- **Team-scoping cloud machines and workspaces in the backend.** This is the
  half that makes members actually share, and it depends on how those resources
  are scoped today (the question on the issue). Nothing here changes where they
  live.
- **The footer row** showing avatar + name + team. cmux's sidebar footer is a
  22pt icon strip shared with Mobile Connect, Help, Upgrade and Extensions;
  making the account control a full-width labelled row redesigns that strip
  rather than changing one control. Left for a separate decision; the active
  team is in the popover header instead.
- **Verifying one source of truth across app, CLI and web dashboard.** The app
  side persists through `CMUXAuthTeamSelectionStore`; I have not verified the
  CLI and dashboard read the same selection.

## What switching actually re-scopes

Every Cloud client — `VMClient`, `RemotesClient`, `CoderouterClient`,
`AIAccountsClient`, `DeviceRegistryClient`, `PresenceHeartbeatClient` — reads
`auth.resolvedTeamID` at call time, so the next cloud request after a switch
lands in the new team. `AuthenticatedTeamScope` advances its generation on the
switch, so work captured under the previous team fails its
`isAuthenticatedTeamScopeCurrent` check and cannot install results into the new
team's view. Both are now pinned by tests.

## Shortcut

`toggleAccountTeamMenu` ships **unbound**, following `.sendFeedback`. Every
nearby combination is already taken, and silently claiming one from a user is
worse than shipping it configurable. It is editable in Settings and in
`cmux.json`. Signed out it routes to sign-in rather than opening an empty
picker, and it dispatches through one notification so the shortcut and the
button share a single open path instead of each keeping its own copy of the
popover state.

The `Settings` row's hint shows the user's real `openSettings` binding rather
than a hardcoded `⌘,`, since that action is rebindable and a hint that
disagrees with the keyboard is worse than none.

## Tests

`AuthCoordinatorTeamMembershipTests` (12): create makes the team active and
persists it, names are trimmed, a blank name never reaches the backend, a
signed-out coordinator cannot create, a created team survives a failed
follow-up list read, refresh picks up a team added elsewhere, refresh moves off
a team you were removed from, a failed refresh keeps known teams, a sign-out
racing a create leaves nothing behind, switching changes what cloud calls are
scoped to, switching fences work captured for the previous team, and a team you
are not a member of never becomes the cloud scope.

`SidebarAccountTeamMenuTests` (10): exactly one row marked active, backend order
preserved, an account with no teams still offers create, an active id outside
the list marks and names nothing, blank-name validation, and the create form's
state transitions.

Full `CmuxAuthRuntime` suite green (254). macOS app builds clean.

## Not yet dogfooded

Verified by build and tests only. The popover layout and the create flow have
not been exercised in a running app.

## Summary by CodeRabbit

* **New Features**
  * Added team creation for signed-in users with name validation and localized error messages.
  * Added team switching and refreshing from the account menu.
  * Newly created teams become active automatically.
  * The account popover now displays the active team and includes a “Create Team…” form.
  * Added an optional configurable shortcut for opening the account/team menu.
  * Added localized support across supported languages.

* **Bug Fixes**
  * Preserves known teams when refreshing fails.
  * Handles stale selections and sign-out during team creation safely.

* **Tests**
  * Added coverage for team creation, selection, validation, refresh behavior, and localized failures.


aliyansajid:
Superseded by #13051, which shipped this and the backend/CLI half I'd flagged as out of scope here. Closing.

Thanks for the pointer on server-side team selection — reading `selectedTeamID()` back from the backend is the piece this was missing.

PR #13045, opened 2026-09-19
feat: add native discovery for configured actions

teamleaderleo:
## Reviewer summary

Shows configured actions in the existing discovery surfaces, with stable identity and the right global-versus-project scope. The change reuses the current action and settings owners.

## What changed

Adds the smallest native discovery/management surface for the existing `cmux.json` action system.

- adds an **Actions · cmux.json…** entry beside Settings/Open cmux.json in the app menu
- adds the same entry to the new-workspace (+) menu management tail
- shows the live resolved custom action registry, including inherited/project-local actions and overridden built-ins
- shows effective placements using the existing config vocabulary: `palette`, `ui.newWorkspace.action`, `ui.newWorkspace.contextMenu`, `ui.surfaceTabBar.buttons`, and `shortcut`
- exposes the existing in-cmux config edit path and the existing custom-commands docs
- keeps execution on the existing action/config paths; this adds no action registry, workflow format, or editor

The discovery model filters shipped built-ins unless they have a config source path, so the panel focuses on user/project customization while still surfacing built-in overrides.

## Lane K / #79

Implements Lane K from https://github.com/teamleaderleo/Tact/issues/79.

Lane J uses **workflows** / **saved layouts** for reusable layout discovery. This PR keeps that vocabulary intact and scopes its UI to resolved `cmux.json` **actions** and their launcher placements; it does not touch Lane J's workflow catalog/docs files.

## Tests

Focused coverage:
- `CmuxConfigWorkspaceActionTests.actionsAndLaunchersDiscoveryReportsEffectivePlacements` covers agent/workspace kinds, palette/default/+ menu/tab-bar/shortcut placement, and hiding unconfigured shipped built-ins.
- existing new-workspace rendering tests now assert the shared Actions entry in the management tail.
- the added Swift source parses cleanly with `swiftc -frontend -parse`.

## Localization audit

No new prose localization keys were introduced. The new native UI reuses existing localized strings for **Actions**, **Open cmux.json**, **Open Docs**, and **OK**. Placement/type rows intentionally display invariant `cmux.json` identifiers and schema enum values so users can map the discovery view directly back to config.

---
<a href="https://app.blacksmith.sh/manaflow-ai/codesmith/cmux/pr/13045?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792433702&installation_model_id=21920&pr_number=13045&ref=codesmith_pr_footer&repository=manaflow-ai%2Fcmux&return_to=https%3A%2F%2Fgithub.com%2Fmanaflow-ai%2Fcmux%2Fpull%2F13045&signature=5a518943dbe9f5de417543d42c33fa601f9acdb4846778dc3ae7a774071cb029"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds a native discovery panel for `cmux.json` actions. Previously there was no in-app view of resolved custom actions; now an **Actions · cmux.json…** entry in the app menu and new-workspace menu tail shows a read-only list of resolved actions with effective placements and source paths. Built-ins are hidden unless overridden. Execution and edit paths are unchanged, and no new localization keys were introduced.

Surface tab bar placement is reported per configured action and only for visible buttons, so hidden buttons are excluded and identical commands are never conflated in the panel.

Implements Lane K of #79 without touching Lane J's workflow/saved-layout vocabulary.

**Tests**
- Pin exact surface-tab-bar placement identity, hidden-button reference dropping, and the shared Actions entry in new-workspace menu rendering.

<sup>Written for commit fc6a118f7313b3f7c3478f53b524b3226598e76b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/13045?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Sources/AppDelegate+WorkspaceActionSave.swift
```diff
@@ -28,7 +28,257 @@ final class WorkspaceDefaultLayoutBox: NSObject {
...
+    }
+
+    private static func surfaceTabBarButton(
+        _ button: CmuxSurfaceTabBarButton,
+        resolves action: CmuxResolvedConfigAction
+    ) -> Bool {
+        if button.id == action.id {
+            return true
+        }
+        guard button.action == action.action else { return false }
```
teamleaderleo: This fallback can over-report surface-tab-bar placement when two actions in the same source file resolve to the same action payload. A button that explicitly references action A is resolved by `CmuxSurfaceTabBarButton.resolved` into A's payload/source and can keep a custom button id; action B with the same payload/source then also satisfies these two comparisons. Since this panel is reporting *which configured action* is placed, can the discovery input preserve/reference the original action id instead of inferring identity from resolved value equality? A focused duplicate-payload test would pin this.

Sources/AppDelegate+WorkspaceActionSave.swift
```diff
@@ -28,7 +28,257 @@ final class WorkspaceDefaultLayoutBox: NSObject {
...
+        scrollView.documentView = textView
+        return scrollView
+    }
+
+    private func handleActionsAndLaunchersResponse(_ response: NSApplication.ModalResponse) {
+        switch response {
+        case .alertFirstButtonReturn:
+            // Reuse the existing in-cmux config editor path used by Settings'
+            // workspace-layout customization.
+            openWorkspaceLayoutsCustomization()
```
teamleaderleo, line 263: The discovery model includes inherited/project-local actions and shows each action's source path, but this button always calls `openWorkspaceLayoutsCustomization()`, which materializes/opens the global `~/.config/cmux/cmux.json`. For a row sourced from project `.cmux/cmux.json`, “Open cmux.json” therefore opens a different file from the one the user needs to edit. A small fix could make the global-only behavior explicit in the label, or expose/open the active local source when one is present.

teamleaderleo (commented):
I found two bounded edge cases in the new discovery/management surface. Both come from the fact that the panel intentionally includes the resolved project-local registry, while some of the UI still loses source/action identity.

Sources/AppDelegate+WorkspaceActionSave.swift
teamleaderleo: Addressed on the current head. `CmuxConfigStore` now preserves `surfaceTabBarActionReferenceIDs` while resolving buttons, and the discovery model consumes those original action IDs instead of comparing resolved payload/source equality. `actionsDiscoveryDoesNotInferSurfacePlacementFromEqualPayloads` pins the duplicate-payload case.

Sources/AppDelegate+WorkspaceActionSave.swift
teamleaderleo: Addressed on the current head by making the global-only edit target explicit. The primary button now includes the abbreviated `globalConfigPath` in its title (for example `Open ~/.config/cmux/cmux.json`), matching the existing global workspace-layout editor path while each discovered action continues to show its own source path.

teamleaderleo (commented):
Follow-up on the two earlier findings: current head `fc6a118f7313b3f7c3478f53b524b3226598e76b` addresses both cleanly.

- `surfaceTabBarActionReferenceIDs` carries the canonical action reference through resolution and is filtered against the final visible button IDs, so equal resolved payloads no longer conflate action placement; the duplicate-payload test pins it.
- the config action now says exactly which global path it will open, while project-local rows continue to display their own source paths.

I did a focused read of the new identity map and found no additional issue in that follow-up.

teamleaderleo:
Replaced by https://github.com/manaflow-ai/cmux/pull/13232: same commits, head branch moved into the org.

PR #13075, opened 2026-09-19
Fix shell integration handoff test setup

teamleaderleo:
Makes the shell integration handoff tests deterministic and renames the tmux refresh test to state its behavior: workspace-scoped values refresh while surface and panel scope is cleared.

Validation: focused app-host CI workflow dispatched for the affected test class. No local Xcode tests were run.


teamleaderleo:
Continued in https://github.com/manaflow-ai/cmux/pull/13125 (same commits, in-org branch `fix-shell-integration-handoff-tests`) so its macOS jobs run on the repository's configured runners. Push further changes to `fix-shell-integration-handoff-tests` on manaflow-ai/cmux, not to the fork.

Issue #13088, opened 2026-09-19
[RFC] Agent PRs are not done until bot review threads are answered


teamleaderleo:
## Problem

An agent-created PR can reach “done” once the branch is pushed and CI is green, while CodeRabbit, Greptile, Cubic, or another review bot still has actionable threads waiting for a response. cmux already surfaces agent attention and local diff comments, but there is no durable PR-level obligation that says: this agent must answer every bot thread before the work is complete.

The current PR template asks contributors to request bot reviews and resolve bot comments, but that is a checklist rather than an enforced state. The repository guidance also currently treats bot reviews as asynchronous after the first pass.

## Proposal

Define an agent PR’s completion contract as:

> The PR is not complete until every actionable bot review thread has been fixed and answered, or explicitly declined with a rationale, then resolved and rechecked against the latest commit.

Model this as a review loop:

`working → PR opened → bot review pending → findings pending → agent addressing → re-review pending → mergeable`

The normal merge gate should require:

- all configured review bots have completed against the latest PR head;
- no actionable bot thread remains unresolved;
- each actionable thread has an agent reply after the latest bot comment;
- “won’t fix” / false-positive decisions include a rationale;
- pushing a new commit invalidates the previous completion result.

GitHub’s “Require conversation resolution before merging” setting is useful but insufficient on its own: it checks resolution, not whether the agent actually answered. Add a required `agent-pr-review-complete` status check for the stronger invariant.

## cmux integration

cmux already has the beginnings of this pattern:

- persistent diff comments in `DiffCommentStore`;
- `cmux comments list`;
- socket delivery of comments into an agent submission;
- PR metadata and agent-attention notifications in the sidebar.

Extend that path to GitHub PR review threads:

- sidebar badge such as `PR · 3 bot threads`;
- an agent-facing `cmux pr review --json` (or equivalent socket method);
- thread payloads containing bot, file/line, body, replies, commit SHA, and state;
- actions to open the thread, focus the owning workspace, reply, and mark a finding fixed/declined;
- notifications when a new bot review arrives or a pushed commit invalidates prior review state.

The agent loop should fetch new threads, patch or explain them, reply under the original thread with the relevant commit or rationale, resolve only after replying, push, and wait for review on the new head.

## Scope and policy questions

- Which bots are in the allowlist, and how are summary/rate-limit comments distinguished from actionable threads?
- Should an agent be allowed to resolve a declined finding, or should that require human confirmation?
[...]

PR #13128, opened 2026-09-20
ci: reuse compiler state from cancelled PR runs

teamleaderleo:
## Summary

Cancelled PR runs can leave useful Xcode compilation-cache objects behind, but the current CI points the compiler CAS at `$RUNNER_TEMP`, so replacement runs lose that local state. This change keeps a branch-and-PR scoped CAS under the runner tool cache, allowing a superseding run to reuse compatible objects from an earlier cancelled run.

The existing read-only GitHub Actions cache remains the cross-run seed. The local cache is scoped by PR number, branch, and compiler fingerprint so objects do not cross contributor or toolchain boundaries. Successful runs report the retained CAS size in the workflow summary for measurement.

The live queue also showed package-test failures allowing the expensive macOS compile/admission and six app-host shards to start anyway. Those macOS jobs now wait for `swift-package-tests` to pass, preventing predictable fan-out waste.

## Validation

- `actionlint .github/workflows/ci.yml`
- `git diff --check`

Refs #13088

## Summary by CodeRabbit

- **Chores**
  - Improved automated build and test validation by ensuring related checks complete successfully before dependent workflows proceed.
  - Added persistent compilation caching for macOS builds to support more consistent and efficient validation runs.
  - Added reporting for reusable compilation state and cache size in workflow summaries.


teamleaderleo:
Review of the approach, before more work goes into it.

**As written this cannot hit.** Every macOS job in this repository runs on a runner that exists for one job. Tonight's runs show a different machine per job (`warp-6x-arm64-w6xi4lr7…`, `warp-6x-arm64-wv2ennr1…` for two consecutive compile admissions), Blacksmith is the same, and the Tart pool in `docs/ci-runners.md` deletes its clone after each job. `$RUNNER_TOOL_CACHE` starts empty every time, so the replacement run never sees what the cancelled run left behind. The "retained on the self-hosted runner" line in the summary describes a runner we do not have. (`Refs #13088` is the bot-review RFC, which looks like the wrong issue.)

**The version that would work is a pull-request-scoped Actions cache saved with `if: always()`**, and I do not think it pays for itself:
- Swift caches per module. A push that touches `Sources/` recompiles the whole `cmux` module (about 340 s of the 945 s compile) whatever the previous push cached, and the package layer already comes from the `main` seed. The only win is a later push that leaves alone a big module an earlier push touched (test-only or CI-only fix-ups).
- The CAS is a few large mapped files, not one file per object, so only whole multi-GB snapshots can be saved. Roughly five runs per pull request and 33 pull requests a day would push the `main` seed out of the cache budget, which is the reason the restore step here is read-only.

**What I would build instead: reuse the compiled product by input fingerprint.** Compile admission already packages the app-host test product and stamps it. Key it by a hash of what the build reads (`Sources/`, `CLI/`, `Packages/`, `cmuxTests/`, the project file and lockfile, the Ghostty revision, the toolchain fingerprint), publish a few-byte cache entry `fingerprint -> run id + artifact id`, and have compile admission download a matching product instead of compiling. That hits in the cases where a run is superseded for reasons that do not touch Swift at all: bot-fix pushes to tests, Python or YAML, merges of `main` that only moved web or docs, re-runs, and merge-queue entries whose build inputs equal what the pull request already built. It costs bytes of cache, and the product itself is already uploaded.
  - Trust rule: only accept a product whose producing run has `head_repository == manaflow-ai/cmux` (in-org branch, `main`, or merge group). A fork run must never seed a product others consume.
  - Measure first: replay last week's runs and count how many consecutive runs on the same pull request had identical build inputs. If it is under about 15%, skip this too.

Separately, tonight's runner move changed the cache backend: the first nightly on Warp (run 35487314929) found no Release compilation cache even by prefix and built cold in 27.6 min, then saved a fresh one. Every existing seed was written on Blacksmith, so all caches are cold until each writer runs once on Warp.

teamleaderleo:
Closing in favour of https://github.com/manaflow-ai/cmux/pull/13139, which gets the same result on single-job runners: a push that changes nothing the build reads skips compile admission by matching a build-input fingerprint against an earlier successful run of the same branch. The reasons the on-disk CAS cannot hit are in the review above.

PR #13149, opened 2026-09-20
Clear deleted Cloud bindings across recoverable workspace managers

lawrencecchen:
Deleting an already-missing Cloud machine can return success while a workspace retained during main-window context replacement stays bound to it. That stale workspace can continue reconnecting.

Cleanup now uses the canonical live-manager traversal, clears remote configuration and the Cloud binding before closing, and retains the final workspace as an unbound local anchor.

Validation:
- On cwgws, a synthetic absent machine returned {ok:true, already_gone:true} while its bound workspace remained. This depends on window lifecycle state; a settled window can pass.
- Patched app b1c734953ddd47464d0a91654eb248c86a6e4c84: three consecutive synthetic absent-machine deletions removed their workspaces; the final-workspace case cleared its binding.
- Fleet job 25b3c7281bf282b5990c7abc compiled and packaged that revision.
- Added a behavioral test for deletion during window context replacement, repeated deletion, and preserving unrelated local workspaces. Hosted fixed and original-cleanup control runs are pending.
- Latest PR-head build job: 35043763c58f5e28de1b3c64.

The earlier production Iroh 404 and nightly route-error traces were not correlated to cwgws and are not evidence for this fix.


lawrencecchen:
Reproduced and verified.

Observed on the old cwgws build:
- vm.destroy returned {ok:true,already_gone:true} for an already-missing VM.
- The bound Cloud workspace remained in workspace.list with its cloud_vm_id.
- It continued attempting reconnects and showed the desktop connection failure.

Root cause:
- deletion cleanup only scanned mainWindowContexts plus the active manager;
- it did not traverse recoverable/windowless managers;
- it closed multi-tab workspaces without clearing the remote configuration and Cloud binding first.

Fix:
- use liveWorkspaceIdentityTabManagers(preferredTabManager:);
- clear remote configuration and cloudVMBinding before close;
- retain the final tab as a local unbound anchor.

Verification:
- fixed fleet build job 35043763c58f5e28de1b3c64 completed successfully;
- fixed app live test passed 3 repeated synthetic missing-VM deletions;
- final-workspace test left one local panel with no Cloud VM binding;
- fixed hosted unit test run 35493434729 passed;
- control run against original cleanup is still running.

PR #13185, opened 2026-09-20
test: separate reftable semantics from production latency budgets

teamleaderleo:
Reftable semantic tests inherit the production two-second Git-read budget while running real subprocesses alongside the rest of the package. The full package reproduces a failure in `metadataUsesGitResolvedWorktreeBranchAndWatchesReftableStorage` even though the suite passes in isolation. Temporary local instrumentation showed a successful `rev-parse` starting with 14 ms remaining and taking 33 ms; the reader correctly rejected it as incomplete after the deadline.

Give these real-process semantic tests an explicit 15-second budget through existing injection points. For the ambient-repository test, use the executable that successfully created the reftable fixture, so environment isolation is tested independently of backend fallback. Preserve branch, commit, watch-path, unborn-branch, and fallback assertions, and add a direct check that an expired reference read remains unreadable without starting Git. Production code and its deadlines are unchanged.

Validation: baseline full-package runs failed locally; the test-first commit also fails the existing linked-worktree test (208 tests, 1 issue). With the fix, the full package passes, including the deadline/process-termination tests. A separate full run with Git 2.53 preferred for fixture creation also passes 208 tests. Builds use two compiler jobs locally, without launching the app. No hosted diagnostic run was dispatched.

This is a scoped reliability fix, not a claim that the original CI failure is conclusively explained: #13163 failed `plumbingIgnoresAmbientRepositorySelection`, which still passes in isolation locally, and CI's Homebrew Git is 2.55 rather than the local versions. Hosted confirmation remains pending.

Related: #13095, #13163.

---
## Summary by cubic
Reftable semantic tests were inheriting the production two-second Git-read budget, so real-process assertions could time out when the full package runs in parallel. They now use an explicit 15-second wall-time budget; production code and deadlines are unchanged.

**Changes**
- Adds a direct assertion that an expired reference read returns `.unreadable` without starting Git.
- Uses the executable that created the reftable fixture for the ambient-repository test so environment isolation is tested independently of backend fallback.
- Preserves the existing branch, commit, watch-path, unborn-branch, and fallback assertions.

**Validation**
- Full package passes locally (208 tests) with default Git and with Git 2.53 preferred for fixture creation.
- Related: #13095, #13163; #13163 remains unexplained by this scoped fix and hosted confirmation is pending.

<sup>Written for commit 23b61feca1b121bf533ecfcb9f2b15b267213e3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/manaflow-ai/cmux/pull/13185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


teamleaderleo:
Duplicate of https://github.com/manaflow-ai/cmux/pull/13186, which merged. Both fixed the same reftable test flake in the same file.

